PR #280, opened 2015-12-10
Port .hgignore to .gitignore

arusahni:
For future iterations, we should switch to using pre-canned ignore lists
from https://gitignore.io.


PCManticore:
Thank you.

PR #287, opened 2015-12-11
Check for flags/enum types before checking for int

dashea:
Running _import_gi_module('gi.repository.GLib') with astroid-1.4.1 fails with a syntax error while trying to parse the generated module stub, and I found that the modcode contained properties such as:

`IO_ERR = <flags G_IO_ERR of type GLib.IOCondition>`

This is due to the switch from `type(x) == y` to `isinstance(x, y)` in a079c3d4abecee064c3ab4c44f086b618f06fa8c. GLib.IO_ERR, for example, returns gi.repository.GLib.GIOCondition for type(), but also inherits from int.

### DIFF
--- a/astroid/brain/brain_gi.py
+++ b/astroid/brain/brain_gi.py
@@ -47,13 +47,13 @@ def _gi_build_stub(parent):
         elif (inspect.ismethod(obj) or
               inspect.ismethoddescriptor(obj)):
             methods[name] = obj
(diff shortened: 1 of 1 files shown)


PCManticore:
Thanks! We could add some tests for this plugin in order to prevent future regressions. I must admit though that this is a black box for me, since I'm not familiar with gi.

PR #295, opened 2015-12-15
Setup a basic AppVeyor deployment.

PCManticore:
Hey folks,

Here's a basic setup of AppVeyor which I'd like to get into astroid as part of its testing suite. It kinda works so far, but depending on numpy takes its toll reflected on the unusual build run time (almost an hour for all the Python versions). I tried multiple solutions, but none worked so far. If you have any idea how we might improve things on this side, please tell me about it. 

cc @The-Compiler would you mind setting up an AppVeyor account for pylint-bot, when you'll get some time? Most probably using it is the best solution.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,34 @@
+version: '{branch}-{build}'
+build: off
+cache:
+  - 'C:\\tmp'
+environment:
+  matrix:
+    - PYTHON: "C:\\Python27"
+      TOXENV: "py27"
+
+    - PYTHON: "C:\\Python33"
+      TOXENV: "py33"
+
+    - PYTHON: "C:\\Python34"
+      TOXENV: "py34"
+
+    - PYTHON: "C:\\Python35"
+      TOXENV: "py35"
+
+init:
+  - ps: echo $env:TOXENV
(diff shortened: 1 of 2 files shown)


ceridwen:
I know next to nothing about installing Python on Windows, so I don't know if I comment about that, though I support testing on it.   This might be a good place to handle IronPython, since .NET is Windows-native?  For numpy, I still think we should split off most of brain.

PCManticore:
I agree with the splitting, but in my opinion we should do it after switching completely to GitHub (having Travis and AppVeyor working for most of the cases).

PCManticore:
Yeah, we could add IronPython as well, after having the basic blocks integrated.

The-Compiler:
https://ci.appveyor.com/project/pylint-bot/pylint and https://ci.appveyor.com/project/pylint-bot/astroid are up and running. :wink:

PCManticore:
Thanks @The-Compiler ! By the way, I think you need to add the hooks for the project. I pushed a commit, but no build was triggered or am I doing something wrong?

PCManticore:
Let's merge it as is and we'll do a test PR shortly.

PR #297, opened 2015-12-16
Add support for inference on threading.Lock

lmedioni:
For now the test fails, due to issue #265. Based on PR #296, everything is OK.

### DIFF
--- a/astroid/brain/brain_stdlib.py
+++ b/astroid/brain/brain_stdlib.py
@@ -424,6 +424,15 @@ def multiprocessing_managers_transform():
             pass
     '''))
 
+def thread_transform():
+    return AstroidBuilder(MANAGER).string_build('''
+
+class Lock(object):
+    def acquire(self, blocking=True):
+        pass
+    def release(self):
+        pass
+''')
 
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_stdlib.py
```diff
@@ -437,3 +446,4 @@ def shutdown(self):
 register_module_extender(MANAGER, 'multiprocessing.managers',
                          multiprocessing_managers_transform)
 register_module_extender(MANAGER, 'multiprocessing', multiprocessing_transform)
+register_module_extender(MANAGER, 'threading', thread_transform)
```
PCManticore: Did you see my comment from https://bitbucket.org/logilab/astroid/pull-requests/98/add-support-for-inference-on-threadinglock/diff regarding transforms vs inference tip?

lmedioni: Sorry. You're right but I'm running out of time... Would it be OK if I modify the module transform to have Lock being a function?

PCManticore:
Merged with a couple of small changes https://github.com/PyCQA/astroid/commit/af0e3f5658097a21b78e3008854fa02ddbfc0ab3. I changed it to a function, which returns an instance of a lock class.

PR #298, opened 2015-12-19
Add /usr/lib and /usr/lib64 to the list of stdlib paths

PCManticore:
This patch adds /usr/lib/python{ver} and /usr/lib64/python{ver} to the list of places where a standard module library could be found.  Issues such as https://github.com/PyCQA/pylint/issues/725 are occurring because those modules are found in /usr/lib64/python2.7, which were not looked before by modutils.is_standard_module. @The-Compiler since you're more familiar with Unix systems that me, do you find this approach safe? Can there be extension packages installed there instead of /usr/local/lib/python{ver} on other OSes? The idea is to have a list of standard library places, so that is_standard_module(module_name) returns true for them, which in certain cases could lead to importing those modules if they are C extensions, so this might not be safe if third party modules get installed there.

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -54,12 +54,7 @@ else:
     PY_SOURCE_EXTS = ('py',)
     PY_COMPILED_EXTS = ('so',)
 
-# Notes about STD_LIB_DIRS
-# Consider arch-specific installation for STD_LIB_DIRS definition
-# :mod:`distutils.sysconfig` contains to much hardcoded values to rely on
-#
-# :see: `Problems with /usr/lib64 builds <http://bugs.python.org/issue1294959>`_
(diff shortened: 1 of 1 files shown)


The-Compiler:
I'm not very familiar with this to be honest... isn't third-party stuff in `dist-packages` or `site-packages` usually?

PCManticore:
Yeah, but on some OSes (Ubuntu for instance), they get installed in /usr/local/lib rather than /usr/lib. I'm not familiar either..

dmand:
IIRC, latest versions of pip install packages into `~/.local/lib`.

PCManticore:
We might solve https://github.com/PyCQA/pylint/issues/536 with a similar change.

PR #303, opened 2016-01-14
added 3 pkg_resources functions to brain

mbdevpl:
[Those 3 are] methods of WorkingSet objects [that] are also available
as module- level functions in pkg_resources that apply to the default
working_set instance. Thus, you can use e.g. pkg_resources.require()
as an abbreviation for pkg_resources.working_set.require()

above taken from:
https://pythonhosted.org/setuptools/pkg_resources.html#basic-workingset-methods

this fixes https://github.com/PyCQA/pylint/issues/780

Signed-off-by: Mateusz Bysiek <mb@mbdev.pl>

### DIFF
--- a/astroid/brain/brain_stdlib.py
+++ b/astroid/brain/brain_stdlib.py
@@ -150,6 +150,15 @@ class deque(object):
 
 def pkg_resources_transform():
     return AstroidBuilder(MANAGER).string_build('''
+def require(*requirements):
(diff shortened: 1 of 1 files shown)


The-Compiler:
I'm not very familiar with astroid brain stuff, but isn't the point of this to *not* call the real code behind it, i.e. should it call stuff from `pkg_resources`?

PCManticore:
That's a string, it doesn't call real functions.

PR #305, opened 2016-01-14
Fix unpack_infer() to fail if results are empty.

dbaum:
As expected, this will also fix pylint so that it correctly deals with "raise ()" in the python3 checker.

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -42,7 +42,7 @@ from astroid import util
 BUILTINS = six.moves.builtins.__name__
 MANAGER = manager.AstroidManager()
 
-
+@decorators.raise_if_nothing_inferred
 def unpack_infer(stmt, context=None):
     """recursively generate nodes inferred by the given statement.
     If the inferred value is a list or a tuple, recurse on the elements
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -42,6 +42,7 @@ from astroid import util
 BUILTINS = six.moves.builtins.__name__
 MANAGER = manager.AstroidManager()
 
+
 @decorators.raise_if_nothing_inferred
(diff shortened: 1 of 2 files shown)


PCManticore:
Merged localy https://github.com/PyCQA/astroid/commit/99a3870d4ba212a036dd629ad9ee020333009425. I also squashed your commits into one and cherry-picked it in multiple branches.

dbaum:
Thanks!

I'm new to using github, etc.  Is there something I should have done to
make the pull easier for you?  Should I have combined the commits
together?  Should I have not tried to synced with upstream/master?

I'm asking because it is likely that I'll have a few more small patches to
astroid in the near future and I want to make things as smooth as possible
for you.

Dave

PCManticore:
No worries. I prefer to have the commits squashed if they are not part of a bigger feature, in that case having some smaller chunks makes more sense. Also you should use rebase for feature branches, so that you won't be needing syncing with the upstream/master. It should be good after that, the cherry-picking is something that I must do, since we have multiple branches of development at the moment.

Thanks for the (future) contributions, let me know if there is something I could help you with.

dbaum:
I'm trying to understand the workflow.  Should I be doing my work on a
brach, keep my master at (or behind) the upstream master, then rebase my
feature branch to master and squash the commits at that point?  Something
like:

git checkout -b new-feature master
git commit -a -m "some work"
git commit -a -m "some more work"

# I'm a little fuzzy on exactly what commands to use here.  The idea is to
bring my
# master up to date with the upstream master.
git pull upsstream/master

# Go back to working on my fix...
git checkout new-feature
git commit -a -m "even more work"

# Clean things up before a pull request.
git rebase -i master

PCManticore:
Exactly, that's the workflow that I'm using mostly.

PR #308, opened 2016-01-15
Fix metaclass infinite loop when a class is its own metaclass.

dbaum:
The unit test shows some very strange source code that can cause an infinite recursion in astroid.  It looks like the context normally provides protection against this sort of thing, but it isn't plumbed through to the check of the assignment.  Here's one possible fix.  I'm completely open to other solutions for fixing this since adding the context arg to a few methods does seem a bit invasive.  Maybe there's a cleaner solution.

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -68,7 +68,7 @@ def _object_type(node, context=None):
     for inferred in node.infer(context=context):
         if isinstance(inferred, scoped_nodes.ClassDef):
             if inferred.newstyle:
-                metaclass = inferred.metaclass()
+                metaclass = inferred.metaclass(context=context)
                 if metaclass:
                     yield metaclass
                     continue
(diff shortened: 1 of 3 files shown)


ceridwen:
This looks fine to me.  @PCManticore?

PCManticore:
I'm not sure we should proliferate the use of the context as a parameter for all these methods. I'll look closely into the problem later today to see if there's another solution we might use for fixing this.

dbaum:
Any ideas on this?  It's not urgent, but if you have suggestions on other ways to fix the problem please let me know.

PCManticore:
Taking a look today or tomorrow, sorry for the delay, I got sidetracked in January with a lot of stuff.

PCManticore:
It seems that this is fixed in 2.0, but it reproduces on master. Unfortunately I couldn't find the commit where this got fixed, but since 2.0 will become the new master in the next couple of weeks, I'm considering this fixed. Sorry for taking me so long to look over this issue and please let me know if it still reproduces for you on 2.0 (please note though that it is in an unstable test and pylint doesn't work yet with it)

PR #314, opened 2016-02-04
Changed the way how parameters are being built

### LINKED ISSUE #215: Add a nodes.Arg to hold argument names, annotations, defaults for nodes.Arguments
Originally reported by: **BitBucket: [ceridwenv](http://bitbucket.org/ceridwenv), GitHub: @ceridwen**

----------------------------------------

The constructors for Arguments nodes now look like:

```python
    def __init__(self, vararg=None, kwarg=None, parent=None):
        self.vararg = vararg
        self.kwarg = kwarg
        self.parent = parent
        self.args = []
        self.defaults = []
        self.kwonlyargs = []
        self.kw_defaults = []
        self.annotations = []
        self.kwonly_annotations = []

    def postinit(self, args, defaults, kwonlyargs, kw_defaults,
                 annotations, kwonly_annotations, varargannotation=None,
                 kwargannotation=None):
        self.args = args
        self.defaults = defaults
        self.kwonlyargs = kwonlyargs
        self.kw_defaults = kw_defaults
        self.annotations = annotations
        self.varargannotation = varargannotation
        self.kwargannotation = kwargannotation
        self.kwonly_annotations = kwonly_annotations
```

Every property for every argument is held in a different field.  I want to move to a different arrangement that's similar to how inspect's Signature objects handle arguments: https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object.  Arguments will have four fields, args, vararg, kwarg, and (on Python 3) kwonlyargs, each of which will hold a sequence of Arg (I'm not sure this is the best name) nodes which will have a name field, a default field, and (on Python 3) an annotation field.  This arrangement makes it easier to look up the value of an argument's annotation or default by name, by first finding the appropriate Arg.node and then looking up the annotation or default.  Some disadvantages to this layout are that on Python 2 the Args node for varargs and kwargs is entirely superfluous, since they can't have default values and there are no annotations on Python 2, and that there will need to be a special constant object representing cases where there's no annotation or default.

Some possible alternative layouts imitate inspect.Signature more closely or imitate the standard library ast module.  inspect.Signature has only one field for function parameters and has parameter objects with a kind field that distinguishes between positional-or-keyword arguments, keyword-only arguments, varargs, and kwargs.  In my experience, this ends up leading to a lot of filtering boilerplate because when processing arguments, the argument's kind almost always matters.  The standard library ast.Arguments has a layout of args, defaults (default values for positional-or-keyword arguments), vararg, kwonlyargs, kw_defaults, and kwargs.  On Python 3, the elements of the appropriate sequences are an ast.Arg node with name and annotation fields, while on Python 2 they're strings.  This arrangement has two disadvantages: writing Python 2/3-compatible code requires constant dispatching on six.PY2/PY3 because the elements of the sequences are different objects with different fields depending on the version, and looking the default value corresponding to a given name requires code like `reversed(list(zip(reversed(args), reversed(defaults)))`, something similar with zip_longest, or the equivalent using indexing to line up the correct name with the correct default.

----------------------------------------
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/215

PCManticore:
The old way consisted in having the parameter names, their defaults and their annotations separated in different components of the Arguments node. We introduced a new Param node, which holds
the name of a parameter, its default value and its annotation. If any of the last two values are missing, then that slot will be filled with a new node kind, Empty, which is used for specifying the
lack of something (None could have been used instead, but that means having non-AST nodes in the Arguments node). We're also having support for positional only arguments, for the moment only in raw_building, but we can add later on for extension modules and other builtins.  I decided to add a flag to the Param nodes which describes it as being positional only or not, because introducing another object to the Arguments node would mean to iterate over both lists (.args and .positional_only_args) when wanting to process all the parameters of a function, so this seems way easier to me to manage.

Close #215

@ceridwen Please do a review when you'll get some time.


ceridwen:
I would use "Parameter" as the name for the new node rather than "Param."  I don't like shortening names into non-words unless it's necessary, and I don't think it is here.  For instance, inspect/funcsigs calls their parameter object "Parameter."

For the implementation of Empty, I was planning to have `__bool__` return False, like None is treated as False.  This will save a lot of refactoring work changing boolean checks to type checks and preserve backwards compatibility for the plugins.  Empty should be a singleton object: there's no reason to consume additional memory or time by creating a new object every time we need an Empty node, since all of them are identical anyways.  I'd call the underlying type `_Empty` or  `_EmptyType` or `EmptyType` (by analogy with NoneType) and then instantiate the object once in node_classes.py.

PCManticore:
Thanks, these make sense! I'll update the PR.

astroid/raw_building.py
```diff
@@ -319,65 +319,73 @@ def ast_from_function(func, built_objects, module, name=None, parent=None):
...
-    def extract_args(parameters, parent):
-        '''Takes an iterator over Parameter objects and returns three
-        sequences, arg names, default values, and annotations.
-
-        '''
-        names = []
-        defaults = []
-        annotations = []
+    def _extract_args(parameters, parent):
+        """Generate an iterator of params from Python parameters."""
```
ceridwen: Since we're now going to have two objects named Parameter (the astroid node and the inspect/funcsigs.Parameter), I'd rewrite this to say something like, "Generates an iterator of Parameter nodes from an iterator of inspect/funcsigs.Parameter objects."

astroid/raw_building.py
```diff
@@ -319,65 +319,73 @@ def ast_from_function(func, built_objects, module, name=None, parent=None):
...
 
-    def extract_vararg(parameter):
-        '''Takes a single-element iterator possibly containing a Parameter and
-        returns a name and an annotation.
+            param.postinit(default=default, annotation=annotation)
+            yield param
 
-        '''
+    def _extract_vararg(parameter, parent):
+        """Build a variadic parameter from the given Python parameter."""
```
ceridwen: This docstring should also be expanded similarly.

astroid/raw_building.py
```diff
@@ -319,65 +319,73 @@ def ast_from_function(func, built_objects, module, name=None, parent=None):
...
+            return node_classes.Empty()
+
+        positional_only = parameter.kind == _Parameter.POSITIONAL_ONLY
+        if parameter.annotation is not _Parameter.empty:
+            annotation = _ast_from_object(parameter.annotation,
+                                          built_objects, module, parent=parent)[0]
+        else:
+            annotation = node_classes.Empty()
+
+        # TODO: check if it is okay to use the parent here?
```
ceridwen: This will hopefully be irrelevant soon, but I don't understand this comment, so you should probably expand it.

ceridwen:
I think that our handling of positional-only and keyword-only arguments ought to be consistent: there should be another field in Arguments nodes to hold positional-only arguments like `kwonlyargs`, or keyword-only arguments should be flagged in the Parameter nodes like the positional-only arguments are.  Doing one with a flag in Parameters and the other with a field in Arguments is going to lead to confusion later, even though Python handles these parameters differently (positional-only parameters are only possible in C, keyword-only arguments are available in pure Python).

astroid/tree/rebuilder.py
```diff
@@ -113,6 +114,53 @@ def _get_context(node):
     return CONTEXTS.get(type(node.ctx), astroid.Load)
 
 
+class ParameterVisitor(object):
```
ceridwen: This is a quibble but this could probably be a subclass of `ast.NodeVisitor`.  In the future, I'd like to reorganize the rebuilder to make it so that the main visitor contains only methods corresponding to the standard library AST's nodes, and move all other nodes to independent functions or something else here.  We might also consider if using the `ast.NodeVisitor` is appropriate for the main visitor: it might offer better forward compatibility when the CPython developers change something in the standard library AST.  On the other hand, it wouldn't work with turning the visitor methods into generators, which I proposed to prevent it from overflowing the call stack and which might also speed it up.  If we're going to end up moving to another parser like 2to3 or Dropbox's altered version of the CPython parser, all this is moot, though.

PCManticore:
Thanks for the review. Here's the updated version. I dropped the positional_only attribute from the Parameter node and went with separate lists of arguments, but I also included a new attribute in the Arguments node, positional_and_keyword, which holds both positional only and positional_or_keyword arguments. That's because when processing the arguments we're interested in both, while trying to separate their processing from kwonlyargs, so whenever .args.args was used, it got replaced with positional_and_keyword list. Decided not to subclass ParameterVisitor from ast.NodeVisitor, it's not a big deal for now and we're planning to change it anyway.

astroid/test_utils.py
```diff
@@ -46,8 +47,18 @@ def _extract_expressions(node):
             child = getattr(node.parent, name)
             if isinstance(child, (list, tuple)):
                 for idx, compound_child in enumerate(child):
-                    if compound_child is node:
+
+                    # Can't find a cleaner way to do this.
```
ceridwen: This comment could be expanded to clarify what this code is doing.  (I'm not clear on its purpose.)

astroid/tests/unittest_python3.py
```diff
@@ -249,6 +250,13 @@ def test_unpacking_in_dict_getitem(self):
             self.assertIsInstance(value, nodes.Const)
             self.assertEqual(value.value, expected)
 
+    @require_version('3.4')
```
ceridwen: Why is this test only run on 3.4?  Shouldn't it at least run on 3.5 as well?

PCManticore: That's the minimum.

astroid/tree/node_classes.py
```diff
@@ -129,54 +147,25 @@ class Name(base.LookupMixIn, base.NodeNG):
...
-        self.defaults = defaults
-        self.kwonlyargs = kwonlyargs
-        self.kw_defaults = kw_defaults
-        self.annotations = annotations
-        self.varargannotation = varargannotation
-        self.kwargannotation = kwargannotation
-        self.kwonly_annotations = kwonly_annotations
+        self.keyword_only = keyword_only
+        self.positional_only = positional_only
+        self.positional_and_keyword = self.args + self.positional_only
```
ceridwen: This should be a read-only property to avoid the possibility of these attributes getting unsynchronized from one another.  If we ever make the nodes immutable, that would also fix the problem.

astroid/tests/unittest_python3.py
ceridwen: Ah, my mistake.  That works then.

ceridwen:
I added a few final comments, otherwise looks good to me.

astroid/tree/node_classes.py
PCManticore: I'm not keen on making this a property for now. It's used in a lot of places in astroid and I don't want to append these lists all the time I'm using positional_and_keyword. The exposed problem (arguments getting unsynchronized) shouldn't happen, since the nodes should be immutable, at least recommending to do so (if someone wants to add a new argument, it should create a new Arguments node instead).

ceridwen: I'm fully onboard with making nodes immutable, I just wasn't sure if that was your plan.  If we do make nodes immutable, though, I'd like to add programmatic safety to ensure they're immutable, replacing mutable structures like lists with tuples, for instance, and documentation.  None of this is urgent, so leaving this as-is is fine as long as we're planning to push for immutability in 2.0.

On February 13, 2016 1:18:07 PM EST, Claudiu Popa <notifications@github.com> wrote:
>> -        self.kwonly_annotations = []
>> -
>> -    def postinit(self, args, defaults, kwonlyargs, kw_defaults,
>> -                 annotations, kwonly_annotations,
>varargannotation=None,
>> -                 kwargannotation=None):
>> -        self.args = args
>> -        self.defaults = defaults
>> -        self.kwonlyargs = kwonlyargs
>> -        self.kw_defaults = kw_defaults
>> -        self.annotations = annotations
>> -        self.varargannotation = varargannotation
>> -        self.kwargannotation = kwargannotation
>> -        self.kwonly_annotations = kwonly_annotations
>> +        self.keyword_only = keyword_only
>> +        self.positional_only = positional_only
>> +        self.positional_and_keyword = self.args +
>self.positional_only
>
>I'm not keen on making this a property for now. It's used in a lot of
>places in astroid and I don't want to append these lists all the time
>I'm using positional_and_keyword. The exposed problem (arguments
>getting unsynchronized) shouldn't happen, since the nodes should be
>immutable, at least recommending to do so (if someone wants to add a
>new argument, it should create a new Arguments node instead).
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/PyCQA/astroid/pull/314/files#r52831455

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PCManticore: That's the plan. :-)

PR #318, opened 2016-02-13
Merge testdata's files into one bundle and revert previous changes to testdata

PCManticore:
This PR reverts the previous changes made to the testdata, bringing it again under astroid.tests.testdata. We're also conflating testdata.python2 and testdata.python3 into one folder, where small modifications were brought to the data files in order to work for both Python 2 and 3. The files with syntax errors though are generated dynamically.

@ceridwen let me know if this looks okay to you.


ceridwen:
This looks good to me.  I tested it against rope, one of my own scripts, and snakefood, and none of them crash, so I think you resolved the problems.  I'd like to clean up the tox.ini with respect to coverage and MANIFEST.in, based on some things I learned from that horrible experience with Python's packaging tools, but I can do that after the merge.

PCManticore:
Done!

PR #319, opened 2016-02-14
Introduce a special attributes model

PCManticore:
Through this model, astroid starts knowing special attributes of certain Python objects, such as functions, classes, super objects and so on. This was previously possible before, but now the lookup and the attributes themselves are separated into a new module, objectmodel.py, which describes, in a more comprehensive way, the data model of each object.

@ceridwen let me know what you think. It's still not perfect, at some point I'd like to move out the lookup completely off the nodes into its separate domain, but that's a medium term plan, not necessarily for now.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,14 @@ Change log for the astroid package (used to be astng)
 
       Closes issue #322
 
+    * Introduce a special attributes model
+
+      Through this model, astroid starts knowing special attributes of certain Python objects,
+      such as functions, classes, super objects and so on. This was previously possible before,
+      but now the lookup and the attributes themselves are separated into a new module,
+      objectmodel.py, which describes, in a more comprehensive way, the data model of each
+      object.
+
     * Changed the way how parameters are being built
 
       The old way consisted in having the parameter names, their
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -1,0 +1,514 @@
+# copyright 2003-2016 LOGILAB S.A. (Paris, FRANCE), all rights reserved.
+# contact http://www.logilab.fr/ -- mailto:contact@logilab.fr
+#
+# This file is part of astroid.
+#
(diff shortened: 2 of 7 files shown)


PCManticore:
PR updated to support some rudimentary descriptor binding (func.__get__(X, Y)). It works only partially, because I had to rebuild the function which got bound for changing the parent to the class which gets bound to when the BoundMethod is created, but I need to do the same for the Arguments node and probably for the components from the body, which quickly gets into a slippery slope of rebuilding ASTs. I'll revisit this when we'll have the zipper, since it's probably going to be easier to make a subpart of the tree to depend on a new parent.

ceridwen:
My general comment so far is that some more high-level documentation would be useful.  I see what you're doing but not how the pieces are going to fit together.  How finished is this?  I'll offer other comments when I have a better understanding of where you're going.

PCManticore:
Sure, I'll add a description. I think it's fairly finished, there are some unimplemented special objects, but I'm not planning to implement them any time soon (__closure__ for instance).

PCManticore:
I added a short description to the objectmodel.py's docstring. Let me know if this seems enough.

PCManticore:
@ceridwen Any comment on this PR? I'd like to merge it in the following days and some feedback would be appreciated (it's a big change and hopefully I didn't miss anything, so it's better to have two pairs of eyes than one)

ceridwen:
The docstring description did help me to understand what was going on, yes.  One early top-level comment: if the object model is part of the lookup mechanism, should it be lookup/ rather than interpreter/?  Are lookup/ and interpreter/ well-enough distinguished to be separate directories?

PCManticore:
No, the lookup will be part of the interpreter namespace. We might have in
the future something as interpreter/lookup, but I am not sure if we need to
add it now.

ceridwen:
Beyond my inline comments, this looks good to me, and an improvement on what we have now.

PCManticore:
Thanks for the review @ceridwen

PR #325, opened 2016-03-29
Zipper pull request 

ceridwen:
See #163.


astroid/__init__.py
```diff
@@ -82,6 +82,8 @@
 from astroid.builder import parse
 from astroid.util import Uninferable, YES
 
+# TODO
```
PCManticore: Why is this a TODO? :-)

astroid/tests/unittest_zipper.py
```diff
@@ -0,0 +1,278 @@
...
+import astroid
+from astroid import nodes
+from astroid.tree import base
+from astroid.tree import zipper
+
+
+# This is a strategy that generates a random file name, screening out
+# the empty init files because they produce 1-element ASTs that aren't
+# useful for testing.
+astroid_file = strategies.sampled_from(os.path.join(p, n) for p, _, ns in os.walk('astroid/') for n in ns if n.endswith('.py') and '__init__.py' not in n)
```
PCManticore: The line is a bit long, can you split it up a bit?

astroid/tree/base.py
```diff
@@ -107,6 +127,10 @@ def __str__(self):
             lines = pprint.pformat(value, indent=2,
                                    width=width).splitlines(True)
 
+            # Some fields, notably source_code for Module nodes, are
+            # too long to display comfortably, so this limits them.
+            if len(lines[0]) > 160:
```
PCManticore: Let's hardcode this value somewhere as a global variable, rather than directly in this function.

astroid/tree/zipper.py
```diff
@@ -0,0 +1,424 @@
...
+import wrapt
+
+from astroid.tree import base
+from astroid.tree import treeabc
+
+# The following are helper functions for working with singly-linked
+# lists made with two-tuples.  The empty tuple is used to denote the
+# end of a linked list.  The zipper needs singly-linked lists for most
+# of its operations to take constant time.
+def linked_list(*values):
```
PCManticore: These needs to be public?

astroid/tree/zipper.py
```diff
@@ -0,0 +1,424 @@
...
+    attributes are accessible through it.
+
+    Attributes:
+        __wrapped__ (base.NodeNG, collections.Sequence): The AST node or
+            sequence at the zipper's focus.
+        _self_path (Path): The Path tuple containing information about the
+            zipper's history.  This must be accessed as ._self_path.
+
+    '''
+    __slots__ = ('path')
```
PCManticore: Does this slot definition actually do anything in combination with a Proxy?

PCManticore:
This looks nice, thanks! The code is quite clean and easy to understand.

astroid/__init__.py
ceridwen: I've been using the base.NodeNG class for debugging, and the TODO is primarily to remind myself to have a discussion with you about exporting it (we need to) and how we ought to do that.

astroid/tests/unittest_zipper.py
ceridwen: Yes, that's on my list for the final clean-up phase.

astroid/tree/zipper.py
ceridwen: Probably not, though I'm not 100% sure so I haven't changed them yet.

astroid/tree/zipper.py
ceridwen: Yes.  wrapt.ObjectProxy has `__slots__` defined: https://github.com/GrahamDumpleton/wrapt/blob/develop/src/wrapt/wrappers.py#L73, and I assume the C class is similar.

ceridwen:
In the last commit I pushed an experimental version of what moving some of the basic methods from NodeNG to the zipper would look like.  This didn't add any new bugs.

PCManticore:
Hey @ceridwen Thank you for this work! Unfortunately I don't think we'll be able to use a zipper in astroid any time soon and without someone to champion this change it will be impossible to land. As such I believe it's best we close this PR for now, and we can revisit later if that changes.

PR #333, opened 2016-04-27
Ignore PyGIWarning

giuseppe:
It solves this error when running PyLint on a file which uses gi introspection:

/usr/lib/python2.7/site-packages/astroid/brain/brain_gi.py:136: PyGIWarning: OSTree was \
imported without specifying a version first. Use gi.require_version('OSTree', '1.0') before \
import to ensure that the right version gets loaded.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


PCManticore:
Thanks. The patch looks good.

PR #345, opened 2016-07-11
Fix quantity of blank lines after code object.

stepshal:
### Fixes / new features
- 
Add missing blank line.
Add missing 2 blank lines.
Remove extra blank lines.


PCManticore:
I'm sorry, we don't do across the board formatting for astroid, only on case by case basis, whenever a function gets modified, we could change the surrounding lines for it. Apart of this, the tests are also failing due to this change.

PR #357, opened 2016-08-20
Add pytest 3 compatibility in brain_pytest

The-Compiler:
This adds compatibility with pytest 3.0 to `brain_pytest.py`, where some attributes were added and others moved.

It also adds some failing attributes which are already present in older pytest versions, like `pytest.xfail` which existed for a long time.

In the second commit, I also re-added pytest 2.9 support (and it also seems to work fine with 2.7 still, didn't test older versions) as 3.0 is quite a big update and I don't expect everyone to upgrade immediately.

### DIFF
--- a/astroid/brain/brain_pytest.py
+++ b/astroid/brain/brain_pytest.py
@@ -18,19 +18,29 @@ try:
     import _pytest.recwarn
     import _pytest.runner
     import _pytest.python
+    import _pytest.freeze_support
+    import _pytest.skipping
+    import _pytest.assertion
+    import _pytest.debugging
+    import _pytest.fixtures
 except ImportError:
     pass
 else:
     deprecated_call = _pytest.recwarn.deprecated_call
+    warns = _pytest.recwarn.warns
+    xfail = _pytest.skipping.xfail
     exit = _pytest.runner.exit
     fail = _pytest.runner.fail
-    fixture = _pytest.python.fixture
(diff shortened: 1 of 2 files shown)


astroid/brain/brain_pytest.py
```diff
@@ -18,19 +18,62 @@ def pytest_transform():
...
+        register_assert_rewrite = _pytest.assertion.register_assert_rewrite
+    except AttributeError:
+        pass
+
+
+# Moved in pytest 3.0
+
+try:
+    import _pytest.freeze_support
+    freeze_includes = _pytest.freeze_support.freeze_includes
```
The-Compiler: I decided to do the assignments here instead of in an `else:` as there's no way this second line could raise an `ImportError`, and this makes the nested `ImportError`s much more readable IMHO.

The-Compiler:
Do I need to add myself to the copyrights, or will you run copyrite before the next release?

PCManticore:
I can run it before the next release, you don't have to worried about that. Do you need a patch release or something? Since the default will go into 1.5.0, it could potentially take a while to be released, together with pylint 2.0, which is why I am asking.

The-Compiler:
I think that would be great to have - I already have some suppressions for a `no-member` with `pytest.xfail`.

I'm not sure what happens after upgrading to pytest 3.0 - the old code snippet would probably raise an `AttributeError`, but I don't get any `no-member` messages anyways?

Issue #358, opened 2016-08-22
Installing: Invalid environment marker: python_version<"3.4"


moylop260:
### Steps to reproduce
1. `pip install --upgrade git+https://github.com/PyCQA/astroid@41a0c`

### Current behaviour
Error
```bash
Collecting git+https://github.com/PyCQA/astroid@41a0c
  Cloning https://github.com/PyCQA/astroid (to 41a0c) to /var/folders/zc/1m37yppx4z5596yfyzg385140000gn/T/pip-Z0zkzq-build
  Could not find a tag or branch '41a0c', assuming commit.
    Complete output from command python setup.py egg_info:
    error in astroid setup command: Invalid environment marker: python_version<"3.4"

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /var/folders/zc/1m37yppx4z5596yfyzg385140000gn/T/pip-Z0zkzq-build/
```

### Expected behavior
No error

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output
Sha: https://github.com/PyCQA/astroid/commit/41a0c77cab88a54787d49913030bebf482fa014c

python --version
Python 2.7.10

python -m pip --version
pip 8.1.2 from /Users/moylop260/Library/Python/2.7/lib/python/site-packages (python 2.7)

python -c "import setuptools;print setuptools.__version__"
1.1.6

Same case for https://travis-ci.org/Vauxoo/pylint-odoo/jobs/154185968#L456

PCManticore:
Try to update setuptools and see if the problem goes away.

moylop260:
```bash
$ pip install --upgrade --user setuptools pip
Collecting setuptools
  Using cached setuptools-26.0.0-py2.py3-none-any.whl
Requirement already up-to-date: pip in /Users/moylop260/Library/Python/2.7/lib/python/site-packages
Installing collected packages: setuptools
Successfully installed setuptools-1.1.6

$ pip install --user --upgrade git+https://github.com/PyCQA/astroid@41a0c
Collecting git+https://github.com/PyCQA/astroid@41a0c
  Cloning https://github.com/PyCQA/astroid (to 41a0c) to /var/folders/zc/1m37yppx4z5596yfyzg385140000gn/T/pip-4l21cO-build
  Could not find a tag or branch '41a0c', assuming commit.
    Complete output from command python setup.py egg_info:
    error in astroid setup command: Invalid environment marker: python_version<"3.4"

[...]

PR #360, opened 2016-08-23
[REF] package: Compatibility with old version of setuptools

moylop260:
### Fixes / new features
- https://github.com/PyCQA/astroid/issues/358

- old version of setuptools
  ```bash
$ python -c "from setuptools import __version__ as setuptools_version;from sys import version_info as py_version;print 'py_version', py_version;print 'setuptools_version', setuptools_version"
  py_version sys.version_info(major=2, minor=7, micro=10, releaselevel='final', serial=0)
  setuptools_version 1.1.6

$ pip install --upgrade git+https://github.com/moylop260/astroid.git@6760ea
  Successfully installed astroid-1.5.0
```

- new version of setuptools
  ```bash
$ python -c "from setuptools import __version__ as setuptools_version;from sys import version_info as py_version;print 'py_version', py_version;print 'setuptools_version', setuptools_version"
py_version sys.version_info(major=2, minor=7, micro=10, releaselevel='final', serial=0)
setuptools_version 26.0.0
  py_version sys.version_info(major=2, minor=7, micro=10, releaselevel='final', serial=0)
  setuptools_version 26.0.0

$ pip install --upgrade git+https://github.com/moylop260/astroid.git@6760ea
  Successfully installed astroid-1.5.0
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,9 @@ install:
   - $PYTHON_EXE -m easy_install --version
   - $PYTHON_EXE -m pip --version
   - $PYTHON_EXE -m tox --version
-  - $PYTHON_EXE -m pip install -U setuptools
 script:
+  - python -m tox -e coverage-erase,$TOXENV
+  - $PYTHON_EXE -m pip install -U setuptools
   - python -m tox -e coverage-erase,$TOXENV
 after_success:
   - tox -e coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ install:
   - $PYTHON_EXE -m pip --version
   - $PYTHON_EXE -m tox --version
 script:
-  - python -m tox -e coverage-erase,$TOXENV
+    # Test install with current version of setuptools
+  - $PYTHON_EXE -m pip install .
+    # Run the tests with newest version of setuptools
   - $PYTHON_EXE -m pip install -U setuptools
   - python -m tox -e coverage-erase,$TOXENV
 after_success:
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -30,10 +30,7 @@ def has_environment_marker_range_operators_support():
     The first known release to support environment marker with range operators
     it is 17.1, see: https://setuptools.readthedocs.io/en/latest/history.html#id113
     """
-    try:
-        return parse_version(setuptools_version) >= parse_version('17.1')
-    except Exception:
-        return False
+    return parse_version(setuptools_version) >= parse_version('17.1')
 
 
 if has_environment_marker_range_operators_support():
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -6,6 +6,12 @@
 # For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
 
 """astroid packaging information"""
+
+from sys import version_info as py_version
+
+from setuptools import __version__ as setuptools_version
+
+
 distname = 'astroid'
 
 modname = 'astroid'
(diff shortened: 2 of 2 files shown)


astroid/__pkginfo__.py
```diff
@@ -15,7 +21,12 @@
 
 extras_require = {}
 install_requires = ['lazy_object_proxy', 'six', 'wrapt']
-extras_require[':python_version<"3.4"'] = ['enum34', 'singledispatch']
+
+if py_version < (3, 4) and setuptools_version < '21.0.0':
```
The-Compiler: You can't just compare strings:

```
>>> '3.0.0' < '21.0.0'
False
```

See what [pytest does](https://github.com/pytest-dev/pytest/blob/master/setup.py#L44) here.

moylop260: Excellent code!
Thanks for feedback

Fixed

jayvdb: The acceptable setuptools version for `python_version` is 3.3 according to https://gsnedders.github.io/python-marker-test/results.html

moylop260: @jayvdb I don't understand your point.
Could you help me to clarify?

jayvdb: setuptools 3.3 is sufficient for the environment marker "`python_version`", so you can use
`setuptools_version < (3, 3)`

moylop260: @jayvdb Thanks for clarify

I added (new commit) the following comment:

> The first known release to support environment marker with range operators
    it is 17.1, see: https://setuptools.readthedocs.io/en/latest/history.html#id113

It is important the **range operators** because we are using a range with `python_version<"3.4"`

.travis.yml
```diff
@@ -34,9 +34,10 @@ install:
   - $PYTHON_EXE -m easy_install --version
   - $PYTHON_EXE -m pip --version
   - $PYTHON_EXE -m tox --version
-  - $PYTHON_EXE -m pip install -U setuptools
```
PCManticore: Why does this have to be moved?

astroid/__pkginfo__.py
```diff
@@ -15,7 +21,26 @@
...
+def has_environment_marker_range_operators_support():
+    """Code extracted from 'pytest/setup.py'
+    https://github.com/pytest-dev/pytest/blob/7538680c/setup.py#L31
+
+    The first known release to support environment marker with range operators
+    it is 17.1, see: https://setuptools.readthedocs.io/en/latest/history.html#id113
+    """
+    try:
+        return parse_version(setuptools_version) >= parse_version('17.1')
+    except Exception:
```
PCManticore: What exception is this expected to raise?

PCManticore:
Looks good. Check those two small comments I've left and we are good to go. Thank you.

.travis.yml
moylop260: We need validate that works with current travis setuptools version and with newest setuptools version.
Then, we need execute the tests without a `pip -U` (old version) and run the tests again after a `pip -U` (new version)
```bash
# test with old setuptools version:
python -m tox -e coverage-erase,$TOXENV

# Update newest setuptools version:
$PYTHON_EXE -m pip install -U setuptools

# test with newest setuptools version:
python -m tox -e coverage-erase,$TOXENV
```

astroid/__pkginfo__.py
moylop260: I don't have the real case
I just use a similar code from: https://github.com/pytest-dev/pytest/blob/7538680c/setup.py#L45-L46

.travis.yml
PCManticore: But why? Now I actually noticed what is happening, this doubles the tests we are running just for testing with older setuptools versions?! Sorry this makes no sense whatsoever.

astroid/__pkginfo__.py
PCManticore: I am pretty sure then this can be simplified. You can try to investigate what errors parse_version is supposed to raise and catch only those, for instance.

.travis.yml
moylop260: You are right!
We don't need run the tests 2 times.
I changed for just test the installation of the package without errors.

astroid/__pkginfo__.py
moylop260: I investigated but I don't found nothing and test with many versions of setuptools without a error.
Maybe is a unnecessary global `except`
I removed and I we find a new issue we could add the explicit except.

PR #364, opened 2016-09-27
Updates for python 3.6

jgarst:
Fixed one regression (inferred._proxied is uninferrable), and added two new ast nodes for python 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,11 @@ install:
   - $PYTHON_EXE -m easy_install --version
   - $PYTHON_EXE -m pip --version
   - $PYTHON_EXE -m tox --version
-  - $PYTHON_EXE -m pip install -U setuptools
 script:
+    # Test install with current version of setuptools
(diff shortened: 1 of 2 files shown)


jgarst:
Apologies for pushing bad code. I'm getting different test results on my local machine - I'll get things consistent sometime this week and resubmit.

astroid/builder.py
```diff
@@ -237,6 +237,8 @@ def delayed_assattr(self, node):
                 try:
                     if inferred.__class__ is bases.Instance:
                         inferred = inferred._proxied
```
PCManticore: I would like to see a test first with this problem, maybe it will have another fix than ignoring the node.

PCManticore:
Thanks for the PR! I left a comment, please check it. I would like to have that change separately, including a test for reproduction. As mentioned, it might have a different fix, depending on the nature of the bug. Without it, it looks fine.

PR #365, opened 2016-10-02
add format string support

jgarst:
Apologies for taking a bit on this.

A problem with circular imports caused me to compare a class name to a string in the joinedstr as_string method. It's fixable, but also felt like another pull request worth of code surgery. 

Without the change to delay_asattr in the second pull request there are a bunch of failed tests with python 3.6b1, specifically:

testExtensionModules
test_multiprocessing_manager
test_from_self_resolve
test_metaclass_generator_hack
test_lru_cache
test_namedtuple_advanced_inference
test_attribute_access

There  are a few other tests that are already failing on my machine, test_namespace_package_pth_support
test_builtin_fromlineno_missing on python3.5+

I took some stabs in the dark at fixing them, and can open more pull requests if you would like. But they might be pretty crude.

### DIFF
--- a/astroid/as_string.py
+++ b/astroid/as_string.py
@@ -481,6 +481,19 @@ class AsStringVisitor3(AsStringVisitor):
     def visit_asyncfor(self, node):
         return 'async %s' % self.visit_for(node)
 
+    def visit_joinedstr(self, node):
+        # Special treatment for constants,
+        # as we want to join literals not reprs
+        string = ''.join(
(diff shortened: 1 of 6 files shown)


PCManticore:
Thank you! Will merge this. You can ignore the builtin_fromlineno test failure, it is a spurious failure for which we know the cause and the test is going to be removed anyway soon.

tgy:
Has this been included in any release yet?

PCManticore:
No.

PR #366, opened 2016-10-02
fix delayed_asattr when proxy is uninferrable

jgarst:
Without this extra check, the following tests fail on python 3.6b1:

testExtensionModules
test_multiprocessing_manager
test_from_self_resolve
test_metaclass_generator_hack
test_lru_cache
test_namedtuple_advanced_inference
test_attribute_access


PCManticore:
The problem was fixed by https://github.com/PyCQA/astroid/commit/1b34d34e87a014d1ad96263169b25ecf34e900cb. Thank you for your PR, even though it didn't get merged.

PR #370, opened 2016-10-24
Python 2.6 compatibility

bocekm:
### Fixes
- Add Python 2.6 compatibility

Justification: On Red Hat Enterprise Linux 6 the only available Python version is 2.6. The compatibility with this version is needed for running pylint as a part of CI tests.
Thanks @misli for help.


PCManticore:
Closing. See the justifications from https://github.com/PyCQA/pylint/pull/1145 for why we don't want to support 2.6 any longer.

Issue #337, opened 2016-05-13
load_module_from_modpath does not fully support namespace packages


NotSqrt:
Hi,

I opened a ticket in PyCQA/pylint#903 but it is probably in astroid itself:
I have the following situation:

```
# package 1, installed with "pip install package1"
mynamespace/
    __init__.py
    pylint_plugins/
        __init__.py
       plugin1.py

# package 2, installed with "pip install -e package2"
mynamespace/
    __init__.py
   # other modules
```

In that situation, `pylint --load-plugins mynamespace.pylint_plugins.plugin1` fails, because `load_module_from_modpath` looks for `pylint_plugins` inside the folder of package2 installed in develop mode.

Cf what Django does to avoid this problem (it used to fail in some cases for django < 1.7): https://github.com/django/django/blob/1.9.6/django/utils/module_loading.py#L9

Steps to reproduce easily:

``` bash
mktmpenv 
pip install pylint
mkdir -p package1/mynamespace/pylint_plugins
mkdir -p package2/mynamespace/
# setup scripts
echo -e "from setuptools import setup, find_packages\nsetup(name='package1', version='0.1', packages=find_packages())" > package1/setup.py
echo -e "from setuptools import setup, find_packages\nsetup(name='package2', version='0.1', packages=find_packages())" > package2/setup.py
# namespace declaration
echo -e "from pkgutil import extend_path\n__path__ = extend_path(__path__, __name__)" > package1/mynamespace/__init__.py
echo -e "from pkgutil import extend_path\n__path__ = extend_path(__path__, __name__)" > package2/mynamespace/__init__.py
# other python files
touch package1/mynamespace/pylint_plugins/__init__.py
echo -e "def register(linter):\n    pass" > package1/mynamespace/pylint_plugins/plugin1.py

pip install package1/
pip install -e package2/
pylint --load-plugins mynamespace.pylint_plugins.plugin1  
# Fails because it looks for pylint_plugins in /absolute/path/to/package2/
# Doing the installs in reverse order does not fix the error

# both with -e
pip uninstall -y package1 package2
pip install -e package1/ -e package2/
pylint --load-plugins mynamespace.pylint_plugins.plugin1  # ok

# both regular install
pip uninstall -y package1 package2
pip install package1/ package2/
pylint --load-plugins mynamespace.pylint_plugins.plugin1  # ok
```

How to fix this ?

Since python2.6 does not seem supported in tox, we might simply use importlib, no ?

Thanks !

[...]

PR #372, opened 2016-11-03
Look under module.__path__ for finding submodules if it exists.

srikiraju:
Need to test this a bit more, but I think we should look under module.`__path__`. I'll try to generate a reproducer and see if this is the same bug in the issue.

### Fixes / new features
-  #337


PCManticore:
I couldn't reproduce the original issue. I'm closing this one.

PR #373, opened 2016-11-06
1.4.5

johnsonzhuangsh:
### Fixes / new features
-


PCManticore:
I don't understand what you intended here, but this is not the way we are doing releases.

PR #382, opened 2016-12-29
infer results of open() and file() calls

rogalski:
Needed for PyCQA/pylint#1245.

It’s a very rough draft, input will be appreciated.

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -490,35 +490,37 @@ def infer_open(node, context=None):
     if not ((six.PY2 and module == '__builtin__') or (six.PY3 and module == '_io')):
         raise UseInferenceDefault
 
-    open_return_classdef_node = _get_open_return_classdef_node(node, context)
+    open_return_classdef_node = _open_return_classdef_node(node, context)
+    if open_return_classdef_node is None:
+        raise UseInferenceDefault
     return open_return_classdef_node.instantiate_class()
 
 
-def _get_open_return_classdef_node(node, context):
+def _open_return_classdef_node(node, context):
     if six.PY2:
-        node = _get_py2_open_return_class(node, context)
+        node = _py2_open_return_class(node, context)
     else:
-        node = _get_py3_open_return_class(node, context)
-    return next(node.infer())
+        node = _py3_open_return_class(node, context)
+    try:
+        return next(node.infer())
+    except InferenceError:
+        return None
 
 
-def _get_py2_open_return_class(node, context):  # pylint: disable=unused-argument
-    code = dedent('''
-        file #@
-    ''')
-    return extract_node(code)
+def _py2_open_return_class(node, context):  # pylint: disable=unused-argument
+    return extract_node('file')
 
 
-def _get_py3_open_return_class(node, context):
+def _py3_open_return_class(node, context):
     code = '''
         import io
         io.{class_name}  #@
     '''
-    class_name = _get_py3_open_class_name(node, context)
+    class_name = _py3_open_class_name(node, context)
     return extract_node(code.format(class_name=class_name))
 
 
-def _get_py3_open_class_name(node, context):
+def _py3_open_class_name(node, context):
     # https://docs.python.org/3/library/functions.html#open
     # TODO: use call_site
     mode = _get_mode(node, context)
@@ -562,23 +564,19 @@ def _get_call_arg(node, context, arg, kwarg, default):
 
 def enter_returns_self(node):
     """ Make Pylint understand context manager protocol for file classes """
-    from astroid import FunctionDef
-    enter_funcdefs = [funcdef for funcdef in node.body
-                      if isinstance(funcdef, FunctionDef)
-                      and funcdef.name == "__enter__"]
-    if not enter_funcdefs:
-        return
-    enter_funcdef = enter_funcdefs[0]
-    if enter_funcdef.body:
-        return
-    code = """
-    class Class(object):
-        def __enter__(self):
-            return self
-    """
-    fake_node = extract_node(code)
-    enter_funcdef.args = fake_node.body[0].args
-    enter_funcdef.body = fake_node.body[0].body
+    from astroid import FunctionDef, extract_node
+    for idx, n in enumerate(node.locals.get('__enter__', [])):
+        if not isinstance(n, FunctionDef) or n.body:
+            continue
+        code = """
+        class Class(object):
+            def __enter__(self):
+                return self
+        """
+        # TODO: nodes in astroid 2.0 should be immutable
+        fake_node = extract_node(code)
+        n.args = fake_node.body[0].args
+        n.body = fake_node.body[0].body
 
 
 # Builtins inference
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -481,77 +481,97 @@ def infer_slice(node, context=None):
     return slice_node
 
 
-def infer_open(node, context=None):  # pylint: disable=unused-argument
+def infer_open(node, context=None):
     """Understand `open` (and `file` in Python 2) calls."""
+    open_return_classdef_node = _get_open_return_classdef_node(node, context)
+    return open_return_classdef_node.instantiate_class()
 
+
+def _get_open_return_classdef_node(node, context):
     if six.PY2:
-        code, module_name = _get_py2_open_call_model(node)
+        node = _get_py2_open_return_class(node, context)
     else:
-        code, module_name = _get_py3_open_call_model(node)
-
-    node = extract_node(code, module_name)
-    return node.instantiate_class()
+        node = _get_py3_open_return_class(node, context)
+    return next(node.infer())
 
 
-def _get_py2_open_call_model(node):  # pylint: disable=unused-argument
-    # https://docs.python.org/2/library/stdtypes.html#file-objects
-    # https://github.com/python/typeshed/blob/master/stdlib/2/_io.pyi
+def _get_py2_open_return_class(node, context):  # pylint: disable=unused-argument
     code = dedent('''
-        class file(object): #@
-            closed = True
-            def __enter__(self): return self
-            def __exit__(self, type, value, traceback): return False
-            def __iter__(self): return self
-            def close(self): return None
-            def fileno(self): return 0
-            def flush(self): return None
-            def isatty(self): return True
-            def next(self): return ""
-            def read(self, size=-1): return ""
-            def readline(self, size=-1): return ""
-            def readlines(self, sizehint=-1): return [""]
-            def xreadlines(self, sizehint=-1): yield ""
-            def seek(self, offset, whence=0): return None
-            def tell(self): return 0
-            def truncate(self, size=0): return None
-            def write(self, s): return None
-            def writelines(self, seq): return None
+        file #@
     ''')
-    module_name = "__builtin__"
-    return code, module_name
+    return extract_node(code)
 
 
-def _get_py3_open_call_model(node):  # pylint: disable=unused-argument
-    # TODO: mimic exact intepreter behavior (e.g. buffered and bytes IO)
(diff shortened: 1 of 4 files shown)


rogalski:
I'm not sure why coverage dropped.

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,79 @@ def infer_slice(node, context=None):
...
 
+def infer_open(node, context=None):  # pylint: disable=unused-argument
+    """Understand `open` (and `file` in Python 2) calls."""
+
+    if six.PY2:
+        code, module_name = _get_py2_open_call_model(node)
+    else:
+        code, module_name = _get_py3_open_call_model(node)
+
+    node = extract_node(code, module_name)
```
rogalski: Extracted class nodes should be kept in module namespace instead of recreating them each time.

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,79 @@ def infer_slice(node, context=None):
...
+            def tell(self): return 0
+            def truncate(self, size=0): return None
+            def write(self, s): return None
+            def writelines(self, seq): return None
+    ''')
+    module_name = "__builtin__"
+    return code, module_name
+
+
+def _get_py3_open_call_model(node):  # pylint: disable=unused-argument
```
rogalski: TODO: inheritance structure is fairly complicated in Py3 case (with BytesIO, TextIO, buffering etc.). For 100% proper inference logic here needs to be significantly improved.

Maybe we should write a simple testcase that generates all valid combos of input arguments and checks if inferred values are equivalent to actually interpreted objects.

astroid/brain/brain_builtin_inference.py
PCManticore: Not sure I'm understanding what you are saying. Can you explain what do you mean?

astroid/brain/brain_builtin_inference.py
PCManticore: I think another good approach, which I considered at some point, is to check the arguments of the ``open`` call. Depending on these, my idea was to return a different object. You can try this idea, in combination with another trick for not regenerating the classes all the time. Instead of building them as you are building them now, you might try to use this:

```
node = extract_node('''
from io import YourIOClass
YourIOClass
''')
inferred = next(n.infer())
return inferred.instantiate_class()
```
What do you think? Can you try this to see if it can work?

PCManticore: This means instead of returning a code string, you can return an instance instead.

PCManticore:
Looks good, but check my suggestion to see if it can improve the situation for Python 3.

astroid/brain/brain_builtin_inference.py
rogalski: Currently, for each inference of "open", mock is recreated each time. It may (and should) be created only once, and on each inference it should just create a Instance node.

astroid/brain/brain_builtin_inference.py
rogalski: Yes, for Python 3 it'll have to check input arguments to figure out correct instance.

Using import in place of hand-written mocks is also a great idea that I wasn't aware of. Thanks!

rogalski:
@PCManticore most definitely, thanks for your comments!

rogalski:
@PCManticore

How would I reasonably infer arguments of function call? Covering only kwargs/only args/unpacking/defaults etc. will be tricky. Is `CallSite` something which may be used here?

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,99 @@ def infer_slice(node, context=None):
...
+def _get_mode(node, context):
+    return _get_call_arg(node, context, arg=1, kwarg='mode', default='r')
+
+
+def _get_buffering(node, context):
+    return _get_call_arg(node, context, arg=2, kwarg='buffering', default=-1)
+
+
+def _get_call_arg(node, context, arg, kwarg, default):
+    """TODO: move to generic solution"""
```
rogalski: Is `CallSite` a way to go here?

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,99 @@ def infer_slice(node, context=None):
...
+            return inferred.value
+    kws = [kw for kw in node.keywords or [] if kw.arg == kwarg]
+    if len(kws) == 1:
+        inferred = helpers.safe_infer(kws[0].value, context)
+        if inferred:
+            return inferred.value
+    return default
+
+
+def enter_returns_self(node):
```
rogalski: I think this is (or may be) a generic operation. We have a "dummy" `ClassDef` built based on C-implemented built-ins where `FunctionDef` in bodies are empty.

General task would be to patch instances (based on predicate)  - based on some fake node. Key statement here is to _patch_ selectively, not to _replace_ with stub.

astroid/brain/brain_io.py
```diff
@@ -35,9 +36,37 @@ def _transform_buffered(node):
     return _generic_io_transform(node, name='raw', cls=FileIO)
 
 
+def _enter_returns_self(node):
```
rogalski: It's duplicated here, because all brain modules weren't importing anything outside of `astroid` root.

astroid/tests/unittest_inference.py
```diff
@@ -811,6 +813,73 @@ def test_builtin_open(self):
...
+            with open("toto.txt") as fh:
+                pass
+        '''
+        ast = parse(code, __name__)
+        node = ast['fh']
+        inferred = list(node.infer())
+        self.assertEqual(len(inferred), 1)
+        self.assertNotEqual(inferred[0], util.Uninferable)
+
+    def test_builtin_open_assignment_infers_proper_class(self):
```
rogalski: As comment says, brute-force, but I guess it does what it should?

astroid/tests/unittest_regrtest.py
```diff
@@ -141,7 +141,8 @@ def test_with_infer_assignnames(self):
         astroid = builder.string_build(data, __name__, __file__)
         # Used to crash due to the fact that the second
         # context manager didn't use an assignment name.
-        list(astroid.nodes_of_class(nodes.Call))[-1].inferred()
+        with self.assertRaises(exceptions.InferenceError):
```
rogalski: Before my changes, inferred value was `Uninferable`. Now it's `InferenceError`.

rogalski:
It needs more work, but conceptually it's much closer to final patch comparing to previous one.

Coverage drop is due to guard clauses not being hit.

astroid/brain/brain_builtin_inference.py
PCManticore: Yes, definitely, we can use CallSite here.

astroid/brain/brain_builtin_inference.py
PCManticore: This use case, dummy ``ClassDef`` built based on C-implemented builtin-ins, sounds like a case for ``raw_building.py``. We have there code for building exactly these dummy classes, in order to provide us with some hints on what might be going on at the built-in level. You can also check the code from 2.0 branch, which is rewritten and far better than what we have currently in master.

Now, leaving that aside, I think it would be better if you replace the enter method in ``node.body`` with a newly generated one and returning the same node from the transform. It should be okay as you did right now, but, for astroid 2.0, we are aiming to have immutable nodes, so every time one wants to patch a node, he will have to return another node instead. Not relying on monkey patching would ease my job for porting this code to 2.0.

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,106 @@ def infer_slice(node, context=None):
...
+    """ Make Pylint understand context manager protocol for file classes """
+    from astroid import FunctionDef
+    enter_funcdefs = [funcdef for funcdef in node.body
```
PCManticore: Why can't you use ``.locals`` here instead?

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,106 @@ def infer_slice(node, context=None):
...
+def _get_open_return_classdef_node(node, context):
+    if six.PY2:
+        node = _get_py2_open_return_class(node, context)
+    else:
+        node = _get_py3_open_return_class(node, context)
+    return next(node.infer())
+
+
+def _get_py2_open_return_class(node, context):  # pylint: disable=unused-argument
+    code = dedent('''
```
PCManticore: No need for dedent, ``extract_node('file')`` is enough

astroid/brain/brain_builtin_inference.py
```diff
@@ -481,6 +481,106 @@ def infer_slice(node, context=None):
...
+def infer_open(node, context=None):
+    """Understand `open` (and `file` in Python 2) calls."""
+    inferred = helpers.safe_infer(node.func)
+    if inferred in (None, util.Uninferable):
+        raise UseInferenceDefault
+    module = inferred.root().name
+    if not ((six.PY2 and module == '__builtin__') or (six.PY3 and module == '_io')):
+        raise UseInferenceDefault
+
+    open_return_classdef_node = _get_open_return_classdef_node(node, context)
```
PCManticore: Protect this with ``except InferenceError``.

PCManticore: Also, drop the ``_get`` from the name.

rogalski:
Quick question - why does astroid seems to be unable to fill arguments in inferred value for `open()` function? Inferred function seems to be correct - at least doc parameter is copied from proper built-in function.

`inspect` module seems to retrieve this signature easily, it looks like there's no reason for AST to not contain these information.

```
>>> print(next(astroid.extract_node('open').infer()).args)
Arguments(vararg=None,
          kwarg=None,
          args=None,
          defaults=[],
          kwonlyargs=[],
          kw_defaults=[],
          annotations=[],
          varargannotation=None,
          kwargannotation=None)
>>> import inspect
>>> inspect.signature(open)
<Signature (file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None)>
>>> 

```

PCManticore:
Because it is a builtin and we have limited support for inferring builtins. Check raw_building, where everything happens (and also check the rewrite from 2.0; unfortunately, in 2.0, we cannot infer open() at all, but that is due to a bug). inspect.signature is not a good comparison, since it is inspecing ``__text_signature__`` for builtins. astroid might access this in the future, since, for builtins, we are operating with the actual objects, so we can take this hints into account.

rogalski:
@PCManticore I can either hardcode information on open() signature in astroid code, or use inspect module to try to fill all objects built in `raw_builder.py`. For me, second option is much better than the first, but I guess it should target 2.0 branch, given huge rewrite was merged in that branch. On the same time, it's such a "big" change that you may be reluctant to do right now.

I see three options:
- hard-code information on open signature in astroid code, merge changes in pylint and astroid, postpone implementing generic feature of inspecting live objects to build more rich AST
- start implementing using instect to build more rich AST (which should target 2.0 branch)
- mark PRs as blocked and start working on them again after 2.0 is released

What you think about it?

PCManticore:
I think we can wait for 2.0, if you don't mind. We can just do the work once, in 2.0's ``raw_building.py``, rather than duplicating it for 1.5 and , then, for 2.0. 
While the code is there for quite some time, after pylint 2.0, I want to have more often releases, so this should be out pretty fast.

ceridwen:
Note that I rewrote raw_building.py in 2.0 to use inspect, including for building the builtin modules.  You should look at the result of that code and start from there, since this is target 2.0 anyways.

rogalski:
@ceridwen in which commit it was implemented?

PCManticore:
You can check the branch, I think there were multiple commits.

ceridwen:
@rogalski, It's a complete rewrite, so many different commits.  As @PCManticore said, the 2.0 branch holds the current version of raw_building.py.

DanielNoord:
I'm going to close this as there are now many merge conflicts and the reviewers are no longer active within the project. It would probably be better to start a new PR if anybody wants to tackle this.

PR #385, opened 2017-01-04
Implement __getitem__ inference for classes (using the metaclass)

### LINKED ISSUE #348: __getitem__ inference for classes (using the metaclass)
Ideally, we should be able to infer this:

```
n = extract_node('''
"""Hello"""
from __future__ import (
    absolute_import,
    division,
    print_function,
    unicode_literals,
)
from six import add_metaclass

class Meta(type):

    def __getitem__(cls, arg):
        return arg

@add_metaclass(Meta)
class Works(object):
    pass

    def __getitem__(self, arg): return arg



x = Works['hello']
x
''')
i = next(n.infer())
print(i)
```

euresti:
Essentially, implement the getitem method in ClassDef which returns the correct value.

Fixes #348

### DIFF
--- a/astroid/scoped_nodes.py
+++ b/astroid/scoped_nodes.py
@@ -25,6 +25,7 @@ from astroid import context as contextmod
 from astroid import exceptions
 from astroid import decorators as decorators_mod
 from astroid.interpreter import objectmodel
+from astroid.interpreter import dunder_lookup
 from astroid import manager
 from astroid import mixins
 from astroid import node_classes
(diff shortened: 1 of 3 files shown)


astroid/scoped_nodes.py
```diff
@@ -1509,6 +1510,29 @@ def _valid_getattr(node):
...
+
+        This is basically looking up the method in the metaclass and calling it.
+        """
+        try:
+            methods = dunder_lookup.lookup(self, '__getitem__')
+        except exceptions.AttributeInferenceError as error:
+            util.reraise(exceptions.InferenceError(**vars(error)))
+
+        method = methods[0]
+        # Rewrap index to Const for this case
```
PCManticore: This comment does not seem true in this context, since we are not wrapping anything in Const.

euresti:
Hmm that's copy pasted from another method, should I fix the comment there
too?

euresti:
Fixed the comment. What do you think?

PCManticore:
Thank you for this! I merged this manually.

PR #389, opened 2017-01-25
mark __init_subclass__ as classmethod [master branch]

rogalski:


rogalski:
Looks like I've tried to modify unittests that are not very "unit" in nature. I'll revisit it tomorrow (same goes to 2.0 change).

PR #392, opened 2017-01-28
RFC: Inject type comments 

euresti:
### Fixes / new features
- Added annotations to the PY2 ast.
- Adds a second pass of the source looking for `# type:` comments.  Adds those into the nodes.
- Some perf improvements.

Just looking for comments on my approach here.

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -17,10 +17,6 @@ import os
 import sys
 import textwrap
 import _ast
-import tokenize
-
-import collections
-from six import StringIO
 
 from astroid import bases
 from astroid import exceptions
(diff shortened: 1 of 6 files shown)


ceridwen:
Python 3.6 adds a new node to the ast, AnnAssign, and the same approach should be used here for type annotations on assignment statements, instead of adding a field to AssignNode nodes, see #394.

More broadly, we don't use type comments in pylint or astroid at the moment, so this pull request should be part of a larger discussion about when or if we want to use type comments.  We have discussed in the past the possibility of making the parser pluggable, so that someone could replace a 2.7 parser that doesn't process type comments with one that does.  That approach might still be better than this one.  As it is, astroid parses the code once with the standard library's ast module, rewrites the tree using rebuilder, and pylint reparses the code using tokenizer to find pragmas.  As-is, this patch adds another tokenizer pass to this process.  Aside from the performance implications, I feel like this is too much code duplication across the two code bases.

euresti:
This is real old.

Issue #393, opened 2017-01-29
Crash when trying to interpret PEP 526 variable type hints


cdunklau:
Looks like astroid doesn't yet support PEP 526 variable type hints, new in Python 3.6

### Steps to reproduce

1. Put the code `myvar: int` in a file.
2. Run pylint against that file

### Current behavior

Pylint crashes with this traceback. Appears to point at the new node introduced in Python 3.6's AST: `AnnAssign(expr target, expr annotation, expr? value, int simple)`

```
Traceback (most recent call last):
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/pylint/lint.py", line 927, in get_ast
    return MANAGER.ast_from_file(filepath, modname, source=True)
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/manager.py", line 86, in ast_from_file
    return AstroidBuilder(self).file_build(filepath, modname)
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/builder.py", line 141, in file_build
    module = self._data_build(data, modname, path)
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/builder.py", line 185, in _data_build
    module = builder.visit_module(node, modname, node_file, package)
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/rebuilder.py", line 154, in visit_module
    newnode.body = [self.visit(child, newnode) for child in node.body]
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/rebuilder.py", line 154, in <listcomp>
    newnode.body = [self.visit(child, newnode) for child in node.body]
  File "/Users/cdunklau/Development/pngdoctor/.tox/style/lib/python3.6/site-packages/astroid/rebuilder.py", line 165, in visit
    visit_method = getattr(self, visit_name)
AttributeError: 'TreeRebuilder3k' object has no attribute 'visit_annassign'
************* Module pylint_breaks_on_AnnAssign
F:  1, 0: <class 'AttributeError'>: 'TreeRebuilder3k' object has no attribute 'visit_annassign' (astroid-error)
```

### Expected behavior

No crash, perhaps a warning about this not being supported yet?

### Pylint and Astroid Versions
- pylint 1.6.5,
[...]

PR #394, opened 2017-01-30
Add support for annotated assignment nodes

rr-:
Crude implementation for #393

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -1051,6 +1051,19 @@ class Assign(mixins.AssignTypeMixin, Statement):
         self.value = value
 
 
+class AnnAssign(mixins.AssignTypeMixin, Statement):
+    """class representing an AnnAssign node"""
+    _astroid_fields = ('targets', 'value', 'annotation',)
+    targets = None
+    value = None
+    annotation = None
+
+    def postinit(self, targets=None, value=None, annotation=None):
+        self.targets = targets
+        self.value = value
+        self.annotation = annotation
+
+
 class AugAssign(mixins.AssignTypeMixin, Statement):
     """class representing an AugAssign node"""
     _astroid_fields = ('target', 'value')
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -1059,7 +1059,7 @@ class AnnAssign(mixins.AssignTypeMixin, Statement):
     value = None
     simple = None
 
-    def postinit(self, target=None, annotation=None, value=None, simple=None):
+    def postinit(self, target, annotation, simple, value=None):
         self.target = target
         self.annotation = annotation
         self.value = value
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -1053,15 +1053,17 @@ class Assign(mixins.AssignTypeMixin, Statement):
 
 class AnnAssign(mixins.AssignTypeMixin, Statement):
     """class representing an AnnAssign node"""
-    _astroid_fields = ('target', 'value', 'annotation',)
+    _astroid_fields = ('target', 'annotation', 'value', 'simple')
     target = None
-    value = None
     annotation = None
+    value = None
+    simple = None
 
-    def postinit(self, target=None, value=None, annotation=None):
+    def postinit(self, target=None, annotation=None, value=None, simple=None):
         self.target = target
-        self.value = value
         self.annotation = annotation
+        self.value = value
+        self.simple = simple
 
 
 class AugAssign(mixins.AssignTypeMixin, Statement):
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -1053,13 +1053,13 @@ class Assign(mixins.AssignTypeMixin, Statement):
 
 class AnnAssign(mixins.AssignTypeMixin, Statement):
     """class representing an AnnAssign node"""
-    _astroid_fields = ('targets', 'value', 'annotation',)
-    targets = None
+    _astroid_fields = ('target', 'value', 'annotation',)
+    target = None
     value = None
     annotation = None
 
-    def postinit(self, targets=None, value=None, annotation=None):
-        self.targets = targets
+    def postinit(self, target=None, value=None, annotation=None):
+        self.target = target
         self.value = value
         self.annotation = annotation
 
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -832,8 +832,8 @@ class TreeRebuilder3(TreeRebuilder):
         """visit a AnnAssign node by returning a fresh instance of it"""
         newnode = nodes.AnnAssign(node.lineno, node.col_offset, parent)
         annotation = _visit_or_none(node, 'annotation', self, newnode)
-        newnode.postinit([self.visit(node.target, newnode)],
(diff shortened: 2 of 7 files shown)


ceridwen:
Can you add a test for this to unittest_nodes.py?  The existing test Python35AyncTest in that file should be a good model.

rr-:
Pushed a basic test

ceridwen:
Sorry I forgot to mention this before, but you should use the _visit_or_none helper function in the rebuilder method, and it should be in TreeRebuilder3 since it's a Python 3 feature.

rr-:
Updated~

astroid/tests/unittest_nodes.py
```diff
@@ -464,6 +464,23 @@ def hello(False):
             self.assertEqual(del_true.name, "True")
 
 
+class AnnAssignNodeTest(unittest.TestCase):
+    def test_ann_assign(self):
```
rogalski: Please extend test suite for more complex cases found in [PEP526](https://www.python.org/dev/peps/pep-0526/):

```
primes: List[int] = []

captain: str  # Note: no initial value!

class Starship:
    stats: ClassVar[Dict[str, int]] = {}
```

astroid/tests/unittest_nodes.py
```diff
@@ -464,6 +464,23 @@ def hello(False):
...
+        """test annotated assignment node"""
+        code = textwrap.dedent("""
+            test: int = 5
+        """)
+        if sys.version_info < (3, 6):
```
rogalski: Consider using `test_utils.require_version` decorator.

astroid/node_classes.py
```diff
@@ -1051,6 +1051,19 @@ def postinit(self, targets=None, value=None):
         self.value = value
 
 
+class AnnAssign(mixins.AssignTypeMixin, Statement):
+    """class representing an AnnAssign node"""
+    _astroid_fields = ('targets', 'value', 'annotation',)
+    targets = None
```
rogalski: this likely should be named `target`, not `targets`. Annotated assign cannot be done anyway:

```
>>> (a, b): (int, int)
  File "<stdin>", line 1
SyntaxError: only single target (not tuple) can be annotated
```

Also, `help(_ast.AnnAssign)` yields:

    _fields = ('target', 'annotation', 'value', 'simple')

Astroid is supposed to be superset of `ast` and should be consistent with it. I'll recommend changing `_astroid_fields` value quoted above.

[PEP](https://www.python.org/dev/peps/pep-0526/#runtime-effects-of-type-annotations) also mentions _simple name_, but I don't think it affects us in any way.

rr-:
Updated~

ceridwen:
I'm going to merge this tomorrow or so if there are no more comments, unless someone else wants to do it.  The remaining CI failure looks like a misconfiguration of pylint somehow.

astroid/node_classes.py
```diff
@@ -1051,6 +1051,19 @@ def postinit(self, targets=None, value=None):
         self.value = value
 
 
+class AnnAssign(mixins.AssignTypeMixin, Statement):
+    """class representing an AnnAssign node"""
+    _astroid_fields = ('target', 'value', 'annotation',)
```
rogalski: we should add `simple` and change orderubg to be in conformance with built-in `ast`.

```
>>> ast.parse('a: int').body[0]._fields
('target', 'annotation', 'value', 'simple')
```

astroid/tests/unittest_nodes.py
```diff
@@ -464,6 +464,51 @@ def hello(False):
...
+        """)
+        assign = builder.extract_node(code)
+        self.assertIsInstance(assign, nodes.AnnAssign)
+        self.assertEqual(assign.target.name, "test")
+        self.assertIsInstance(assign.annotation, astroid.Subscript)
+        self.assertIsInstance(assign.value, astroid.Dict)
+
+    @test_utils.require_version(minver='3.6')
+    def test_as_string(self):
+        code = '''print()
```
rogalski: maybe actually indenting it and using `textwrap.dedent` will look slightly nicer in a code.

astroid/tests/unittest_nodes.py
```diff
@@ -464,6 +464,51 @@ def hello(False):
...
+test: int = 5
+test2: str
+test3: List[Dict[str]] = []
```
rogalski: Just a nitpick, test cases should be correct, let's use type specification for both keys and values: `Dict[str, str]`

astroid/rebuilder.py
```diff
@@ -828,6 +828,15 @@ def visit_try(self, node, parent):
         elif node.handlers:
             return self.visit_tryexcept(node, parent)
 
+    def visit_annassign(self, node, parent):
+        """visit a AnnAssign node by returning a fresh instance of it"""
```
PCManticore: visit an*

astroid/rebuilder.py
```diff
@@ -828,6 +828,19 @@ def visit_try(self, node, parent):
...
             return self.visit_tryexcept(node, parent)
 
+    def visit_annassign(self, node, parent):
+        """visit an AnnAssign node by returning a fresh instance of it"""
+        newnode = nodes.AnnAssign(node.lineno, node.col_offset, parent)
+        annotation = _visit_or_none(node, 'annotation', self, newnode)
+        newnode.postinit(self.visit(node.target, newnode),
+                         annotation,
+                         _visit_or_none(node, 'value', self, newnode),
+                         nodes.Const(node.simple,
```
rogalski: @PCManticore just before merging it - is it correct way to do it (creation of fake Const node)?

I encountered a very similar situation during implementation of [async comprehensions] support for astroid.

PCManticore: Mm, now that you mention this, we can probably call ``visit_const`` instead of building it on our own. Thanks for finding this.

astroid/node_classes.py
```diff
@@ -1053,15 +1053,17 @@ def postinit(self, targets=None, value=None):
...
-    _astroid_fields = ('target', 'value', 'annotation',)
+    _astroid_fields = ('target', 'annotation', 'value', 'simple')
     target = None
-    value = None
     annotation = None
+    value = None
+    simple = None
 
-    def postinit(self, target=None, value=None, annotation=None):
+    def postinit(self, target=None, annotation=None, value=None, simple=None):
```
ceridwen: Something I just noticed: which of these fields are optional in the standard-library ast nodes?  I'm assuming that the annotation, target, and value fields aren't, but I haven't looked at 3.6 myself.  Whatever fields aren't optional should have required arguments.  There's legacy code that requires that it be possible to build some AST nodes without having all the mandatory fields set, but this is bad practice, and we shouldn't be carrying it forward to new nodes.

rr-: https://docs.python.org/3/library/ast.html
```
          -- 'simple' indicates that we annotate simple name without parens
          | AnnAssign(expr target, expr annotation, expr? value, int simple)
```

Looks like only `value` is optional

PCManticore:
I merged this manually, with a couple of small changes, such as putting ``simple`` in ``_other_fields``, instead of ``_astroid_fields``. Thank you so much @rr- for your contribution!

PR #396, opened 2017-01-30
Infer starred expressions in tuple, list, set and dict literals

rogalski:
Just a POC, needs:

1) Guidance on if it’s a correct way to solve a problem
2) Extra tests
3) Docs entries

Aims to solve PyCQA/pylint#957

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -42,8 +42,6 @@ nodes.ClassDef._infer = infer_end
 nodes.FunctionDef._infer = infer_end
 nodes.Lambda._infer = infer_end
 nodes.Const._infer = infer_end
-nodes.Dict._infer = infer_end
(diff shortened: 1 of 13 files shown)


rogalski:
Lack of coverage is due to guard except clause not being hit. May be easily improved by adding additional tests.

astroid/inference.py
```diff
@@ -42,13 +42,78 @@ def infer_end(self, context=None):
...
+    return values
+
+
+nodes.List._infer = infer_seq
+nodes.Tuple._infer = infer_seq
+nodes.Set._infer = infer_seq
+
+
+def infer_map(self, context=None):
+    if not any(isinstance(k, nodes.DictUnpack) for k, v in self.items):
```
PCManticore: We don't use ``v``.

PR #397, opened 2017-02-05
Fix namespace population when wildcard import is used

rogalski:
Fix PyCQA/pylint#1312


PCManticore:
The current behavior is expected. You can check the previous commit where ``public_names`` was introduced: https://github.com/rogalski/astroid/commit/82c6ef644a2efb77217a23d9b8a6cfb5caffb4ba. As explained there, we cannot provide strong guarantees over ``__all__`` support, so we started relying on all the public names, instead of filtering them through ``__all__``. Cannot check right now the Pylint issue, since I am on a phone, but probably there is another solution to fix it.

PR #400, opened 2017-02-07
Asynchronous comprehensions

### LINKED ISSUE #399: Support for asynchronous comprehensions
See https://www.python.org/dev/peps/pep-0530/ for details

rogalski:
Fixes #399.

### DIFF
--- a/astroid/as_string.py
+++ b/astroid/as_string.py
@@ -132,7 +132,7 @@ class AsStringVisitor(object):
 
     def visit_comprehension(self, node):
         """return an astroid.Comprehension node as string"""
-        ifs = ''.join(' if %s' % n.accept(self) for n in node.ifs)
+        ifs = ''.join([' if %s' % n.accept(self) for n in node.ifs])
         return 'for %s in %s%s' % (node.target.accept(self),
                                    node.iter.accept(self), ifs)
 
@@ -505,10 +505,8 @@ class AsStringVisitor3(AsStringVisitor):
 
     def visit_comprehension(self, node):
         """return an astroid.Comprehension node as string"""
-        ifs = ''.join(' if %s' % n.accept(self) for n in node.ifs)
-        return '%sfor %s in %s%s' % ('async ' if node.is_async else '',
-                                     node.target.accept(self),
-                                     node.iter.accept(self), ifs)
+        return '%s%s' % ('async ' if node.is_async else '',
+                         super(AsStringVisitor3, self).visit_comprehension(node))
 
 
 def _import_string(names):
--- a/astroid/as_string.py
+++ b/astroid/as_string.py
@@ -132,7 +132,7 @@ class AsStringVisitor(object):
 
     def visit_comprehension(self, node):
         """return an astroid.Comprehension node as string"""
-        ifs = ''.join([' if %s' % n.accept(self) for n in node.ifs])
+        ifs = ''.join(' if %s' % n.accept(self) for n in node.ifs)
         return 'for %s in %s%s' % (node.target.accept(self),
                                    node.iter.accept(self), ifs)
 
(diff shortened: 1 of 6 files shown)


astroid/node_classes.py
```diff
@@ -1192,20 +1192,22 @@ def last_child(self):
 
 class Comprehension(NodeNG):
     """class representing a Comprehension node"""
-    _astroid_fields = ('target', 'iter', 'ifs')
+    _astroid_fields = ('target', 'iter', 'ifs', 'is_async')
```
PCManticore: Let's put ``is_async`` in ``_other_fields``. ``_astroid_fields`` is only for objects which are AST objects.

astroid/as_string.py
```diff
@@ -494,6 +494,13 @@ def visit_joinedstr(self, node):
     def visit_formattedvalue(self, node):
         return '{%s}' % node.value.accept(self)
 
+    def visit_comprehension(self, node):
+        """return an astroid.Comprehension node as string"""
+        ifs = ''.join([' if %s' % n.accept(self) for n in node.ifs])
```
PCManticore: ``' '.join('if %s' % n.accept(self) for i in ...)`` feels more natural to me.

astroid/rebuilder.py
```diff
@@ -357,7 +357,11 @@ def visit_comprehension(self, node, parent):
         newnode.postinit(self.visit(node.target, newnode),
                          self.visit(node.iter, newnode),
                          [self.visit(child, newnode)
-                          for child in node.ifs])
+                          for child in node.ifs],
+                         nodes.Const(getattr(node, 'is_async', None),
```
PCManticore: Check my previous comment regarding ``is_async``. Sorry, I think I misled you in a different PR, with a similar use case.

astroid/tests/unittest_python3.py
```diff
@@ -258,7 +258,74 @@ def test_underscores_in_numeral_literal(self):
...
+            extract_node("async def f(): return [i async for i in aiter() if i % 2]"),
+            extract_node("async def f(): return {i async for i in aiter() if i % 2}"),
+            extract_node("async def f(): return (i async for i in aiter() if i % 2)"),
+            extract_node("async def f(): return {i: i async for i in aiter() if i % 2}")
+        ]
+        return_non_async_comp_funcs = [
+            extract_node("async def f(): return {i: i for i in iter() if i % 2}")
+        ]
+
+        for func in return_async_comp_funcs:
```
PCManticore: You can use ``@`` to extract a particular node, instead of doing it manually, such as in this example:

```
f = extract_node('''
async def f():
   return [i async for i in...] #@
''')
# now f is the Return
```
Or you can use ``__()`` for a fine grained approach, as in:

```
f = extract_node('''
async def f():
   return __([i async for i in...])
''')
# now f is the comprehension
```

astroid/tests/unittest_python3.py
```diff
@@ -258,7 +258,74 @@ def test_underscores_in_numeral_literal(self):
...
+            comp = func.body[-1].value.generators[0]
+            self.assertTrue(comp.is_async.value)
+        for func in return_non_async_comp_funcs:
+            comp = func.body[-1].value.generators[0]
+            self.assertFalse(comp.is_async.value)
+
+    @require_version('3.7')
+    def test_async_comprehensions_outside_coroutine(self):
+        """
```
PCManticore: A comment should be enough here. If I remember correctly, docstrings have special meaning in unittests.

rogalski:
@PCManticore thanks for the review, see new commits for fixes

rogalski:
@PCManticore since dunder for async iterables is `__aiter__` and not `__iter__`, I'd expect that by default it would fail. I'll try to come up with a test case, but I haven't used asyncio at all, so it may take a while.

PCManticore:
@rogalski I think we can merge this as is, and create a new issue for adding support for ``__aiter__`` and async comprehension's ``assigned_stmts``.

rogalski:
@PCManticore agreed. I'll do a merge commit, just so all tests can complete successfully and _squash and merge_ it afterwards.

PR #402, opened 2017-02-08
Fix metaclass detection when multiple keyword arguments are used in ClassDef

rogalski:
Needed for PyCQA/pylint#1320


astroid/rebuilder.py
```diff
@@ -317,7 +317,7 @@ def visit_classdef(self, node, parent, newstyle=None):
             for keyword in node.keywords:
                 if keyword.arg == 'metaclass':
                     metaclass = self.visit(keyword, newnode).value
-                break
+                    break
```
PCManticore: Ah, wonder how the tests were passing.

PR #390, opened 2017-01-26
Add inference for typing.NamedTuple

euresti:
### Fixes / new features
- Adds support so typing.NamedTuple(...) actually infers something useful.

PR #403, opened 2017-02-08
Fix typing.NamedTuple being not callable

rogalski:
Needed for PyCQA/pylint#1320

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -233,7 +233,7 @@ class Instance(BaseInstance):
 
     def callable(self):
         try:
-            self._proxied.getattr('__call__', class_context=False)
+            self._proxied.getattr('__call__')
             return True
         except exceptions.AttributeInferenceError:
             return False
--- a/astroid/tests/unittest_inference.py
+++ b/astroid/tests/unittest_inference.py
@@ -3338,7 +3338,6 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
 
     @test_utils.require_version('3.5')
     def test_typing_namedtuple_is_callable(self):
-        import typing
         node = extract_node("""
         import typing
         MyNamedTuple = typing.NamedTuple('Named', [('foo', int), ('bar', int)])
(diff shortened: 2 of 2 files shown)


astroid/tests/unittest_inference.py
```diff
@@ -3336,6 +3336,20 @@ def blurb(self):
...
+    @test_utils.require_version('3.5')
+    def test_typing_namedtuple_is_callable(self):
+        import typing
+        node = extract_node("""
+        import typing
+        MyNamedTuple = typing.NamedTuple('Named', [('foo', int), ('bar', int)])
+        MyNamedTuple
+        """)
+        inferred = list(node.infer())
+        if util.Uninferable in inferred:
```
rogalski: I don't like this early exit, but on the other hand `typing` is known to be actively developed, so I don't see much value in chasing issues related to older version of library code...

astroid/bases.py
```diff
@@ -233,7 +233,7 @@ def __str__(self):
 
     def callable(self):
         try:
-            self._proxied.getattr('__call__', class_context=False)
+            self._proxied.getattr('__call__')
```
PCManticore: What is the reason for this change?

astroid/tests/unittest_inference.py
PCManticore: I don't like it either. Does this happen with older versions of ``typing``? We should, probably, pin it to a version that doesn't give us too much trouble.

astroid/bases.py
rogalski: typing.NamedTuple was inferred as not callable when class_context was False

PCManticore: Yeah, but I'm asking from a protocol perspective. ``class_context`` is False, since the lookup is done on the instance, thus it should not go through the metaclass lookup and so on. What I'm asking is why typing.NamedTuple is special. Is it an instance or a class? If it is an Instance, how come it has a ``__call__`` defined in the metaclass (I presume), that it can access in this place? Where does its ``__call__`` come from?

rogalski:
I'll close it - it's much better solved by #390

PR #404, opened 2017-02-09
object.__new__(cls) calls fails to infer correct type when used in inherited classmethod

rogalski:
These tests are MCVE that exposes root cause of PyCQA/pylint#224

TODO
- [x] implement tests
- [ ] implement fixes
- [ ] add changelog entries

### DIFF
--- a/astroid/tests/unittest_inference.py
+++ b/astroid/tests/unittest_inference.py
@@ -3336,6 +3336,52 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         self.assertIsInstance(inferred, nodes.Const)
         self.assertEqual(inferred.value, 25)
 
+    def test_factory_methods_cls_call(self):
+        ast = extract_node("""
+        class C:
+            @classmethod
+            def factory(cls):
+                return cls()
+
+            def m(self):
+                return self.__class__.factory()
+
+        class D(C):
+            pass
+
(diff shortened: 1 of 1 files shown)


PCManticore:
@rogalski Need any help over here?

brycepg:
It's interesting that `return cls()` and `return object.__new__(D)` both work properly

brycepg:
Well, I've found the source of the issue right [here](https://github.com/PyCQA/astroid/blob/master/astroid/bases.py#L310). The problem is I'm not sure how to fix it because of another bug.

`astroid` currently has a UnboundMethod to instantiate object.__new__ calls. The problem is that the `context.boundnode` is now pointing to `object`. `cls` gets inferred to the wrong class because no context is given upon inference (and if you add the context it will point to object)

PCManticore:
It sounds to me that we'd need to provide a correct `boundnode` when inferring a `Call` node (https://github.com/PyCQA/astroid/blob/b604b267f4876298d142f165713d6be1bd77790e/astroid/inference.py#L189), paired with passing a context to the `infer_call_result` in the `UnboundMethod.infer_call_result` method.

brycepg:
@PCManticore so for `UnboundMethod` calls the `boundnode` should be the type of the first argument?

brycepg:
I think this may be related to #177

PR #405, opened 2017-02-11
Improve inference of namedtuples

rogalski:
Closes PyCQA/pylint#1033


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -113,8 +113,12 @@ def infer_named_tuple(node, context=None):
     """Specific inference function for namedtuple Call node"""
     class_node, name, attributes = infer_func_form(node, nodes.Tuple._proxied,
                                                    context=context)
+    field_def = "    {name} = property(lambda self: self[{index:d}], doc='Alias for field number {index:d}')"
```
PCManticore: Let's make this a bit shorter. Maybe we can put it in parens, so that we won't have lines longer than 100.

PR #408, opened 2017-02-28
builder: fix detection of whether found file is a package when there is a __init__.py in file name

rogalski:
Fixes PyCQA/pylint#1348


PCManticore:
I don't think that this is the root cause of that bug. In the reproducing example provided there, the package had a ``__init__.py`` file. We should check where astroid is raising the ImportError in the first place, in order to know what triggers the bug.

rogalski:
@PCManticore Well, all I know is that it fixes reported issue. I'm just no sure where is a place for an automated test.

PR #409, opened 2017-02-28
No longer install tox / pip twice in travis

asottile:


asottile:
Seems to be left over from this commit: https://github.com/PyCQA/astroid/commit/8ed837d828692db5b95b3e906bebeacbc68c256f

PCManticore:
Thank you!

PR #410, opened 2017-03-08
Quickstart _infer_binary_operation to raise errors that occur during startup

cpennington:
Since https://github.com/PyCQA/astroid/commit/3e27213914271309a4716662b09fda91fca9efa1,
`is_subtype` and `is_supertype` raise exceptions when they can't find a relationship between types. This is caught by the callers of `_infer_binary_operation`. However, the error is only thrown after the generator return by `_infer_binary_operation` is started.

Prior to this patch, the generator is only started inside the `else` clause of:

```
try:
    results = _infer_binary_operation(lhs, rhs, self, context, _get_binop_flow)
except exceptions._NonDeducibleTypeHierarchy:
    yield util.Uninferable
else:
    for result in results:
        yield result
```

To address this, the `quickstart` decorator pulls the first value from the generator when it's called, and then buffers that first result until it is needed by the later iteration.

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -611,6 +611,22 @@ def _get_aug_flow(left, left_type, aug_opnode, right, right_type,
     return methods
 
 
+def quickstart(func):
+    """
+    Run a generator enough to get the first value, and then buffer that value.
+    This will raise any exceptions that happen during the generator startup
+    when func is called, rather than when it is used.
+    """
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        result_generator = func(*args, **kwargs)
+        first_value = next(result_generator)
+
+        return itertools.chain([first_value], result_generator)
+    return wrapper
+
+
+@quickstart
 def _infer_binary_operation(left, right, binary_opnode, context, flow_factory):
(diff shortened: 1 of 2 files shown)


cpennington:
@degustaf: Yup, I added a test case that demonstrates the issue.

ceridwen:
"However, the error is only throw after the generator return by _infer_binary_operation is iterated over, which means the current except:/else: structure doesn't work."  I'm having trouble understanding what you mean here.  Could you revise that sentence?

cpennington:
@ceridwen: I've update the commit message and the PR body. Please let me know if you want any additional clarification.

ceridwen:
@cpennington, thanks for rewriting the description, that's much clearer.  @PCManticore, I haven't run the tests yet but that code looks like it ought to fix the bug.  If that's sufficient, I suggest we do that, though we should also add @cpennington's test.

PCManticore:
@ceridwen Yep, if the code works, we should include @cpennington test.

PCManticore:
Merged manually, with the iteration of results inside the try branch. Thank you @cpennington !

PR #411, opened 2017-03-08
Add a brain for the json module.

cpennington:
This implements the fix suggested in https://github.com/PyCQA/pylint/issues/922 for the result of `json.load` being inferred as a `bool`.

### DIFF
--- a/astroid/brain/brain_json.py
+++ b/astroid/brain/brain_json.py
@@ -1,0 +1,21 @@
+import astroid
+
+
+def _json_transform():
+    # pylint always thinks that json.loads returns a bool.
+    # Message: Instance of 'bool' has no 'get' member (no-member)
+    return astroid.parse('''
+    def loads(s, encoding=None, cls=None, object_hook=None, parse_float=None,
+        parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
+        if 1 in kw: return {}
+        if 2 in kw: return []
+        if 3 in kw: return True
+        if 4 in kw: return ""
+        if 5 in kw: return u""
+        if 6 in kw: return None
+        if 7 in kw: return 3.14
+    ''')
+
+
+astroid.register_module_extender(astroid.MANAGER, 'json', _json_transform)
+astroid.register_module_extender(astroid.MANAGER, 'simplejson', _json_transform)
--- a/astroid/tests/unittest_brain.py
+++ b/astroid/tests/unittest_brain.py
@@ -625,6 +625,27 @@ class TypingBrain(unittest.TestCase):
         self.assertIsInstance(inferred, astroid.Instance)
         self.assertEqual(inferred.qname(), "typing.NamedTuple")
 
+
+class JsonBrainTest(unittest.TestCase):
+    def test_load_type(self):
+        result = builder.extract_node('''
+        import json
+        json.load('path')
+        ''')
+        inferred = list(result.infer())
+        for inferred in result.infer():
+            self.assertIsInstance(inferred, (nodes.Const, nodes.Dict, nodes.List))
+
+    def test_loads_type(self):
+        result = builder.extract_node('''
+        import json
+        json.loads('{}')
+        ''')
+        inferred = list(result.infer())
+        for inferred in result.infer():
+            self.assertIsInstance(inferred, (nodes.Const, nodes.Dict, nodes.List))
+
+
 class ReBrainTest(unittest.TestCase):
     def test_regex_flags(self):
         import re


ceridwen:
Someone else reported on IRC that pyilnt thought that json.loads was returning a bool in their code, but when they tried to give us a MWE, I tried it and couldn't replicate.  Under what conditions is this showing up?  Do you have a MWE to demonstrate the bug?

cpennington:
With master of both pylint and astroid:

```
import json

foo = json.loads('bar')
foo.get('baz')
```

degustaf:
I can replicate the error with master branches of astroid and pylint, and python 2.7.6.

ceridwen:
I can finally replicate this error.  I'm not sure what changed, but between the last release and the current master, the inference result for json.loads changed from `[Uninferable]` to `[Uninferable, <Const.NoneType l.42 at 0x7fa032b9b6d8>, <Const.bool l.44 at 0x7fa032b9bc88>, <Const.bool l.46 at 0x7fa032b10748>]`.  These come from this code in the standard library `json/scanner.py`:

```
        elif nextchar == 'n' and string[idx:idx + 4] == 'null':
            return None, idx + 4
        elif nextchar == 't' and string[idx:idx + 4] == 'true':
            return True, idx + 4
        elif nextchar == 'f' and string[idx:idx + 5] == 'false':
            return False, idx + 5
```
The tuples are split in `json/decoder.py` by `JSONDecoder.raw_decode` and `JSONDecoder.decode`.  Meanwhile, it's giving up on all the other returns in that function from `scanner.py`.  Those are functions passed in from `json.loads`, and they all have default values which (I think) astroid isn't finding.  Figuring out where precisely it's failing will take more work.

I'm leery of this fix, since it's essentially patching the underlying problem, which is that astroid isn't finding the defaults correctly, for one specific standard library module.  The problem with this approach is that for any particular inference false positive, the easiest patch is going to be to write a brain module for it.  Taken to its logical conclusion, there won't be any limitation on how many patches get put into brain, which (IMO) is already a maintenance headache.  As far as I'm concerned, we should be trying to shrink brain, not grow it.  If we decide we want to allow this particular patch in, we should write up a policy for what patches to accept/not accept for brain and announce in the docs, to prevent unlimited growth.  Also, if taken, it needs to be applied to `json.load`, which will suffer from the same bug.

For pylint, how is the checker for this bug defined, inclusively or exclusively?  I mean, if astroid's inference returns a list of possible types to pylint, does pylint complain if *any* of those types don't have a particular attribute, or if *all* of those types don't have a particular attribute?  Since this problem resulted from an inference improvement, I'm concerned that improving the inference will start creating more false positives like this, so maybe we should be trying to complain less when astroid really has no idea what the return value is because so many of the code paths are opaque.

The actual message pylint is returning is also misleading, because as far as astroid knows, it could be either None or a bool.  I definitely think mentioning all the possible types that could cause problems would improve the pylint message.

ceridwen:
I traced this problem all the way down to `py_make_scanner()` in `json/scanner.py`.  The `_default_encoder` in `__init__.py` is a `decoder.JSONDecoder` object.  Inference on the `scanner.make_scanner(self)` line returns Uninferable, as does inference on all the lines in `py_make_scanner` that involve assigning from `context`.  Here, `context`is the `_default_encoder` object.  Thus, the inference is failing to find the functions assigned to this object.

rogalski:
While I agree with an idea of having brain shrinking instead of growing, JSON is too popular of the module to have such spurious errors. I'd vote to have it merged for 1.5.0 release of astroid and figure out proper way to fix it later.

ceridwen:
Do we know how many other standard library modules have easy-to-find false positives?  I'm asking because if we make this the policy, it would be nice to know how many other modules will need to be included in brain.

PCManticore:
I agree with @ceridwen that we need to find a compromise between an ever increasing brain and not having false positives in pylint. Here's what I'm using as a rule for adding brain plugins:

- does it fix something that is implemented in a dynamic way, that cannot be understood by pylint? e.g enum and namedtuple, where they use ``exec``. Then a brain plugin is okay.
- does the module in question changes a lot over time, having different signatures and whatnot for each release? This is the case for ``subprocess.Popen`` for example. Then a brain plugin is okay as well.
- does the brain plugin **fixes** a deficiency in astroid's inference? This one is the trickiest and I think is the case of this json plugin as well. The answer for this subpoint can be different if:
   * the feature that the inference needs is hard to implement, but is implementable. E.g. control flow. If it takes too much time to implement that feature, e.g. 6 months - 1 year, then having a brain plugin to fix it in the meantime is acceptable.
   * the feature that the inference needs is probably not implementable, e.g fully interpreting the code for some reason. In that case, a brain plugin is probably the only acceptable solution.

For this one, I think it's simple enough that it warrants being merged. But at the same time, I do agree with @ceridwen that if astroid is returning opaque inference results to pylint, then pylint **shouldn't** assume it can go ahead and emit a warning, since it *might* be wrong. To that effect, I added a new fix in pylint which makes obsolete this current patch: https://github.com/PyCQA/pylint/commit/bf3e06fcb86fef4d0273ba40f8254377a050aa00
With this new change, whenever pylint gets an ``Uninferable`` from astroid, it won't emit ``no-member`` any longer, unless explicitly configured to do the opposite. This will hide a plethora of warnings where astroid couldn't infer some branch of the code. I think it is a good compromise for now, between having an useful linter and a fully fledged interpreter/inference engine written in Python

PR #412, opened 2017-03-08
Fix an infinite loop that could occur during inference of __nonzero__.

cpennington:
For example:

```
class AccessResponse(object):
    def __init__(self, has_access):
        self.has_access = has_access

    def __nonzero__(self):
        return self.has_access

not AccessResponse(False)
```

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -258,7 +258,8 @@ class Instance(BaseInstance):
              all its instances are considered true.
         """
         context = contextmod.InferenceContext()
-        context.callcontext = contextmod.CallContext(args=[self])
+        context.callcontext = contextmod.CallContext(args=[])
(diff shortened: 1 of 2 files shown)


ceridwen:
Could you add a test case for this issue that passes with your changes?  Same for the other pull requests.  It should probably go in unittest_regrtest.py.

cpennington:
@ceridwen: Yeah, I just added a test for the other PRs (and for this one, actually). I'm happy to move it to `regrtest.py` if that would be better than `unittest_inference.py` (which is where I just put it).

ceridwen:
It's probably fine in `unittest_inference`.

PCManticore:
Thanks for the pull request @cpennington !

PR #413, opened 2017-03-10
Improving caching

cpennington:
Before:
```
Command exited with non-zero status 30
305.85user 11.24system 5:17.37elapsed 99%CPU (0avgtext+0avgdata 584308maxresident)k
0inputs+888outputs (0major+168864minor)pagefaults 0swaps
```

After:
```
Command exited with non-zero status 30
296.86user 11.82system 5:09.93elapsed 99%CPU (0avgtext+0avgdata 584008maxresident)k
208inputs+480outputs (0major+169539minor)pagefaults 0swaps
```

### DIFF
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -35,8 +35,12 @@ def has_environment_marker_range_operators_support():
 
 if has_environment_marker_range_operators_support():
     extras_require[':python_version<"3.4"'] = ['enum34', 'singledispatch']
-elif py_version < (3, 4):
-    install_requires.extend(['enum34', 'singledispatch'])
+    extras_require[':python_version<"3.2"'] = ['functools32']
+else:
+    if py_version < (3, 4):
+        install_requires.extend(['enum34', 'singledispatch'])
+    if py_version < (3, 2):
+        install_requires.append('functools32')
 
 
 # pylint: disable=redefined-builtin; why license is a builtin anyway?
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -20,6 +20,11 @@ attribute. Thus the model can be viewed as a special part of the lookup
 mechanism.
 """
 
+try:
+    from functools import lru_cache
+except ImportError:
+    from functools32 import lru_cache
+
 import pprint
 import os
 import types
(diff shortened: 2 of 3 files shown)


cpennington:
Oops. My latest set of changes seems to have changed the semantics.

ceridwen:
Caching finders looks like it produces a major speed up, but I'm not immediately sure this is the right way to do it.  I'm assuming the slow steps are the calls to the standard library import machinery?  Is the search_path key unique across modules?  (I assume yes, but I don't completely understand this code.)  Rather than the cache as a nested dictionary accessed by a global variable, can we move it into separate caches on the finders?  @PCManticore, any comments?

Using an lru cache on ObjectModel.attributes looks good to me.

Based on @cpennington's performance results, caching MROs doesn't get us much.  If we want to do it, adding another mutable attribute to ClassDef nodes is the wrong way, because the MRO is a property not only of the ClassDef node itself but of the tree it's embedded in.  Any memoization would have to take place at the tree level and require cache invalidation in the case of tree mutation.

cpennington:
@ceridwen: IIRC (and from looking at the set of finders), I think the biggest culprit in the Finders was the `ZipFinder`. We could maybe just do an `lru_cache` on `_precache_zipimporters`. Although... that depends on `sys.path` and the `sys.path_importer_cache`.

I'm going to do a bit of digging to see if there's a more targeted approach.

astroid/__pkginfo__.py
```diff
@@ -35,8 +35,12 @@ def has_environment_marker_range_operators_support():
 
 if has_environment_marker_range_operators_support():
     extras_require[':python_version<"3.4"'] = ['enum34', 'singledispatch']
-elif py_version < (3, 4):
-    install_requires.extend(['enum34', 'singledispatch'])
+    extras_require[':python_version<"3.2"'] = ['functools32']
```
PCManticore: Sorry for being late to the party, I thought this was still work-in-progress.

Why don't we use https://pypi.python.org/pypi/backports.functools_lru_cache/ instead? this is what we use in pylint and it seems official, rather than functools32.

ceridwen: Good point, thanks.  We had some conversation about it in IRC this afternoon, and I've been trying to prevent any more pull requests than we already have from languishing.

PR #417, opened 2017-03-15
Cache the computation of which modules haven't been added to the path_importer_cache yet

cpennington:
On https://github.com/edx/edx-platform/tree/master/common/lib/xmodule/xmodule, this resulted in the following performance improvement (with no change in the errors detected):

Before:

```
Command exited with non-zero status 30
308.18user 11.97system 5:22.79elapsed 99%CPU (0avgtext+0avgdata 583616maxresident)k
0inputs+440outputs (0major+169620minor)pagefaults 0swaps
```

After:

```
Command exited with non-zero status 30
233.37user 12.33system 4:09.24elapsed 98%CPU (0avgtext+0avgdata 611180maxresident)k
0inputs+416outputs (0major+175975minor)pagefaults 0swaps
```

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -13,6 +13,11 @@ try:
 except ImportError:
     _HAS_MACHINERY = False
 
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
+
 from . import util
 
 ModuleType = enum.Enum('ModuleType', 'C_BUILTIN C_EXTENSION PKG_DIRECTORY '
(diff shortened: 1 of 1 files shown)


ceridwen:
Assuming @cpennington answers my question, I'm going to merge this tomorrow if there are no other comments.

cpennington:
I'll try it out and see. I'm not sure if the expensive part in the existing code is the conversion to sets, or the set difference.

cpennington:
@ceridwen: Ok, I ran the numbers, and converting to tuples and then caching is faster than converting to sets and then caching.

ceridwen:
Can you add a comment to that effect?

cpennington:
@ceridwen: done.

PR #418, opened 2017-03-20
Catch exceptions raised in running pylint on enum.py.

degustaf:
### Fixes / new features
- Catch exceptions raised in running pylint on enum.py from the standard library.

### DIFF
--- a/astroid/tests/unittest_inference.py
+++ b/astroid/tests/unittest_inference.py
@@ -2303,6 +2303,27 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
             error = errors[0]
             self.assertEqual(str(error), expected_value)
 
+    def test_unary_type_errors_crash(self):
+        ast_nodes = extract_node('''
+        class EnumMeta(type):
+            def __new__(mcs, cls, bases, classdict):
+                __new__ = object.__new__
+                enum_class = super().__new__(mcs, cls, bases, classdict)
+        
+                enum_member = __new__(enum_class)
+                not hasattr(enum_member, '_value_') #@
+                for _ in classdict._member_names:
+                    enum_member = __new__(enum_class)
+                    if not hasattr(enum_member, '_value_'): #@
+                        enum_member._value_ = 0
+        ''')
+        ast_nodes[1] = next(ast_nodes[1].get_children())
+        for node in ast_nodes:
+            try:
+                node.type_errors()
+            except TypeError:
+                self.fail("node.type_errors raises unexpected TypeError.")
+
     def test_unary_empty_type_errors(self):
         # These aren't supported right now
         ast_nodes = extract_node('''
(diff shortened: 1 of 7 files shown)


degustaf:
@ceridwen, I think a helper function (probably in utils) that actually raises the error will minimize the duplicated code.  I'll see what I can put together for you to take a look at.

degustaf:
@ceridwen Something appears to be wrong with coveralls.  It looks like it is reporting the Python 2.7 results as the total coverage.

degustaf:
@ceridwen All the tests, including coverage, are green.  Can we get this merged in now?

PCManticore:
I need to make some time to review this first

ceridwen:
Sorry for the delay.  I've been trying to confirm or refute whether [stack introspection](http://pypy.org/performance.html#micro-tuning-tips) has anything to do with why PyPy runs pylint more slowly than CPython.  Currently, the only stack introspection we're doing is in `util.py`, calling `sys.exc_info()`when we reraise exceptions.  I haven't been able to prove that it's the problem, but I haven't been able to rule it out, either.  However, reraising exceptions is essential to getting useful tracebacks for debugging, so I've been looking for a better way to handle this that doesn't involve stack introspection.  So far, I haven't found anything that works with Python 2, which means this may be a permanent problem.

This makes me reluctant to add more stack introspection.    Regardless, we currently have clutter in our stack traces from `util.reraise()`, and I'd rather have clutter than introduce more stack introspection.

My only other comment is that because astroid is almost completely undocumented (this is a problem we inherited, unfortunately), it won't be apparent that `**kws` is only being used here to collect spurious arguments if something of the wrong type is being called.  Putting comments or docstrings in each function would be most transparent, but would require a lot of duplication.  Anyone else have a better idea?

ceridwen:
To be clear: I'm willing to merge after `inspect.stack` is removed, but I still want to wait for @PCManticore's review.

degustaf:
@ceridwen I removed the added use of inspect.stack.

astroid/tests/unittest_inference.py
```diff
@@ -2303,6 +2303,27 @@ class A(object): pass
             error = errors[0]
             self.assertEqual(str(error), expected_value)
 
+    def test_unary_type_errors_crash(self):
+        ast_nodes = extract_node('''
```
PCManticore: @degustaf what exactly is the problem here? Running pylint over this snipper raises indeed a TypeError, due to ``class_context`` being passed, but why was it passed to ``getattr`` in the first place? This pull request seems to address the effect rather than the cause. Let's investigate first in what circumstances ``class_context`` gets passed and why astroid does this.

PCManticore:
I'm closing this. Let's investigate first why _proxied isn't a ClassDef and how did that AST node got there. A more appropriate fix might be needed after figuring out these details.

PR #420, opened 2017-03-28
Fix assigned_stmts() for AnnAssign nodes without value

rogalski:


rogalski:
@PCManticore @ceridwen @degustaf it would be nice to get any form of input before it'll be merged.

PCManticore:
This looks simple enough.

PR #423, opened 2017-04-21
Allow pylint/astroid to correctly interpret a class whose baseclass has been assigned to

cpennington:
If a class is derived from a base class that had been assigned to elsewhere in the code, `_is_metaclass` will see an `AssignAttr` or an `AssignName` as an inferred `baseobj`. These don't have `qname()`, so we use the first assigned value as the `baseobj` instead.

### DIFF
--- a/astroid/scoped_nodes.py
+++ b/astroid/scoped_nodes.py
@@ -999,6 +999,9 @@ def _is_metaclass(klass, seen=None):
     for base in klass.bases:
         try:
             for baseobj in base.infer():
+                if isinstance(baseobj, mixins.AssignTypeMixin):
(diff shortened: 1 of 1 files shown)


rogalski:
Hey,
Can we have a test case that shows an issue and confirms that fix actually works?

cpennington:
This one was one that I had trouble reproducing. Not sure when I'll get a chance to figure out how to write a test case for it.

PCManticore:
If I remember correctly, there was a conversation about this change with @ceridwen on IRC. The result of the conversation was that the original AST was malformed, since it's not possible to have an assign statement as a base class of the node (as this patch might suggest). I'm closing this one.

PR #429, opened 2017-05-10
Fix an inference for trying to unpack Uninferable value in context manager

rogalski:
Closes PyCQA/pylint#1463

### TODO
- [x] write a test case
- [x] implement a fix
- [x] simplify a test case to absolute minimum


PCManticore:
@rogalski a more simple test case would be wonderful to have. I'm not sure what triggers the problem just by looking over this test case.

PR #436, opened 2017-05-18
Add tests for ImportFrom aliasing

rogalski:


PCManticore:
Hey Łukasz, I'm closing this right now since merging it without the fix will cause the CI to fail. But feel free to send it together with the change!

PR #440, opened 2017-06-03
Fix finder for namespace packages

ioanatia:
Fixes https://github.com/PyCQA/pylint/issues/1489

#### Issue description
Run pylint (with a version `>= 1.7.0`) on a file `test.py` with the following content:

```
import google.auth
```

Provided that `google-auth` is installed, it outputs the following warning:
```
W:  1, 0: Relative import 'google.auth', should be 'test.google.auth' (relative-import)
```

#### Fix description

In `ExplicitNamespacePackageFinder.find_module` the value of the `submodule_path` parameter is not used. Since `test` is a valid module and `google` a valid namespace, it concludes that `test.google` is valid, returns a `ModuleSpec`, instead of returning `None`.

Checking for `submodule_path` seems to fix the pylint issue.

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -121,6 +121,8 @@ class ExplicitNamespacePackageFinder(ImpFinder):
     """A finder for the explicit namespace packages, generated through pkg_resources."""
 
     def find_module(self, modname, module_parts, processed, submodule_path):
+        if processed:
+           return None
         if util.is_namespace(modname) and modname in sys.modules:
             submodule_path = sys.modules[modname].__path__
             return ModuleSpec(name=modname, location='',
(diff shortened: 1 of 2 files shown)


astroid/interpreter/_import/spec.py
```diff
@@ -121,6 +121,8 @@ class ExplicitNamespacePackageFinder(ImpFinder):
     """A finder for the explicit namespace packages, generated through pkg_resources."""
 
     def find_module(self, modname, module_parts, processed, submodule_path):
+        if submodule_path:
```
PCManticore: I'm not sure this is the problem or at least, I can't demonstrate it so far. From what I'm noticing, ``submodule_path`` contains the filename provided by the ``ImpFinder``, which doesn't make sense, since a file is not a valid search location. Secondly, it seems that in this test case, the submodule_path provided is the current directory. Instead of just returning when ``submodule_path`` is not empty, we could check if the submodule_path passed in this function points to different root directories than the submodule path found by looking in the module itself. Do you think this makes sense? can you check it?

PR #448, opened 2017-08-13
Correctly infer typing.X() and typing.X[] constructs as classdefs

rogalski:


rogalski:
Failure is inherited from master.

astroid/brain/brain_typing.py
```diff
@@ -78,6 +78,57 @@ def looks_like_typing_namedtuple(node):
...
+
+    if func.qname() not in {'typing.TypeVar',  'typing.NewType'}:
+        raise UseInferenceDefault
+
+    if not node.args:
+        raise UseInferenceDefault
+
+    typename = node.args[0].as_string().strip("'")
+    node = extract_node(TYPING_TYPE_TEMPLATE.format(typename))
+    return node.infer(context=context)
```
PCManticore: A TypeVar seems to act like an instance, not a class. Why do we infer it as a class though?

astroid/brain/brain_typing.py
```diff
@@ -78,6 +78,57 @@ def looks_like_typing_namedtuple(node):
...
+    func = node.func
+    if isinstance(func, nodes.Attribute):
+        return func.attrname in {'TypeVar', 'NewType'}
+    if isinstance(func, nodes.Name):
+        return func.name in {'TypeVar', 'NewType'}
+    return False
+
+
+TYPING_TYPE_TEMPLATE = """
+class Meta:
```
PCManticore: This is not a valid metaclass, you'll have to inherit from ``type``.

astroid/brain/brain_typing.py
```diff
@@ -78,6 +78,57 @@ def looks_like_typing_namedtuple(node):
     return False
 
 
+def looks_like_typing_typevar_or_newtype(node):
+    func = node.func
+    if isinstance(func, nodes.Attribute):
+        return func.attrname in {'TypeVar', 'NewType'}
```
PCManticore: Let's just save these 2 in a variable and check it against it.

astroid/brain/brain_typing.py
```diff
@@ -78,6 +78,57 @@ def looks_like_typing_namedtuple(node):
...
+    if isinstance(func, nodes.Name):
+        return func.name in {'TypeVar', 'NewType'}
+    return False
+
+
+TYPING_TYPE_TEMPLATE = """
+class Meta:
+    def __getitem__(self, item):
```
PCManticore: A TypeVar doesn't seem to support indexing. Why do we implement ``__getitem__`` for it?

astroid/tests/unittest_brain.py
```diff
@@ -812,5 +812,21 @@ def test_uuid_has_int_member(self):
...
 
+@test_utils.require_version('3.6')
+class BrainTypingTest(unittest.TestCase):
+    def test_typing_types(self):
+        ast_nodes = builder.extract_node("""
+        from typing import TypeVar, Iterable, Tuple, NewType, Dict, Union
+        TypeVar('MyTypeVar', int, float, complex) #@
+        Iterable[Tuple[MyTypeVar, MyTypeVar]] #@
+        TypeVar('AnyStr', str, bytes) #@
+        NewType('UserId', str) #@
```
PCManticore: NewType seems to be a function rather than a ClassDef, I'm confused to why we have to infer it as a class. And nonetheless, I'm not sure implementing it in this way brings up a lot of benefits, we also have to make it act as a subclass of the original types that were passed in the constructor, which we don't do currently.

PCManticore:
Merged manually. Thanks @rogalski !

PR #449, opened 2017-08-19
Trying to solve pylint bug #1477 concerning false positive with numpy

hippo91:
pylint bug #1477 deals with numerous false positive concerning ``numpy``. 
The functions concerned are those of ``numpy.core.ufunc`` module except those that have been imported through the file under numpy/lib/function_base.py. 
The module ``numpy.core.ufunc`` has no ``__all__`` members. 
The proposed solution is really not elegant but for now i don't see any other solution.


rogalski:
FYI, failures from CI are unrelated.

hippo91:
OK thanks @rogalski!

PCManticore:
Thanks @hippo91 !

PR #452, opened 2017-10-08
Proposal for fighting against pylint false positive ``no-member`` messages when code uses ``numpy``

hippo91:
This PR tries to solve ``pylint`` false positive ``no-member`` messages as encountered in PyCQA/pylint#779 and PyCQA/pylint#1477. 
A high number of false positive ``no-member`` messages, emitted when ``numpy`` is used, are due to the fact that three ``numpy`` modules are written in c and thus can't be correctly parsed by ``astroid`` (i.e they end up being built as a *stub* module):

  - ``numpy.core.umath``;
  - ``numpy.core.multiarray``;
  - ``numpy.random.mtrand``.

Instead of listing all missing members in ``brain_numpy.py``, as it was done in PR #449 (for ``numpy.core.umath``), it seems more efficient to allow those three modules to be dynamically loaded by ``astroid``. The main lock avoiding this was the ``_can_load_extension`` method of ``AstroidManager``. I added a test to know if the module to be loaded is one of the three listed above. If it is the case then ``_can_load_extension`` returns ``True``.

I also cleaned up ``brain_numpy.py`` but i added in it, the members of ``numpy.core.numerictypes`` that 
are not correctly handled by ``astroid`` (although it is not very obvious why). 

I wanted to add a unit test to ensure all ``numpy`` members are correctly detected but i don't really now how to do this in a clean way. @PCManticore, if this PR seems to be interesting,  can you explain me a little bit how to do add unit tests please?

### DIFF
--- a/astroid/brain/brain_numpy.py
+++ b/astroid/brain/brain_numpy.py
@@ -13,54 +13,19 @@ import astroid
 
 def numpy_core_transform():
     return astroid.parse('''
-    from numpy.core import numeric
-    from numpy.core import fromnumeric
-    from numpy.core import defchararray
-    from numpy.core import records
-    from numpy.core import function_base
-    from numpy.core import machar
-    from numpy.core import getlimits
-    from numpy.core import shape_base
-    from numpy.core.umath import (
-        absolute, add, arccos, arccosh, arcsin, arcsinh,
-        arctan, arctan2, arctanh, bitwise_and, bitwise_or,
-        bitwise_xor, cbrt, ceil, conj, conjugate, copysign,
-        cos, cosh, deg2rad, degrees, divide, e, equal,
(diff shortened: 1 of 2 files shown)


PCManticore:
I'm sorry but I cannot agree with this. C extensions cannot be loaded for a reason, which is that it creates a security risk, which is why they have to be exempted with ``--extension-pkg-whitelist``. The problem with `numpy` though stems from the fact that it's relying heavily on dynamic features, such as generating members dynamically, which cannot be inferred correctly by pylint. What we need to do is to introduce proper hints through brain modules for these inner modules where numpy uses those dynamic features.

hippo91:
Ok, thanks @PCManticore for these explanation!
I will try to make something in brain_numpy.py to handle those dynamic features.

Issue #454, opened 2017-10-12
Tests: AssertionError: <ClassDef.instance l.2 at 0x7f9fdf2f8290> is not an instance of <class 'astroid.bases.Instance'>


hugovk:
### Steps to reproduce
1. Test latest master (eg. on Travis CI or Appveyor)

### Current behavior
```
FAIL: test_object_dunder_new_is_inferred_if_decorator (unittest_inference.ObjectDunderNewTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/PyCQA/astroid/.tox/py27/lib/python2.7/site-packages/astroid/tests/unittest_inference.py", line 4269, in test_object_dunder_new_is_inferred_if_decorator
    self.assertIsInstance(inferred, Instance)
AssertionError: <ClassDef.instance l.2 at 0x7f9fdf2f8290> is not an instance of <class 'astroid.bases.Instance'>
```

### Expected behavior
Tests pass.

### More info

I think https://travis-ci.org/PyCQA/astroid/builds/257015909 is the first failing build, from commit https://github.com/PyCQA/astroid/commit/e348d6396a6849cc2b73397dce92c3d1a7921802.

Tip: rather than committing directly to master, always create a PR, that way you can see if the new changes pass or fail the CI before merging.

hugovk:
On further investigation, commit c2e8ca524793905018e0e4adc5e29e1d3b0446c6 to fix #172, which introduced that test, has been failing from the start.

Here's the failing commit to fix #172:
https://travis-ci.org/hugovk/astroid/builds/286902750

And here's the passing commit (except for pylint) before that one:
https://travis-ci.org/hugovk/astroid/builds/286903238

In the short term, I suggest at least removing the test so the CI is useful for new PRs. And then revisit the test to make sure the fix for #172 is working as expected.

PCManticore:
This is caused by https://github.com/PyCQA/astroid/issues/444. The PR looks good to me, plus we also track the error in the linked issue.

PR #455, opened 2017-10-12
Mark test which has never passed as an expected failure

### LINKED ISSUE #172: Understand object.__new__ decorator

---

The object.**new** decorator creates an instance under the hood.

```
#!python

@object.__new__
class A:
   pass

A #@

```

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/172

hugovk:
Resolves https://github.com/PyCQA/astroid/issues/454.

This test was added in commit c2e8ca524793905018e0e4adc5e29e1d3b0446c6 to fix #172, but it's never been passing.

Rather than having all CI builds failing, I recommend removing this test so the CI can be of more use for future PRs. And then revisit the test to make sure the fix for #172 is working as expected.

Here's the failing commit to fix #172:
https://travis-ci.org/hugovk/astroid/builds/286902750

And here's the passing commit (except for pylint) before that one:
https://travis-ci.org/hugovk/astroid/builds/286903238

### DIFF
--- a/astroid/tests/unittest_inference.py
+++ b/astroid/tests/unittest_inference.py
@@ -4273,6 +4273,7 @@ class CallSiteTest(unittest.TestCase):
         self.assertIn('f', site.duplicated_keywords)
 
 
+@unittest.expectedFailure
 class ObjectDunderNewTest(unittest.TestCase):
 
     def test_object_dunder_new_is_inferred_if_decorator(self):


astroid/tests/unittest_inference.py
```diff
@@ -4273,17 +4273,5 @@ def test_duplicated_keyword_arguments(self):
         self.assertIn('f', site.duplicated_keywords)
 
 
-class ObjectDunderNewTest(unittest.TestCase):
```
PCManticore: Rather than removing it entirely, this test should be marked as expected failure.

hugovk: @PCManticore Thanks, updated.

PCManticore:
Thanks @hugovk ! The build still fails, but due to some pylint warnings

PR #456, opened 2017-10-13
Add brain_curses.py for curses attributes defined at runtime

mathieui:
### Fixes / new features
- Add a ncurses brain 

Explanation for the need of curses brain: https://github.com/PyCQA/pylint/issues/1703
Most of the curses module attributes are defined at runtime (for portability reasons, apparently), so  there is a need for an astroid brain in order to infer correctly.

The names currently in the brain were extracted from the python documentation.
If required, I can pull the real values from the C header.


PCManticore:
Thanks for the PR!

PR #458, opened 2017-10-16
Public facing node documentation

AWhetter:
This ended up being bigger that I thought...
If this needs splitting up into smaller chunks then let me know!

ast_objects.inv contains the [Green Tree Snakes](http://greentreesnakes.readthedocs.io/en/latest/) documentation inventory, but with the classes prepended with `ast` so that they don't conflict with astroid nodes and therefore we can still link to `ast` documentation.


astroid/node_classes.py
```diff
@@ -1025,70 +1530,235 @@ def _format_args(args, defaults=None, annotations=None):
...
+        :type lineno: int or None
+
+        :param col_offset: The column that this node appears on in the
+            source code.
+        :type col_offset: int or None
+
+        :param parent: The parent node in the syntax tree.
+        :type parent: NodeNG or None
+        """
+        self.op = op
```
PCManticore: Initially I thought this was removed! But it was just lost through these docstrings. I think it would help if we could add an empty line between these blocks (attribute + docstring + newline). What do you think?

AWhetter: I could put the docstring for the attribute instead. That's maybe a bit more conventional, and it won't spread out the code too much?

PR #460, opened 2017-10-26
Fix nested namespace package import causing `import-error`

macbeth322:
### Fixes / new features
- Importing a module which is in more than one namespace package, eg `foo.bar.baz` where `foo` and `bar` are both namespace packages, causes an error to be raised and no module to be found
- This merge request only addresses half of the issue; https://github.com/PyCQA/astroid/pull/440/commits/6099e8526f38dd7f1437ec7225c673a7be444f14
- This specific issue is also documented here https://github.com/PyCQA/pylint/issues/1686
- Also added a test that breaks without this change, and tested in 2.7.14 and 3.6.2


brycepg:
I'm seeing astroid importing namespace packages properly on master:

```
mkdir -p foo/bar/baz
touch foo/bar/baz/__init__.py
```
```
import astroid
code = """
import foo.bar.baz
foo.bar.baz #@
"""
astroid.extract_node(code).inferred()
```
Returns a module.

Additionally, linting this module doesn't return any error. Can you provide a code or repo example which reproduces the error which this pull request fixes?

PCManticore:
@brycepg This PR is for the "old" style namespace packages, not for PEP 420 (that one?) namespace packages. That is, it fixes the problem for the old pkg_resources hack where in `__init__` one has to add `__import__('pkg_resources').declare_namespace(__name__)` to declare a namespace package.

I just tested this PR locally and it works great! Thanks @macbeth322 and super sorry for merging this so late!

PR #461, opened 2017-10-29
Numpy brain refinement

hippo91:
Taking into account @PCManticore remarks concerning the preceding PR #452, all improvements concerning numpy are taken into account through the numpy brain module.
It consists essentially in introducing the signature of the numpy functions that are in `numpy.core.umath` and `numpy.random.mtrand` modules.
This way, not only pylint doesn't emit 'no-member' message anymore but it is also able to check function arguments.


astroid/tests/unittest_brain_numpy.py
```diff
@@ -0,0 +1,246 @@
+#-*- encoding=utf-8 -*-
+# Copyright (c) 2013-2014 Google, Inc.
```
PCManticore: If you wrote this yourself and it doesn't contain code from other files, feel free to just add yourself here!

hippo91:
Hey @PCManticore, the CI passes but just because i added a decorator to skip the untittest concerning `numpy` if the `numpy` module is not present. Does it mean that those unittests are never executed in the CI (because tox env doesn't install numpy)?

PCManticore:
@hippo91 apparently that's true! https://github.com/PyCQA/astroid/blob/7bfb5e83b162d2126c9929b6d863724176269b6d/tox.ini#L14 we weren't installing numpy for pypy and py3.3

PCManticore:
Merged since the missing numpy doesn't block this PR. We'll handle this in its own change.

PR #462, opened 2017-10-30
Add Transform Plugin for attr to fix false positives due to instance attribute rewriting

brycepg:
### Fixes / new features
This commit fixes https://github.com/PyCQA/pylint/issues/1698 unsupported-assignment-operation issues
I implemented a transform plugin as @PCManticore suggested

### DIFF
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -1,0 +1,40 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
+"""
+Astroid hook for the attrs library
+
+Without this hook pylint reports unsupported-assignment-operation
+for atrrs classes
+"""
+
+import astroid
+from astroid import MANAGER
+
+
+def is_decorated_with_attrs(
+        node, decorator_names=('attr.s', 'attr.attrs', 'attr.attributes')):
+    """Return True if a decorated node has
+    an attr decorator applied."""
+    if not node.decorators:
+        return False
+    for decorator_attribute in node.decorators.nodes:
+        if decorator_attribute.as_string() in decorator_names:
+            return True
+    return False
+
+
+def attr_attributes_transform(node):
+    """Given that the ClassNode has an attr decorator,
+    rewrite class attributes as instance attributes
+    """
+    for cdefbodynode in node.body:
+        if not isinstance(cdefbodynode, astroid.Assign):
+            continue
+        for target in cdefbodynode.targets:
+            node.locals[target.name] = [astroid.Attribute(target.name)]
+
+
+MANAGER.register_transform(
+    astroid.Class,
+    attr_attributes_transform,
+    is_decorated_with_attrs)
--- a/astroid/tests/unittest_brain.py
+++ b/astroid/tests/unittest_brain.py
@@ -44,6 +44,13 @@ try:
     HAS_PYTEST = True
 except ImportError:
     HAS_PYTEST = False
+
+try:
+    import attr as attr_module # pylint: disable=unused-import
+    HAS_ATTR = True
+except ImportError:
+    HAS_ATTR = False
+
 import six
 
 from astroid import MANAGER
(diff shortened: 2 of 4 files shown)


astroid/brain/brain_attrs.py
```diff
@@ -0,0 +1,38 @@
+"""Astroid hook for the attrs library
```
PCManticore: Let's add the Licensing header like the one from the other brain tips

astroid/brain/brain_attrs.py
```diff
@@ -0,0 +1,38 @@
...
+
+Without this hook pylint reports unsupported-assignment-operation
+for atrrs classes
+"""
+import astroid
+from astroid import MANAGER
+
+
+def register(linter):
```
PCManticore: No need to have this, since it is not a pylint plugin. We already do the registration through the manager

astroid/brain/brain_attrs.py
```diff
@@ -0,0 +1,38 @@
...
+    for decorator_node in node.decorators.nodes:
+        if not isinstance(decorator_node, astroid.Attribute):
+            continue
+        if decorator_node.as_string() in ("attr.s",
+                                          "attr.attributes",
+                                          "attr.attrs"):
+            _attr_assign_attributes(node)
+
+
+MANAGER.register_transform(astroid.Class, attr_attributes_transform)
```
PCManticore: We should pass the filter from the transform here instead. Something along the lines of:

```
MANAGER.register_transform(astroid.Class, attr_attributes_transform, is_decorated_with_attrs)
```
where `is_decorated_with_attrs` returns True or False

astroid/brain/brain_attrs.py
```diff
@@ -0,0 +1,38 @@
...
+def attr_attributes_transform(node):
+    """If the given ClassNode has an attr decorator,
+    then rewrite class attributes as instance attributes
+    """
+    if not node.decorators:
+        return None
+    for decorator_node in node.decorators.nodes:
+        if not isinstance(decorator_node, astroid.Attribute):
+            continue
+        if decorator_node.as_string() in ("attr.s",
```
PCManticore: Please use `.qname()` rather than `as_string()`

brycepg: qname isn't available for Attribute nodes. I could do decorator_node.inferred()[0].qname() however that returns `attr._make.attributes`; this fully qualified name refers to a module that could be changed since `_make` isn't part of the API, so I thought that using the names would be better.

What do you think?

brycepg:
Thanks for the feedback @PCManticore 

I've made the requested changes,

astroid/brain/brain_attrs.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def is_decorated_with_attrs(node):
+    """Return True if a decorated node has
+    an attr decorator applied."""
+    if not node.decorators:
+        return False
+    for decorator_attribute in node.decorators.nodes:
+        for decorator_def in decorator_attribute.infer():
+            if decorator_def.qname() == "attr._make.attributes":
```
PCManticore: Ah, I see, @brycepg, maybe we should revert back to using `as_string()`. The problem is that we should avoid doing `infer()` in transform predicates, since this could lead to circular inferences across the board, rendering everything unusable (e.g an `infer()` in a predicate could trigger a circular inference for the same object, leading to a recursion error). Sorry for misleading you down this path!

astroid/tests/unittest_brain.py
```diff
@@ -50,6 +50,13 @@
     HAS_PYTEST = True
 except ImportError:
     HAS_PYTEST = False
+
+try:
+    import attr as attr_module # pylint: disable=unused-import
```
PCManticore: We should expose this in tox so that the tests could run in CI: https://github.com/PyCQA/astroid/blob/be694f811f7df89c4fcc672ef09c0e1fc499cf23/tox.ini#L14

astroid/brain/brain_attrs.py
PCManticore: Sounds good, we can use `as_string`

brycepg:
Done

The-Compiler:
Oh, awesome work - thanks!

PR #463, opened 2017-11-01
treat coroutines as generators

rogalski:
I know it’s technically not true, but it’s good enough approximation
for initial support of asyncio.

Initial effort to fix PyCQA/pylint#1715

### DIFF
--- a/astroid/scoped_nodes.py
+++ b/astroid/scoped_nodes.py
@@ -932,14 +932,17 @@ class FunctionDef(node_classes.Statement, Lambda):
             return True
 
     def is_generator(self):
-        """return true if this is a generator function (or coroutine)"""
-        yield_nodes = (node_classes.Yield, node_classes.YieldFrom, node_classes.Await)
+        """return true if this is a generator function (or old-style coroutine)"""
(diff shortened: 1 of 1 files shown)


PCManticore:
Closing as I think that this would be better to handle at `AsyncFunctionDef` level. Thanks for the patch @rogalski !

nicktimko:
What do you mean by "better to handle at `AsyncFunctionDef` level"; the patch added an `is_coroutine` method, what's the better way?

I could try to add some tests to this (or just in general, then let someone else fix the code to match) if that would help move it along.

PCManticore:
@nicktimko The problem is that `is_coroutine` was an attribute of FunctionDef, which as a data model is not a reflection of the data representation at Python level. I was suggesting that this might be addressed at the `AsyncFunctionDef` level, with a custom `infer_call_method` for `AsyncFunctionDef` which would yield a Generator (or better something that indicates it's a coroutine), falling back to the parent's implementation in the contrary case. Does that make sense?

nicktimko:
Makes sense; I'll take a look and see if I can put together a PR unless you're planning on it soon (in which case don't let me stop you, I'd probably do it crudely anyway)

PR #474, opened 2017-12-28
Assign __module__ and __qualname__ to class locals implicitly

brycepg:
This prevents the locals `__qualname__` and `__module__` in the class definition from raising a false-positive no-member by implicitly creating them.

There is a problem that pylint thinks that locals reference variables upward in the mro without a namespace (class name), which is not the case, but I this fix is more of an improvement than before.

fixes PyCQA/pylint#1753

### DIFF
--- a/astroid/scoped_nodes.py
+++ b/astroid/scoped_nodes.py
@@ -1797,6 +1797,21 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG,
         if parent is not None:
             parent.frame().set_local(name, self)
 
+        for local_name, node in self.implicit_locals():
+            self.add_local_node(node, local_name)
+
+    def implicit_locals(self):
+        """Get implicitly defined class definition locals.
+
+        :returns: the the name and Const pair for each local
+        :rtype: tuple(tuple(str, node_classes.Const), ...)
+        """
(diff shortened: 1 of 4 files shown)


brycepg:
How do I make __qualname__ and __module__ refer to only 1 object in mro?
possible solution:
Use poor man's cache in ClassModel to refer to only one node, then modify used-before-assignment to ignore implicitly assigned locals
problem: Using del locally (`del __module__`) does not delete the value from the class namespace.

brycepg:
The only failure right now is pylint failing #475

astroid/scoped_nodes.py
```diff
@@ -1797,6 +1797,17 @@ def __init__(self, name=None, doc=None, lineno=None,
         if parent is not None:
             parent.frame().set_local(name, self)
 
+        for name, node in self.implicit_locals():
+            self.add_local_node(node, name)
+
+    def implicit_locals(self):
+        """Get implicitly defined locals locals"""
+        locals_ = (('__module__', self.special_attributes.py__module__),)
```
PCManticore: Can we have this a static, with the exception of the `if` statement? e.g no method for it.

PCManticore: Ah, I see, nevermind!

PCManticore:
Thank you @brycepg !

PR #478, opened 2018-01-07
Fix submodule import in six.moves

brycepg:
This commit fixes import errors where modname
started with but was not equal to six.moves

Fixes  PyCQA/pylint#1640


PCManticore:
Thank you so much @brycepg ! This is great!

Issue #472, opened 2017-12-19
test_object_dunder_new_is_inferred_if_decorator unexpected success in 1.6.0


felixonmars:
### Steps to reproduce
1. Run the test suite in Python 3.6.3

### Current behavior
```
test_object_dunder_new_is_inferred_if_decorator (unittest_inference.ObjectDunderNewTest) ... unexpected success
...
Ran 764 tests in 8.279s

FAILED (skipped=52, expected failures=8, unexpected successes=1)
```

### Expected behavior
Test passes.

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output
1.6.0

Issue #444, opened 2017-07-24
Can't stack multiple inference tips


PCManticore:
Each inference tip is overriding the previous set inference tip for the corresponding node. This shouldn't be the case. Maybe we can stack them up. Also, we should probably support predicates for inference tips, as we do for transforms.

PCManticore:
Also this causes https://github.com/PyCQA/astroid/issues/454 (note that the test will be marked as an expected failure, so remove that once this gets fixed)

brycepg:
I could make _explicit_inference an container of transforms. infer could call the next result if the first _explicit_inference returned UsedDefaultInference.

This wouldn't solve a potential issue what order should the inferences be called in

PCManticore:
I'm on board with the idea of making `_explicit_inference` a container of transforms. For the rest, I would say that we should call the remainder of inference tips for that node even when a former inference tip raises `UsedDefaultInference`. I think this is the correct approach since multiple transforms can modify the same node, but with different conditions / filters. As long as the transform is idempotent, this should also solve the problem of ordering, which can't be relied upon (that is we cannot guarantee any ordering whatsoever, so it's best to inform the users that any transform they write has to be idempotent). What will happen once all transforms have raised `UseInferenceDefault` is that we'll default to the default inference, as we currently do. How does this sound?

PR #482, opened 2018-01-22
Fix inference causing existing node locations to be changed

brycepg:
Only set the inferred node lineno and col_offset if they are unset upon
inference.

Fixes PyCQA/pylint#1839


brycepg:
These CI tools are getting really flakey

brycepg:
AppVeyor is failing with

```
Command exited with code 1
dir "env:
```

astroid/tests/unittest_inference.py
```diff
@@ -4273,7 +4291,6 @@ def test_duplicated_keyword_arguments(self):
         self.assertIn('f', site.duplicated_keywords)
 
 
-@unittest.expectedFailure
```
PCManticore: Why was this removed? I'm not sure we fixed this.

brycepg: I was trying to stop TravisCI from failing. I'll drop that line and re-push

brycepg:
For some reason #472 is now affecting the Travis CI build

PCManticore:
Interesting. I restarted the jobs to see if they work now. I imagine this might be caused by non-deterministic runs of the transforms, that means if inference transforms are suddenly applied in a different order than what was happening until now, we might see the original problem of the issue you linked gone (but as I mentioned there or in some other issue, if I recall correctly, the problem is that only one inference tip gets successful if we have multiple for the same node class, since the latter gets reset)

PCManticore:
This is the issue I was thinking about: https://github.com/PyCQA/astroid/issues/444. Anyway, we can remove the expectedfailure decorator for now. Can you do that if you get a chance? also don't forget to rebase this branch so we can merge it.

brycepg:
Thanks for the review @PCManticore, I've removed the expected failure. AppVeyor is acting up again.

I'll try to work on inference_tip stacking, I don't know much about it yet however.

brycepg:
So it looks like this expected failure test is passing on TravisCI and Failing on AppVeyor.....

brycepg:
I'll try to work on inference tip stacking

PCManticore:
Thank you @brycepg ! Let's merge this, we can revisit the inference stacking afterwards. Let me know if I can help you with anything regarding that.

PR #484, opened 2018-01-28
Fix shared InferenceContext path causing uninferable results

### LINKED ISSUE #483: Layered DictUnpack nodes become uninferable
### Steps to reproduce
```python
import astroid
code = """
base = {'data': 0}
new = {**base, 'data': 1}
new3 = {**base, **new}
b = new3 #@
"""
ass = astroid.extract_node(code)
ass.value.inferred()
```

### Current behavior

The above code returns `[Uninferable]`

### Expected behavior

return `[Dict]`

2.0.0

brycepg:
Upon InferenceContext clone, the `path` attribute set is now shallow copied instead of shared (unintentionally?) among multiple diverged `InferenceContext` objects.

Fixes #483

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,11 @@ Change log for the astroid package (used to be astng)
 
      Close #473
 
+   * Fix multiple objects sharing the same InferenceContext.path causing
+   uninferable results
+
+     Close #483
+
 2017-12-15 -- 1.6.0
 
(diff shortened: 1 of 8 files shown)


PCManticore:
Wow @brycepg this is phenomenal! Thank you for this fix! It should improve the inference quite a bit.

PR #487, opened 2018-02-02
Update subprocess brain to know about Popen args member in python>=3.3

brycepg:
Fixes PyCQA/pylint#1860

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -28,6 +28,11 @@ Change log for the astroid package (used to be astng)
 
      Close PyCQA/pylint#1839
 
+   * Subprocess.Popen brain now knows of the args member
+
+     Close PyCQA/pylint#1860
+
+
 2017-12-15 -- 1.6.0
 
    * When verifying duplicates classes in MRO, ignore on-the-fly generated classes
--- a/astroid/tests/unittest_brain.py
+++ b/astroid/tests/unittest_brain.py
@@ -860,5 +860,24 @@ class RandomSampleTest(unittest.TestCase):
         self.assertEqual(elems, [1, 2])
 
 
+class SubprocessTest(unittest.TestCase):
+    """Test subprocess brain"""
+    # TODO Add more tests so that we can some day
+    # Remove this brain when all the tests work without the brain
+    @unittest.skipIf(sys.version_info < (3, 3),
+                     reason="Python 2.7 subprocess doesnt have args")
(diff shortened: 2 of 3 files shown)


astroid/tests/unittest_brain.py
```diff
@@ -908,5 +908,27 @@ def test_inferred_successfully(self):
...
+        """Make sure the args attribute exists for Popen
+
+        Test for https://github.com/PyCQA/pylint/issues/1860"""
+        name = astroid.extract_node("""
+        import subprocess
+        p = subprocess.Popen(['ls'])
+        p #@
+        """)
+        [inst] = name.inferred()
+        try:
```
PCManticore: This can be written as:
```
with self.assertRaises(Exception):
   inst.getattr(...)
```

brycepg: I actually don't want to assert that the test raises anything. I just want to make sure that the test fails properly on the occasion that it does fail: (F) instead of (E)

PCManticore: Ah, I see, I read it wrongly. In this case, I find it confusing, the `except` clause should be removed and just assert that the `args` value points to something that you expect.

astroid/tests/unittest_brain.py
```diff
@@ -908,5 +908,24 @@ def test_inferred_successfully(self):
...
+    def test_subprocess_args(self):
+        """Make sure the args attribute exists for Popen
+
+        Test for https://github.com/PyCQA/pylint/issues/1860"""
+        name = astroid.extract_node("""
+        import subprocess
+        p = subprocess.Popen(['ls'])
+        p #@
+        """)
+        [inst] = name.inferred()
```
PCManticore: Thanks for changing the test @brycepg

ChangeLog
```diff
@@ -86,6 +86,10 @@ Change log for the astroid package (used to be astng)
 
       Close PyCQA/pylint#1328
 
+	* Subprocess.Popen brain now knows of the args member
```
PCManticore: This seems to be placed in the wrong section, it needs to go in 2.0

PCManticore:
@brycepg Left one more comment for you to address. Also it would be great if you could rebase your branch on top of master, I fixed the `random` tests so it should be easier to spot if this PR fails or not. Once it's green, please merge :)

PR #486, opened 2018-01-28
Enhancement of numpy brain

hippo91:
This PR addresses the problem of false positive ``no-member`` emission when using ``numpy`` and specific types (last part of #https://github.com/PyCQA/pylint/issues/779).
The missing types have been added to the numpy brain and a corresponding unit test has been created.

PR #488, opened 2018-02-03
Backport of PR #486

hippo91:
+ Add of ChangeLog entry.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1387,6 +1387,3 @@ Change log for the astroid package (used to be astng)
       longer be maintained (this explains that this package is starting with
       the 0.13 version number, since the fork occurs with the version
       released in logilab-common 0.12).
-
-
-
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,9 @@
 Change log for the astroid package (used to be astng)
 =====================================================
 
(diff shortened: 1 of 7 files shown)


hippo91:
@PCManticore is there a problem with Travis and python3.x or did i miss something?

ceridwen:
Looks like it's failing because of pylint failures and an unexpected success.  Unfortunately the test output doesn't tell us what the unexpected success *is*.  Can you try running the tests locally to figure out what the unexpected success is?

ceridwen:
Actually, it might be this bug: https://github.com/PyCQA/astroid/issues/472

In which case it's a preexisting issue and nothing to do with your pull request.

brycepg:
We have this issue in master, except I removed the unexpected success. The test is unreliably passing and failing depending on the CI and something else (we think inference tips being overwritten by each other). I think marking this test as ignored is the best option right now.

PCManticore:
@brycepg is right, we can disable the test for now. @hippo91 this is failing the linting, otherwise it is good to go.

hippo91:
Thanks @ceridwen , @brycepg  and @PCManticore .  I replace the decorator @unittest.expectedfailure with @unittest.skip and the CI is ok!.
I merge it.

PR #489, opened 2018-02-05
Fix inference tip collisions

### LINKED ISSUE #472: test_object_dunder_new_is_inferred_if_decorator unexpected success in 1.6.0

brycepg:
Set predicates for transforms which use inference_tips to prevent
_explicit_inference being overwritten

Fixes #472

### DIFF
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -97,9 +97,12 @@ class AsStringRegexpPredicate(object):
         # pylint: disable=no-member; github.com/pycqa/astroid/126
         return self.regexp.search(node.as_string())
 
-def inference_tip(infer_function):
+def inference_tip(infer_function, raise_on_overwrite=False):
     """Given an instance specific inference function, return a function to be
     given to MANAGER.register_transform to set this inference function.
+
+    :param bool raise_on_overwrite: Raise an `InferenceOverwriteError`
+        if the inference tip will overwrite another. Used for debugging
 
     Typical usage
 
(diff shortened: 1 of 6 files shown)


astroid/__init__.py
```diff
@@ -107,8 +107,22 @@ def inference_tip(infer_function):
 
        MANAGER.register_transform(Call, inference_tip(infer_named_tuple),
                                   predicate)
+
+    .. Note::
+
+        ALWAYS use a predicate with an inference_tip
+        since registering a transform will
+        influence inference: an interance tip
```
PCManticore: Typo here for `inference`

astroid/__init__.py
```diff
@@ -107,8 +107,22 @@ def inference_tip(infer_function):
...
+        overwrites the _explicit_inference
```
PCManticore: I find this a bit too technical, let's just say that currently when we define an inference tip, it will override any previously set inference tip for the same kind of node, leading to some of them not being called.

astroid/__init__.py
```diff
@@ -107,8 +107,22 @@ def inference_tip(infer_function):
...
+        attribute
     """
     def transform(node, infer_function=infer_function):
+        if (raise_on_overwrite
+                and node._explicit_inference is not None
+                and node._explicit_inference is not infer_function):
+            raise Exception(
```
PCManticore: Let's raise a custom astroid exception, by adding one in `astroid/exceptions.py`

brycepg:
Thanks @PCManticore ! I made those changes

PR #490, opened 2018-02-10
Add move_to_end to OrderedDict collections brain

ioanatia:
### Fixes / new features
- Close https://github.com/PyCQA/pylint/issues/1872

### DIFF
--- a/astroid/brain/brain_collections.py
+++ b/astroid/brain/brain_collections.py
@@ -6,6 +6,7 @@ import sys
 
 import astroid
 
+PY34 = sys.version_info >= (3, 4)
 PY35 = sys.version_info >= (3, 5)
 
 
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_collections.py
```diff
@@ -62,5 +58,16 @@ def __imul__(self, other): pass
         def __rmul__(self, other): pass'''
     return base_deque_class
 
+
+def _ordered_dict_mock():
+    base_ordered_dict_class = '''
+    class OrderedDict(dict):
+        def __reversed__(self): return self[::-1]
+    '''
+    if PY35:
```
brycepg: Nice! `move_to_end` was added in 3.2 however (https://docs.python.org/3.6/library/collections.html#collections.OrderedDict.move_to_end) so we at least want to include 3.4 (the earliest python3.x we support atm) in this transform.

brycepg:
Looks good @ioanatia can you either squash these commits or separate out the unittest changes from the changelog entry?

PR #491, opened 2018-02-11
Stop generated builtin str methods from referencing a dummy class

brycepg:
This class, called 'whatever' ended up appearing in pyreverse
due to a local called `__class__` which pointed to it. pyreverse
seemed to have found this class by recursively visiting nodes
and calling the values() method of each node.

Closes PyCQA/pylint#1875

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -74,6 +74,11 @@ def _extend_str(class_node, rvalue):
     code = code.format(rvalue=rvalue)
     fake = AstroidBuilder(MANAGER).string_build(code)['whatever']
     for method in fake.mymethods():
+        method.parent = class_node
+        method.lineno = None
+        method.col_offset = None
+        if '__class__' in method.locals:
+            method.locals['__class__'] = [class_node]
         class_node.locals[method.name] = [method]
         method.parent = class_node
 


astroid/brain/brain_builtin_inference.py
```diff
@@ -74,6 +74,11 @@ def ljust(self, width, fillchar=None):
     code = code.format(rvalue=rvalue)
     fake = AstroidBuilder(MANAGER).string_build(code)['whatever']
     for method in fake.mymethods():
+        method.parent = class_node
+        method.lineno = None
+        method.col_offset = None
+        if '__class__' in method.locals:
+            method.locals['__class__'] = [class_node]
```
PCManticore: Hmm, I was expecting `__class__` to point to `fake` here not to `class_node`. In this case, `__class__` inside a method should point to its parent class. Although I would say we should move away from putting `__class_` in locals since it doesn't reflect the actual code (we patched locals to make sure that `__class__` can be found by pylint, but we should switch to something along the lines mentioned in this comment https://github.com/PyCQA/pylint/issues/1825#issuecomment-355815607). Can you test to see if removing `__class__` helps in this case if it is present?

brycepg: `fake` is `ClassDef.whatever`  `class_node` is `ClassDef.Bytes`. I'm just swapping out parent classes since we are injecting methods to the builtin classes. The issue is that pyreverse has a whilelist for builtin classes, however the class `whatever` isn't a bulitin class. I think this issue would keep popping up if we keep references to this class

brycepg: I can confirm removing https://github.com/PyCQA/astroid/blob/master/astroid/scoped_nodes.py#L1289 stops the pyreverse tests from failing

PCManticore: I was referring at removing `__class__` from the locals of `method`, if it works with that. If it works, we can go with this for now, and then we should revisit the patching of `locals` with `__class__`.

brycepg: Yes removing `__class__` from the locals of the method does fix this issue.

PR #495, opened 2018-02-16
Run python 3.6 in AppVeyor

ioanatia:
This is already done for travis:

https://github.com/PyCQA/astroid/blob/aae8d207b65c27dc04e4a8cfab176e06d03065b2/.travis.yml#L15-L18

Unfortunately, version 3.7 is not available in AppVeyor https://www.appveyor.com/docs/build-environment/#python, so I have only included python 3.6.


PCManticore:
This is nice!

PR #497, opened 2018-02-18
Making Pylint faster

nickdrozd:
I recently learned how to do Python profiling with `yappi`. `pylint` has
always seemed slow to me, so I decided to see if it could be sped up.

Here is the call graph from running `pylint` against
https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py:

pycodestyle-master

On these graphs, nodes represent function calls, and the brighter the
node, the more time spent in that function. Each node has three
numbers: 1) the total time spent in the function, including its
subcalls; 2) the total time spent in that function but not its
subcalls; and 3) the total number of times the function was called.

I was somewhat surprised to learn (although it seems obvious in
retrospect) that the `pylint` code itself is not especially slow and
that most time is being spent in `astroid` functions. Looking at that
graph, it's obvious that `nodes_of_clas` and `get_children` are
bottlenecks, and optimizing those functions could have a big impact.
(To explain the numbers a bit, `nodes_of_class` is taking up almost
60% of total CPU time, and more than a third of that time is being
spent in `get_children`.)

First, I timed three runs of `pylint` with `astroid` on master to get
a benchmark:

```
real	1m17.996s
user	0m31.987s
sys	0m45.806s

real	1m17.828s
user	0m31.704s
sys 	0m45.936s

real	1m16.928s
user	0m31.327s
sys 	0m45.416s
```

Next, I timed three runs with astroid on the commit `Add type-specific
get_children`. This commit gives each (or almost each) node class its
own `get_children` method instead of having them all use the same
generic function. This sped things up:

```
real	1m11.266s
user	0m30.778s
sys	0m40.306s

real	1m11.503s
user	0m31.215s
sys	0m40.036s

real	1m11.761s
user	0m31.376s
sys	0m40.194s
```

Here is the call graph with that change:

pycodestyle-children

As compared with the previous graph, `nodes_by_class` takes slightly
less total CPU time, but of the time it does take, less of it is spent
in its subcalls. This is because the type-specific `get_children`
calls go much faster.

Okay, so `nodes_of_class` is now the sole bottleneck. First, I applied
the commit `Move nodes_of_class null check out of inner loop`, which,
as the name suggests, just shuffles around the null check logic in
that function. This provided a modest speedup:

```
real	1m10.510s
user	0m30.712s
sys	0m39.624s

real	1m11.590s
user	0m31.072s
sys	0m40.328s

real	1m11.274s
user	0m31.055s
sys	0m40.049s
```

According to the call graph (which I won't bother to post), this
change saved about .2% of total CPU time, which is not bad for such a
small change.

Finally, I applied a much larger change, the commit `Add type-specific
nodes_of_class`. This commit adds a few functions that are just like
`nodes_of_class` except that they apply only to specific node
searches. This eliminates the need to do expensive runtime type
checking. These replacing calls to `nodes_of_class` within `astroid`
with these new functions sped things up significantly:

```
real	0m58.005s
user	0m25.634s
sys	0m32.200s

real	0m57.874s
user	0m25.870s
sys	0m31.838s

real	0m58.177s
user	0m26.026s
sys	0m31.984s
```

Here is the call graph

pycodestyle-nodes

It's a little hard to interpret, but by my count `nodes_of_class` went
from taking ~58% of total CPU time to ~48%.

Note that all of this data comes from running `pylint`, as I said
earlier, against just a single file, `pycodestyle.py`. Larger `pylint`
targets (projects with lots of subdirectories, for instance) have
different profiles, and different functions become more prominent
(`get_children` and `nodes_of_class` are slow in all circumstances,
however). I have ideas for more optimizations, which I will come back
with after these first changes are taken care of.

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -627,8 +627,16 @@ class NodeNG(object):
         """
         if isinstance(self, klass):
             yield self
+
+        if skip_klass is None:
+            for child_node in self.get_children():
+                for matching in child_node.nodes_of_class(klass, skip_klass):
+                    yield matching
+
+            return
+
         for child_node in self.get_children():
-            if skip_klass is not None and isinstance(child_node, skip_klass):
+            if isinstance(child_node, skip_klass):
                 continue
             for matching in child_node.nodes_of_class(klass, skip_klass):
                 yield matching
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -220,6 +220,7 @@ class NodeNG(object):
 
     :type: bool
     """
+    is_lambda = False
     # Attributes below are set by the builder module or by raw factories
     lineno = None
     """The line that this node appears on in the source code.
@@ -639,6 +640,30 @@ class NodeNG(object):
             if isinstance(child_node, skip_klass):
                 continue
             for matching in child_node.nodes_of_class(klass, skip_klass):
+                yield matching
+
+    def _get_assign_nodes(self):
+        for child_node in self.get_children():
+            for matching in child_node._get_assign_nodes():
+                yield matching
+
+    def _get_name_nodes(self):
+        for child_node in self.get_children():
+            for matching in child_node._get_name_nodes():
+                yield matching
+
+    def _get_return_nodes_skip_functions(self):
+        for child_node in self.get_children():
+            if child_node.is_function:
+                continue
+            for matching in child_node._get_return_nodes_skip_functions():
+                yield matching
+
+    def _get_yield_nodes_skip_lambdas(self):
+        for child_node in self.get_children():
+            if child_node.is_lambda:
+                continue
+            for matching in child_node._get_yield_nodes_skip_lambdas():
                 yield matching
 
     def _infer_name(self, frame, name):
@@ -1267,6 +1292,13 @@ class Name(LookupMixIn, NodeNG):
     def get_children(self):
         yield from ()
 
+    def _get_name_nodes(self):
+        yield self
+
+        for child_node in self.get_children():
+            for matching in child_node._get_name_nodes():
+                yield matching
+
 
 class Arguments(mixins.AssignTypeMixin, NodeNG):
     """Class representing an :class:`ast.arguments` node.
@@ -1701,6 +1733,13 @@ class Assign(mixins.AssignTypeMixin, Statement):
         yield from self.targets
 
         yield self.value
+
+    def _get_assign_nodes(self):
+        yield self
+
+        for child_node in self.get_children():
+            for matching in child_node._get_assign_nodes():
+                yield matching
 
 
 class AnnAssign(mixins.AssignTypeMixin, Statement):
@@ -2792,7 +2831,7 @@ class ExceptHandler(mixins.AssignTypeMixin, Statement):
         """
         if self.type is None or exceptions is None:
             return True
-        for node in self.type.nodes_of_class(Name):
+        for node in self.type._get_name_nodes():
             if node.name in exceptions:
                 return True
         return False
@@ -3582,7 +3621,7 @@ class Raise(Statement):
         """
         if not self.exc:
             return False
-        for name in self.exc.nodes_of_class(Name):
+        for name in self.exc._get_name_nodes():
             if name.name == 'NotImplementedError':
                 return True
         return False
@@ -3620,6 +3659,15 @@ class Return(Statement):
     def get_children(self):
         if self.value is not None:
             yield self.value
+
+    def _get_return_nodes_skip_functions(self):
+        yield self
+
+        for child_node in self.get_children():
+            if child_node.is_function:
+                continue
+            for matching in child_node._get_return_nodes_skip_functions():
+                yield matching
 
 
 class Set(_BaseContainer):
@@ -4260,6 +4308,15 @@ class Yield(NodeNG):
     def get_children(self):
         if self.value is not None:
             yield self.value
+
+    def _get_yield_nodes_skip_lambdas(self):
+        yield self
+
+        for child_node in self.get_children():
+            if child_node.is_function_or_lambda:
+                continue
+            for matching in child_node._get_yield_nodes_skip_lambdas():
+                yield matching
 
 
 class YieldFrom(Yield):
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -965,6 +965,9 @@ class _BaseContainer(mixins.ParentAssignTypeMixin,
         :rtype: str
         """
 
+    def get_children(self):
+        yield from self.elts
+
 
 class LookupMixIn(object):
     """Mixin to look up a name in the right scope."""
@@ -1170,6 +1173,9 @@ class AssignName(LookupMixIn, mixins.ParentAssignTypeMixin, NodeNG):
 
         super(AssignName, self).__init__(lineno, col_offset, parent)
 
+    def get_children(self):
+        yield from ()
+
 
 class DelName(LookupMixIn, mixins.ParentAssignTypeMixin, NodeNG):
     """Variation of :class:`ast.Delete` represention deletion of a name.
@@ -1206,6 +1212,9 @@ class DelName(LookupMixIn, mixins.ParentAssignTypeMixin, NodeNG):
         """
 
         super(DelName, self).__init__(lineno, col_offset, parent)
+
+    def get_children(self):
+        yield from ()
 
 
 class Name(LookupMixIn, NodeNG):
@@ -1246,6 +1255,9 @@ class Name(LookupMixIn, NodeNG):
         """
 
         super(Name, self).__init__(lineno, col_offset, parent)
+
+    def get_children(self):
+        yield from ()
 
 
 class Arguments(mixins.AssignTypeMixin, NodeNG):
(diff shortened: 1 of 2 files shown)


rogalski:
Wow, an effort put to this is very impressive. Thank you!

astroid/node_classes.py
```diff
@@ -630,12 +631,44 @@ def nodes_of_class(self, klass, skip_klass=None):
...
         for child_node in self.get_children():
-            if skip_klass is not None and isinstance(child_node, skip_klass):
+            if isinstance(child_node, skip_klass):
                 continue
             for matching in child_node.nodes_of_class(klass, skip_klass):
                 yield matching
 
+    def get_assign_nodes(self):
+        for child_node in self.get_children():
+            for matching in child_node.get_assign_nodes():
```
brycepg: You can use `yield from` now that we're using Python 3 only (unless you want to backport this to the 1.6 branch). That actually might give a speedup

astroid/node_classes.py
```diff
@@ -1172,6 +1209,9 @@ def __init__(self, name=None, lineno=None, col_offset=None, parent=None):
 
         super(AssignName, self).__init__(lineno, col_offset, parent)
 
+    def get_children(self):
+        return (_ for _ in ())
```
brycepg: I think you could do `yield from ()` or `return iter([])` here

brycepg:
Sorry ignore my comments if you saw them, I remembered that we haven't remove 2.7 yet from tox (however `yield from` might help with speedups)

brycepg:
@nickdrozd What do you think we can do to help prevent performance regressions? At first glance, some of this code would make me want to refactor it away because of DRY (but I understand the trade off)

astroid/node_classes.py
```diff
@@ -644,6 +645,30 @@ def nodes_of_class(self, klass, skip_klass=None):
             for matching in child_node.nodes_of_class(klass, skip_klass):
                 yield matching
 
+    def get_assign_nodes(self):
```
PCManticore: I would prefer these to be private. Some comments before them mentioning why they are like this would also help in the future in case someone wonders why these couldn't have been made more DRY

PCManticore:
Thank you for doing this amazing work @nickdrozd ! There's not much to be commented about this PR, left a couple of comments with things that we can improve, but overall looks like a pretty good gain in performance. Also I noticed from your yippi output that there might be other places where there are tons of calls, where we should have less, such as in the transforms for instance.
Also I support Bryce's question. Before we start optimising further, we should think about measuring the performance first; I imagine it would be super easy to introduce regressions for it.

nickdrozd:
Thanks for the kind words! I was somewhat worried that these changes
wouldn't be well-received, so I'm glad that everyone is on board :cat2:

A few points:

#### `yield from`

I was going to say that I don't feel comfortable being the one to
cause the Python 2 tests to fail, but it looks like all that stuff is
already getting ripped out. I will update everything to use `yield
from`.

#### Performance Regression Testing

I have no experience testing for performance regression, so I don't
have any clever ideas. This may be a naive approach, but the simplest
thing I can think to do would be to add a pass in the CI that runs
some fixed version of Pylint (say, 1.8) with the version of Astroid in
master or the PR being tested against some fixed file (say,
`pycodestyle.py`). If the approximate speed of that run in the test
environment can be determined, then the test will fail if the run of
Pylint is not within some reasonable bound of the target time.

That's a general (and again, possibly naive) approach to performance
regression testing. As far as preventing regression in the specific
areas affected by this PR, there are a few options. One possibility
would be to delete the generic `NodeNG.get_children` entirely and
require that each node type implements its own. A variation of this
idea would have `NodeNG.get_children` raise a `NotImplementedError`,
accompanied by a comment that says something like

```
# DON'T REFACTOR THIS, I KNOW IT LOOKS INELEGANT
# BUT IT'S LIKE THIS FOR PERFORMANCE REASONS
```

(`NodeNG` could be made into an actual abstract base class, but that
seems a little dramatic to me.)

Use of `nodes_of_class` within Astroid code is always bad, so that
should be prohibited. Astroid's `pylintrc` could probably be modified
to ensure this. In the worst case, a custom Pylint plugin could be
written to check for its use.

#### Making `get_assign_nodes`, etc, private

@PCManticore suggests that the type-specific `nodes_of_class`
replacements be made private. This is reasonable, as making them
"public" functions increases the API surface. On the other hand,
Pylint makes lots of calls to `nodes_of_class`, and these calls are,
of course, slow:

```
~/pylint $ grep -r nodes_of_class pylint
pylint/checkers/async.py:        for child in node.nodes_of_class(astroid.Yield):
pylint/checkers/base.py:    for import_node in try_block.nodes_of_class((astroid.ImportFrom, astroid.Import)):
pylint/checkers/base.py:        _node for _node in loop.nodes_of_class(loop_nodes,
pylint/checkers/base.py:        _node for _node in loop.nodes_of_class(astroid.Break,
pylint/checkers/base.py:        starred = list(node.targets[0].nodes_of_class(astroid.Starred))
pylint/checkers/base.py:        returns = node.nodes_of_class(astroid.Return,
pylint/checkers/base.py:            for child in node.nodes_of_class(astroid.Global)
pylint/checkers/base.py:        for node_name in node.nodes_of_class(astroid.Name):
pylint/checkers/base.py:            child.names for child in node.nodes_of_class(astroid.Nonlocal)
pylint/checkers/base.py:            child.names for child in node.nodes_of_class(astroid.Global)
pylint/checkers/base.py:        if any(expr.nodes_of_class(astroid.Call)):
pylint/checkers/classes.py:            for call in infer_method.nodes_of_class(astroid.Call):
pylint/checkers/classes.py:    for call in fundef_node.nodes_of_class(astroid.Call):
pylint/checkers/classes.py:        for stmt in node.nodes_of_class(astroid.Call):
pylint/checkers/imports.py:                any(node.nodes_of_class((astroid.Import, astroid.ImportFrom))):
pylint/checkers/imports.py:            if any(root.nodes_of_class((astroid.Import, astroid.ImportFrom))):
pylint/checkers/newstyle.py:        for stmt in node.nodes_of_class(astroid.Call):
pylint/checkers/refactoring.py:        for defined_argument in scope.args.nodes_of_class(astroid.AssignName):
pylint/checkers/refactoring.py:        for name in node.target.nodes_of_class(astroid.AssignName):
pylint/checkers/refactoring.py:            for name in names.nodes_of_class(astroid.AssignName):
pylint/checkers/refactoring.py:        return_nodes = node.nodes_of_class(astroid.Return)
pylint/checkers/refactoring.py:            for subscript in child.nodes_of_class(astroid.Subscript):
pylint/checkers/typecheck.py:    for name in statement.nodes_of_class(astroid.Name):
pylint/checkers/typecheck.py:        returns = list(function_node.nodes_of_class(astroid.Return,
pylint/checkers/utils.py:            for ass_node in _node.nodes_of_class(astroid.AssignName):
pylint/checkers/utils.py:            for ass_node in _node.target.nodes_of_class(astroid.AssignName):
pylint/checkers/utils.py:        for ass_node in _node.nodes_of_class(astroid.AssignName):
pylint/checkers/utils.py:        for imp_node in _node.nodes_of_class((astroid.ImportFrom, astroid.Import)):
pylint/checkers/utils.py:            for default_name_node in default_node.nodes_of_class(astroid.Name):
pylint/checkers/utils.py:        if node in base.nodes_of_class(astroid.Name):
pylint/checkers/variables.py:    imports = frame.nodes_of_class((astroid.Import, astroid.ImportFrom))
pylint/checkers/variables.py:    assign_stmts = name_node.scope().nodes_of_class(astroid.AssignName)
pylint/checkers/variables.py:        assigned_to = [var.name for var in node.target.nodes_of_class(astroid.AssignName)]
pylint/checkers/variables.py:        global_names = _flattened_scope_names(node.nodes_of_class(astroid.Global))
pylint/checkers/variables.py:        nonlocal_names = _flattened_scope_names(node.nodes_of_class(astroid.Nonlocal))
pylint/extensions/_check_docs_utils.py:            for ret in target.nodes_of_class(astroid.Return):
pylint/extensions/docparams.py:        return_nodes = node.nodes_of_class(astroid.Return)
pylint/extensions/redefined_variable_type.py:                          redef_parent in orig_parent.nodes_of_class(astroid.If)):
```

A whole slew of these functions could be added to the API:
`get_assign_nodes`, `get_assignname_nodes`, `get_global_nodes`, etc.
Using these instead of the generic `nodes_of_class` would make Pylint
faster still. Doing that is a lot more work, however, and would
require coordinating changes in the two repos.

As I write this, I've convinced myself that they should be made
"private" for now, but this decision should certainly be revisited in
the near future.

nickdrozd:
The PR has been updated to

* add `get_children` to every node,
* use `yield from`, and
* make the `nodes_of_class` replacements "private".

`NodeNG.get_children` could be deleted at this point.

brycepg:
I think it's a good idea to keep `NodeNG.get_children` for yet to be added nodes and as an interface

PCManticore:
I agree with Bryce that `NodeNG.get_children` could still be here for newly added nodes. Regarding pylint's usage of `nodes_of_class` I would say to not add those here, but rather implement them right into pylint, astroid should only provide fast `nodes_of_class` alternatives for what it's already using, while pylint could have its own variants.

Nevertheless, this was a fun patch @nickdrozd ! Thank you so much for contributing this work.
Regarding the performance regression, that sounds good, but I wonder if we're going to have potential varying results due to, say, CI being too slow at a given time. I wonder if there is any service that out there that provides this information somehow and that's free for open source projects. Anyway, I think we can merge this PR as is and we can revisit later how to measure performance.

PR #502, opened 2018-02-20
Provide support for isinstance builtin

### LINKED ISSUE #98: Understand isinstance builtin

---

Just as we do with super and other builtins. Will be used for flow control analysis.

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/98

brycepg:
isinstance now properly infers `True` or `False` based on its arguments.

Close #98

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -46,6 +46,10 @@ Change log for the astroid package (used to be astng)
 
 	Close PyCQA/pylint#1884
 
+    * Inference now understands the 'isinstance' builtin
+
+     Close #98
+
 
 2017-12-15 -- 1.6.0
 
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -12,7 +12,8 @@ from textwrap import dedent
 
 import six
 from astroid import (MANAGER, UseInferenceDefault, AttributeInferenceError,
-                     inference_tip, InferenceError, NameInferenceError)
+                     inference_tip, InferenceError, NameInferenceError,
+                     AstroidTypeError, AstroidError)
 from astroid import arguments
 from astroid.builder import AstroidBuilder
 from astroid import helpers
@@ -515,6 +516,64 @@ def _infer_object__new__decorator_check(node):
     return False
 
 
+def infer_isinstance(callnode, context=None):
+    """Infer isinstance calls
+
+    :param nodes.Call callnode: an isinstance call
+    :param InferenceContext: context for call
+        (currently unused but is a common interface for inference)
+    :rtype nodes.Const: Boolean Const value of isinstance call
+
+    :raises UseInferenceDefault: If the node cannot be inferred
+    """
+    call = arguments.CallSite.from_call(callnode)
+    if call.keyword_arguments:
+        # isinstance doesn't support keyword arguments
+        raise UseInferenceDefault("TypeError: isinstance() takes no keyword arguments")
+    if len(call.positional_arguments) != 2:
+        raise UseInferenceDefault(
+            "Expected two arguments, got {count}"
+            .format(count=len(call.positional_arguments)))
+    # The left hand argument is the obj to be checked
+    obj_node, class_or_tuple_node = call.positional_arguments
+    # The right hand argument is the class(es) that the given
+    # obj is to be check is an instance of
+    try:
+        class_container = _class_or_tuple_to_container(
+            class_or_tuple_node, context=context)
+    except InferenceError:
+        raise UseInferenceDefault
+    try:
+        isinstance_bool = helpers.object_isinstance(
+            obj_node, class_container, context)
+    except AstroidTypeError as exc:
+        raise UseInferenceDefault("TypeError: " + str(exc))
+    except AstroidError:
+        raise UseInferenceDefault
+    return nodes.Const(isinstance_bool)
+
+
+def _class_or_tuple_to_container(node, context=None):
+    # Move inferences results into container
+    # to simplify later logic
+    # raises InferenceError if any of the inferences fall through
+    node_infer = next(node.infer(context=context))
+    # arg2 MUST be a type or a TUPLE of types
+    # for isinstance
+    if isinstance(node_infer, nodes.Tuple):
+        class_container = [
+            next(node.infer(context=context))
+            for node in node_infer.elts
+        ]
+        class_container = [
+            klass_node for klass_node
+            in class_container if klass_node is not None
+        ]
+    else:
+        class_container = [node_infer]
+    return class_container
+
+
 # Builtins inference
 register_builtin_transform(infer_bool, 'bool')
 register_builtin_transform(infer_super, 'super')
@@ -528,6 +587,7 @@ register_builtin_transform(infer_dict, 'dict')
 register_builtin_transform(infer_frozenset, 'frozenset')
 register_builtin_transform(infer_type, 'type')
 register_builtin_transform(infer_slice, 'slice')
+register_builtin_transform(infer_isinstance, 'isinstance')
 
 # Infer object.__new__ calls
 MANAGER.register_transform(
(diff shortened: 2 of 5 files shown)


ChangeLog
```diff
@@ -42,6 +42,10 @@ Change log for the astroid package (used to be astng)
 
      Close #437, #447, #313, PyCQA/pylint#1642, PyCQA/pylint#1805, PyCQA/pylint#1854, PyCQA/pylint#1452
 
+  * Infernece now understands the 'isinstance' builtin
```
PCManticore: A typo here. Also the following line has a weird indentation, maybe it was indented with tabs instead of spaces?

astroid/brain/brain_builtin_inference.py
```diff
@@ -515,6 +516,62 @@ def _infer_object__new__decorator_check(node):
     return False
 
 
+def infer_isinstance(callnode, context=None):
+    """Infer isinstance calls
+
+    :param nodes.Call callnode: an isinstance call
```
PCManticore: Maybe we should add `context` as well?

astroid/brain/brain_builtin_inference.py
```diff
@@ -515,6 +516,62 @@ def _infer_object__new__decorator_check(node):
...
+    :rtype nodes.Const: Boolean Const value of isinstance call
+    """
+    call = arguments.CallSite.from_call(callnode)
+    if call.keyword_arguments:
+        # isinstance doesn't support keyword arguments
+        raise TypeInferenceError("isinstance() takes no keyword arguments")
```
PCManticore: Since this is part of the inference, we raise `UseDefaultInference` instead of an error. By raising an error, every time we infer `isinstance` as part of, say pylint, we're going to crash rather than throwing an `InferenceError` or returning `Uninferable`. For the cases where something is invalid in the inference engine, like this one, we usually add pylint rules.

astroid/brain/brain_builtin_inference.py
```diff
@@ -515,6 +516,62 @@ def _infer_object__new__decorator_check(node):
...
+    if len(call.positional_arguments) != 2:
+        raise TypeInferenceError(
```
PCManticore: Ditto.

astroid/brain/brain_builtin_inference.py
```diff
@@ -515,6 +516,62 @@ def _infer_object__new__decorator_check(node):
...
+    try:
+        class_container = _class_or_tuple_to_container(
+            class_or_tuple_node, context=context)
+    except InferenceError:
+        raise UseInferenceDefault
+    try:
+        isinstance_bool = helpers.object_isinstance(
+            obj_node, class_container, context)
+    except AstroidTypeError as exc:
+        raise TypeInferenceError(str(exc))
```
PCManticore: Ditto

astroid/brain/brain_builtin_inference.py
```diff
@@ -515,6 +516,62 @@ def _infer_object__new__decorator_check(node):
...
+    if isinstance_bool is util.Uninferable:
+        raise UseInferenceDefault
+    if isinstance_bool:
```
PCManticore: Could be simplified as `return nodes.Const(bool(isinstance_bool))`

astroid/helpers.py
```diff
@@ -89,6 +89,38 @@ def object_type(node, context=None):
...
+    class_seq = [item if not isinstance(item, bases.Instance)
+                 else util.Uninferable for item in class_seq]
+    # strict compatibility with isinstance
+    # isinstance(1, (int, 1)) evaluates to true
+    # isinstance(1, (1, int)) raises TypeError
+    for klass in class_seq:
+        if klass is util.Uninferable:
+            raise exceptions.AstroidTypeError(
+                "isinstance() arg 2 must be a type or tuple of types")
+        for obj_subclass in obj_type.mro():
```
PCManticore: You should catch the exceptions that `mro()` could raise: DuplicateBasesError and InconsistentMroError. Otherwise if these happen, we're going to raise an error rather than continue the inference/linting

astroid/tests/unittest_brain.py
```diff
@@ -921,5 +918,134 @@ def test_subprocess_args(self):
...
+        """) == "True" == str(isinstance(mainbar, type))
+
+    def test_isinstance_edge_case(self):
+        """isinstance allows bad type short-circuting"""
+        assert _get_result("isinstance(1, (int, 1))") == "True" == str(isinstance(1, (int, 1)))
+
+    def test_uninferable_bad_type(self):
+        """The second argument must be a class or a tuple of classes"""
+        # Should I subclass
+        with pytest.raises(astroid.TypeInferenceError):
```
PCManticore: Ditto here, these should get back InferenceError/Uninferable, not an astroid exception.

astroid/brain/brain_builtin_inference.py
brycepg: Alright. I don't like astroid silently failing (it makes debugging harder), but I'm not sure what to do about it at the moment.

PCManticore: Yeah, I agree with the sentiment, but since pylint is the main consumer of astroid, it's not easy to make these changes.

As an example of a place where I wanted to get type error information out of astroid is in the binary operations inference (https://github.com/PyCQA/astroid/blob/3fae32f99c7f432a1281b151ea17e163a2e8dc47/astroid/inference.py#L659), where the type errors hints can later on be accessed by a `type_errors` method (https://github.com/PyCQA/astroid/blob/8425d915c0186f87da66ad6223588ce70c49031c/astroid/node_classes.py#L1779). It's super ugly, by pylint can acess this one without having the entire inference crash down on a type error somewhere in a random file.

astroid/helpers.py
brycepg: I'll add these to the mro docstring

astroid/brain/brain_builtin_inference.py
brycepg: I guess I would have to do this in `Call`. I might come back to this later

astroid/brain/brain_builtin_inference.py
brycepg: Alright, did it

ChangeLog
```diff
@@ -46,6 +46,10 @@ Change log for the astroid package (used to be astng)
 
 	Close PyCQA/pylint#1884
 
+  * Inference now understands the 'isinstance' builtin
+
+	 Close #98
```
PCManticore: Tab-based indentation here?

astroid/helpers.py
```diff
@@ -83,6 +83,38 @@ def object_type(node, context=None):
...
+
+    :raises AstroidTypeError: if the given ``classes_or_seq`` are not types
+    """
+    if not isinstance(class_or_seq, (tuple, list)):
+        class_seq = (class_or_seq,)
+    else:
+        class_seq = class_or_seq
+    obj_type = object_type(node, context)
+    if obj_type is util.Uninferable:
+        return util.Uninferable
```
PCManticore: I see that this returns boolean but also an Uninferable object, maybe we could either raise an exception here, that the caller could handle, or we could hardcode the meaning of Uninferable into a `None`. I would prefer the first one since it seems more obvious. Nevertheless, it's a nitpick.

ChangeLog
brycepg: My bad, looks like vim has a 'changelog' filetype that I needed to override

astroid/helpers.py
brycepg: So I am trying to make these helpers consistent. `object_type` also returns Uninferable

brycepg: Hmm but I already raise an exception

PCManticore: Yeah, it's fine if `object_type` already does that. It was just a small pet peeve regarding two different return types that should have been handled by `object_type` as well when it was introduced.

PCManticore:
Cool, thanks Bryce!

PR #504, opened 2018-03-05
Fix contextmanager transform for nested contextmanagers

brycepg:
Close PyCQA/pylint#1746

The contextmanager transform was incorrectly using the 'first' yield instead of the yield in the current scope.

### DIFF
--- a/astroid/tests/unittest_inference.py
+++ b/astroid/tests/unittest_inference.py
@@ -2218,6 +2218,36 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         self.assertRaises(InferenceError, next, module['other_decorators'].infer())
         self.assertRaises(InferenceError, next, module['no_yield'].infer())
 
+    def test_nested_contextmanager(self):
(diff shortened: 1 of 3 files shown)


astroid/tests/unittest_inference.py
```diff
@@ -2218,6 +2218,31 @@ def no_yield_mgr():
...
+        def outer():
+            @contextmanager
+            def inner():
+                yield 2
+            yield inner
+
+        with outer() as foo:
+            foo #@
+        """
+        value = next(extract_node(code).infer())
```
PCManticore: Maybe we should also do an inference test, where we'd infer the value returned by the inner context manager, something along the lines of:
```
n = extract_node('''
import contextlib
@contextlib.contextmanager
def test():
  @contextlib.contextmanager
  def inner():
      yield 2
  yield inner

with test() as ctx:
   with ctx as y:
       y #@
''')
assert isinstance(next(n.infer()), Const)
```

brycepg: Good idea

PR #506, opened 2018-03-10
Builtin len

### LINKED ISSUE #112: Understand len builtin

---

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/112

brycepg:
Implement inference for len builtin 

Close #112

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -222,3 +222,45 @@ def class_instance_as_index(node):
     except exceptions.InferenceError:
         pass
     return None
+
+
+def object_len(node, context=None):
+    """Infer length of given node object
+
+    :param Union[nodes.ClassDef, nodes.Instance] node:
+    :param node: Node to infer length of
+
+    :raises AstroidTypeError: If an invalid node is returned
+        from __len__ method or no __len__ method exists
+    :raises InferenceError: If the given node cannot be inferred
+        or if multiple nodes are inferred
+    :rtype int: Integer length of node
+    """
+    inferred_node = safe_infer(node, context=context)
+    if inferred_node is None or inferred_node is util.Uninferable:
+        raise exceptions.InferenceError(node=node)
+    if inferred_node.qname() in ('builtins.str', 'builtins.bytes'):
+        return len(inferred_node.value)
+    if isinstance(inferred_node, (nodes.List, nodes.Set, nodes.Tuple, nodes.FrozenSet)):
+        return len(inferred_node.elts)
+    if isinstance(inferred_node, nodes.Dict):
+        return len(inferred_node.items)
+    try:
+        node_type = object_type(inferred_node, context=context)
+        len_call = next(node_type.igetattr("__len__", context=context))
+    except exceptions.AttributeInferenceError:
+        raise exceptions.AstroidTypeError(
+            "object of type '{}' has no len()"
+            .format(len_call.pytype()))
+
+    try:
+        result_of_len = next(len_call.infer_call_result(node, context))
+        # Remove StopIteration catch when #507 is fixed
+    except StopIteration:
+        raise exceptions.InferenceError(node=node)
+    if (isinstance(result_of_len, nodes.Const) and result_of_len.pytype() == "builtins.int"):
+        return result_of_len.value
+    else:
+        raise exceptions.AstroidTypeError(
+            "'{}' object cannot be interpreted as an integer"
+            .format(result_of_len))
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -2837,6 +2837,21 @@ class ExceptHandler(mixins.AssignTypeMixin, Statement):
         return False
 
 
+class FrozenSet(_BaseContainer):
+    """class representing a FrozenSet composite node"""
+
+    def pytype(self):
+        return '%s.frozenset' % BUILTINS
+
+    def _infer(self, context=None):
+        yield self
+
+    @decorators.cachedproperty
+    def _proxied(self): # pylint: disable=method-hidden
+        builtins = MANAGER.astroid_cache[BUILTINS]
+        return builtins.getattr('frozenset')[0]
+
+
 class Exec(Statement):
     """Class representing the ``exec`` statement.
 
--- a/astroid/tests/unittest_brain.py
+++ b/astroid/tests/unittest_brain.py
@@ -1092,5 +1092,144 @@ def _get_result(code):
     return _get_result_node(code).as_string()
 
 
+class TestLenBuiltinInference:
+    def test_len_list(self):
+        # Uses .elts
+        node = astroid.extract_node("""
+        len(['a','b','c'])
+        """)
+        node = next(node.infer())
+        assert node.as_string() == '3'
+        assert isinstance(node, nodes.Const)
+
+    def test_len_tuple(self):
+        node = astroid.extract_node("""
+        len(('a','b','c'))
+        """)
+        node = next(node.infer())
+        assert node.as_string() == '3'
+
+    def test_len_var(self):
+        # Make sure argument is inferred
+        node = astroid.extract_node("""
+        a = [1,2,'a','b','c']
+        len(a)
+        """)
+        node = next(node.infer())
+        assert node.as_string() == '5'
+
+    def test_len_dict(self):
+        # Uses .items
+        node = astroid.extract_node("""
+        a = {'a': 1, 'b': 2}
+        len(a)
+        """)
+        node = next(node.infer())
+        assert node.as_string() == '2'
+
+    def test_len_set(self):
+        node = astroid.extract_node("""
+        len({'a'})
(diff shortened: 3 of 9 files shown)


brycepg:
Something in package_resources might be inferred as len for some reason. (seeing from the lint errors), will checkout out.

astroid/helpers.py
```diff
@@ -199,3 +199,44 @@ def class_instance_as_index(node):
...
     return None
+
+
+def object_len(node, context=None):
+    """Infer length of given node object
+
+    :param Union[nodes.ClassDef, nodes.Instance] node:
+    :param node: Node to infer length of
+
+    :raises AstroidTypeError: If an invalid node is given
```
PCManticore: I can't quite parse this sentence, can you rephrase it?

astroid/helpers.py
```diff
@@ -199,3 +199,44 @@ def class_instance_as_index(node):
...
+        for an invalid type is returned
+    :raises InferenceError: If the given node cannot be inferred
+        or if multiple nodes are inferred
+    :rtype int: Integer length of node
+    """
+    inferred_node = safe_infer(node, context=context)
+    if inferred_node is None or inferred_node is util.Uninferable:
+        raise exceptions.InferenceError(node=node)
+    if inferred_node.qname() in ('builtins.str', 'builtins.bytes'):
```
PCManticore: I'd suggest to check against concrete types, e.g `Dict, List, Tuple` rather than verifying their inner nodes. This way it will be more clear what types are these.

astroid/helpers.py
```diff
@@ -199,3 +199,44 @@ def class_instance_as_index(node):
...
+    try:
+        node_type = object_type(inferred_node, context=context)
+        len_call = next(node_type.igetattr("__len__", context=context))
+    except exceptions.AttributeInferenceError:
+        raise exceptions.AstroidTypeError(
+            "object of type '{}' has no len()"
+            .format(len_call.pytype()))
+
+    try:
+        result_of_len = next(len_call.infer_call_result(node, context))
```
PCManticore: The StopIteration is odd, we should never return it per se, but rather raise an astroid exception. Can you check what exactly is triggering a StopIteration?

astroid/tests/unittest_brain.py
```diff
@@ -1011,5 +1011,144 @@ def _get_result(code):
     return _get_result_node(code).as_string()
 
 
+class TestLenBuiltinInference:
+    def test_len_list(self):
+        # Uses .elts
+        node = astroid.extract_node("""
+        len(['a','b','c'])
+        """)
+        node = next(node.infer())
```
PCManticore: Let's also verify in a test that we return a Const node.

astroid/tests/unittest_brain.py
```diff
@@ -1011,5 +1011,144 @@ def _get_result(code):
...
+
+    def test_len_failure_missing_variable(self):
+        node = astroid.extract_node("""
+        len(a)
+        """)
+        with pytest.raises(astroid.InferenceError):
+            next(node.infer())
+
+    def test_len_bytes(self):
+        """Use .length method to find bytes value"""
```
PCManticore: What's the `.length` method?

astroid/helpers.py
brycepg: I'll double check, but I think the StopIteration is coming from classes that don't have a __len__ method

astroid/tests/unittest_brain.py
brycepg: I think that was a typo, thanks for the catch

astroid/helpers.py
brycepg: More qname?

astroid/helpers.py
brycepg: yeah

astroid/helpers.py
brycepg: next raises `StopIteration` when the given iterator is empty

PCManticore: What I'm trying to say is that if someone does `next(node.infer_call_result())` that should either fail with InferenceError/return Uninferable or raise some other exception. It shouldn't fail with `StopIteration` explicitly, which usually shows a potential bug. If we put this in some other way, tools that use astroid, for instance pylint, should never care about catching `StopIteration` but only exceptions that astroid provide.

astroid/helpers.py
PCManticore: `qname()` could work, as long as it's clear what nodes are we handling here.

astroid/helpers.py
brycepg: Oh I see. infer_call_result shouldn't fail silently. I'm thinking it has to do with methods that are defined via the interpreter (written in C)

brycepg:
@PCManticore Heres a reproduction of the StopIteration issue

    import astroid
    code = """
    a = "fast"
    a
    """
    node = astroid.extract_node(code)
    lenmeth = next(node.inferred()[0].igetattr("__len__"))
    next(lenmeth.infer_call_result(None, None))

PCManticore:
Thanks for the reproduction example, Bryce! Given this is not introduced by this PR, there's no reason to be blocked by it. Let's just create an issue with it and it can be investigated in parallel.

astroid/helpers.py
brycepg: I think maybe in the future I could transform builtin methods to return correct values instead of being empty. Do you think this is easily do-able?

astroid/node_classes.py
```diff
@@ -2837,6 +2837,21 @@ def catch(self, exceptions): # pylint: disable=redefined-outer-name
         return False
 
 
+class FrozenSet(_BaseContainer):
```
PCManticore: Hey Bryce, this was actually in objects because there is no frozenset AST node in Python, if I recall correctly. Can you check and revert this change if so?

astroid/helpers.py
PCManticore: Sorry I don't understand what you mean :)

astroid/node_classes.py
brycepg: I mean, there's no Instance node in the AST as well

brycepg:
@PCManticore I mean I would like to transform `str.__len__` and make it return `self.value`. Is there a way to transform methods like this?

astroid/node_classes.py
brycepg: Hm but `Instance` is in `Bases`

brycepg: @PCManticore I could put `FrozenSet` into its own file?

brycepg: Yeah I'll revert it

PCManticore:
If I understand correctly, you mean to apply this for the builtin methods? What would be the rationale? I'm trying to figure out what you're trying to accomplish. In any case, you could look into the `objectmodel` module, where we do this sort of things, but for this particular use case, I don't quite understand from where the `self.value` is coming from when you infer what `bytes.__len__` is returning, because after all, you already have an instance of `Const` when we detect a `bytes` or `dict` object. If you have a code example, even if it doesn't work, that would be great, since right now I don't quite get your intention with this.

Also, for reference, there is also a `#pylint` channel on IRC where I hang from time to time. If you have any questions that are not necessarily appropriate for an issue, you can also write there and I'll try to respond when I get a chance (might not be completely in sync due to timezone differences)

astroid/node_classes.py
PCManticore: Just wanted to reply to the deleted comments: yes, Instance is in `bases`, but that's because it's needed as a base class for some of the nodes. It's also a bit more generic than what we have in `objects`, since those classes are more specialised versions of an instance of *something*.
Regarding the circular import, in the 2.0 branch that we started some years ago with @ceridwen  there are some changes to alleviate a part of these circular dependencies, although not all of them were addressed. Once I'll finish what I have in store for 3.7 support, I'll focus on bringing those changes to master.

brycepg:
For example, lets say someone's python code calls `str.__len__("f")` or `"f'.__len__()`. Currently this call will fail even with my implementation of the `len` builtin since astroid can't access the C code that implements this length call. I'd like to instead rely astroid providing the machinery to determine the result.

I'd like

```
const = astroid.extract_node("""
a = "f"
a
""").inferred()[0]
bound_method = next(const.igetattr("__len__"))
next(bound_method.infer_call_result(const))
```
`infer_call_result` to yield `nodes.Const(1)` here instead of `StopIteration`

PCManticore:
Gotcha, thanks for the example. It's doable, but not in a simple way. You could take a look at `objectmodel`, where you could implement a custom `__len__` method, that just returns a specialised version per each implementation, which when called, can operate on the argument that was given to.  Here's an example for instance for how we implement the `__get__` method: https://github.com/PyCQA/astroid/blob/8425d915c0186f87da66ad6223588ce70c49031c/astroid/interpreter/objectmodel.py#L280. You might use this one as an inspiration.

PR #510, opened 2018-03-12
Make object.__new__(cls) in classmethod uninferable instead of incorrect

brycepg:
InferenceContext does not provide enough information to instantiate the
correct class for `object.__new__(cls)` inside a classmethod

This is a patch to fix potential false positives.

If this pull gets merged I'll open a new issue about contexts not providing enough stack information.

Close #404 
Close PyCQA/pylint#224


PCManticore:
@brycepg I'd rather prefer to have the fix for this bug instead of disabling the inference path for this particular use case.

PR #518, opened 2018-03-17
Fix call context propagation

### LINKED ISSUE #177: Call contexts aren't propagated properly

---

```
#!python

n = extract_node('''
def chest(a):
    return a * a

def best(a, b):
    return chest(a)

def test(a, b, c):
    return best(a, b)

test(4, 5, 6) #@
''')
print(n)
print(n.infered())
```

---

brycepg:
Fix inference of call arguments

Close #177
Close #404 
Close PyCQA/pylint#224
Close PyCQA/pylint#1660

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -519,6 +519,8 @@ def _is_not_implemented(const):
 def _invoke_binop_inference(instance, opnode, op, other, context, method_name):
     """Invoke binary operation inference on the given instance."""
     methods = dunder_lookup.lookup(instance, method_name)
+    if context is not None:
+        context.boundnode = instance
     method = methods[0]
     inferred = next(method.infer(context=context))
     return instance.infer_binary_op(opnode, op, other, context, inferred)
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -186,6 +186,7 @@ def infer_call(self, context=None):
     callcontext = context.clone()
     callcontext.callcontext = contextmod.CallContext(args=self.args,
                                                      keywords=self.keywords)
+    _set_explicit_context_for_call(self, context.clone())
     callcontext.boundnode = None
     for callee in self.func.infer(context):
         if callee is util.Uninferable:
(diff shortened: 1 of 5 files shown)


brycepg:
Note: `pytest -k test_numpy_crash` took multiple minutes to complete for me locally.

brycepg:
I may have to do some performance tuning before this gets accepted

brycepg:
A problem I'm seeing is that arguments with multiple inferences cause the function result to become `Uninferable`. Do you think I should cover this case?

brycepg:
I think I'm going to try to change nodes to allow an explicit context, which will override regular inference contexts to avoid Inference too soon

brycepg:
This is causing a recursion error in home-assistant/home-assistant. I'll have to fix that
Integration tests are already paying off :)

brycepg:
I think I'm going to rewrite this

astroid/inference.py
```diff
@@ -518,6 +519,8 @@ def _is_not_implemented(const):
 def _invoke_binop_inference(instance, opnode, op, other, context, method_name):
     """Invoke binary operation inference on the given instance."""
     methods = dunder_lookup.lookup(instance, method_name)
+    if context is not None:
```
PCManticore: I think this could definitely be in another PR.

astroid/inference.py
```diff
@@ -832,3 +835,17 @@ def instance_getitem(self, index, context=None):
             node=self, index=index, context=context))
 
 bases.Instance.getitem = instance_getitem
+
+
+def _set_explicit_context_for_call(call, context):
```
PCManticore: One problem I have with this is that it mutates the underlying nodes with a context. While we do that for the explicit inference, I'd like us to start considering the nodes as immutable, with as few as possible places where we'd have to mutate them (right now another one apart of `explicit_inference` would be mutating `.locals`). It would make more sense to push these values inside the context itself, after which we could modify the inference to look for the nested values of a context when inferring values. This way we keep nodes untouched, we only pass the modified context.

astroid/inference.py
brycepg: I'll see if this passes any of the tests in a new PR. I think it depends on the explicit context

PCManticore:
Closed as we already merged the improvement based on a new parameter.

PR #519, opened 2018-03-17
Making Pylint faster 2

nickdrozd:
My [last PR](https://github.com/PyCQA/astroid/pull/497) contained what I would describe as compiler-style
optimizations, mainly loop unrolling and getting rid of runtime type
checking and dynamic attribute handling. I applied those changes more
or less mechanically, without really understanding the purpose of all
the functions involved.

In spite of those improvements, the [last yappi graph](https://user-images.githubusercontent.com/17630138/36357123-f391482a-14be-11e8-9266-c48b5eb17140.png) still showed
major hotspots in `_get_return_nodes_skip_functions` and
`_get_yield_nodes_skip_lambdas` (and also, to a lesser extent, in
`_get_assign_nodes`). Thinking about how those functions were being
used, I realized that a lot of unnecessary work was still being done.
Consider `return` statements, for instance. Most node types (e.g.
function calls, assign statements, comprehensions) cannot legally
contain them, so if we're searching for `return` statements, we can
pass those nodes by immediately. The only nodes that can contain
`return` statements are nodes with multi-line bodies, like `If` and
`FunctionDef`, and thus they are the only ones that need to implement
`_get_return_nodes_skip_functions`. Similar reasoning holds for the
other changes here. (I'm sure there is plenty of cleverness to be applied
further.)

To test these changes, I ran `pylint` (with `pylint`'s `pylintrc`
file) against `pycodestyle.py`, as before, as well as against `pylint`
itself. These cases differ in two import ways: 1) `pylint` is a
multi-file project with nested directories, while `pycodestyle.py` is
just a single file, and 2) `pylint` is written to pass `pylint`
without any errors, while `pycodestyle.py` raises lots of `pylint`
errors.

Despite their differences, `pylint` ran about twenty seconds faster in
both cases:

# pycodestyle.py

## Before

    real	0m55.362s
    user	0m24.856s
    sys	0m30.283s
    
    real	0m54.343s
    user	0m24.282s
    sys	0m29.905s
    
    real	0m54.996s
    user	0m24.679s
    sys	0m30.169s

## After

    real	0m34.826s
    user	0m16.159s
    sys	0m18.539s
    
    real	0m35.294s
    user	0m16.434s
    sys	0m18.723s
    
    real	0m35.089s
    user	0m16.328s
    sys	0m18.658s

# pylint

## Before

    real	3m34.382s
    user	1m38.719s
    sys	1m55.555s
    
    real	3m34.185s
    user	1m38.867s
    sys	1m55.192s
    
    real	3m34.986s
    user	1m38.878s
    sys	1m55.964s

## After

    real	3m13.449s
    user	1m29.708s
    sys	1m43.619s
    
    real	3m13.627s
    user	1m29.739s
    sys	1m43.785s
    
    real	3m13.680s
    user	1m29.919s
    sys	1m43.661s

Now, as these optimizations become increasingly specialized, there
comes a risk that the code will become increasingly inelegant.
Certainly some elegance has already been sacrificed by replacing
`nodes_of_class` with type-specific variations. But I think that
`pylint`'s slowness is a major usability issue, especially for large
projects, and therefore the tradeoff is definitely worth it.

### DIFF
--- a/astroid/mixins.py
+++ b/astroid/mixins.py
@@ -131,3 +131,38 @@ class ImportFromMixin(FilterStmtsMixin):
         raise exceptions.AttributeInferenceError(
             'Could not find original name for {attribute} in {target!r}',
             target=self, attribute=asname)
+
+
+class MultiLineBlockMixin:
+    """Mixin for nodes with multi-line blocks, e.g. For and FunctionDef.
+    Note that this does not apply to every node with a `body` field.
+    For instance, an If node has a multi-line body, but the body of an
+    IfExpr is not multi-line, and hence cannot contain Return nodes,
+    Assign nodes, etc.
+    """
+
+    @decorators.cachedproperty
+    def _multi_line_blocks(self):
+        return tuple(
+            getattr(self, field)
+            for field in self._multi_line_block_fields
+        )
+
+    def _get_return_nodes_skip_functions(self):
+        for block in self._multi_line_blocks:
+            for child_node in block:
+                if child_node.is_function:
+                    continue
+                yield from child_node._get_return_nodes_skip_functions()
+
+    def _get_yield_nodes_skip_lambdas(self):
+        for block in self._multi_line_blocks:
+            for child_node in block:
+                if child_node.is_lambda:
+                    continue
+                yield from child_node._get_yield_nodes_skip_lambdas()
+
+    def _get_assign_nodes(self):
+        for block in self._multi_line_blocks:
+            for child_node in block:
+                yield from child_node._get_assign_nodes()
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -654,11 +654,7 @@ class NodeNG(object):
         yield from ()
 
     def _get_yield_nodes_skip_lambdas(self):
-        for child_node in self.get_children():
-            if child_node.is_lambda:
-                continue
-            for matching in child_node._get_yield_nodes_skip_lambdas():
-                yield matching
+        yield from ()
 
     def _infer_name(self, frame, name):
         # overridden for ImportFrom, Import, Global, TryExcept and Arguments
@@ -2708,6 +2704,10 @@ class Expr(Statement):
     def get_children(self):
         yield self.value
 
+    def _get_yield_nodes_skip_lambdas(self):
+        if not self.value.is_lambda:
+            yield from self.value._get_yield_nodes_skip_lambdas()
+
 
 class Ellipsis(NodeNG): # pylint: disable=redefined-builtin
     """Class representing an :class:`ast.Ellipsis` node.
@@ -2833,6 +2833,12 @@ class ExceptHandler(mixins.AssignTypeMixin, Statement):
             if child_node.is_function:
                 continue
             yield from child_node._get_return_nodes_skip_functions()
+
+    def _get_yield_nodes_skip_lambdas(self):
+        for child_node in self.body:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
 
 
 class Exec(Statement):
@@ -2994,6 +3000,17 @@ class For(mixins.BlockRangeMixIn, mixins.AssignTypeMixin, Statement):
             if child_node.is_function:
                 continue
             yield from child_node._get_return_nodes_skip_functions()
+
+    def _get_yield_nodes_skip_lambdas(self):
+        for child_node in self.body:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
+
+        for child_node in self.orelse:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
 
 
 class AsyncFor(For):
@@ -3285,6 +3302,17 @@ class If(mixins.BlockRangeMixIn, Statement):
             if child_node.is_function:
                 continue
             yield from child_node._get_return_nodes_skip_functions()
+
+    def _get_yield_nodes_skip_lambdas(self):
+        for child_node in self.body:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
+
+        for child_node in self.orelse:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
 
 
 class IfExp(NodeNG):
@@ -4017,6 +4045,17 @@ class TryExcept(mixins.BlockRangeMixIn, Statement):
             for matching in child_node._get_return_nodes_skip_functions():
                 yield matching
 
+    def _get_yield_nodes_skip_lambdas(self):
+        for child_node in self.body:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
+
+        for child_node in self.orelse:
+            if child_node.is_lambda:
+                continue
+            yield from child_node._get_yield_nodes_skip_lambdas()
+
 
 class TryFinally(mixins.BlockRangeMixIn, Statement):
     """Class representing an :class:`ast.TryFinally` node.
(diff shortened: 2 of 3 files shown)


astroid/node_classes.py
```diff
@@ -3252,6 +3285,35 @@ def get_children(self):
         yield from self.body
         yield from self.orelse
 
+    def _get_assign_nodes(self):
```
brycepg: A lot of these functions are copy-pasted between these classes. Do you think we could use a Mixin class to prevent code duplication?

astroid/scoped_nodes.py
```diff
@@ -2701,6 +2716,16 @@ def get_children(self):
         if self.decorators is not None:
             yield self.decorators
 
+    def _get_assign_nodes(self):
+        for child_node in self.body:
+            yield from child_node._get_assign_nodes()
+
+    def _get_yield_nodes_skip_lambdas(self):
```
brycepg: Do you think it would make sense to make some of these duplicated functions into Mixins? I see `_get_yield_nodes_skip_lamdas` is placed exactly the same here twice

PCManticore:
Thanks for tackling this @nickdrozd ! My main concern is, as you said, that these changes starts to sacrifice the readability of the code for the performance, with the code becoming extremely specialized and too clever for its sake. We should try to strike a balance between these two, so some degree of code generalisation should be kept around. 

I have two points for discussion for this PR. As Bryce mentioned this patch introduces some duplicate methods which are also a bit hard to understand without having the context of why they were implemented like this (for new contributors for instance). I suggest we should also keep them separated from the node implementations themselves, since right now performance tricks mingle with the rest of the code. I don't have something in particular in mind, maybe we can implement these in, say, something like `astroid.node_specialisation` or something along these lines? I imagine for now these methods could be part of mixins or even specialized functions per each kind of node. Also this code has to be heavily documented on why a function is implemented in one way or another.

The other point is that even with these specialisations, as mentioned in your other PR, I think that us not having a way to verify in an automated way that the performance is not hurt, through some new feature or change we're adding, will definitely be a disadvantage. We should invest some effort into this before coming up with new specialisations, since we're not going to run yappi ourselves periodically.

A third point that came in my mind during writing this comment is more of a question really. I wonder if it's worth looking into rewriting some of the astroid's stack in a more efficient language rather than  doing it all in Python. After a point, there won't be a lot of specialized changes that could be applied to get a significant boost in performance. I know @ceridwen also thought about that at some point. If we could rewrite the AST in C (maybe a CPython extension or Cython, something along these lines), we could introduce run these tree functions from C rather than from pure Python. This could lead to the biggest performance gain with the effect that the code could be kept super simple and not overly specialized. Unfortunately I definitely don't have time for a big project like this one.

ceridwen:
Something that I would like to do (but I also don't have time for big projects at this point) is to create a corpus of Python code for testing astroid against, or least run pytest on all the standard library modules to make sure it doesn't crash.  The same corpus could be used for profiling/performance testing.  As it is, we have two problems.  First, all the benchmarks that anyone has ever run astroid/pylint against are awfully specific, which is fine for users who are really only concerned how fast astroid is on their particular projects but not good for astroid development because we don't know if optimizing for one project might hurt performance in others.  Second, we don't have any continuous integration set up for performance regressions so any future changes in the code base could easily slow down astroid without their authors realizing it.

As for this patch in particular, have we dropped support for Python 2.7?  If so, someone should really update the readme and PyPi tags because they still claim support for 2.7.  (Dropping support for 2.7 before it officially EOLs in 2020 seems like a bad idea to me, but I won't reopen that if it's decided.). Beyond that, like @brycepg says, there's a lot of code duplication that could probably be reduced, at least, with mixins.

PCManticore:
I agree with @ceridwen on the fact that we need performance regressions.

@ceridwen regarding Python 2 support, we decided to drop it in https://github.com/PyCQA/pylint/issues/1763. This is restricted to the future major releases, pylint 2.0 and astroid 2.0 We'll still release bug fixes for the current astroid 1.6 and pylint 1.8 until Python 2 is EOL, so probably until January 2020.

brycepg:
One problem I've found with CI performance regression tests is that Appveyor/Travis CI are not made to provide performance consistency: [Benchmarks](https://bheisler.github.io/post/benchmarking-in-the-cloud/), [travis CI issue](https://github.com/travis-ci/travis-ci/issues/352)  (plus they would be too slow, the CI is already nearly unbearably slow)

I think It wouldn't be too difficult for me to set up a folder in the project for integration tests and curate some number of projects to run pylint against using `pytest-benchmark`, which could then be run locally on a developer's machine (which would need some performance tuning such as setting the CPU governor for consistency) when needed, with an automated setup step for downloading the project to run the benchmarks.

I'll open a new issue for performance integration tests, but I don't think they should block this issue

brycepg:
I'd say this change generally makes code faster (definitely not slower):
```
----------------------------- benchmark: 14 tests ------------------------------
Name (time in s)                        Mean            StdDev            Rounds
--------------------------------------------------------------------------------
test_pycodestyle (0004_nickdro)       6.5699 (1.0)      0.6506 (1.0)           3
test_pycodestyle (0003_master)        8.7538 (1.33)     0.9468 (1.46)          3
test_yapf (0004_nickdro)              9.7190 (1.48)     1.0315 (1.59)          3
test_yapf (0003_master)              10.0176 (1.52)     0.9679 (1.49)          3
test_pyfunctional (0004_nickdro)     10.9815 (1.67)     1.2713 (1.95)          3
test_pyfunctional (0003_master)      10.8861 (1.66)     2.1725 (3.34)          3
test_utilisnips (0004_nickdro)       10.2468 (1.56)     0.7364 (1.13)          3
test_utilisnips (0003_master)        10.6300 (1.62)     0.8745 (1.34)          3
test_requests (0004_nickdro)         14.8844 (2.27)     1.5987 (2.46)          3
test_pgcli (0004_nickdro)            15.1486 (2.31)     1.5961 (2.45)          3
test_requests (0003_master)          15.8045 (2.41)     1.8531 (2.85)          3
test_pgcli (0003_master)             16.3026 (2.48)     1.6357 (2.51)          3
test_lektor (0004_nickdro)           27.9670 (4.26)     3.4107 (5.24)          3
test_lektor (0003_master)            29.9270 (4.56)     4.0418 (6.21)          3
--------------------------------------------------------------------------------
```
The corpus is here:
https://github.com/brycepg/pylint-corpus

PCManticore:
Thanks for tackling the performance benchmarks Bryce! Regarding the output, what are the values from before and after this patch? Would be great if you'd offer a small breakdown on those values.

Regarding the PR itself, this is not blocked by the performance benchmarks, but by my other point regarding the specialization, separation & code duplication. I'm waiting on @nickdrozd to follow up. Overall I think the patch and its improvements are great to have, just needs some polishing before getting it in.

brycepg:
`0004_nickdro` is nickroid's branch
`0003_master` is the master branch
`test_<repo>` is the benchmark for a python repository (from most downloaded, or most starred on github with a pylintrc). pylint is run three times on each repository and the mean time in seconds is taken. (lower being better).

It appears `pycodestyle` is the most optimized out of all the repos in terms of percent difference, which makes sense since that's where nick took the call graphs.  Only the `pyfunctional` didn't see improvements (both results were really close).

I unfortunately do not have access to a dedicated server for proper performance testing and these measurements were taken in a Paravirtual machine.

nickdrozd:
Okay, I figured out what I think is a reasonably elegant way to consolidate all that repeated code, the `MultiLineBlockMixin`. In fact you're all witnesses to history, as this is the first mixin I've written in my entire career.

@PCManticore It would certainly be nice if a bird flew by and dropped Pylint-rewritten-in-C into our hands, but short of that I doubt it will ever happen. And is it really needed anyway? I think there are still plenty of performance gains to be had just from changes to the existing Python code.

PCManticore:
@nickdrozd Definitely, this would be a huge project for which none of us has time nor the incentive to do it. Although it would be interesting to think about it if we're aiming for pylint to become as fast as pyflakes & co with regard to linting. That is, even if we optimize pylint as much as we can, it will never be as fast as the likes of pyflakes & co.

astroid/mixins.py
```diff
@@ -131,3 +131,41 @@ def real_name(self, asname):
         raise exceptions.AttributeInferenceError(
             'Could not find original name for {attribute} in {target!r}',
             target=self, attribute=asname)
+
+
+class MultiLineBlockMixin:
+    def _get_multi_line_blocks(self):
```
PCManticore: We could use a `decorators.cachedproperty` here.

astroid/mixins.py
```diff
@@ -131,3 +131,41 @@ def real_name(self, asname):
         raise exceptions.AttributeInferenceError(
             'Could not find original name for {attribute} in {target!r}',
             target=self, attribute=asname)
+
+
+class MultiLineBlockMixin:
```
PCManticore: Can you add a docstring on why we have this mixin?

nickdrozd:
@PCManticore done

PCManticore:
Thanks @nickdrozd ! Looking forward to the next improvements! :)

PR #525, opened 2018-03-26
Update Unknown Node

### LINKED ISSUE #523: Reconsider the usefulness of Unknown
Pylint is failing on `pandas` because of `Unknown` not having a name attribute.
```
    def visit_call(self, node):
        """Visit a Call node."""
        try:
            for inferred in node.func.infer():
                if inferred is astroid.Uninferable:
                    continue
>               if inferred.root().name == OPEN_MODULE:
E               AttributeError: 'Unknown' object has no attribute 'name'

../pylint/pylint/checkers/stdlib.py:209: AttributeError
```

brycepg:
- Add name attribute to Unknown class
 This allows .root().name calls to work for Unknown objects
 - add qname to Unknown node to avoid error in pylint

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -4421,6 +4421,10 @@ class Unknown(mixins.AssignTypeMixin, NodeNG):
     introspection failed.
     """
     name = "Unknown"
+
+    def qname(self):
+        return "Unknown"
+
     def infer(self, context=None, **kwargs):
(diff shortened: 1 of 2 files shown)


PCManticore:
Sounds good, but check my comment from #523

brycepg:
Alright, I'll merge this and keep #523 open

PR #526, opened 2018-03-29
Fix calls for object.__new__(cls) inside a classmethod

brycepg:
Like #518 but doesn't modify nodes. It instead uses an additional argument to determine the correct inference

Close #404
Close PyCQA/pylint#224
Close PyCQA/pylint#1660

If this gets accepted I'll try to use `context_lookup` to finish off #177

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -139,6 +139,18 @@ Change log for the astroid package (used to be astng)
 
       Close #472
 
+    * Fix object.__new__(cls) calls in classmethods by using
+        a context which has the proper boundnode for the given
+        argument
+
+        Close #404
+
+    * Fix Pathlib type inference
+
+        Close PyCQA/pylint#224
+        Close PyCQA/pylint#1660
+
+
 2017-06-03 -- 1.5.3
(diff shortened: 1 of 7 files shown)


ChangeLog
```diff
@@ -129,6 +129,16 @@ Change log for the astroid package (used to be astng)
 
       Close #472
 
+    * Fix object.__new__(cls) calls in classmethods
```
PCManticore: Maybe add some extra details on how this is fixed.

astroid/bases.py
```diff
@@ -302,14 +302,20 @@ def igetattr(self, name, context=None):
             return iter((self.special_attributes.lookup(name), ))
         return self._proxied.igetattr(name, context)
 
-    def infer_call_result(self, caller, context):
+    def infer_call_result(self, caller, context, context_lookup=None):
```
PCManticore: Can you add a short docstring on how context_lookup is used?

brycepg:
Thanks for the review! Fixed

PR #531, opened 2018-04-03
Further use of context_lookup to fix nested calls

### LINKED ISSUE #177: Call contexts aren't propagated properly

brycepg:
Deliver the correct context with the correct boundnode
for function argument nodes

Close #177

### DIFF
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -23,14 +23,20 @@ class CallSite(object):
     In order to infer what an argument represents, call
     :meth:`infer_argument` with the corresponding function node
     and the argument name.
+
+    The context/call_context are mainly for unpacking Starred nodes
     """
 
-    def __init__(self, callcontext):
+    def __init__(self, callcontext, context_lookup=None):
+        if context_lookup is None:
+            context_lookup = {}
+        self.context_lookup = context_lookup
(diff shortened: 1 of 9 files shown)


PCManticore:
Thanks for the PR Bryce, and sorry for taking so long to have a look. I am a bit worried with these new changes since they introduce a new parameter to all the inference functions, only to handle a couple of edge cases in the inference. I wonder though if it is possible to move the parameter at the context level. e.g instead of passing it to every function, we can add an extra attribute to the inference context, `extra_context` or something more suggestive. Later on, instead of looking for `extra_context` as we do now, we look in the context itself to get the corresponding extra data. This has the nice effect that we only have to modify the context where we need, e.g in `infer_call` if I recall, and the additional data gets propagated automatically down the line. What do you think?

brycepg:
I think that might work. I'll give it a try at some point on a separate PR

brycepg:
I wasn't able to get this working as an attribute for context. I might try again sometime

PR #534, opened 2018-04-11
fix StopIteration raising for python3.7

HoverHell:
NOTE: I'm not suggesting this as a proper fix; mostly putting it out there in case anyone runs into the same problem.

The error this is trying to fix is `RuntimeError: generator raised StopIteration`

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -95,13 +95,12 @@ def path_wrapper(func):
                 if ares not in yielded:
                     yield res
                     yielded.add(ares)
-        except (exceptions.MyStopIteration, StopIteration) as error:
+        except StopIteration as error:
             # Explicit StopIteration to return error information, see
             # comment in raise_if_nothing_inferred.
             if error.args:
-                raise exceptions.MyStopIteration(error.args[0])
-            else:
-                return
+                return error.args[0]
+            return
 
     return wrapped
 
@@ -141,7 +140,7 @@ def raise_if_nothing_inferred(func, instance, args, kwargs):
         while True:
             yield next(generator)
             inferred = True
-    except (exceptions.MyStopIteration, StopIteration) as error:
+    except StopIteration as error:
         if not inferred:
             if error.args:
                 # pylint: disable=not-a-mapping
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -95,13 +95,13 @@ def path_wrapper(func):
                 if ares not in yielded:
                     yield res
                     yielded.add(ares)
-        except StopIteration as error:
+        except (exceptions.MyStopIteration, StopIteration) as error:
             # Explicit StopIteration to return error information, see
             # comment in raise_if_nothing_inferred.
             if error.args:
-                raise StopIteration(error.args[0])
+                raise exceptions.MyStopIteration(error.args[0])
             else:
-                raise StopIteration
+                return
 
     return wrapped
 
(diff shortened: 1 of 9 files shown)


PCManticore:
Hey @HoverHell thanks for the PR! I think we could just return the corresponding value instead of raising `StopIteration` or a custom exception. If I recall correctly, the return value gets pushed in the `StopIteration` exception, so we'd only have to catch that instead for getting the additional data.

Also thanks for tackling this!

HoverHell:
> the return value gets pushed in the StopIteration exception

Right, I see py2.7 support was dropped a month ago. In that case it would indeed be a viable solution.

PCManticore:
Yup, the master is 3+ only. We're planning to have support for 3.7, this change would be super great to have as right now the inference doesn't work at all.

HoverHell:
Rewrote the change to use the py3.3+ feature

HoverHell:
There are still lots of cases where a generator does `next()` on another generator. What is the recommended way of handling those? `try: ... except StopIteration: return`?

brycepg:
Slightly relevant: How has Travis CI been passing with 3.7-dev if inference is completely broken? Just looking at the last commit on master seems to be passing: https://github.com/PyCQA/astroid/commits/master

astroid/decorators.py
```diff
@@ -99,9 +99,8 @@ def wrapped(node, context=None, _func=func, **kwargs):
             # Explicit StopIteration to return error information, see
             # comment in raise_if_nothing_inferred.
             if error.args:
-                raise StopIteration(error.args[0])
-            else:
-                raise StopIteration
+                return error.args[0]
+            return
```
brycepg: The tests seem to be passing on 3.7 with these changes. The only thing that is failing is some linting errors:
`return None` here

HoverHell:
> How has Travis CI been passing with 3.7-dev if inference is completely broken by raising StopIteration

I suspect it is using an earlier version of py3.7 branch which didn't have that change yet. Although that would be a months-earlier version, which is probably suspicious.

astroid/decorators.py
PCManticore: Yup, as mentioned, I think we run on an "old" version of 3.7 on Travis, probably alpha 1 or so. I have it locally and the beta 1 seems to have this change, so this could explain why the tests are passing on Travis. From a short look, it doesn't seem that there is an update on Travis.

PCManticore:
@HoverHell Regarding the `next()` question, your approach sounds good to me! Thanks for tackling this PR! It looks pretty good! Let me know once you add the `next()` guards and we can merge this.

HoverHell:
> Let me know once you add the next() guards

Added a few.

HoverHell:
Just to dump the barely relevant results here:

One of the remaining errors in py3.7 tests (worthy of a different pull request) is `AssertionError: '_py_abc.ABCMeta' != 'abc.ABCMeta'`; as far as I've found out, the `lib/python3.7/abc.py` contains two relevant pieces:
1. `from _abc import ...` appears to be ignored (assumed to raise ImportError),
2. The ABCMeta object is made by`from _py_abc import ABCMeta, get_cache_token` and `ABCMeta.__module__ = 'abc'`; however, `metaclass.qname()` returns `'_py_abc.ABCMeta'`, ignoring the `.__module__ = ...` hack.
What was the status of supporting the `.__module__ = ...` hack, in general?

HoverHell:
bump.

PCManticore:
Hey @HoverHell thank you so much for the patch! 
Regarding your quetions:
1. I presume so, but I think the test might be accomodated to support the correct module name. We do support extension packages as well, is `_ast` is builtin we should be able to construct an AST from it.
2. What's the `__module__` hack?

HoverHell:
> What's the __module__ hack?

    >>> class A:
    ...     pass
    ... 
    >>> A
    <class '__main__.A'>
    >>> A.__module__ = 'stuff'
    >>> A
    <class 'stuff.A'>
    >>>

PR #548, opened 2018-05-20
[WIP] Add support for type comments

PCManticore:
### Fixes / new features
Uses `typed_ast` to get type comments into the AST as new attributes to a couple of nodes. We can then use these attributes in the corresponding `visit` methods in PyLint in order to not warn about unused imports.

### DIFF
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -6,6 +6,7 @@
 # For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
 
 """astroid packaging information"""
+import platform
 
 distname = 'astroid'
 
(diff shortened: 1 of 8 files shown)


astroid/__pkginfo__.py
```diff
@@ -19,7 +19,8 @@
     'lazy_object_proxy',
     'six',
     'wrapt',
-    'typing;python_version<"3.5"'
+    'typing;python_version<"3.5"',
+    'typed_ast;python_version<"3.7"',
```
PCManticore: Need to look into how to make it work on PyPy

astroid/rebuilder.py
```diff
@@ -11,13 +11,16 @@
 """
 
 import sys
-import _ast
+import typed_ast as _ast
+
+# TODO: allow for picking Python 2
```
PCManticore: We also want to allow parsing Python 2 code while we're on Python 3.

gvanrossum:
Thank you!

PR #559, opened 2018-06-14
Add cPickle to the list of available imports for six.moves in Python2

brycepg:
close PyCQA/pylint#2185


PCManticore:
Thanks Bryce!

PR #560, opened 2018-06-14
Changes for running pylint without __init__ file

mzfr:
Related to [PR #2190](https://github.com/PyCQA/pylint/pull/2190) on pylint

I think this won't require any other changes since `get_module_files` is directly called from the `expand_module` function in `/pylint/utils.py`.

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -442,7 +442,7 @@ def get_module_part(dotted_name, context_file=None):
     return dotted_name
 
 
-def get_module_files(src_directory, blacklist, list_all=False):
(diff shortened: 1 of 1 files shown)


mzfr:
@Razzeee @PCManticore The tests are failing because in `tests/unittest_modutils.py` [Line#258](https://github.com/PyCQA/astroid/blob/master/astroid/tests/unittest_modutils.py#L258) we'll have to pass another argument i.e `lint_all=True`

mzfr:
@Razzeee I was thinking that once we are sure with the changes and everything else is finalized I'll make the sufficient changes in all the required tests.

PR #561, opened 2018-06-15
Reworking astroid numpy brain

hippo91:
### Fixes / new features
- This PR is a complete reworking of the ```numpy.core.numerictypes``` part of the ```numpy``` brain.
I used the inspect module to analyze deeply the structure and the class hierarchy of the module.
- It should fix quite a lot of bugs dealing with the ```numpy numerictypes``` including pyCQA/pylint#2140
- @PCManticore i thought it was a good idea to store the module i designed and used to analyze the ```numpy.core.numerictypes```, that's why i created the '''utils""" directory. It's maybe not a good idea neither the right place to do that. What do you think about it?


hippo91:
@PCManticore maybe i have been misunderstood.
In the ```numpy``` brain i do not import anything. The brain consist only in a string.
However to build this string, i used ```inspect``` and ```numpy.core.numerictypes``` modules and the purpose of the file in ```utils/numpy_types_analysis``` is to help with the inspection of ```numpy.core.numerictypes``` through the use of ```inspect```.

PCManticore:
@hippo91 ah, yes, you are right, I didn't notice you are not importing numpy in the brain file. But I'm not sure it is worth adding the inspect part as well, as it will technically be dead code until we run it again to get updated brain tips.

hippo91:
@PCManticore you are right for the file in ```utils```, i will keep it in my own repository.
Should i make another PR with just the brain and unittest associated?

PCManticore:
Thank you for working on this @hippo91 !

PR #567, opened 2018-06-21
Prevent Const copy resulting in recursion

### LINKED ISSUE #565: RecursionError when copying constants
### Steps to reproduce
Run
```
copy.copy(astroid.nodes.Const(1))
```

### Current version

cdd9756b0f9ac270a2ca365ecde972d47a891e94

brycepg:
Calling ``object.__new__`` on a Const node would result in infinite recursion

Fix infinite recursion by explicitly raising AttributeError if
``__getattr__`` is called for value

See https://nedbatchelder.com/blog/201010/surprising_getattr_recursion.html
for more details

Close #565


astroid/raw_building.py
```diff
@@ -391,9 +391,6 @@ def _astroid_bootstrapping(astroid_builtin=None):
 
 _astroid_bootstrapping()
 
-# TODO : find a nicer way to handle this situation;
```
PCManticore: Why was this TODO removed, is this the same bug?

astroid/node_classes.py
```diff
@@ -2295,6 +2295,17 @@ def __init__(self, value, lineno=None, col_offset=None, parent=None):
 
         super(Const, self).__init__(lineno, col_offset, parent)
 
+    def __getattr__(self, name):
+        # This is needed becuase of Proxy's __getattr__ method.
```
PCManticore: Typo here for `because`

astroid/bases.py
```diff
@@ -71,7 +71,13 @@ def _is_property(meth):
 
 
 class Proxy(object):
-    """a simple proxy object"""
+    """a simple proxy object
+
+    Note:
+
+    Subclasses of this object will need a custom __getattr__
```
PCManticore: I wonder if we should try https://github.com/PyCQA/astroid/issues/205 to get rid of this `Proxy` implementation instead.

astroid/raw_building.py
brycepg: I thought so because of the semicolon referencing the recursion issue on the following line

brycepg:
Thanks for the review

I'll take a look at #205

Issue #331, opened 2016-04-19
as_string() is still buggy


ceridwen:
Running

```python
ast = parse(source_file.read())
ast == ast # True
ast == parse(ast.as_string()) # False
```

with the commented-out equality-checking code for ASTs in base.py enabled indicates that as_string() isn't preserving the invariant it should.  I've also tried parsing the same file twice and comparing those ASTs and that evaluates to True, so the problem is probably in as_string() rather than the equality code.

brycepg:
I'm seeing that whitespace is being stripped from `as_string`:

\>`parse("\n\n\n\n").as_string()`
`'\n\n'`

ceridwen:
as_string() in general doesn't preserve whitespace because it's operating on ASTs, not concrete syntax trees.  The invariant that should be preserved is that if you convert an existing AST to code and then back to an AST, you should get the same AST back.  If it's processing whitespace in a way that breaks Python semantics, that's a bug that's possibly related to this issue.

brycepg:
Oh I see what you mean.

`NodeNG` doesn't overload the equality operator so the `ast == parse(ast.as_string())` *shouldn't* be true, since those `ast`s are not in the same memory location.

However I am seeing that as_string isn't immediately convergent which may be an issue:
```
`ast.as_string() == parse(ast.as_string()).as_string()` is `False`.

Here are some examples diffing the two as_strings from pycodestyle:

```
2110,2112c1971,1973
<                 Find all globally visible functions where the first argument name
<                 starts with argument_name and which contain selected tests.
<                 """
---
>             Find all globally visible functions where the first argument name
>             starts with argument_name and which contain selected tests.
>             """
2116c1977
<             if any(((not code) and (self.ignore_code(code)) for code in codes)):
---
>             if any((not (code) and (self.ignore_code(code)) for code in codes)):
```
```
<         self.options.report = (reporter) or (self.options.reporter(self.options))
---
>         self.options.report = (reporter) or (self.options.reporter)(self.options)
```

```
1254,1257c1118,1119
<     E713: if not X.B in Y:
<     pass
<     E714: if not X is Y:
<     pass
```

I'm wondering if this has to do with astpeephole being fussy

ceridwen:
You're right, the correct form with respect to the current code is
`ast.as_string() == parse(ast.as_string()).as_string()`.  I actually
defined an equality operation on the nodes that checked for AST
equivalence rather than memory location for what I did.

[...]

PR #569, opened 2018-06-22
Improve as_string output of operators, elif, with, return & docs

brendanator:
### Fixes / new features
- The precedence and associativity rules of operators are respected and parens are only wrapped around child nodes when needed
- A single If node inside the else block is output as `elif`
- Unneccesary parens in with statements are removed
- Unneccesary parens in tuple returns are removed
- Doc strings in classes and functions no longer get additional indenting


brycepg:
Closes #331 I think

brycepg:
I've looked this over multiple times @PCManticore, and I don't think there's anything wrong with this PR. `as_string` is close to as good as it can be. Can I merge?

PCManticore:
I have one concern about importing node_classes in as_string, which creates a cyclical dependency that wasn't there before. I want to review this before getting the PR in, but I'm currently travelling for work so I can't dedicate a lot of time to it.

brycepg:
I did see that.

My thoughts: The class type needs to be inspected to determine precedence to see whether given parenthesis are needed to retain the code's original meaning since `BinOp`, `BoolOp`, and `UnaryOp` have difference precedence depending on their `op` value. A way around a circular import I can see is giving the node classes the ability to determine their precedence via a new method on `NodeNG` which gives a numerical precedence kind of like `__cmp__`

PCManticore:
@brycepg Yes, that approach could work in removing this cyclical dependency!

astroid/as_string.py
```diff
@@ -15,9 +15,33 @@
 """
 import sys
 
+from astroid import node_classes
```
PCManticore: @brycepg mentioned how this improvement might work without importing `node_classes` here, by providing a precedence method for the base `NodeNG` class which can then be implemented by the rest of nodes that you are interested in.

But now I see that we also use `node_classes` for other use cases as well, such as to see if the first node from `orelse` is an `If` node. Instead of doing a type check against the class itself, we prefer doing a `.__class__.__name__` check instead. There's a third option to get past this cyclical dependency, which is to use the `lazy_import` function from `astroid.util`, but let's see if the first two solutions work before that.

astroid/tests/testdata/python2/data/module.py
```diff
@@ -45,17 +46,17 @@ def __init__(self):
...
         try:
             MY_DICT = {}
             local = None
             autre = [a for (a, b) in MY_DICT if b]
             if b in autre:
                 return b
-            else:
-                if a in autre:
-                    return a
+            elif a in autre:
```
PCManticore: This is a good improvement.

astroid/tests/testdata/python2/data/module.py
```diff
@@ -81,9 +82,9 @@ def four_args(a, b, c, d):
     else:
         b += -2
     if c:
-        d = ((a) and (b)) or (c)
+        d = a and (b or c)
     else:
-        c = ((a) and (b)) or (d)
+        c = a and b or d
```
PCManticore: This is so much better!

astroid/tests/unittest_nodes.py
```diff
@@ -173,6 +173,26 @@ def test_slice_and_subscripts(self):
         ast = abuilder.string_build(code)
         self.assertEqual(ast.as_string(), code)
 
+    def test_operator_precedence(self):
+        with open(resources.find('data/operator_precedence.py')) as f:
+            for code in f.readlines():
```
PCManticore: You can drop the `readlines` code, just iterate over the file with `for code in f`.

astroid/tests/unittest_nodes.py
```diff
@@ -173,6 +173,26 @@ def test_slice_and_subscripts(self):
...
+        Check that as_string produces source code with exactly the same
+        semantics as the source it was originally parsed from
+        """
+        pre = builder.parse(code)
+        post = builder.parse(pre.as_string())
+
+        pre_repr = pre.repr_tree()
+        post_repr = post.repr_tree()
+
+        assert pre_repr == post_repr
```
PCManticore: That's a good assertion to have, thanks for adding it.

brendanator:
Cool, will have a look. Pretty sure I can do this without introducing nasty `node.__class__.__name__ == 'If'` checks

brendanator:
I think the CI failure in pypy already exists in master as I can reproduce it locally

Other than that I think this PR is ready

brendanator:
Great stuff. Thanks to both of you for your review

brycepg:
Oh note: This is going to change some of the expected messages in pylint tests

PCManticore:
Yup, we already have some test failures on master but should be easy to fix.

PR #572, opened 2018-06-26
Cut legacy Python syntax nodes

nickdrozd:
In the change from Python 2 to Python 3, `print` and `exec` were
demoted from keywords to functions, while repr backticks were
eliminated entirely.


PCManticore:
Let's leave these in for now. We want to have a mode that will run `pylint` over Python 2 using `typed_ast.py27` parser, but didn't get to do that yet.

PR #576, opened 2018-06-29
Fix issue with inherited __call__ improperly inferencing self

brycepg:
Instance infer_call_results needed boundnode set for its context.

Close PyCQA/pylint#2199


brycepg:
Thanks for the reminder

PR #578, opened 2018-06-30
Fix __call__ precedence for classes with custom metaclasses

brycepg:
Call the dunder call method if available from a custom metaclass

Note: I had to give ClassDef metaclass lookup a context to prevent a
Recursion

Close PyCQA/pylint#2159

### DIFF
--- a/astroid/scoped_nodes.py
+++ b/astroid/scoped_nodes.py
@@ -2036,7 +2036,27 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG,
                 and len(caller.args) == 3):
             result = self._infer_type_call(caller, context)
             yield result
+            return
+
+        dunder_call = None
+        try:
+            metaclass = self.metaclass(context=context)
+            if metaclass is not None:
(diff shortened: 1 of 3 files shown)


astroid/scoped_nodes.py
```diff
@@ -2036,7 +2036,27 @@ def infer_call_result(self, caller, context=None, context_lookup=None):
...
+        dunder_call = None
+        try:
+            metaclass = self.metaclass(context=context)
+            if metaclass is not None:
+                dunder_call = next(metaclass.igetattr("__call__", context))
+        except exceptions.AttributeInferenceError:
+            pass
+        if (dunder_call is not None and
+                dunder_call.qname() != "builtins.type.__call__"):
+            if context is not None:
```
PCManticore: We have this pattern in a couple of places, what do you think about making it a function? It can receive a `context` and a node to bound, and returns the said context or a new one but with the `boundnode` set to that given node?

brycepg: Yeah I've noticed that as well. Sounds good

brycepg:
Made the changes. Is this what you had in mind?

PR #582, opened 2018-07-04
Nested call args2

### LINKED ISSUE #177: Call contexts aren't propagated properly

brycepg:
Add context_lookup to the context class as `extra_context`

Deliver the correct context with the correct boundnode
for function argument nodes

Close #177

This removes use of `context_lookup` as per your previous recommendation @PCManticore 

---

I had to make some changes to `type.__new__` tests because my changes caused them to fail. These tests were faulty anyway.

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -376,8 +376,7 @@ class BoundMethod(UnboundMethod):
 
         In order for such call to be valid, the metaclass needs to be
         a subtype of ``type``, the name needs to be a string, the bases
-        needs to be a tuple of classes and the attributes a dictionary
-        of strings to values.
+        needs to be a tuple of classes
         """
         from astroid import node_classes
         # Verify the metaclass
@@ -419,13 +418,10 @@ class BoundMethod(UnboundMethod):
         for key, value in attrs.items:
             key = next(key.infer(context=context))
             value = next(value.infer(context=context))
-            if key.__class__.__name__ != 'Const':
-                # Something invalid as an attribute.
-                return None
-            if not isinstance(key.value, str):
-                # Not a proper attribute.
-                return None
-            cls_locals[key.value].append(value)
+            # Ignore non string keys
+            if (key.__class__.__name__ == 'Const' and
+                    isinstance(key.value, str)):
+                cls_locals[key.value].append(value)
 
         # Build the class from now.
         cls = mcs.__class__(name=name.value, lineno=caller.lineno,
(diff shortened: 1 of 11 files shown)


astroid/bases.py
```diff
@@ -424,13 +420,10 @@ def _infer_type_new_call(self, caller, context):
...
             key = next(key.infer(context=context))
             value = next(value.infer(context=context))
-            if key.__class__.__name__ != 'Const':
-                # Something invalid as an attribute.
-                return None
-            if not isinstance(key.value, str):
-                # Not a proper attribute.
-                return None
-            cls_locals[key.value].append(value)
+            if key.__class__.__name__ == 'Const':
```
PCManticore: This change will add in `locals` keys that are integers, which is the reason why the original code was there.

astroid/tests/unittest_inference.py
```diff
@@ -3309,6 +3309,8 @@ class Entity(object):
         self.assertIsInstance(attributes[0], nodes.Const)
         self.assertEqual(attributes[0].value, 1)
 
+
+    @pytest.mark.xfail(reason="These call should be uninferable because they are invalid calls")
```
PCManticore: Do we still expect the test to fail even if we modified the test to check for an exception instead?

astroid/tests/unittest_inference.py
```diff
@@ -4616,5 +4625,41 @@ class Sub(Base):
     assert val.name == "Sub"
 
 
+class TestInferencePropagation:
```
PCManticore: This is a great improvement!

astroid/bases.py
brycepg: Hmm ` type.__new__(type, 'a', (), {1:2, "a":5})` is a valid class in `CPython` but returns an error in `pypy`. Do you think we should disallow non string keys regardless?

astroid/tests/unittest_inference.py
brycepg: It's more of a `TODO` thing that I can fix in another issue where these tests should be raising errors because they are invalid `type.__new__` calls. The previous asserts were testing for incorrect behavior

astroid/bases.py
PCManticore: Gotcha, it's great that `pypy` errors on this, since if you do add integers on CPython, then your class will be broken anyway:
```
----> 1 dir(a)

TypeError: '<' not supported between instances of 'str' and 'int'
```

Let's just skip those, so instead of returning, don't add those attributes.

Issue #583, opened 2018-07-04
astroid.modutils.load_module_from_file failing on path when using multiple symlinks


mariocj89:
### Steps to reproduce
1. Create a folder and symlinked it to another one. Create a module within it.
```
mkdir /tmp/deployment
ln -s /tmp/deployment /tmp/path_to_include
touch /tmp/secret.py
ln -s /tmp/secret.py /tmp/deployment/secret.py
```
2. Validate ```  PYTHONPATH=/tmp/path_to_include python3.6 -c "import secret" ``` works.
3. Astroid wont be able to load it:
```
PYTHONPATH=/tmp/path_to_include python3.6 
>>> import astroid
>>> astroid.modutils.load_module_from_file('/tmp/path_to_include/secret.py')
```

Note `/tmp/path_to_include/secret.py` exists and we are passing `/tmp/path_to_include` in the path, but as the module resolves it to realpath `load_module_from_file` wont work.

### Current behavior

Astroid is unable to load a module it the python path where it is present is on is a symlink (the one on sys.path, not the file path itself)

### Expected behaviour

To be able to import it the same way you would with python.

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output

2.0.0.dev3

### version information:

Present in the latest versions of 1.6 and 1.5.
Works in the latest version of 1.4: 1.4.9

Related issue: https://github.com/PyCQA/astroid/issues/538

Related commit: https://github.com/PyCQA/astroid/commit/788bb41c9eb1078237724933e3d33ca98419c41f

This manifests as well for the `extra-path` argument on the same function in modutils.

Proposal: Lookup on both realpath and abspath the same way #538 was resolved.

mariocj89:
I’m trying to put a test case through and will send a PR with the proposal unless you think it is not worth fixing or not a real issue.

mariocj89:
Fixed via #584

PR #584, opened 2018-07-04
Include abspath of sys.path on searches at load_from_module

mariocj89:
At the moment `load_from_module` is resolving all paths in `sys.path` to validate that the module being imported is in a folder part of the path. This is done to resolve symlinks.

This can be an issue though it both the file being imported and the path in `sys.path` resolve to different locations. Even if the symlinks are importable, the resolution of the paths in `sys.path` might not.

The fix just makes that when checking if a file is within one of the existing paths in sys.path or extrapath to check both before and after expanding the symlinks.

### Fixes / new features
- #583

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,12 @@ What's New in astroid 2.0?
 
 Release Date: Unknown
 
+   * Fix loading files with `modutils.load_from_module` when
+     the path that contains it in `sys.path` is a symlink and
+     the file is contained in a symlinked folder.
+
+     Close #583
+
    * Reworking of the numpy brain dealing with numerictypes
      (use of inspect module to determine the class hierarchy of
       numpy.core.numerictypes module)


astroid/tests/unittest_modutils.py
```diff
@@ -132,6 +133,37 @@ def test_import_symlink_both_outside_of_path(self):
...
+        real_secret_path = os.path.join(tmp, 'secret.py')
+        symlink_secret_path = os.path.join(path_to_include, 'secret.py')
+
+        # setup double symlink
+        # /tmp/deployment
+        # /tmp/path_to_include (symlink to /tmp/deployment)
+        # /tmp/secret.py
+        # /tmp/deployment/secret.py (points to /tmp/secret.py)
+        os.mkdir(deployment_path)
+        self.addCleanup(shutil.rmtree, deployment_path)
```
PCManticore: Nice use of `addCleanup` here.

mariocj89:
Let me know how https://github.com/PyCQA/astroid/pull/584/commits/8fabcf7e34afc0c9f90c2efbe7d22f15991f17b9 looks. Happy to make any changes.

For further context, this manifested as we had our python site-packages path as a symlink and the pylint plugin folder was a symlink within the realpath as well.

Thanks a lot! I know it is quite a "peculiar" case :P

Issue #465, opened 2017-11-05
node_classes.py has too many lines


AWhetter:
I am about to merge #458 but it is failing the pylint check because the docstrings have made the module too big. I didn't want to start moving things around so for now I have disabled the message, until we figure out how we want things laid out.

PCManticore:
We could definitely start splitting it up, but I don't mind having the error disabled for now.

brycepg:
What do you think about making `nodes` a subpackage, moving `nodes.py` into `nodes/__init__.py`
Move `node_classes` and `scope_nodes` into this subpackage. For important nodes like `NodeNG` we could single files per class e.g. `nodes/nodeng.py` , and move some of the assist functions into a `nodes/util.py`.
This would help with organization, split up large files, and prepare for #329

PCManticore:
I wouldn't mind some of these refactorings, @brycepg although I'm not sure we'll ever do #329

PCManticore:
I think this is fine as it is right now, as per my reasoning from #587. If the file would grow larger than it is right now, I'd say we should split it, but not the nodes themselves, rather any functionality that doesn't necessarily belong there.

Pierre-Sassoulas:
Reopening because of the argument here : https://github.com/PyCQA/astroid/pull/587#issuecomment-403326307

PR #587, opened 2018-07-07
Move NodeNG to its own file

brycepg:
NodeNG is getting really big. Time to refactor node_classes.py

If this is accepted, then I will continue to split up `node_classes.py` for #465


PCManticore:
After seeing what this refactoring would entail, I honestly think this kind of refactoring is going to generate an unreasonable amount of churn, especially resulting in more confusing git blames. I'm not entirely sure this is actually worth doing it. Is that file even that big, considering that a big chunk of its size are the actual docstrings? If I recall, it was around 2000 or so lines, which is kind of big, but not necessarily unmanageable. There's also the question on what exactly would be break out after we get `NodeNG` to its own file. Would we move every node into its own file? That's going to become unmanageable as well, since now we'd have to jump to one of the 30-40 new files in order to look up a node, instead of looking it up in a single place. I don't think this makes sense, as it feels a bit a refactoring for the refactoring's sake.

brycepg:
possible pros to refactoring node_classes.py:

* Easier keep track of what method belongs to what class
* Related classes could be logically grouped together (List, Tuple, Set, go in `container_nodes.py` or `nodes/containers.py`)

Pierre-Sassoulas:
I'm convinced by @brycepg argument, the node_classes and scope_nodes are unmanageable right now and in particular fixing the circular import is a nightmare without a clear understanding of what needs what and literally hundreds of import per file.

Pierre-Sassoulas:
Closing because I did not realize it was a MR and not an issue. Rebasing would cost much more than just redoing it.

PR #591, opened 2018-07-09
Fix attributes of ancestor classes from appearing in attribute inference

### LINKED ISSUE #581: type.__new__ attribute infers to both type.__new__ and object.__new__
Example:

```python
import astroid
print(astroid.extract_node("""
type.__new__
""").inferred())
```

Results in `[<BoundMethod __new__ of builtins.type at 0x126350295329928, <BoundMethod __new__ of builtins.object at 0x126350078306680]`
The above code should only infer to BoundMethod of `type.__new__`

This is causing some issues and hacks in the code base including https://github.com/PyCQA/astroid/blob/master/astroid/bases.py#L452 I think

brycepg:
Close #581

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -343,9 +343,7 @@ class UnboundMethod(Proxy):
         # instance of the class given as first argument.
         if (self._proxied.name == '__new__' and
                 self._proxied.parent.frame().qname() == '%s.object' % BUILTINS):
-            # XXX Avoid issue with type.__new__ inference.
-            # https://github.com/PyCQA/astroid/issues/581
-            if caller.args and len(caller.args) == 1:
+            if caller.args:
                 node_context = context.extra_context.get(caller.args[0])
                 infer = caller.args[0].infer(context=node_context)
             else:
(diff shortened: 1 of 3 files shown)


brycepg:
I found this change caused an error in pylint tests, but I found that it was a false-positive that was previously hidden by a safe_infer, (after this change the result was only one value) so I'll fix that test if this gets merged.

I can't just get the first value of igetattr in infer_attribute because igetattr does some other work that isn't done by `getattr(...).infer(context)` (a valid test will fail). Plus there isn't really any apparent reason to look at masked attributes given by getattr.

PCManticore:
The approach looks sane but I wonder if it's going to inhibit any of pylint's inference capabilities. But since only one test fails, we can probably find a solution.

brycepg:
I was thinking of adding a bool parameter on `igetattr` toggling whether to take all values of `getattr` instead or just the first. Might still do that.

The only way I think this could limit inference capabilities is if the getattr mro isn't perfect (and the first value of getattr isn't actually the first value), since getting all the inferred ancestors isn't something that I think is a very common use case.

Fixing up getattr/igetattr is something on my short order TODO list anyway

Issue #585, opened 2018-07-05
`NodeNG.get_children` yields nodes in inconsistent order


brendanator:
### Steps to reproduce
1. Write function with args, default args, annotations, varargs, kwargs
2. Parse and call get_children

### Current behavior
Nodes are yielded in this order
- args
- defaults
- kwonlyargs
- kw_defaults
- annotations
- varargannotation
- kwargannotation
- kwonlyargs_annotations

### Expected behavior
- Nodes are yielded in order they appear in source code
- Varargs and kwargs are also yielded

### Other example
`IfExp.get_children` yields `test` node before `body` node

### Possible Solution
I would expect all fields that aren't identifiers in the [python3 AST](https://docs.python.org/3/library/ast.html#abstract-grammar) to be yielded in the order they appear in the source code

It seems as though astroid nodes are based on [python2.7 AST](https://docs.python.org/2.7/library/ast.html#abstract-grammar) with additional information from the later version added. I think it would be better to base the astroid nodes on the latest python AST and get the builder code to convert from python2.7 AST. This solution means that more information is retained in the AST and cleanly solves this issue

I may provide a PR for this but want to check:
- are there any problems with this solution
- what needs looking out for when implementing this?
- how would other libraries be affected?
- is this likely to be merged?

nickdrozd:
I messed with `get_children` in https://github.com/PyCQA/astroid/commit/52e67656d7728f8ba958c9c35bef3a44013ad6dd, and https://github.com/PyCQA/astroid/commit/228e34db1fdc4935df0d3d63f24e1e484dddf7e4 made some corrections to address your issue. I don't know if it fixed all of it.

That said, there does not seem to be a requirement that the child nodes get yielded in any particular order:
```
    def get_children(self):
        """Get the child nodes below this node.

        :returns: The children.
        :rtype: iterable(NodeNG)
        """
```
What is the use case for requiring that the children are yielded "in the order they appear in the source code"? Why not yield them alphabetically (e.g. `body`, `orelse`, `test` for `While` nodes)? I have no objection to changing the order to match some convention, but making a particular order a requirement seems a little odd to me.

PCManticore:
As @nickdrozd I'm also not sure we were promoting `get_children` as traversing the tree in any given order, but we can address this if it doesn't make sense right now, as long as it is feasible.

[...]

PR #595, opened 2018-07-16
Ensure NodeNG.get_children yield nodes in source code order

brendanator:
### Fixes
- In Arguments, an argument, its annotation and its default are yielded before the next argument
- In IfExp, the body nodes is yielded before the test node

Fixes the ordering issue from #585

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -399,7 +399,7 @@ class NodeNG:
     def get_children(self):
         """Get the child nodes below this node.
 
-        :returns: The children.
+        :returns: The children in order they appear in the source code.
         :rtype: iterable(NodeNG)
         """
         for field in self._astroid_fields:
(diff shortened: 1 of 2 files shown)


nickdrozd:
I have no objection to changing the order in `IfExp.get_children`, but
the change in `Arguments.get_children` looks suspicious to me. In
general, `get_children` implementations are not elegantly written, but
that's because the function is a hotspot and the loops have been
unrolled in the name of speed. Adding the complexity of `_yield_args`
could slow things down, and we need to be careful to avoid that.

But how about this for a compromise? Add a new function
`get_children_in_order`. In most cases, this would just be
`get_children`:

```
class NodeNG:
	def get_children(self):
		'''Yield children in unspecified order'''
		# ...

	def get_children_in_order(self):
		'''Yield children in order they appear in the code'''
		# in most cases the children already appear in order
		yield from self.get_children()


class Return(NodeNG):
	def get_children(self):
		if self.value is not None:
			yield self.value

	# inherits get_children_in_order from NodeNG
```

But with `Arguments`, the implementation would be more particular:
```
class Arguments(NodeNG):
	def get_children(self):
		# as it is now

	def get_children_in_order(self):
		# new implementation with _yield_args
```

This way, we don't need to mess with the spec for `get_children`, and
implementations of `get_children_in_order` can be altered without
concern for general slowdown.

brendanator:
You are right @nickdrozd 

I profiled it and there is definitely some overhead from `_yield_args`. I tried 3 versions
* Original - baseline
* `_yield_args` version (the current PR) - 2.88 times as long to execute `get_children`
* Inline the yield args calls - 1.71 times as long to execute `get_children`

I'm happy to inline the calls if that's what's wanted.

It seems like adding a new function `get_children_in_order` would clutter the API a bit needlessly to avoid a smallish slowdown in a single node class, after all I could just monkey patch it to get the functionality I need.

I can do whichever though so let me know

PCManticore:
I agree with @nickdrozd on this one. astroid and pylint are so slow that performance should trump most of the other initiatives, as long as those initiatives are not related to furthering the capabilities of astroid itself (e.g. inference or control flow). 
The question now is why it's so important to have the iteration done in the order of the definition. Do you have a particular case that the current behaviour is impeding or it's more of an irk? I feel that a `get_children_in_order` could make sense but I'm not sure if it's so worth having without an use case.

brendanator:
OK, doesn't seem like it's worth it then

PR #598, opened 2018-07-20
Stop repeat inference attempt causing a RuntimeError in Python3.7

brycepg:
Empty generators and next calls within a generator without a default argument
now cause a cascade effect which results in a RuntimeError.

Raise InferenceError instead of return None to avoid the above problem.

Close PyCQA/pylint#2317

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 2.0.2?
 
 Release Date: |TBA|
 
+   * Stop repeat inference attempt causing a RuntimeError in Python3.7
+
+     Close PyCQA/pylint#2317
+
 
 What's New in astroid 2.0.1?
 ============================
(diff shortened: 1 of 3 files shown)


brycepg:
Duplicate CI tasks because I accidentally created a branch on PyCQA/astroid and then deleted

PCManticore:
I'll check this PR these days, thanks! My main concern is why we now need to raise an inference error.

brycepg:
because returning an empty generator would cause `next` calls to raise stopiteration which causes a runtimeerror in 3.7. The alternative would be to catch StopIteration on every `next` call inside of a generator

PCManticore:
@brycepg The reason I was concerned is that `path_wrapper` should not care about the fact that in its upper call sites we're raising InferenceError to propagate an inference issue down to the consumers. But from a short look it seems that the corresponding inference methods weren't decorated at all with `raise_if_nothing_inferred`, which should deal with this problem at a more general level. I've pushed https://github.com/PyCQA/astroid/commit/8841891b2337c8166450cc84c1e503e22d49daa7 which seems to fix the original problem from `pylint`. I merged your PR locally to get the text without the raising of the InferenceError.

brycepg:
Alright, sounds good

PR #602, opened 2018-07-25
Use final Python 3.7 in Travis

scop:
Requiring sudo: true is some kind of (temporary?) anomaly: https://github.com/travis-ci/travis-ci/issues/9815


PCManticore:
Thanks!

PR #603, opened 2018-07-26
Cut obsolete "explicit StopIteration" comments

nickdrozd:
The explicit StopIterations were themselves were cut in ceeee097.

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -99,8 +99,6 @@ def path_wrapper(func):
                     yield res
                     yielded.add(ares)
         except StopIteration as error:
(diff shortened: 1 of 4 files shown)


PCManticore:
Thanks for doing this! Regarding the returning inside generators with error information, the value that either gets returned or the value that's passed to `StopIteration` when raised gets extracted and added into the `InferenceError` for inference functions decorated with `raise_if_nothing_is_inferred`. This propagates any particular reason why the inference stopped up to a handler that cares, but usually the consumers of those exceptions hide them behind `safe_infer` et al. It's more of a mechanism to debug `astroid` by astroid developers rather than a public API.

PR #615, opened 2018-09-20
brain_gi: use Exception as a base class for classes inheriting from Exception

lazka:
All gi classes use object as a base which leads to pylint complaining in
case a class which is really an exception is used in a try/except construct.
See PyCQA/pylint#2504

In case the object is a subclass of Exception use Exception as a base instead
of object to make pylint happy.

This fixes pylint warnings with pygobject code such as:
    E0712: Catching an exception which doesn't inherit from Exception: GError (catching-non-exception)

Close PyCQA/pylint#2504


PCManticore:
Great stuff, @lazka Thanks for the sleuthing!

PR #624, opened 2018-11-06
Cut generator flags

nickdrozd:
This simplifies the code, and may slightly improve performance.


PCManticore:
@nickdrozd Looks good! Can we get the CI green before merging?

nickdrozd:
What's the current method for formatting?

PCManticore:
We use `black` for that with `pre-commit` hooks, so the following should do it:

- install ``pre-commit`` using ``pip install pre-commit``
- then run ``pre-commit install`` in the ``astroid`` root directory to enable the git hooks.

PCManticore:
@nickdrozd I merged this locally, thanks for the PR!

nickdrozd:
Thanks! Btw, I've been out of the Python world for a while, but I'm back in it now, so hopefully I'll be contributing more.

PR #629, opened 2018-12-11
Extend detection of classes defined with attr

gavento:
## Steps

- [x] Add a ChangeLog entry describing what your PR does
- [x] Write a good description on what the PR does

## Description

Extends detection of classes defined with attr to `attr.attrs`, `attrs`, `attr.attrib` and `attrib`.
Extends the `test_attr_transform` test.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #628

PR #631, opened 2018-12-11
Remove else after return/raise to make pylint happy

gavento:
## Steps

- Not adding a changelog entry - only cosmetic changes
- [x] Write a good description on what the PR does

## Description

The PR removes one `else` and changes `elif` to `if` to address the following pylint errors in travis. No functional change, the code should be equivalent in all cases to the old one.

```
************* Module astroid.decorators
astroid/decorators.py:132:12: R1720: Unnecessary "else" after "raise" (no-else-raise)
************* Module astroid.rebuilder
astroid/rebuilder.py:977:8: R1705: Unnecessary "elif" after "return" (no-else-return)
```

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -132,7 +132,6 @@ def raise_if_nothing_inferred(func, instance, args, kwargs):
             if error.args:
                 # pylint: disable=not-a-mapping
                 raise exceptions.InferenceError(**error.args[0])
-            else:
-                raise exceptions.InferenceError(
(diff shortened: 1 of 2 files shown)


gavento:
@PCManticore this fixes the pylint issues in travis that I hit in #629.

PCManticore:
@gavento thank you so much for fixing this!

PR #635, opened 2019-01-12
Correcting commit 74d79ef

hippo91:
# Description

This PR is here to fix the pb arising after commit 74d79ef. This commit breaks the unittest `InferenceTest.test_two_parents_from_same_module`.

Proceeding by bisection, i revert the strictly necessary to have all unittest ok.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |


PCManticore:
I don't think this is the right fix, left a comment on why is that.

PR #637, opened 2019-01-13
Proof of Concept on using a global inference context

PCManticore:
It's currently not working as expecting and lots of tests are crashing, including `pylint` but this should be the ideal solution on getting `pylint` fast and with improved inference.

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -84,21 +84,21 @@ def path_wrapper(func):
     def wrapped(node, context=None, _func=func, **kwargs):
         """wrapper function handling context"""
         if context is None:
-            context = contextmod.InferenceContext()
-        if context.push(node):
+            context = contextmod.global_context
+
+        execution = context.push(node)
+        if execution.is_recursing():
             return None
 
-        yielded = set()
-        generator = _func(node, context, **kwargs)
-        for res in generator:
-            # unproxy only true instance, not const, tuple, dict...
-            if res.__class__.__name__ == "Instance":
-                ares = res._proxied
-            else:
-                ares = res
-            if ares not in yielded:
+        execution.increment()
+        inferred = execution.inferred
+        if inferred:
+            yield from list(inferred)
+        else:
+            generator = _func(node, context, **kwargs)
+            for res in generator:
                 yield res
-                yielded.add(ares)
+                inferred.add(res)
 
     return wrapped
 
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -90,21 +90,15 @@ def path_wrapper(func):
 
         yielded = set()
         generator = _func(node, context, **kwargs)
-        try:
-            while True:
-                res = next(generator)
-                # unproxy only true instance, not const, tuple, dict...
-                if res.__class__.__name__ == "Instance":
-                    ares = res._proxied
-                else:
-                    ares = res
-                if ares not in yielded:
-                    yield res
-                    yielded.add(ares)
-        except StopIteration as error:
-            if error.args:
-                return error.args[0]
-            return None
+        for res in generator:
+            # unproxy only true instance, not const, tuple, dict...
+            if res.__class__.__name__ == "Instance":
+                ares = res._proxied
+            else:
+                ares = res
+            if ares not in yielded:
+                yield res
+                yielded.add(ares)
 
     return wrapped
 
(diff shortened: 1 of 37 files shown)


kodonnell:
I assume you're just using a global while testing it out? To me this seems to be inviting trouble (e.g. forgetting to `clear` - as you had to do to make the tests work).

Also - what's the status of this? I wouldn't mind helping out, but as per #670, I'm still not fully understanding what's going on with inference etc. However, if you think this is 99% there, and just needs some debugging, then I might run with it. If it's just a PoC to be thrown away, then I wont.

PCManticore:
@kodonnell The inference unfortunately depends on some sort of global context to be useful in practice. Right now the context is passed for every inference call, but this has some side effects:

- it's not very performant. We're instantiating an `InferenceContext` probably thousands of time during linting a relatively small file, and for large projects this can mean hundreds of thousands of instantiations. I've got these numbers while trying to figure out why a simple pandas script was taking minutes to lint, and the high number of instantiations was pretty obvious there. Similar to the instantiations is the `path` copying. The `path` shows through where the inference went just to avoid inferring an object a second time. I'll explain in a next bullet point why a global context might alleviate this. The `path` needs to diverge in certain points of the inference (e.g. branching), which means that we need to copy it. But turns out that when you have thousands of instantiations, copying it all the time  results in significant performance penalties, and this was in fact the major culprit in the aforementioned pandas issue. This PR proposes some junction points in which we can copy the path, which reduces the amount of copying that was needed, but it's still not ideal.

- contexts cannot actually be reused.  Here's a long standing issue https://github.com/PyCQA/astroid/issues/108 with some examples, but the idea is that because of the way contexts were implemented so far, if you infer a second time passing the context that was already used once for inference, you'll get an `InferenceError` instead of getting the actual object that was inferred the first time. This means contexts are not that useful right now from the point of view of performance & inference as well, as we'd have to pass fresh contexts thus beating the purpose of having the contexts in the first place. This one of the main reasons why there is a global context in this PR, so that we can keep track of what was inferred already, even as a side effect this means an increased memory consumption. If we'd do that, there's also a chance of getting rid of various recursion errors that are plaguing pylint.

Regarding the PR status, it was mostly done when I stopped working on it, but there were a couple of test failures here in `astroid` and a bunch more in `pylint`. I haven't been able to pick it up again, but if you want to help, that would be great. Sadly it's been a while since the PR is up so it most likely diverged from `master` but with a clean rebase should be ready to be picked up again.

kodonnell:
Thanks @PCManticore. As per #670 I'm still not 100% on how the inference works, but since this is hopefully just a case of updating to master, and then fixing bugs, it may be something I can handle - and maybe I'll figure out more in the process (and add documents for #670!). That said, I'm thinking about toying with parso instead (and you know, a minor thing like converting pylint to using it ... !), so we'll see what happens.

FYI, what I meant was instead of using a module level global (which e.g. may cause cache collisions if you run inference multiple times on different bits of code?), I'd leave it to the user to manage i.e. they create their own `InferenceContext` to be used as the 'global' one, and this then gets passed around etc. (Though ... this behaviour was automatic in the old approach anyway ... so maybe only the reduced branching will help.) Anyway, I can always test this out if I continue the PR.

astroid/decorators.py
```diff
@@ -84,28 +84,22 @@ def path_wrapper(func):
...
 
+        execution.increment()
+        inferred = execution.inferred
+        if inferred:
+            yield from list(inferred)
+        else:
+            generator = _func(node, context, **kwargs)
+            for res in generator:
+                yield res
+                inferred.add(res)
```
kodonnell: I might be wrong here ... but say "generator" has length 5. If `next` is only called once, then `execution.inferred` will only have 1 element, not 5 - which will then be reused next time as a single item iterator, and the remaining 4 items will be forever lost. Then again ... I've only got 3 failed tests, and fixing this doesn't seem to help the one I'm focusing on.

kodonnell:
OK, I tried to rebase here: https://github.com/kodonnell/astroid/tree/global_context

Only 3 tests failing. I've spent a couple of hours on the first one: `test_augassign_recursion` i.e.

```python
code = """
def rec():
  a = 0
  a += rec()
  return a
rec()
"""
cls_node = extract_node(code)
print(next(cls_node.infer()))
```
Currently it's inferring const 0 as opposed to uninferrable. I've messed round with this a bunch, and you know what seems to 'fix' it? (Not sure about breaking other things.) See change line 789 below.

image

Likewise if you write a manual crossproduct with two for loops. I discovered this by just trying to see what `lhs` and `rhs` were - and normally `rhs` (i.e. `rec()`) is const 0 (so somehow it's skipping all the recursion stuff). I'm not sure why (maybe testing to see all values of `rhs_iter` - which when you iterate through, correctly recurses), but I came across the above 'fix'. Now, it's really weird, because this should be the first step of `itertools.product` right? Put another way, how can `itertools.product` not be triggering this? Maybe [this](https://bugs.python.org/issue10109)? Not an exact fit, but similar ish ... though I didn't see OOM.

Anyway, sorry if I'm barking up the wrong tree, but I figured I'd post it here anyway.

degustaf:
Since `test_augassign_recursion` is marked as xfail in the master branch, would it make sense to mark this as xfail since it doesn't seem to actually be related to making the context global?

kodonnell:
I thought so, except [it was removed in this PR](https://github.com/PyCQA/astroid/pull/637/files#diff-f3bac76dbbae7c2bb2321ce5729fdeeeL4767), and I assume there was a reason. This PR has introduced a new way of handling recursion, which may be why.

kodonnell:
I should also have been more clear - it looks like the test *should* pass, if it weren't for the weird behaviour above.

PR #643, opened 2019-02-09
Bug pylint 2436

hippo91:
## Steps

- [Ok ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [Ok] Write a good description on what the PR does.

## Description
This PR solve the bug reported as PyCQA/pylint#2436.
In fact two problems were present. 
The first one was a lack of unary operators definition for some numeric types in `brain_numpy`. Thus the inference of such operators failed and a message `invalid-unary-operand-type` was emitted.

The second one was a wrong inference of call to some `numpy` functions. A `List` or `Tuple`instance was present in the inference generator. It has been corrected by adding some transformations in the `brain_numpy` module. Such transformation creates a closure in which the call to classical inference method is made and then filtered to remove undesirable types.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Close PyCQA/pylint#2436


hippo91:
@PCManticore, no pb i'll do a PR just for the first point.
For the second however i'wonder how it will be possible to suppress spurious message such as `E1130: bad operand type for unary -: list (invalid-unary-operand-type)` without filtering `List` or `Tuple` from inference results.
Moreover, from the `numpy` documentation, i understand, but i'm maybe wrong, that those function (`array` and `linspace`) return `ndarray` and not another type.

hippo91:
@PCManticore , the PR for the first part is available under #645.

PCManticore:
@hippo91 Gotcha, that makes sense. I approved the PR, let's fix the conflicts and merge it.

hippo91:
@PCManticore i fixed the conflicts. Only two unit tests are failing but they are identical to the ones that failed in previous commits. I merged the branch.

PR #648, opened 2019-02-24
Added 2 more missing _ssl constants

S3ntinelX:
## Description
Adds 2 more missing _ssl constants which were introduced in Python 3.6.
These are designed to depreceate the protocols on the line above.

(Sorry if this isn't done right it is my first pull request)

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |


PCManticore:
Thanks for your PR @S3ntinelX !

Issue #529, opened 2018-03-30
Improving inference caching


brycepg:
From investigating the code I'm getting the impression that context `cache_generator` isn't working very well for caching, and may have some low hanging fruit for improving performance.

Possible improvements:

- Allow partial caching. Currently an inference value is only cached if the inference is completely exhausted. Since there are many instances of only retrieving the first value of the inference, caching is ignored because there is no partial caching.
- Centralize caching. Currently caching is tied to the context. Since pylint does many disparate inferences travelling the inference tree, this may have a significant speedup. One possible downside would be a larger memory-speed tradeoff.
- Have [context.push](https://github.com/PyCQA/astroid/blob/master/astroid/decorators.py#L82) attempt to find inference results in the cache. This could potentially reduce the amount of `context.clones` required to stop inference from failing due to already traveled values in the context path.

PCManticore:
These all sound good to me, but I would move the third point into its own issue, as I don't see it a trivial  low hanging fruit. We'd also have to monitor that with these changes we don't get a too much increase in the used memory, but with your dataset this should be relatively easy to figure out.

jacobtylerwalls:
> Allow partial caching. [Currently](https://github.com/PyCQA/astroid/blob/master/astroid/context.py#L101) an inference value is only cached if the inference is completely exhausted. Since there are many instances of only retrieving the first value of the inference, caching is ignored because there is no partial caching.

I looked into this. I don't see a way to do it without eagerly evaluating `_infer()` and moving away from the paradigm of generators.

```py
infer1 = node.infer()
next(infer1)  # yields first result, we cache it
infer2 = node.infer()
next(infer1)  # yield first result from cache
next(infer2)  # uh oh...
```
In that example, we would need to get the second result from an iterator that hasn't been evaluated yet. As far as I know, the best ways to do that with `islice` still involve evaluating the entire iterator, trading away the performance savings.

Also, I'm thinking that energy might be better spent moving away from `next(node.infer())` anyway, see #136.

> Have [context.push](https://github.com/PyCQA/astroid/blob/master/astroid/decorators.py#L82) attempt to find inference results in the cache. This could potentially reduce the amount of context.clones required to stop inference from failing due to already traveled values in the context path.

This is intriguing. Especially after #1009 and now #2257 I'm sure there are a lot of unnecessary context clones all over the place. Would be good to attack this with #670.

[...]

PR #654, opened 2019-03-09
Cut cache_generator

nickdrozd:
Is cache_generator doing anything? Cutting it doesn't break anything
as far as I can tell, and it Pylint even seems to go a little faster
without it.

### DIFF
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -105,17 +105,6 @@ class InferenceContext:
         clone.extra_context = self.extra_context
         return clone
 
-    def cache_generator(self, key, generator):
-        """Cache result of generator into dictionary
-
-        Used to cache inference results"""
-        results = []
-        for result in generator:
-            results.append(result)
-            yield result
-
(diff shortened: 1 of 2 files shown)


nickdrozd:
Someone else should try this on a large repo and see what happens.

PCManticore:
It's possible `cache_generator` is not having the performance impact we expected it to have, but having some tests with `pylint` on top of libraries like `numpy` or `pandas` could prove that.

kodonnell:
As per [here](https://github.com/PyCQA/astroid/issues/529#issue-310116375): 

> Allow partial caching. Currently an inference value is only cached if the inference is completely exhausted. Since there are many instances of only retrieving the first value of the inference, caching is ignored because there is no partial caching.

So - probably want to fix it (easy?) as opposed to getting rid of it - and hopefully that'll actually improve performance.

PCManticore:
Yeah, that makes sense, it is possible that addressing some of its flaws could result in a performance improvement albeit minor.

Issue #658, opened 2019-03-27
Extension modules incorrectly classified as belonging to stdlib.


markmcclain:
### Steps to reproduce
1. Install a platform specific extension.
2. Run pylint (errors on potentially unsafe extension being scanned)

### Current behavior
Extensions in the platform specific location are misclassified by is_standard_module() as being in the standard library because the path shares prefix with standard library. 

### Expected behavior
The extension should not be considered part of the stdlib.

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output
master

PR #659, opened 2019-03-27
make is_standard_module() to properly classify extensions

markmcclain:
Extensions can be installed in either a platform independent or specific
location[1]. The search is updated to include both paths. Previously, an
extension in the platform specific location would be misclassified because
only the independent location was considered and the platform specific location
shared a prefix with the standard lib.

[1] http://docs.python.org/3/distutils/apiref.html#distutils.sysconfig.get_python_lib

Fixes Issue: #658
Co-authored-by: Matt Story <s.matt.story@gmail.com>

## Steps

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |


PCManticore:
Thank you for the great patch @markmcclain !

PR #664, opened 2019-04-21
Bug pylint 2721

hippo91:
## Steps

## Description
This PR is deep rework of the `numpy` brain.
The original brain has been split into numerous one. Each `numpy` brain module corresponds to 
a `numpy`one. There are:

  * `brain_numpy_core_fromnumeric` => `numpy.core.fromnumeric`
  * `brain_numpy_core_function_base` => `numpy.core.function_base`
  * `brain_numpy_core_multiarray` => `numpy.core.multiarray`
  * `brain_numpy_core_numeric` => `numpy.core.numeric`
  * `brain_core_numerictypes` => `numpy.core.numerictypes`
  * `brain_core_umath` => `numpy.core.umath`
  * `brain_numpy_random_mtrand` => `numpy.random.mtrand`

Last but not least, the `numpy.ndarray` class has its corresponding `astroid`'s brain (`brain_numpy_ndarray`).

Each `astroid`'s brain has its own associated unit test. 

The major improvement, besides a clearer organisation, is the fact that each method of the `ndarray` class should now be correctly inferred. Especially those returning `ndarray` objects.

It has been made possible by ensuring that the return type of those methods is inferred as `ndarray` and nothing else. To achieve this, the `inference_tip` function has been used.

When a `numpy` callable may return two different kinds of objects, for example `bool` or `ndarray`, the choice to prefer `ndarray` has been made in the corresponding `astroid` brain method.

There is still work to do, especially correcting the return types of the `brain_numpy_random_mtrand` and `brain_numpy_core_numerictypes` modules. The last one should also be reworked to correct method signature of the `generic` class.
But this PR should strongly improve the `numpy` support for `pylint`.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes PyCQA/pylint#2747
Closes PyCQA/pylint#2721
Closes PyCQA/pylint#2326
Closes PyCQA/pylint#2021

Related to PyCQA/pylint#2767
Related to PyCQA/pylint#2694
Related to PyCQA/pylint#2784
Related to PyCQA/pylint#2759


hippo91:
This PR should also prevent infinite recursion for bug PyCQA/pylint#2865

willsALMANJ:
For this file:
```
import numpy as np
a = np.append([0], 0)
-a*1
```

I still see:

```
/home/wshanks/tmp/test.py:3:4: E1130: bad operand type for unary -: tuple (invalid-unary-operand-type)
/home/wshanks/tmp/test.py:3:4: E1130: bad operand type for unary -: list (invalid-unary-operand-type)
```

Should that have been fixed by this PR?

hippo91:
@willsALMANJ sorry for the delay.
The `append` function is not taken into account through `astroid numpy` brains. 
I don't understand yet why this problem appears. I'll try to have a look ASAP.
Can you, please, open a specific issue for that?
Thanks again for your feedback.

.gitignore
```diff
@@ -13,3 +13,6 @@ astroid.egg-info/
 .cache/
 .eggs/
 .pytest_cache/
+.python-version
```
PCManticore: @hippo91 These should go into your global gitignore, as they are specific to tools you use on your dev environment rather than specific to tools used by the project itself.

ChangeLog
```diff
@@ -6,6 +6,24 @@ What's New in astroid 2.3.0?
...
+  
+    - ``numpy.core.numeric`` 
+    - ``numpy.core.function_base`` 
+    - ``numpy.core.multiarray`` 
+    - ``numpy.core.numeric`` 
+    - ``numpy.core.numerictypes`` 
+    - ``numpy.core.umath`` 
+    - ``numpy.random.mtrand``
+
+  Close PyCQA/pylint2865
```
PCManticore: That's a nice set of closed issues!

astroid/brain/brain_numpy_utils.py
```diff
@@ -0,0 +1,23 @@
...
+def infer_numpy_member(src, node, context=None):
+    node = astroid.extract_node(src)
+    return node.infer(context=context)
+
+
+def looks_like_numpy_member(member_name, node):
+    return (
+        isinstance(node, astroid.Attribute)
+        and node.attrname == member_name
+        and node.expr.inferred()[0].name.startswith("numpy")
```
PCManticore: @hippo91 Unfortunately using inference in the transform filters could have certain recursion error side effects, given that `brain` is not fully ready by the time the transform filters are called. Is there any way to prevent having this inference call here?

astroid/tests/unittest_brain_numpy_core_fromnumeric.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class SubTestWrapper(unittest.TestCase):
+    """
+    A class for supporting all unittest version wether or not subTest is available
+    """
+
+    def subTest(self, msg=None, **params):
+        try:
+            # For python versions above 3.5 this should be ok
```
PCManticore: ❓ Isn't `subTest` available from 3.4? If so, we could use it today without the need of `subtestMock`.

astroid/tests/unittest_brain_numpy_core_function_base.py
```diff
@@ -0,0 +1,86 @@
...
+    import numpy  # pylint: disable=unused-import
+
+    HAS_NUMPY = True
+except ImportError:
+    HAS_NUMPY = False
+
+from astroid import builder
+
+
+class SubTestWrapper(unittest.TestCase):
```
PCManticore: Even if we can't `subTest` right now (check my previous question on this topic), we should move these definitions into a common test file and use them from there.

.gitignore
hippo91: Yes ! Off-course.

astroid/brain/brain_numpy_utils.py
hippo91: @PCManticore , i changed the way the attribute is meant to be a `numpy` member.
I also introduced docstrings and type annotations.

astroid/tests/unittest_brain_numpy_core_fromnumeric.py
hippo91: Well done @PCManticore . You are totally right this mock is useless now.
I deleted it.

astroid/tests/unittest_brain_numpy_core_function_base.py
hippo91: All occurrences of this mock have been deleted.

hippo91:
@PCManticore ,

Thanks for you review and kind words. 
I made the changes you suggested.
The Travis CI failed due to the `pylint` step and i don't really understand why.
I used `pylint` with an old version of `pylint`'s master branch (approximately 2 months old) and everything went fine. However after updating to last version i got the same problem as in `Travis`. 
Have you any clue about it?

PCManticore:
Hi @hippo91 This should work now if you try again, there was a bug in pylint's master.

hippo91:
Thanks @PCManticore !
It seems that i'v got a problem with `python3.8`. I will investigate this ASAP.

PCManticore:
@hippo91 Seems to me that the Python 3.8 tests started to fail for both pylint and astroid. We should fix it separate of this PR though.

PCManticore:
Thank you again @hippo91 for all the hard work! 👏

hippo91:
You are welcome @PCManticore. Thanks for merging it.

PR #672, opened 2019-05-18
Fixed being unable to find distutils submodules by name when in a virtualenv

AWhetter:
## Description
When in a virtualenv, distutils is patched to pick up submodules from the original Python and not from the virtualenv. This is done by prepending the location of the outside distutils to the `__path__` of the module.
Because astroid already uses distutils, it is already imported. Therefore we can use the already imported module to find what `__path__` should be, and use that to look for submodules.

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |
|    | :scroll: Docs |

## Related Issue

Closes PyCQA/pylint#73

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -59,6 +59,9 @@ Release Date: TBA
 
   Close PyCQA/pylint#1628
 
+* Fixed being unable to find distutils submodules by name when in a virtualenv.
+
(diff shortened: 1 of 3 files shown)


PCManticore:
Great fix! 👏

PR #678, opened 2019-06-18
Fix finding of docstring under python3.8

keszybz:
It is stored as a Constant, not a Str, so we get the value a bit differently.

Fixes one test failure in https://github.com/gristlabs/asttokens/issues/28.


PCManticore:
Thank you @keszybz ! I merged this locally to fix the formatting error.

keszybz:
Thanks! With 28fc86f all tests pass on Fedora rawhide with python3.8-beta2.

Issue #681, opened 2019-07-01
Importing astroid issues a DeprecationWarning


pquentin:
### Steps to reproduce

Import astroid.

### Current behavior

There's a DeprecationWarning:

```
$ python -W error -c 'import astroid'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "../astroid/__init__.py", line 63, in <module>
    from astroid.nodes import *
  File "../astroid/nodes.py", line 23, in <module>
    from astroid.node_classes import (
  File "../astroid/node_classes.py", line 38, in <module>
    from astroid import bases
  File "../astroid/bases.py", line 33, in <module>
    MANAGER = manager.AstroidManager()
  File "../astroid/util.py", line 26, in <lambda>
    lambda: importlib.import_module("." + module_name, "astroid")
  File ".../importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "../astroid/manager.py", line 24, in <module>
    from astroid.interpreter._import import spec
  File "../astroid/interpreter/_import/spec.py", line 12, in <module>
    import imp
  File ".../imp.py", line 33, in <module>
    DeprecationWarning, stacklevel=2)
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
$ 
```

We've been using `-W error` to run the trio test suite for a long time, but this warning only became an issue when [migrating trio to pytest 5](https://github.com/python-trio/trio/pull/1133) which [has changes related to warnings](https://docs.pytest.org/en/latest/changelog.html#pytest-5-0-0-2019-06-28).

### Expected behavior

No warning:

```
$ python -W error -c 'import astroid'
$ 
```

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output

2.2.5

Thank you!

PCManticore:
Thanks for the report! We had https://github.com/PyCQA/astroid/issues/594 for moving away from `imp` to something like `importlib` but it fell off the radar for some time. This means we definitely need to prioritize the switch sooner.

hippo91:
I close this issue because it should have been solved thanks to #857.

Issue #594, opened 2018-07-15
Move away from imp


PCManticore:
Now that we only support Python 3, we should move away from `imp` to `importlib`.

Check https://bugs.python.org/issue34056 and https://github.com/PyCQA/astroid/issues/593#issuecomment-405079163.

sushobhit27:
@PCManticore can I work on this as this seems to be simple find replace issue.

PCManticore:
@sushobhit27 Please do!

pkolbus:
I'm running into the `DeprecationWarning` as well, as enabling this warning is recommended by the [Python 3.9 release notes](https://docs.python.org/3.9/whatsnew/3.9.html); [PEP 594](https://www.python.org/dev/peps/pep-0594/#imp) indicates a desire to remove `imp` in 3.10 which is now starting development. @degustaf has made excellent headway on this in #686 but this work has stalled.

@PCManticore @degustaf any objection if I take this (and #681) over?

degustaf:
@pkolbus I have no issue with you taking this over. I have been busy on other things.

PR #686, opened 2019-07-24
Partially remove dependency on imp.

degustaf:
## Steps

## Description

This partially removes the dependency on the deprecated imp module.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
This is related to issues #594 and #681

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -10,17 +10,13 @@ import abc
 import collections
 import distutils
 import enum
-import imp
 import os
 import sys
 import zipimport
 
-try:
-    import importlib.machinery
+import importlib.machinery
 
-    _HAS_MACHINERY = True
-except ImportError:
-    _HAS_MACHINERY = False
+_HAS_MACHINERY = True
 
 try:
     from functools import lru_cache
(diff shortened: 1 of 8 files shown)


degustaf:
Note that removing the final dependency on imp requires making a decision. It is located in `modutils.load_module_from_modpath` and is a pair of calls to `imp.find_module` and `imp.load_module`. The official documentation recommends replacing these with `importlib.import_module`. For this use case it does not allow for ignoring `sys.modules`, which the API allows. 

I see 2 options for how to solve this:

1. Change the API to remove the option to ignore `sys.module`. This would affect the functions `modutils.load_module_from_modpath`, `modutils.load_module_from_name`, and `modutils.load_module_from_file`.  This option does not appear to be used anywhere in astroid or pylint, except for 1 unittest.

2. Copy the code from imp.py so that we maintain the functionality by maintaining the code.

astroid/interpreter/_import/spec.py
```diff
@@ -116,23 +96,62 @@ class ImpFinder(Finder):
...
-        if stream:
-            stream.close()
-
-        return ModuleSpec(
-            name=modname,
-            location=mp_filename,
-            module_type=_imp_type_to_module_type(mp_desc[2]),
+        else:
+            try:
+                loader = importlib.util.find_spec(modname)
```
gyermolenko: Just curious. The function returns `ModuleSpec` object, why did you name it `loader`?

gyermolenko:
Don't know if it would help, but here is ref to similar improvement in pytest-dev/pytest#5468

PCManticore:
This is on my todo list for reviewing. Unfortunately might take me a bit as I'll be in vacation for the next two weeks.

astroid/interpreter/_import/spec.py
```diff
@@ -116,23 +96,62 @@ class ImpFinder(Finder):
...
+                        return ModuleSpec(
+                            name=modname,
+                            location=None,
+                            module_type=ModuleType.PY_FROZEN,
+                        )
+            except ValueError:
+                pass
+            submodule_path = sys.path
+
+        suffixes = (
```
PCManticore: Is it possible to build this list globally instead of building it every time this function is called?

PCManticore:
@degustaf Can you rebase your PR when you get a chance? I'd like to pull this in. Thank you!

degustaf:
@PCManticore I should be able to do that tonight.

stonebig:
any hope on this ?

hippo91:
I close this issue because #857 has been merged.

PR #692, opened 2019-09-02
Remove NodeNG.nearest method

### LINKED ISSUE #691: Removing NodeNG.nearest method
EDIT:
Based on discussions below we have decided to remove this method altogether because it's not being used in `astroid` or `pylint` and that it doesn't seem to have any use in the near future.

### Steps to reproduce
Run this script:
```
src = """
print(before)
print(random)
print(middle)
print(after)
"""
mod = astroid.parse(src)
before, _, middle, after = mod.body
print(middle.nearest([before, after]))
``` 

### Current behavior
`middle.nearest([before, after])` returns `before`.

### Expected behavior
`middle.nearest([before, after])` should return `after` since it's closer

### Issue Description
The issue is that for a NodeNG node `n`, `n.nearest` method only returns the nearest node that occurs on lines before `n` and disregards any nodes on the lines after.

https://github.com/PyCQA/astroid/blob/33488e28677a1e980361757cdb1db0eb13c99435/astroid/node_classes.py#L582-L607

I see two possible fixes for this:

1. Fix the method implementation so that the node that is the closest (either before or after) is returned (where precedence is given to the node before).
2. Update the docstring so that it is in tandem with how the method currently works.

I think that fix (1) is in line with method description so I will be making a PR for this.

Note: There is no usage of `NodeNG.nearest` in both `pylint`'s  and `astroid`'s source code.
2.3.0

kavins14:
EDIT:
Based on the discussion on #691, we have decided to remove the `NodeNG.nearest` method instead of fixing it.

## Steps

## Description
The `NodeNG.nearest` method was not working properly as outlined in #691 and so this PR is a fix to the method.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #691

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -579,34 +579,6 @@ class NodeNG:
         """
         return self.parent.previous_sibling()
 
-    def nearest(self, nodes):
-        """Get the node closest to this one from the given list of nodes.
-
-        :param nodes: The list of nodes to search. All of these nodes must
-            belong to the same module as this one. The list should be
-            sorted by the line number of the nodes, smallest first.
-        :type nodes: iterable(NodeNG)
-
-        :returns: The node closest to this one in the source code,
-            or None if one could not be found.
-        :rtype: NodeNG or None
-        """
-        myroot = self.root()
-        mylineno = self.fromlineno
-        nearest = None, 0
-        for node in nodes:
-            assert node.root() is myroot, (
-                "nodes %s and %s are not from the same module" % (self, node)
-            )
-            lineno = node.fromlineno
-            if lineno > mylineno:
-                if nearest[1] > 0 and lineno-mylineno >= mylineno-nearest[1]:
-                    break
-            if lineno > nearest[1]:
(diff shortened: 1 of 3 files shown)


astroid/node_classes.py
```diff
@@ -599,8 +599,9 @@ def nearest(self, nodes):
                 "nodes %s and %s are not from the same module" % (self, node)
             )
             lineno = node.fromlineno
-            if node.fromlineno > mylineno:
-                break
+            if lineno > mylineno:
+                if nearest[1] > 0 and lineno-mylineno >= mylineno-nearest[1]:
```
kavins14: nearest[1] > 0 check here is necessary so that None is not returned because mylineno is closer to 0 compared to lineno.

kavins14:
Based on the discussion on #691, we have decided to remove the NodeNG.nearest method instead of fixing it.

PCManticore:
Thank you @kavins14 !

PR #694, opened 2019-09-15
Allow annotated assign with attrs

### LINKED ISSUE #656: Unsupported object typing when using `attr.ib` with `default`
This issue seems to be similar/continuation to #628. 
`pylint` misunderstood typing when using `attr.ib(default=...)`.

I've tried `typing.cast` & attr.ib with `type` attr and the result is the same.

### Tested on
* Python 3.6.8
* pylint 2.3.1
* astroid 2.2.5

### Steps to reproduce
1. Sample python file:
    ```python
    # tst.py
    import attr

    class ApiKlass:
        def a_method(self):
            pass

    @attr.s
    class D1:
        d = attr.ib()
        api: ApiKlass = attr.ib(default=ApiKlass())
    
        def test(self):
            return self.api.a_method
    ```
2. Run `pylint tst.py`.

### Current behavior
```
[skip]
tst.py:13:15: E1101: Instance of '_CountingAttr' has no 'a_method' member (no-member)
[skip]
```

### Expected behavior
Line 13 (class `D1`) should not raise `no-member` warning.

Funth0mas:
## Steps

## Description
Update attrs brain to naively support annotated assigns as well (relates to #656). Annotated assigns are different astroid class.
Moreover every attr.ib() creates new instance attribute - let add it. (the class level attributes shall be deleted but it would produce "attribute-defined-outside-init" so I left it there).

The attr library in fact builds the `__init__` function with assignments to `self.given_attribute` and _deletes_ the original class level attributes. If I understand well, the best solution would be to synthetise such `__init__` function instead of writing node.locals (what is moreover wrong I think) or node.instance_attrs. If you have a good example how to do it, I can try to do something simple. But in another branch obviously.

Any feedback is welcome, thank you.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

## Related Issue
Closes #656

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -116,6 +116,10 @@ Release Date: TBA
 
   Close PyCQA/pylint#73
 
+* Update attr brain to partly understand annotated attributes
+
+  Close #656
+
 What's New in astroid 2.2.0?
 ============================
 Release Date: 2019-02-27
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -44,7 +44,11 @@ def attr_attributes_transform(node):
                 continue
         else:
             continue
-        targets = cdefbodynode.targets if hasattr(cdefbodynode, "targets") else [cdefbodynode.target]
+        targets = (
+            cdefbodynode.targets
+            if hasattr(cdefbodynode, "targets")
+            else [cdefbodynode.target]
+        )
         for target in targets:
 
             rhs_node = astroid.Unknown(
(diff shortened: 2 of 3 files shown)


PCManticore:
Thanks for the PR! I merged this locally to make the linting errors go away.

Funth0mas:
@PCManticore still if you have any hint to my question in the PR description, let me know, I'd try to improve the brain once more.

PCManticore:
@Funth0mas Sorry, missed that comment when I reviewed the PR. You can check https://github.com/PyCQA/astroid/blob/master/astroid/brain/brain_namedtuple_enum.py for some inspiration on how we approached the problem for `namedtuple` and `enum`. It would require generating some sort of fake class, which is then passed to `extract_node()` to transform it into the AST that replaces the `attrs` instance.

PR #702, opened 2019-10-04
Provides annotations for scipy.signal module

vallsv:
Hi.

Here is annotations for `scipy.signal` module.

I checked all the functions containing `*args` and `**kwargs` in order to provide this list. Maybe there is better way to check.

Closes https://github.com/PyCQA/pylint/issues/3154


vallsv:
Re-pushed with black formatting

PCManticore:
Hey @vallsv ! Thank you for the PR!

PR #710, opened 2019-10-18
Fix async function definitions with decorators

alexmojaki:
:bug: Bug fix

Currently as_string places the `async` keyword before the decorators. This fixes that.


PCManticore:
Nice, thank you!

PR #711, opened 2019-10-18
Fix as_string for attribute nodes with integer values

alexmojaki:
:bug: Bug fix

This PR fixes as_string producing invalid syntax like `3.real` (similar code exists in the standard library).


PCManticore:
Thank you, I appreciate all these PRs!

PR #719, opened 2019-11-05
Update six version to 1.13

### LINKED ISSUE #718: Update six version
### Steps to reproduce
1. pip install -U six
2. pip install astroid

### Current behavior
    ERROR: astroid 2.3.2 has requirement six==1.12, but you'll have six 1.13.0 which is incompatible.

uilianries:
## Steps

## Description

- Update six library version to the latest version available.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

## Related Issue

closes #718

### DIFF
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -23,7 +23,7 @@ numversion = tuple(int(elem) for elem in version.split(".") if elem.isdigit())
 extras_require = {}
 install_requires = [
     "lazy_object_proxy==1.4.*",
-    "six==1.13",
+    "six~=1.13",
     "wrapt==1.11.*",
     'typed-ast>=1.4.0,<1.5;implementation_name== "cpython" and python_version<"3.8"',
 ]
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -23,7 +23,7 @@ numversion = tuple(int(elem) for elem in version.split(".") if elem.isdigit())
 extras_require = {}
 install_requires = [
     "lazy_object_proxy==1.4.*",
-    "six==1.12",
+    "six==1.13",
     "wrapt==1.11.*",
     'typed-ast>=1.4.0,<1.5;implementation_name== "cpython" and python_version<"3.8"',
 ]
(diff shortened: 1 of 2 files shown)


michael-k:
Why the hard pin and not `six~=1.12`?  The hard pin just causes interoperability issues¹ with other packages that have either a hard pin or an upper bound (`six<1.13`).

Imho, six has proven stable enough in the past to refrain from hard pins/upper bounds.

¹ Not necessarily at runtime, but during install time when using `pip check` to catch those issues early.

uilianries:
> Why the hard pin and not `six~=1.12`? The hard pin just causes interoperability issues¹ with other packages that have either a hard pin or an upper bound (`six<1.13`).
> 
> Imho, six has proven stable enough in the past to refrain from hard pins/upper bounds.
> 
> ¹ Not necessarily at runtime, but during install time when using `pip check` to catch those issues early.

I agree with you, but this project is using fixed versions for any package. I don't know if the Devs will approve this idea.

PCManticore:
We can definitely relax `six` to `~=1.12`.

astroid/__pkginfo__.py
```diff
@@ -23,7 +23,7 @@
 extras_require = {}
 install_requires = [
     "lazy_object_proxy==1.4.*",
-    "six==1.12",
+    "six==1.13",
```
lasote: Should be `six~=1.13` otherwise, it will break the whole world again.
I've released Conan `1.7.6`, `1.8.5`, `1.9.3`, `1.10.3`, `1.11.3`, `1.12.4`, `1.13.4`, `1.14.6`, `1.15.5`, `1.16.2`, `1.17.3`, `1.18.6` and `1.19.4` fixing this, so please, not again.

tox.ini
```diff
@@ -24,7 +24,7 @@ deps =
   pytest
   python-dateutil
   pypy: singledispatch
-  six==1.12
+  six==1.13
```
lasote: Idem

astroid/__pkginfo__.py
uilianries: done

tox.ini
uilianries: done

PCManticore:
Thanks folks!

PR #726, opened 2019-11-26
Avoid a call to `getcwd` if it is not needed

rtpg:
## Description

When profiling locally, approximately 7% of a full pylint run is spent in `os.getcwd` (within `ast_from_module_name`).

This result is not actually used in cases where the `context_file` parameter is set to `None`, though (or rather, a no-op directory change gets sent in).

A quick grep over this project and others seem to indicate that `context_file` is rarely used, so this should almost entirely get rid of this system call. Local profiling seems to indicate this.

## Type of Changes

|   | Type |
| ✓  | :sparkles: Performance fix  |

### DIFF
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -128,8 +128,8 @@ class AstroidManager:
             return self.astroid_cache[modname]
         if modname == "__main__":
             return self._build_stub_module(modname)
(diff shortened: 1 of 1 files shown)


PCManticore:
This is amazing, thank you @rtpg for your PR!

PR #731, opened 2019-12-04
Handle StopIteration error in infer_int.

david-yz-liu:
## Steps

## Description

This fixes https://github.com/PyCQA/pylint/issues/3274. Prior to Python 3.7, `StopIteration` errors bubbled up through generators, so this wasn't an issue for astroid because the `StopIteration` error was handled later. Now `StopIteration` errors are converted to `RuntimeError`s ([PEP 479]), so the same error handling doesn't work.

**Note**: based on the above issue, I was able to come up with a minimal code example where calling `.inferred()` raises an error on Python 3.7+ (see the test I added), but I wasn't actually able to determine out *why* a `StopIteration` error occurred at all. I'm not familiar enough with astroid's inference to tell whether raising `StopIteration` is the intended behaviour or not, so there might be a better fix.

[PEP 479]: https://www.python.org/dev/peps/pep-0479/

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -69,6 +69,10 @@ Release Date: TBA
 
   Clean up the setup.py file, resolving a handful of minor warnings with it.
 
+* Handle StopIteration error in infer_int.
+
+  Close PyCQA/pylint#3274
+
 What's New in astroid 2.3.2?
 ============================
 Release Date: TBA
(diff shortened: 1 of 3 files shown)


PCManticore:
Thanks for the PR @david-yz-liu !

PR #733, opened 2019-12-22
Make use of cache while transform builtin containers

stanislavlevin:
## Description
As of now, the transformation of builtin containers which
members, in turn, are containers relies on `safe_infer`
helper. `safe_infer` tries to infer the node to get its
values. Doing this without a cache containing a previously
inferred nodes lead to infinite recursion, for example, on
resolving a class attribute.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes https://github.com/PyCQA/pylint/issues/3245

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -154,7 +154,7 @@ def _container_generic_inference(node, context, node_type, transform):
     if len(node.args) > 1:
         raise UseInferenceDefault()
 
-    arg, = args
+    (arg,) = args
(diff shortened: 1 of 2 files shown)


PCManticore:
Thank you for the great PR @stanislavlevin !

DanielNoord:
This seems like another issue. Do you use any `qgis` plugins for `pylint`?

couteau:
I don't know of any `qgis` plugin for `pylint`. I will open a new issue for this.

PR #737, opened 2019-12-27
Bug 666

### LINKED ISSUE #666: Add numpy brain tips for `numpy.append`
### Steps to reproduce
astroid currently does not account for `numpy.append` as mentioned in #664. An example of the problem is the following file:

```
import numpy as np
a = np.append([0], 0)
a = -a
```

which produces the following errors in pylint:

```
/tmp/test.py:3:4: E1130: bad operand type for unary -: tuple (invalid-unary-operand-type)
/tmp/test.py:3:4: E1130: bad operand type for unary -: list (invalid-unary-operand-type)
```

These errors are generated for pylint=1.7.2, astroid=1.5.3 and numpy=1.16.2. The errors are not generated for numpy=1.15.4.

hippo91:
## Steps

- [ * ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [ * ] Write a good description on what the PR does.

## Description

This PR solves the issue #666.
The problem was that `astroid` could not infer the result of a call to `append` function because this function, itself, calls the `concatenate` function.
This last function is inferred thanks to the `brain_numpy_core_multiarray` module but only when the corresponding node is an `astroid.Attribute` (for example `numpy.concatenate`). It turns out that in the source of the `append` function the node that realises the call to `concatenate` is a `astroid.Name`. Thus the correction proposed here is to register the `concatenate` inference tip function in order to apply it, also, to `astroid.Name`.

This PR contains also a refactor of the `brain_numpy_core_fromnumeric` module in order to minimize the interaction with the inference engine. There is no need to redefine the `sum` function, but instead help `astroid` to correctly infer the variable `out` in the function `sum`.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes #666

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@ What's New in astroid 2.4.0?
 ============================
 Release Date: TBA
 
+* Added a call to ``register_transform`` for all functions of the ``brain_numpy_core_multiarray``
+  module in case the current node is an instance of ``astroid.Name``
+
+  Close #666
+
 * Added some functions of the ``numpy.core.multiarray`` module
 
   Close PyCQA/pylint#3208
--- a/astroid/brain/brain_numpy_core_fromnumeric.py
+++ b/astroid/brain/brain_numpy_core_fromnumeric.py
@@ -10,7 +10,10 @@ import functools
 import astroid
 from brain_numpy_utils import infer_numpy_member
 
-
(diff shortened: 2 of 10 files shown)


hippo91:
@PCManticore 

Please do not take into account this pull request. I just realized some problems with it. Sorry

PR #740, opened 2019-12-30
Sum and multiply

### LINKED ISSUE #669: Wrong inference maybe due to inference cache
This bug has been found during the setup of the `numpy` brains. 
Fixing it could improve our inference system for `numpy` functions.

The directory described hereafter reflects a similar structure inside the `numpy` sources.
### Steps to reproduce
1. Create a directory named `bug_inf` with the following structure:
```
$ tree bug_inf
bug_inf
├── core
│   ├── __init__.py
│   └── shape_base.py
├── __init__.py
└── lib
    ├── __init__.py
    └── shape_base.py
```
2. `bug_inf/core/__init__.py` content is:
```
$ cat bug_inf/core/__init__.py 
from .shape_base import *
```
3. `bug_inf/core/shape_base.py` content is:
```
$ cat bug_inf/core/shape_base.py 
def vstack():
    return 42
```
4. `bug_inf/__init__.py` content is:
```
$ cat bug_inf/__init__.py 
from .core import *
from .lib import *
```
5. `bug_inf/lib/__init__.py` content is:
```
$ cat bug_inf/lib/__init__.py 
from .shape_base import *
```
6. `bug_inf/lib/shape_base.py` content is:
```
$ cat bug_inf/lib/shape_base.py 
from bug_inf.core import vstack
```
7. Create a file `bug_inf.py` which content is:
```

SRC = """
from bug_inf import vstack
vstack()
"""

NODE = extract_node(SRC)
print(NODE.inferred())
```
8. Run `python3 bug_inf.py`

### Current behavior
```
 $ python3 bug_inf.py 
[<Const.int l.2 at 0x7f08471338d0>, Uninferable]
```

### Expected behavior
```
 $ python3 bug_inf.py 
[<Const.int l.2 at 0x7f08471338d0>]
```

### Astroid version
```
python3 -c "from astroid import __pkginfo__; print(__pkginfo__.version)"
2.3.0
```

hippo91:
## Steps

- [ * ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [ * ] Write a good description on what the PR does.

## Description
This PR tries to solve the issue #669. In fact the call to some functions of the `numpy` brains are inferred not only as `numpy.ndarray` but also as `Uninferable`. This lead to false negative for such function calls. 

The origin of this bug has been hard to find but seems to be due to the fact that the `context.push` method return `True` as soon as the node in argument has been visited once for the current context. 
However for some particular cases, such as those encountered in the `numpy` library, a node could be visited more than once due to peculiar imports. 
The modification made here ensure that all calls to `numpy ufunc` functions are inferred only as `numpy.ndarray` and modifies two inference unit tests. Those ones seem more pertinent but a close look from the reviewer would be appreciated.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes #669


PCManticore:
Hey @hippo91 The code looks good. I tested this against `pylint` and it causes a regression test to fail. Can you determine why that would be the case with this patch? It's `regression_property_no_member_2641`. I believe it's something to do with the new inference limit but haven't been able to investigate more closely.

hippo91:
@PCManticore thanks for looking at this. I confirm that there is a problem with this unit test. I'm trying to fix it.

hippo91:
@PCManticore , 

i had a look at the `regression_property_no_member_2641` unittest and i think that the problem is not directly linked to this PR.
In fact, the following code triggers also the `no-member` message :  
```
# pylint: disable=missing-docstring,unused-argument,too-few-public-methods
class Person:
    def __init__(self, name):
        self.name = name

    @property
    def name(self):
        return self.__name

    @name.setter
    def name(self, value):
        self.__name = value

A = Person('toto')
Person.name.fset(A, 'toto')
```
In case `fset` is replaced by `fget` then no problem arises. I think the problem is a lack of `attr_fset` method in the `PropertyModel` class (`objectmodel` module).
@PCManticore do you thinks there is a reason for the absence of `attr_fset` and `attr_del` method in this class?
Is it wise to add them? 
If yes i'll make another PR. 
Thanks in advance.

PCManticore:
@hippo91 Yeah, that makes sense, let's add that in separate PRs.

PCManticore (approved):

hippo91:
Replaced by `sum_and_multiply_clean`. Same sources with a cleaner log history

PR #743, opened 2020-01-10
Fix for no-member regression in pylint

sodul:
## Description

This fixes the regression documented in PyCQA/pylint#3134 and introduced in PyCQA/pylint#2841.

When running pylint on code that imports some third party package such as boto3 we get a `no-member` error as documented in the pylint bug report. This PR rolls back a single line that introduced the regression.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes PyCQA/pylint#3134

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -176,10 +176,7 @@ def _container_generic_transform(arg, context, klass, iterables, build_elts):
         if all(isinstance(elt, nodes.Const) for elt in arg.elts):
             elts = [elt.value for elt in arg.elts]
         else:
-            # TODO: Does not handle deduplication for sets.
-            elts = filter(
-                None, map(partial(helpers.safe_infer, context=context), arg.elts)
-            )
+            raise UseInferenceDefault()
     elif isinstance(arg, nodes.Dict):
         # Dicts need to have consts as strings already.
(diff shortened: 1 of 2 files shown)


PCManticore:
Hey @sodul Thanks for the PR. This looks to me that it reverts a functionality rather than actually fixing the underlying problem, I'm not entirely convinced that's the way to go. We probably need to investigate a bit more to figure out a better way without reverting the said capability.

sodul:
@PCManticore The regression is a bug that seems to impact many and this patch has allowed us to upgrade to 3.8 by letting us install the latest pylint. Since there is no timeline as to when the new feature will be fixed our recommendation is to rollback the regression for now so that there can move forward with newer versions of pylint instead of pining to an older version.

We do prefer to fall forward but for regressions that cannot be fixed promptly a rollback is reasonable and buys time for a proper long term fix of the feature.

mcallaghan-bsm:
My 2 Cents:

Without being in the weeds on the details here, at a high level if a new feature introduces a fairly widely impacting regression bug, and blocks users from upgrading to Python3.8; we should definitely opt for removing whatever feature introduced it.

Then as time permits, authors and stakeholders of said feature can work through the problem properly without urgency.

PCManticore:
Closed in favour of https://github.com/PyCQA/astroid/pull/763.

Issue #745, opened 2020-01-23
Exception when invoking `infer_lhs()` on `AssignAttr`


ltcmelo:
### Steps to reproduce
I got the exception below in a large project and I don't have a concise snippet to reproduce it. Yet, I'd like to report the basics of the issue; then you can take it under consideration.

*a.py*
```python
class Configuration:
    def __init__(self):
        self.input_dir = None
```

*b.py*
```python
config = Configuration()
config.input_dir = p  # where p is a for variable
```

At some point in my code, I get the [AssignAttr](http://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.AssignAttr) node (let's call it `n`) for `config.input_dir` and invoke `n.infer_lhs()`.

### Current behavior

I get this exception:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/astroid/inference.py", line 310, in infer_attribute
    context.boundnode = owner
AttributeError: 'NoneType' object has no attribute 'boundnode'
```

<img width="1131" alt="Screen Shot 2020-01-23 at 15 54 16" src="https://user-images.githubusercontent.com/2905588/73014262-99d0c080-3df8-11ea-8a1b-2ae4ef09da7c.png">

### Expected behavior

No exception, and the usual possible results from the `infer` function(s).

ltcmelo:
Very succinct (possibly, incomplete) [PR for this](https://github.com/PyCQA/astroid/pull/746).

PCManticore:
Thanks for the report! The PR seems like it might fix the issue, we just need a regression test to confirm it.

PR #746, opened 2020-01-23
Don't crash upon invalid contex on attr. inference

ltcmelo:
## Description

This is an attempt to fix [this issue](https://github.com/PyCQA/astroid/issues/745). Sorry for the succinct PR... I'm not an astroid developer and this still needs testing and (likely) further elaboration. I'd just like to give some primary input.

 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -307,6 +307,8 @@ def infer_attribute(self, context=None):
                 except exceptions._NonDeducibleTypeHierarchy:
                     # Can't determine anything useful.
                     pass
+        elif not context:
+            context = contextmod.InferenceContext()
 
         try:
             context.boundnode = owner


PCManticore:
Thanks for the PR! Is it possible to add a regression test for it as well?

ltcmelo:
Hi @PCManticore , as [I explained in the issue](https://github.com/PyCQA/astroid/issues/745), I couldn't, within a few minutes, come up with a simple self-contained snippet that reproduces the problem. Perhaps, you (or other astroid developers) would have a better insight on how to achieve that?

PCManticore:
@ltcmelo Without a full traceback, there's no way even for us to come up with a test 😅 Regardless, the fix is a simple one and it makes sense, so we can skip the tests for now.

PR #747, opened 2020-01-24
Bugfix/enum crash/pylint-3365

martimlobao:
## Steps

## Description
Fixes a `pylint` crash caused by an `IndexError` exception when running `pylint` on any file that has a `with ... as` in an enum class.

Example:
```python
import enum

class Error(enum.Enum):
    Foo = "foo"
    Bar = "bar"
    with "error" as err:
        pass
```

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

## Related Issue

Closes PyCQA/pylint#3365

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 2.4.0?
 Release Date: TBA
 
 
+* Fix a `pylint` crash when using `with ... as` in an enum class.
+
+  Close PyCQA/pylint#3365
+
 * Added a call to ``register_transform`` for all functions of the ``brain_numpy_core_multiarray``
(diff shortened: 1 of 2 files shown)


PCManticore:
@martimlobao Thanks for the PR! Unfortunately the root cause of the issue was our Enum transform, which was dropping the attributes that it could not understand from the class's locals., such as those for `With` nodes. 

https://github.com/PyCQA/astroid/commit/78e56c64cb016e372a5ae017798ebd7cf83e1f26 should fix this properly.

martimlobao:
@PCManticore Awesome, thanks for fixing the real underlying issue then!

PR #752, opened 2020-02-10
Bug pylint 3332

hippo91:
## Steps

- [*] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [*] Write a good description on what the PR does.

## Description
This PR fixes PyCQa/pylint#3332 issue. It was  due to the fact that `datetime64` class inherited from `generic` one and in this last class the `astype` method was returning `uninferable`. Now it returns `ndarray` object.

This fix should be generalized to all other methods of the `generic` class as the `numpy` doc specifies that the `generic` object API should be the same as the `ndarray` one. A TODO has been added.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
PyCQa/pylint#3332


hippo91:
@PCManticore the travis CI has failed for `pypy`but it is not related to my development.

PCManticore:
@hippo91 Just retried the job, maybe it'll work now. Most likely it was a flakey environment.

PCManticore:
Thanks for the fix! Seems `pypy` doesn't work but that can be fixed separately.

PR #771, opened 2020-04-27
If a generic class is subscripted, infer that class itself

mthuurne:
## Description

Ideally, when a generic class is subscripted, the value for the type variable would be remembered and used in later inference. However, just interring the subscripted class itself already solves multiple false positives in pylint, so this seems like a useful first step.

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |
|    | :scroll: Docs |

## Related Issue

Closes PyCQA/pylint#3131
Closes PyCQA/pylint#3505

It might also solve PyCQA/pylint#3129, but I'm having trouble reproducing that issue on Python 3.8 even without this PR applied.

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -84,7 +84,9 @@ def infer_typing_attr(node, context=None):
     except InferenceError as exc:
         raise UseInferenceDefault from exc
 
-    node = extract_node(TYPING_TYPE_TEMPLATE.format(value.qname().split(".")[-1]))
+    node = extract_node(
+        TYPING_TYPE_TEMPLATE.format(value.qname().split(".")[-1]), "typing"
+    )
     return node.infer(context=context)
 
 
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -75,12 +75,14 @@ def _looks_like_typing_subscript(node):
 def infer_typing_attr(node, context=None):
     """Infer a typing.X[...] subscript"""
     try:
-        value = next(node.value.infer())
+        for value in node.value.infer():
+            if value.qname().startswith("typing."):
+                break
+        else:
+            raise UseInferenceDefault
+
     except InferenceError as exc:
         raise UseInferenceDefault from exc
-
-    if not value.qname().startswith("typing."):
-        raise UseInferenceDefault
 
     node = extract_node(TYPING_TYPE_TEMPLATE.format(value.qname().split(".")[-1]))
     return node.infer(context=context)
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -368,6 +368,10 @@ def infer_subscript(self, context=None):
         if value is util.Uninferable:
             yield util.Uninferable
             return None
+        if isinstance(value, nodes.ClassDef):
+            if value.is_subtype_of("typing.Generic"):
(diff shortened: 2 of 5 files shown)


mthuurne:
I just got an `InconsistentMroError` when checking [SoftFab](https://github.com/boxingbeetle/softfab/) with this PR applied. So please do review the PR, but be warned that it either directly or indirectly breaks pylint.

mthuurne:
Part of the problem seems to be that one `ClassDef` instance for `typing.Generic` is not equal to another `ClassDef` instance for `typing.Generic`. This leads to the C3 algorithm not discarding some duplicate bases.

How is this supposed to work? Should `ClassDef` override `__eq__()` and compare `qname()`? Or maybe `qname()` plus source file and line, to be able to tell apart redefinitions?

Or should there some kind of mechanism that ensures there is a single representative instance for each member of the `typing` module? Perhaps `astroid.MANAGER` could be used for this, but it's unclear to me how to properly populate and query its cache.

NeilGirdhar:
might also close https://github.com/PyCQA/pylint/issues/3684 ?

NeilGirdhar:
@PCManticore This pull request would be very helpful.  Would you be able to privde some feedback about how you want it implemented?

astroid/brain/brain_typing.py
```diff
@@ -75,14 +75,18 @@ def _looks_like_typing_subscript(node):
 def infer_typing_attr(node, context=None):
     """Infer a typing.X[...] subscript"""
     try:
-        value = next(node.value.infer())
+        for value in node.value.infer():
```
PCManticore, line 78: Not sure why we were not passing the `context` parameter to `value` inference. Can you add it here and see if it has any unintended side effects? (e.g. if certain tests start to fail, let's keep it removed)

astroid/inference.py
```diff
@@ -368,6 +368,10 @@ def infer_subscript(self, context=None):
         if value is util.Uninferable:
             yield util.Uninferable
             return None
+        if isinstance(value, nodes.ClassDef):
+            if value.is_subtype_of("typing.Generic"):
```
PCManticore, line 372: Can you also pass the `context` to `is_subtype_of`? This should prevent potential circular inference that could cause `RecursionErrors` to pop up.

PCManticore (commented):
Overall the change looks good, but I wonder the following:

1. Can we define a `Generic` object in [objects](https://github.com/PyCQA/astroid/blob/master/astroid/objects.py) that acts as a future placeholder for a more thorough inference? This would be similar to the other objects that we have defined there, like `Super`, `FrozenSet`, `DictKeys` and whatnot. The way it could work would be that when we detect the `typing.Generic`, we emit instead a `Generic` instance after the inference. The `Generic` instance could contain the original type and for all intents and purposes could act as a `Instance` or `ClassDef` (I'm not sure what `typing.Generic` is in Python)
2. You mentioned some possible errors that break pylint. I'd like to see those reflected in tests over here in `astroid`, so it would be easier to detect what's broken and potentially fix the problems before shipping this change.
3. Same here with regards to the MRO issues, it would be great if those could be replicated in tests over here.

Pierre-Sassoulas:
Probably outdated as it aim to fixes three issues that are closed. Seems like #927 is what fixed them.

PR #790, opened 2020-05-17
Remove unneeded python2 test data

DavidCain:
## Description
Now that astroid no longer supports Python 2, there's reason to keep
this test data around. The `tests/testdata/python2` directory is not
referenced at all, so we may safely delete it.

It's perhaps just worth moving all of `testdata/python3` into a new
directory, but this is a start.

## Type of Changes


PCManticore:
Thanks for the cleanup!

PR #801, opened 2020-06-10
Allow lazy-object-proxy >= 1.5.0

### LINKED ISSUE #800: lazy-object-proxy update to 1.5.0 breaks astroid
### Steps to reproduce
1. locally install lazy-object-proxy==1.5.0
2. locally install astroid

### Current behavior
Exception
> Requirement.parse('lazy_object_proxy==1.4.*'), {'astroid'})

### Expected behavior
Works as before, as I don't see any breaking changes in lazy-object-proxy
Before e1b4e11e383bf281544f1300e41eea2ac6ab251c this would have worked

priv-kweihmann:
by allowing every version >= 1.4.0

Fixes #800

## Steps

## Description

Allow lazy-object-proxy version >= 1.5.0

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|    | :scroll: Docs |

## Related Issue

Closes #800

### DIFF
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -26,7 +26,7 @@ numversion = tuple(int(elem) for elem in version.split(".") if elem.isdigit())
 
 extras_require = {}
 install_requires = [
-    "lazy_object_proxy==1.4.*",
+    "lazy_object_proxy>=1.4.0",
     "six~=1.12",
     "wrapt~=1.11",
     'typed-ast>=1.4.0,<1.5;implementation_name== "cpython" and python_version<"3.8"',


ViktorHaag:
If the purpose of the original pinning was to lower the risk of breaking changes in a dependency, would it not be better for this PR to pin to versions <2 for the dependency than admit for future, potentially breaking, versions?

priv-kweihmann:
Don't know, never heard anything back about the original reasoning of the version pinning.
And to be honest a breaking API change could happen with any version update (as nobody seems to care really about a proper versioning scheme) - so I personally think that setting a base version and accepting any version higher than that, is the way it should be done - as fixes for newer versions have to be done anyway (if not now than at some point in the future)

priv-kweihmann:
Any news of this item? Can I do something to get this merged?

stonebig:
I would hope if you change it to "lazy_object_proxy>=1.4,<1.6" it may be accepted

FelixMoelder:
A merge would be nice, as `lazy-object-proxy 1.4.x` is not available on conda-forge for osx-arm64.

efwahl:
The ==1.4.* pinning is breaking with other packages which have an explicit 1.5 requirement and the new pip resolver behavior. It would be nice to have a fix promptly.

AadamZ5:
I'm also running into issues with the `==1.4.*` pinning having conflicts with other packages in my project. It seems like this PR would be fine to move in with `>=1.50`. The sooner the better 😊, and thank you for your help!

PCManticore:
Thanks for the PR.

PR #805, opened 2020-06-20
Fix recursion error from occuring on self referential length calls

### LINKED ISSUE #777: Astroid Infinite loop when `__len__(self)` is mistakenly defined as `return len(self)`
### Steps to reproduce
Lint the following program with pylint
```python
class Crash:
    def __len__(self) -> int:
        return len(self)
```

### Current behavior

The following error is thrown and pylint crashes
```
RecursionError: maximum recursion depth exceeded
```

### Expected behavior

It should not crash.

### Asteroid version
2.4.0

brycepg:
## Steps

## Description

Fix recursion error from occuring on self referential length calls

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
## Related Issue
Closes #777

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -237,13 +237,31 @@ def object_len(node, context=None):
     :raises AstroidTypeError: If an invalid node is returned
         from __len__ method or no __len__ method exists
     :raises InferenceError: If the given node cannot be inferred
-        or if multiple nodes are inferred
+        or if multiple nodes are inferred or if the code executed in python
+        would result in a infinite recursive check for length
     :rtype int: Integer length of node
     """
     # pylint: disable=import-outside-toplevel; circular import
     from astroid.objects import FrozenSet
 
     inferred_node = safe_infer(node, context=context)
+
+    # prevent self referential length calls from causing a recursion error
+    # see https://github.com/PyCQA/astroid/issues/777
+    node_frame = node.frame()
+    if (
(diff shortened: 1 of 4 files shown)


tests/unittest_brain.py
```diff
@@ -2020,5 +2020,24 @@ class Other:
     assert isinstance(name[0], astroid.Unknown)
 
 
+def test_no_recursionerror_on_self_referential_length_check():
+    """
+    Regression test for https://github.com/PyCQA/astroid/issues/777
+    """
+    with test_utils.pytest_assert_not_raises(RecursionError):
```
PCManticore, line 2027: Let's remove this check entirely and simply rely on checking for inference errors. Checking against `RecursionError` itself does not provide any value.
```
     next(node.infer())
```

astroid/helpers.py
```diff
@@ -244,6 +246,23 @@ def object_len(node, context=None):
...
+        and node_frame.name == "__len__"
+        and inferred_node._proxied == node_frame.parent
+    ):
+        message = (
+            "Self referential __len__ function will "
+            "cause a RecursionError on line {} of {}".format(
+                node.lineno, node.root().file
+            )
+        )
+        raise exceptions.AstroidRecursionError(message)
```
PCManticore, line 264: Rather than adding a new exception, let's stop with `InferenceError` like the rest of the function is doing.

brycepg:
Yeah that makes testing easier. Ok I made those changes

PR #811, opened 2020-07-12
replace deprecated importlib methods

### LINKED ISSUE #703: ImportWarning due to usage of deprecated method
### Steps to reproduce
It seems that the ``importer.find_module`` calls in the ``_search_zip`` function emit an ImportWarning when namespace packages are encountered:

https://github.com/PyCQA/astroid/blob/bff51e9d41a3084b9313fa2108ae13f079bc0b58/astroid/interpreter/_import/spec.py#L276-L292

This then breaks pytest-pylint runs in which warnings are turned into errors.

According to the [importlib documentation](https://docs.python.org/3/library/importlib.html#importlib.abc.PathEntryFinder.find_module) ``find_module`` is deprecated and one should use ``find_spec`` instead. When patching the library locally this seems to have solved the issue with the namespace packages.

### Current behavior
An ImportWarning is raised and the namespace package does not get imported. For example:
``ImportWarning: Not importing directory /home/.venv/lib/python3.5/site-packages/google: missing __init__``

### Expected behavior
The package gets imported and no ImportWarning is issued.

### Version
astroid == 2.3.1

GergelyKalmar:
## Steps

## Description
Astroid seems to be using deprecated methods in importlib which causes pytest-pylint runs to fail (with a certain configuration) when working with namespace packages. The linked issue described the problem in detail.

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #703

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -278,9 +278,10 @@ def _precache_zipimporters(path=None):
 def _search_zip(modpath, pic):
     for filepath, importer in list(pic.items()):
         if importer is not None:
-            found = importer.find_spec(modpath[0])
+            found = getattr(importer, 'find_spec', importer.find_module)(modpath[0])
             if found:
-                if not importer.find_spec(os.path.sep.join(modpath)):
+                name = os.path.sep.join(modpath)
+                if not getattr(importer, 'find_spec', importer.find_module)(name):
                     raise ImportError(
                         "No module named %s in %s/%s"
                         % (".".join(modpath[1:]), filepath, modpath)
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -240,7 +240,7 @@ def _is_setuptools_namespace(location):
         with open(os.path.join(location, "__init__.py"), "rb") as stream:
             data = stream.read(4096)
     except IOError:
-        pass
+        return None
     else:
         extend_path = b"pkgutil" in data and b"extend_path" in data
         declare_namespace = (
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -294,7 +294,9 @@ def _search_zip(modpath, pic):
             found = getattr(importer, "find_spec", importer.find_module)(modpath[0])
             if found:
                 module_path = os.path.sep.join(modpath)
-                if not getattr(importer, "find_spec", importer.find_module)(module_path):
+                if not getattr(importer, "find_spec", importer.find_module)(
+                    module_path
+                ):
                     raise ImportError(
                         "No module named %s in %s/%s"
                         % (".".join(modpath[1:]), filepath, modpath)
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -280,8 +280,8 @@ def _search_zip(modpath, pic):
         if importer is not None:
             found = getattr(importer, "find_spec", importer.find_module)(modpath[0])
             if found:
-                name = os.path.sep.join(modpath)
-                if not getattr(importer, "find_spec", importer.find_module)(name):
+                module_path = os.path.sep.join(modpath)
+                if not getattr(importer, "find_spec", importer.find_module)(module_path):
                     raise ImportError(
                         "No module named %s in %s/%s"
                         % (".".join(modpath[1:]), filepath, modpath)
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -278,9 +278,9 @@ def _precache_zipimporters(path=None):
 def _search_zip(modpath, pic):
     for filepath, importer in list(pic.items()):
         if importer is not None:
-            found = importer.find_module(modpath[0])
+            found = importer.find_spec(modpath[0])
             if found:
-                if not importer.find_module(os.path.sep.join(modpath)):
+                if not importer.find_spec(os.path.sep.join(modpath)):
                     raise ImportError(
                         "No module named %s in %s/%s"
                         % (".".join(modpath[1:]), filepath, modpath)
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -12,17 +12,11 @@ import abc
 import collections
 import distutils
 import enum
-import imp
 import os
 import sys
 import zipimport
 
-try:
-    import importlib.machinery
-
-    _HAS_MACHINERY = True
-except ImportError:
-    _HAS_MACHINERY = False
+import importlib.machinery
 
 try:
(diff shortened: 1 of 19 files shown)


GergelyKalmar:
@PCManticore very strange, when applying this exact fix on the example case in issue #703 it fixes the issue. Do you have an idea why do the tests fail at the same time? In my example case ``importer`` is an instance of ``<class '_frozen_importlib_external.FileFinder'>`` and not ``zipimport.zipimporter``, how could that be?

GergelyKalmar:
@PCManticore anyways, I could work around this. Could you review this PR? It's a tiny change that would help a lot. Thank you!

GergelyKalmar:
This issue still persists and blocks the usage of `filterwarnings = error` with `pytest-pylint`. Could someone review & possibly merge?

astroid/interpreter/_import/spec.py
```diff
@@ -291,9 +291,10 @@ def _precache_zipimporters(path=None):
 def _search_zip(modpath, pic):
     for filepath, importer in list(pic.items()):
         if importer is not None:
-            found = importer.find_module(modpath[0])
+            found = getattr(importer, "find_spec", importer.find_module)(modpath[0])
```
hippo91, line 294: Why not simply replacing `find_module` with `find_spec`?

astroid/interpreter/_import/spec.py
```diff
@@ -291,9 +291,10 @@ def _precache_zipimporters(path=None):
...
             if found:
-                if not importer.find_module(os.path.sep.join(modpath)):
+                name = os.path.sep.join(modpath)
```
hippo91, line 296: I found that the name `name` is not well suited here because the variable is definitely more a path than a name. Maybe something along `module_path`?

hippo91 (changes requested):

hippo91:
@GergelyKalmar thanks for this PR! 
In order to test this PR i'm trying to reproduce #703 but i'm unable to. I sent a comment in the discussion dedicated to.

astroid/interpreter/_import/spec.py
GergelyKalmar: I tried that first but the tests were failing - it seems that the `importer` object that appeared here did not have the `find_spec` attribute, I'm not sure why. This way the change is backwards compatible because it will simply fall back to `find_module` if `find_spec` doesn't exist, so I think it's anyways a safer choice.

astroid/interpreter/_import/spec.py
GergelyKalmar: That's a good point, I've updated the variable name.

astroid/interpreter/_import/spec.py
hippo91: @GergelyKalmar i investigated a bit the need for `find_spec` or `find_module` methods.
In fact the `pic.items()` container is made of `FileFinder` objects and `ZipFinder`objects. While the first have the `find_spec` method (that should be used instead of deprecated `find_module`) the latter don't.
The `FileFinder` objects come from the `sys.path_importer_cache`, in the method `_precache_zipimporters`, while the `ZipFInder` objects are added in the same method.
The `sys` module documentation explains that the objects in `sys.path_importer_cache` can be of two types, both of them having the `find_spec` method so i think it is ok to think of them as the same `FileFinder` type.
I wonder if it wouldn't be clearer to handle a container dedicated to `FileFinder` objects and another to `ZipFinder` objects, to distinguish clearly both types.
What do you think about it?

GergelyKalmar: Hm, as far as I see, the `sys.path_importer_cache` objects are all subclasses of `importlib.abc.Finder` (which is deprecated though) while `ZipFinder` is a class defined in Astroid and is a subclass of `Finder` which is defined in `interpreter/_import/spec.py`. We could certainly do an `isinstance` check instead of checking for the existence of the attribute, so this would become

```suggestion
            if isinstance(importer, (importlib.abc.MetaPathFinder, importlib.abc.PathEntryFinder)):
                found = importer.find_spec(modpath[0])
            elif isinstance(importer, spec.Finder):  # alternatively, we could use spec.ZipFinder
                found = importer.find_module(modpath[0])
```
and then we could do the same below. It is more explicit but less elegant in my opinion. Alternatively, we could modify `ZipFinder` (or perhaps `Finder`?) to expose a `find_spec` method which just refers to `find_module`, so the interfaces are better aligned and then we can just replace `find_module` with `find_spec` in this function.

Either solution works, but perhaps the last one is the most elegant one. What do you think?

hippo91: @GergelyKalmar i'm not fond either with the `isinstance` check.
By the way, i'm sorry because i made a mistake in my previous post. We are not dealing here with `ZipFinder` but with `zipimport.zipimporter` objects. Thus the second proposition you made doesn't hold.
Maybe the simplest thing to do is to let the `getattr` solution by explicitly commenting it to explain the two kinds of objects inside `pic.items()`...

GergelyKalmar: I agree – I think this is the most appropriate solution for now. I've added a comment explaining the situation and linking to this discussion for future reference accordingly. Thank you for your feedback!

hippo91:
@GergelyKalmar while i was about to merge this into master, i decided to test it against `pylint`. Two tests were failing in the `test_self.py` module (namely `test_can_list_directories_without_dunder_init` and `test_duplicate_code_raw_strings`). 
In those two tests, `pylint` tries to import two specific modules (located in `pylint/tests/regrtest_data` for the second test and in `/tmp` for the first one). Those two modules are NOT contained into Zip files.
The way it works without your patch:
the `find_module` method in the `_search_zip` function returns None for those module. This way the `ZipFinder` instance cannot find the module and this leads to the try of another finder which will be able to find the module.
The way it works with your patch:
the `find_spec` method does not return None but the corresponding module cannot be imported. As the return is not None no other finder will be called to find the module. 

After investigating a bit, i think that the core of the problem is the mix of `zipimporter` and `FileFinder` instances inside the collection returned by `pic.items()` whereas we are explicitly trying to import a module from a Zip file.
If this collection was filled with only `zipimporter` objects, then the problem will not arise.

I think we can get back to the original call to `find_module` inside `_search_zip` function but we have to ensure that `pic.items()` method returns only `zipimporter` instances. What i propose is simply to change a bit the return of the `_precache_zipimporters` function in a way similar to `return {key: pic.get(key) for key in new_paths}`.

I tried this and both `astroid` and `pylint` tests suite seem to be ok. 

What do you think about it?

GergelyKalmar:
If it fixes the issue at hand, I'm perfectly fine with it! Could you perhaps create a new PR with your suggestion? Then I could try and test that too to see if it works with my use cases.

hippo91:
@GergelyKalmar i opened the #869 PR if you want to review it and make tests.

PR #820, opened 2020-07-20
Feature: teach astroid about Hypothesis' `@st.composite` decorator

Zac-HD:
Hypothesis' `@st.composite` decorator removes (ie supplies for the caller) the first argument of the decorated function.  This has been a small frustration for many people, and caused a large number of `# pylint: disable=no-value-for-parameter` comments.

Recently, with https://github.com/HypothesisWorks/hypothesis/issues/2498, I personally got annoyed on behalf of our users (and, OK, at being repeatedly asked about it) enough to look up Pylint's plugin system, and a few hours later here we are!  I must say it's a very smooth extension API :smile:

Imitating the boto3 brain there aren't any tests, but I've confirmed it's working locally - just let me know what if anything else I need to do.

### DIFF
--- a/astroid/brain/brain_hypothesis.py
+++ b/astroid/brain/brain_hypothesis.py
@@ -1,0 +1,53 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
+"""
+Astroid hook for the Hypothesis library.
+
+Without this hook pylint reports no-value-for-parameter for use of strategies
+defined using the `@hypothesis.strategies.composite` decorator.  For example:
+
+    from hypothesis import strategies as st
+
+    @st.composite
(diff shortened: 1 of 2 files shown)


astroid/brain/brain_hypothesis.py
```diff
@@ -0,0 +1,54 @@
...
+    def a_strategy(draw):
+        return draw(st.integers())
+
+    a_strategy()
+
+"""
+
+import astroid
+
+COMPOSITE_NAMES = (
```
DRMacIver, line 21: This doesn't seem to detect when people are importing `composite` directly rather than referencing it via a module import. I bet a lot of people are doing that. Is there any way to test for that case?

Zac-HD: Yep... I just have to add the bare name to the list of decorators we treat.  I think this is OK since I've added "is the first argument named `draw`" to the detection heuristic.

PCManticore (approved):
Thanks for the clear patch, folks.

PR #829, opened 2020-08-04
Skip test for | in dictionaries due to PEP-584 in Python 3.9+

### LINKED ISSUE #826: Dict union operator in Python 3.9 causes test failure
### Steps to reproduce

The below line where | is used for dictionaries is valid now with https://www.python.org/dev/peps/pep-0584/ in Python 3.9

 https://github.com/PyCQA/astroid/blob/25384d4bebf0187b6704c818c7df64945793362c/tests/unittest_inference.py#L2458

### Current behavior

Error is raised in Python 3.9

### Expected behavior

No error should be raised.

tirkarthi:
## Steps

## Description

Skip check for | in dictionaries since PEP 584 is implemented in Python 3.9 that adds support for dictionary union using | operator.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #826

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -2455,7 +2455,6 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         1 ** (lambda x: x) #@
         {} * {} #@
         {} - {} #@
-        {} | {} #@
         {} >> {} #@
         [] + () #@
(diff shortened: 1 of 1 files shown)


mgorny:
ping.

PR #830, opened 2020-08-05
Changes the way `flat` attribute of `numpy` array is inferred

hippo91:
## Steps

## Description

This PR solves the pylint bug PyCQA/pylint#3640. 
It was due to the fact that in the `brain_numpy_ndarray` the `flat` attribute was `None`.
Rigorously it should be a `numpy.flatiter` instance, but as for now this class doesn't exist in the brain, i choose to replace it by just a `numpy.ndarray`. This way, iteration and indexation on the return value of the `flat` attribute is allowed.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes PyCQA/pylint#3640

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,10 @@ astroid's ChangeLog
 What's New in astroid 2.5.0?
 ============================
 Release Date: TBA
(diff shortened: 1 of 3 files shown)


PCManticore (approved):
Nice fix.

PR #838, opened 2020-09-06
Run tests with deprecation warnings enabled

atodorov:
## Description

Enables testing with deprecation warnings .

## Type of Changes

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,6 +30,7 @@ deps =
 
 setenv =
     COVERAGE_FILE = {toxinidir}/.coverage.{envname}
+    PYTHONWARNINGS = d
 
 commands =
     ; --pyargs is needed so the directory astroid doesn't shadow the tox


hippo91:
@atodorov thanks for this PR. When running tox with the modification you propose, i can't see any warning concerning deprecated functionality. Is it normal?
Moreover, in the description you made reference to a fix concerning deprecated module but i didn't find it.

atodorov:
> @atodorov thanks for this PR. When running tox with the modification you propose, i can't see any warning concerning deprecated functionality. Is it normal?

Not sure TBH. On some test suites I've worked you can see the deprecation warnings even when running through tox on others you don't see them with this configuration. I'm not sure what's causing this.

> Moreover, in the description you made reference to a fix concerning deprecated module but i didn't find it.

I've updated the description. I was looking into replacing the imp module with importlib but it looked harder than what I could do ATM so left only the changes for tox.

FTR the warning I'm seeing is:

```
2020-09-07T16:17:31.0670396Z /opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/astroid/interpreter/_import/spec.py:15: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
```

GergelyKalmar:
Not sure if it is related, but I'm also having issues with the import mechanism issuing deprecation warnings in some cases, see https://github.com/PyCQA/astroid/issues/703 and https://github.com/PyCQA/astroid/pull/811. Interestingly it only happens with ``pytest-pylint`` but not when ``pylint`` is used directly, I assume because they use a different import approach.

cdce8p (commented):
`pytest` prints deprecation warnings by default, so I'm not sure this is really necessary.
See: https://github.com/PyCQA/astroid/runs/2656259183#step:6:56

Pierre-Sassoulas:
Closing because of what @cdce8p said.

PR #847, opened 2020-10-26
Drop tracebacks on Astroid import errors for caches

rtpg:
This is a variant on #837, which drops tracebacks instead of introducing a messy workaround
 (thanks @pstch @char101 for the suggestion!)

 Before this change, when there were import failures by Astroid, it
 would store the error in the module cache (for example a "module not
 found" error), including tracebacks. These tracebacks would take up a
 lot of memory, and don't seem to be used for any sort of purpose
 beyond debugging value.

 On one project stripping this value entirely reduces memory usage by
 75%.

For the memory analysis itself, this variant uses a _bit_ more memory than my original solution, but I think that the non-weirdness of the solution makes this solution better

image

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,7 @@ Release Date: TBA
 
 * Fixed exception-chaining error messages.
 
+* Reduce memory usage of astroid's module cache.
 
 What's New in astroid 2.4.3?
 ============================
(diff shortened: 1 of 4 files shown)


rtpg:
Ok there were some unrelated pylint errors (not 100% sure on the flow that got me to these failures showing up) but I pulled them in and things are passing in CI now

@Pierre-Sassoulas is there anything else needed for a merge on this?

Pierre-Sassoulas (approved):
Thank you for the MR, can you just add a quick summary in the changelog, please ?

Pierre-Sassoulas:
Thanks again, great analysis resulting in a huge memory boost with almost no drawback !

hippo91:
@rtpg thanks for the great work! :+1:

PR #854, opened 2020-11-21
Fix python39

hippo91:
## Steps

- [ x ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [ x ] Write a good description on what the PR does.

## Description
This PR aims to enable python 3.9 in `astroid`. 4 unit tests where failing when passing from python 3.8 to python 3.9.
One concerning dataclasses. It was due to the fact that in python 3.9 the function `replace` in the `dataclasses` module is using positional only arguments  which was not the case before. Thus it appears that there was probably a lack in the `infer_argument` method of the `arguments` module.
The other failing tests concern the `NamedTuple` object. Prior to python 3.9 it is just a class inside the `typing` module. With python 3.9 it is now a function that create a new type. It is much more difficult to infer. In fact i spent few days trying to achieve something smart to enable astroid to infer the result of this function but i failed. Instead i decided to generalise the NamedTuple brain.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ jobs:
     env: TOXENV=py37
   - python: 3.8
     env: TOXENV=py38
+  - python: 3.9
+    env: TOXENV=py39
 before_install:
 - python --version
 - uname -a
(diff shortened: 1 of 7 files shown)


hippo91:
There is a clever way to handle `NamedTuple` in python3.9. Another PR will follow.

Issue #855, opened 2020-11-22
deque misses __class_getitem__


JulienPalard:
In `brain_collections.py`, deque misses its `__class_getitem__`, implemented in Python 3.9.

Should it be added with a test version, so astroid "knows" that it's not present in Python 3.8 but is in Python 3.9?

Related to: https://github.com/PyCQA/pylint/issues/3951

PR #856, opened 2020-11-22
Add missing __class_getitem__ to deque.

JulienPalard:
## Description

Add a missing (introduced in Python 3.9) classmethod to deque.

Feel like it should be marked as introduced in 3.9 so astroid won't tell it exists while reading 3.8 code, any hint?

## Type of Changes

Single line of code, won't name it a bug fix nor a new feature.

## Related Issue

Closes: https://github.com/PyCQA/astroid/issues/855

### DIFF
--- a/astroid/brain/brain_collections.py
+++ b/astroid/brain/brain_collections.py
@@ -60,7 +60,9 @@ def _deque_mock():
         def __iadd__(self, other): pass
         def __mul__(self, other): pass
         def __imul__(self, other): pass
-        def __rmul__(self, other): pass"""
+        def __rmul__(self, other): pass
+        @classmethod
+        def __class_getitem__(self, item): pass"""
     return base_deque_class
 
 


hippo91:
@JulienPalard thanks for this PR! :+1:  Could you add a test and an entry in the ChangeLog please?

PCManticore:
Thanks for the PR! If this needs to be marked as 3.9, then we can add it to the class string only on that version, so a test asserting it does not exist in 3.8 would be achievable in that case. Other than that, this looks good.

hippo91:
@JulienPalard when do you think having time to finish this PR?
If you prefer i can handle it.

JulienPalard:
@hippo91 yeah, my free time melted down in january :( you can handle it if you have more than me. Consider the code in this PR as open/free as possible (wow there's only 3 lines :D).

hippo91:
I'm closing this PR as it is replaced by #885. Thanks @JulienPalard !

PR #857, opened 2020-11-22
Remove dependency on imp.

### LINKED ISSUE #594: Move away from imp

pkolbus:
## Steps

## Description

Removes all usage of the `imp` module, which has been deprecated since python 3.4.

This is a continuation of @degustaf's excellent work in #686. I've rebased, addressed review comments from that PR, fixed pylint warnings, and implemented the solution for modutils as discussed in #686.

The test_qqch test is removed as it seems to behave differently depending on whether modutils imports anything from stdlib, and it’s not clear what the objective of that test actually is.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #594 and #681

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -541,18 +541,6 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         self.assertEqual(ancestor.root().name, BUILTINS)
         self.assertRaises(StopIteration, partial(next, ancestors))
 
-    def test_qqch(self):
-        code = """
-            from astroid.modutils import load_module_from_name
-            xxx = load_module_from_name('__pkginfo__')
-        """
-        ast = parse(code, __name__)
-        xxx = ast["xxx"]
-        self.assertSetEqual(
-            {n.__class__ for n in xxx.inferred()},
-            {nodes.Const, util.Uninferable.__class__},
-        )
-
     def test_method_argument(self):
         code = '''
             class ErudiEntitySchema:
--- a/ChangeLog
+++ b/ChangeLog
@@ -55,6 +55,11 @@ Release Date: TBA
 
 * Reduce memory usage of astroid's module cache.
 
+* Remove dependency on `imp`.
+
+  Close #594
+  Close #681
+
 What's New in astroid 2.4.3?
 ============================
 Release Date: TBA
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -12,17 +12,11 @@ import abc
 import collections
 import distutils
 import enum
-import imp
 import os
 import sys
 import zipimport
 
-try:
-    import importlib.machinery
-
-    _HAS_MACHINERY = True
-except ImportError:
-    _HAS_MACHINERY = False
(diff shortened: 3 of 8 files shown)


tests/unittest_inference.py
```diff
@@ -549,8 +549,7 @@ def test_qqch(self):
         ast = parse(code, __name__)
         xxx = ast["xxx"]
         self.assertSetEqual(
-            {n.__class__ for n in xxx.inferred()},
-            {nodes.Const, util.Uninferable.__class__},
+            {n.__class__ for n in xxx.inferred()}, {util.Uninferable.__class__}
```
kolbus-garmin, line 552: The change in behavior in this test is due to the removal of 'import imp' in modutils.py. I'm having trouble understanding what 'qqch' is and this test has been part of astroid since the initial commit. If this adaptation of expected behavior is incorrect or perhaps the input code should be adapted to something else, let me know.

pkolbus:
This is now ready for review.

tests/unittest_inference.py
```diff
@@ -541,18 +541,6 @@ class Warning(Warning):
         self.assertEqual(ancestor.root().name, BUILTINS)
         self.assertRaises(StopIteration, partial(next, ancestors))
```
hippo91, line 543: I agree these test is not specially clear, but i do not understand why do you delete it. Can you explain a little bit more?

hippo91 (changes requested):
@pkolbus thank you very much for this PR. 
I just need some explanations about the `test_qqch` deletion. Otherwise it's all good for me.
By the way thanks also to @degustaf !

pkolbus:
> @pkolbus thank you very much for this PR. 
> 
> I just need some explanations about the `test_qqch` deletion. Otherwise it's all good for me.
> 
> By the way thanks also to @degustaf !

You're welcome!

The `test_qqch` deletion is summarized in the PR description. The test now gives different results on Python 3.8 and earlier versions. Oddly, simply re-adding the `import imp` to modutils changes the result, so the fact that it worked before seemed accidental and fragile. And it's not clear to me from the test name or the repository history what this is aiming to cover or why.

If you think the deletion is a problem and can offer any clues about the intent of the test, I'd be happy to re-add it, or an equivalent.

hippo91:
@pkolbus i investigated a bit this unit test. With current master (i.e using `imp` module) the inferred value for `xxx` node is `[Uninferable, <Const.NoneType l.236 at 0x7f601f5c5050>, <Const.NoneType l.326 at 0x7f601f801ad0>]`. Thus even with the current implementation this import is not inferred precisely.
The first `NoneType` comes from the function `load_module_from_modpath` (probably line 236). This one doesn't exist any more with your implementation because this function has been considerably redefined.
The second `NoneType` comes from the `imp` module itself.
Thus IMHO this unit test doesn't means anything anymore. Let's remove it.

pkolbus:
Thanks!

PR #858, opened 2020-11-22
Fix python39

hippo91:
## Steps

## Description
This PR deals with python 3.9. The biggest change concerns the `NamedTuple` brain. With `python 3.9` `NamedTuple` is no more a class but instead it is now a function. The class `NamedTuple` is built dynamically that's why it is necessary to enhance the brain module to take this into account. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Linked issues
Closes PyCQA/pylint#3876 and partially PyCQA/pylint#3895

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -399,6 +399,21 @@ def infer_typing_namedtuple_class(class_node, context=None):
     return iter((generated_class_node,))
 
 
+def infer_typing_namedtuple_function(node, context=None):
+    """
+    Starting with python3.9, NamedTuple is a function of the typing module.
+    The class NamedTuple is build dynamically through a call to `type` during
+    initialization of the `_NamedTuple` variable.
+    """
+    klass = extract_node(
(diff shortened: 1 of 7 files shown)


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -399,6 +399,21 @@ def infer_typing_namedtuple_class(class_node, context=None):
     return iter((generated_class_node,))
 
 
+def infer_typing_namedtuple_function(node, context=None):
```
Pierre-Sassoulas, line 402: Are we keeping the compat to 3.5 ? If that's not the case we could add typing on new functions :)

Pierre-Sassoulas (approved):
Nice evolution. Should we also remove python 3.5 ? What about in pylint ? python3.5 prevent some thing from being done easily and the end of life was September 2020.

hippo91:
@Pierre-Sassoulas you are right, it's time to drop support for python3.5 but i would prefer to do this in another PR.

PR #864, opened 2020-12-15
Drop six support, drop Py35 support

### LINKED ISSUE #863: Please bump dependency version of six
The current 1.12 dependency is two years old, and there are some other packages out in the Python ecosystem that are starting to depend on newer versions of six. For us, this was google-api-core and related Google cloud packages, but as time goes on that will only increase.

Alternatively, your package now is python 3 only. It would be good to drop the dependency entirely.

dgilman:
Drop support for six in the codebase.

Add CI integration to test environments guaranteed to have and not have six in them.

While making all of this work, I found that a dependency of pytest was pulling in six on python 3.5. It appears the latest pytest, only released a few days ago, dropped support for python before 3.6. I didn't set out to deprecate 3.5 support here but I figure if you can't run unit tests on a platform you can't really support it. So I went ahead and bumped the minimum Python version for astroid to 3.6.

If this PR gets accepted it would be good to do a release of astroid, as it would be good to get this fix out there but I also see you have a few other interesting bug fixes in source control like the infamous exception/stacktrace memory leak.

## Steps

TODO: changelog, once the PR passes review

## Type of Changes
|   | Type |

## Related Issue

Closes #863

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -123,6 +123,13 @@ Release Date: TBA
 
   Close PyCQA/pylint#3904
 
+* Corrected the parent of function type comment nodes.
+
+  These nodes used to be parented to their original ast.FunctionDef parent
+  but are now correctly parented to their astroid.FunctionDef parent.
+
+  Close PyCQA/astroid#851
+
 
 What's New in astroid 2.4.2?
 ============================
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -238,7 +238,7 @@ class TreeRebuilder:
 
         return type_object.value
 
-    def check_function_type_comment(self, node):
+    def check_function_type_comment(self, node, parent):
         type_comment = getattr(node, "type_comment", None)
         if not type_comment:
             return None
@@ -251,10 +251,10 @@ class TreeRebuilder:
 
         returns = None
         argtypes = [
-            self.visit(elem, node) for elem in (type_comment_ast.argtypes or [])
+            self.visit(elem, parent) for elem in (type_comment_ast.argtypes or [])
         ]
         if type_comment_ast.returns:
-            returns = self.visit(type_comment_ast.returns, node)
+            returns = self.visit(type_comment_ast.returns, parent)
 
         return returns, argtypes
 
@@ -615,7 +615,7 @@ class TreeRebuilder:
             returns = None
 
         type_comment_args = type_comment_returns = None
-        type_comment_annotation = self.check_function_type_comment(node)
+        type_comment_annotation = self.check_function_type_comment(node, newnode)
         if type_comment_annotation:
             type_comment_returns, type_comment_args = type_comment_annotation
         newnode.postinit(
--- a/tests/unittest_nodes.py
+++ b/tests/unittest_nodes.py
@@ -1181,6 +1181,19 @@ def test_type_comments_posonly_arguments():
                 assert actual_arg.as_string() == expected_arg
 
 
+@pytest.mark.skipif(not HAS_TYPED_AST, reason="requires typed_ast")
+def test_correct_function_type_comment_parent():
+    data = """
+        def f(a):
+            # type: (A) -> A
+            pass
+    """
+    astroid = builder.parse(data)
+    f = astroid.body[0]
+    assert f.type_comment_args[0].parent is f
+    assert f.type_comment_returns.parent is f
+
+
 def test_is_generator_for_yield_assignments():
     node = astroid.extract_node(
         """


tests/unittest_brain.py
```diff
@@ -400,6 +401,12 @@ def test_nose_tools(self):
 
 @unittest.skipUnless(HAS_SIX, "These tests require the six library")
 class SixBrainTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        tox_env = os.environ.get("TOX_ENV_NAME")
+        if tox_env and "six" not in tox_env and HAS_SIX:
```
hippo91, line 403: `tox_env.endswith('-six')` instead of `"six" in tox_env` wouidn't be safer here?

hippo91 (changes requested):
@dgilman thanks for this PR!
It is all clear for me except a probably stupid question : why should we keep tests and test environments dealing with `six`?
By the way , do not forget to add an entry in the ChangeLog.
Thanks again

dgilman:
There is a brain file for six and it has unit tests so it makes sense to keep testing it in CI. Likewise it makes sense for a astroid developer to install six to their local venv for when they run unit tests. The thinking behind the CI work to run tests where six is guaranteed to not be installed is that this guards against someone who codes in astroid against six when it is no longer a requirement of the package. That breakage is not going to show up in their local running of tests, nor will it show up when CI is run on GitHub, but will blow up in the user's face when they try to use the released package, which would be unfortunate.

But maybe the risk of this happening is too small to care about. I totally understand if you want to boot them, let me know and I will fix up the PR.

hippo91:
@dgilman to my mind we can drop all what deal with `six` but i would have @PCManticore opinion before doing this.

PCManticore:
We still need `six` for Python 2/3 compatibility though, but we don't need (as far as I know) the `six` dependency itself.

tox.ini
```diff
@@ -18,13 +18,13 @@ deps =
...
-  py35,py36,py37,py38,py39: typed_ast>=1.4.0,<1.5
+  py{35,36,37,38,39}: numpy
+  py{35,36,37,38,39}: attrs
+  py{35,36,37,38,39}: typed_ast>=1.4.0,<1.5
   pytest
-  python-dateutil
+  !py{35,36,37,38,39}-six: python-dateutil
   pypy: singledispatch
-  six~=1.12
+  six: six
```
hippo91, line 26: @dgilman i'am sorry but i don't understand how do you ensure that in environments suffixed with `-six` the `six` module is really absent...
In fact it seems to be the opposite. But can you explain how do you manage this?

tests/unittest_scoped_nodes.py
```diff
@@ -1499,13 +1507,10 @@ class A(object): pass
...
         module = builder.parse(
             """
-        import six
-
         class Metaclass(type):
             foo = lala
 
-        @six.add_metaclass(Metaclass)
-        class B(object): pass
+        class B(object, metaclass=Metaclass): pass
```
hippo91, line 1507: I do not understand why in some tests `six` is removed whereas in other it is kept.

hippo91 (approved):
@dgilman i still have some questions. I'm sorry if some of them are a bit silly but i'm not yet very used to `travis` coupled to `tox`...

tests/unittest_brain.py
```diff
@@ -79,6 +79,13 @@
 except ImportError:
     HAS_ATTR = False
 
+try:
+    import six  # pylint: disable=unused-import
+
```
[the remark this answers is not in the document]
dgilman: I was following the line break pattern used in the other import tests above it. If you'd like me to clean all of them up I can do that.

dgilman:
I was not familiar with the `--py3k` option to pylint, and in fact I didn't attempt to run this PR's version of astroid with pylint at all. I will try running some pylint testing this weekend.

tox.ini
dgilman: I am not sure what to say about the -tox line, that seems to be incorrect and I will fix it, but tox appears to not care. The root of the configuration happens on this line though, and is documented upstream here https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings

Just by declaring `six: six` a new factor called six has been created (the factor name is the part before the colon), and any tox environment that contains that factor will install the dependency.

The setUpClass that raises an exception is only run when you're under tox (`TOX_ENV_NAME` envvar is defined), and six is installed (`HAS_SIX`), and the six factor was not declared in this environment (`tox_env.endswith('-six')`). If you run the unit test case locally you won't have TOX_ENV_NAME declared, so you don't have to worry about this conditional. If you are running under tox you have tox setting up and tearing down the virtualenv, which because of the above `deps=` configuration makes sure six is or is not installed, allowing us to do an extra round of double checking that six did not sneak in either as an explicit dependency in our tox.ini or because one of our dependencies decided to pull it in. That's why python-dateutil is blacklisted for six, it's pulling in six for its own purposes.

tests/unittest_scoped_nodes.py
dgilman: `SixBrainTest` keeps six because it is meant to work with six.  But other than that one-off exception, my understanding is that the other tests used six to implement 2/3 compatibility within the unit test itself. Python 2 and six support was dropped from the codebase in an earlier commit, but wasn't fully purged from the unit tests, likely because six was still getting installed by tox. The purpose of this PR is to purge all these forgotten uses of six which were likely just oversights from the earlier PR that dropped Python 2.

tests/unittest_brain.py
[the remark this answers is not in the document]
hippo91: You are totally right. Let's keep it.

tox.ini
hippo91: @dgilman thanks a lot for this explanation and the like toward the doc. Now i understand. :+1:

tests/unittest_scoped_nodes.py
hippo91: @dgilman i understand and agree with you. However what about the `test_metaclass_generator_hack` which is not part of `SixBrainTest` but keep using `six`? Two other tests (`test_using_invalid_six_add_metaclass_call`, `test_with_metaclass_mro`) are in the same case but given the name or commit comment i'm ok to keep `six` in them.

dgilman: See the comment in scoped_nodes.py:1684 . This is the "metaclass hack" that this is all referring to. In order to create a metaclass that exercises these code paths you have to jump through a bunch of hoops, which is what [six does in its with_metaclass()](https://github.com/benjaminp/six/blob/master/six.py#L840) . Going back to the scoped_nodes.py function, you see it tests for `self.name == "with_metaclass"` before applying the hack, which is six's name for the function that is doing all of this. So the unit test and code conditionals were written with six in mind, so it makes sense to keep the unit test relying on six.

hippo91: @dgilman thanks for this clear explanation and sorry for all those questions.

hippo91:
@dgilman IMHO this PR can be merged into master. However i would have @PCManticore opinion because we maybe want to release an `astroid` version with `python 3.9` support + `python 3.5` deprecation warning before definitely throwing away `python 3.5` support.

Pierre-Sassoulas (commented):
The cleanup of all the ` @test_utils.require_version` is pretty nice !

AWhetter:
Given that Python 3.5 is end of life, I'm happy for us to drop support for that in the next version without releasing a version with deprecation warnings first. We're correctly setting `python_requires` so it won't break things for people. However we should sync this up with dropping support in pylint as well. Thankfully the lovely @dgilman has got a pull request in for that too.

PCManticore (approved):

PCManticore:
I concur with Ashley. We should still keep the `six` transform in `astroid` but otherwise, feel free to drop Python 3.5 and `six`  code paths from the code base.

hippo91:
@dgilman i'm sorry i did a wrong manipulation when trying to fix merge problems. I think i broke this PR. 
Hopefully i got a copy on my local machine and will make a new PR.
Sorry for that.

hippo91:
See PR #889

PR #869, opened 2020-12-24
Fix deprecated importlib methods

### LINKED ISSUE #703: ImportWarning due to usage of deprecated method

hippo91:
## Steps

## Description
This PR is a variant from #811 . Instead of having different types of objects inside the `pic.items()` collection returned by the `_precache_zipimporters` function, the return value of this last one is filled only with `zipimporter` instances.
Thus the warning dealing with `find_spec/find_module` is no more present because `zipimporter` object have only `find_module` method.
For more details please read the conversation of #811.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related issue

Closes #703

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -285,7 +285,7 @@ def _precache_zipimporters(path=None):
             pic[entry_path] = zipimport.zipimporter(entry_path)
         except zipimport.ZipImportError:
             continue
-    return pic
+    return {key: value for key, value in pic.items() if isinstance(value, zipimport.zipimporter)}
 
 
 def _search_zip(modpath, pic):
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -276,7 +276,7 @@ def _precache_zipimporters(path=None):
     """
     For each path that has not been already cached 
     in the sys.path_importer_cache, create a new zipimporter
-    instance and store it into the cache.
+    instance and add it into the cache.
     Return a dict associating all paths, stored into the cache, to corresponding
     zipimporter instances
 
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -281,7 +281,7 @@ def _precache_zipimporters(path=None):
     zipimporter instances.
 
     :param path: paths that has to be added into the cache
-    :return: association between paths stored into the cache and zipimporter instances
+    :return: association between paths stored in the cache and zipimporter instances
     """
     pic = sys.path_importer_cache
 
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -270,6 +270,16 @@ def _cached_set_diff(left, right):
 
 
 def _precache_zipimporters(path=None):
+    """
+    For each path that has not been already cached 
+    in the sys.path_importer_cache, create a new zipimporter
+    instance and store it into the cache.
+    Return a dict associating all paths, stored into the cache, to corresponding
+    zipimporter instances
+
+    :param path: paths that has to be added into the cache
+    :return: association between paths stored into the cache and zipimporter instances
+    """
     pic = sys.path_importer_cache
 
     # When measured, despite having the same complexity (O(n)),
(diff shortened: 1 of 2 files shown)


astroid/interpreter/_import/spec.py
```diff
@@ -270,6 +271,16 @@ def _cached_set_diff(left, right):
 
 
 def _precache_zipimporters(path=None):
+    """
+    For each path that has not been already cached 
+    in the sys.path_importer_cache, create a new zipimporter
+    instance and store it into the cache.
+    Return a dict associating all paths, stored into the cache, to corresponding
```
GergelyKalmar, line 278:
```suggestion
    Return a dict associating all paths, stored in the cache, to corresponding
```

astroid/interpreter/_import/spec.py
```diff
@@ -270,6 +271,16 @@ def _cached_set_diff(left, right):
...
+    zipimporter instances
```
GergelyKalmar, line 279:
```suggestion
    zipimporter instances.
```

astroid/interpreter/_import/spec.py
```diff
@@ -270,6 +271,16 @@ def _cached_set_diff(left, right):
...
+
+    :param path: paths that has to be added into the cache
+    :return: association between paths stored into the cache and zipimporter instances
```
GergelyKalmar, line 282:
```suggestion
    :return: association between paths stored in the cache and zipimporter instances
```

astroid/interpreter/_import/spec.py
```diff
@@ -270,6 +271,16 @@ def _cached_set_diff(left, right):
 
 
 def _precache_zipimporters(path=None):
+    """
+    For each path that has not been already cached 
+    in the sys.path_importer_cache, create a new zipimporter
+    instance and store it into the cache.
```
GergelyKalmar, line 277:
```suggestion
    instance and add it into the cache.
```

astroid/interpreter/_import/spec.py
```diff
@@ -285,7 +296,7 @@ def _precache_zipimporters(path=None):
             pic[entry_path] = zipimport.zipimporter(entry_path)
         except zipimport.ZipImportError:
             continue
-    return pic
+    return {key: value for key, value in pic.items() if isinstance(value, zipimport.zipimporter)}
```
GergelyKalmar, line 299: I'm not sure if that's as intended, but this simply filters out everything from `pic` that is not a `zipimport.zipimporter` class (i.e. the other types of importers). This may be related to the failing test too.

GergelyKalmar:
@hippo91 It worked fine for me, I just added a few cosmetical suggestions. As this is supposed to be just a cache I think it should be fine to potentially throw out things which are not of zipimporter type, but I must add too that I'm not super familiar with the inner workings of astroid.

astroid/interpreter/_import/spec.py
hippo91: @GergelyKalmar it is exactly as intended. The goal is to have a container with only one type of objects. The failing tests (dealing with pypy) are not due to this PR.

hippo91:
@GergelyKalmar thanks for your wise suggestions!

GergelyKalmar:
@hippo91 Awesome! Thank *you* for the help with this quite tricky issue. Feel free to merge & release!

PR #872, opened 2020-12-27
Astroid 2.5.0

hippo91:
## Steps

## Description
This is a PR for future 2.5.0 version.
When merged we wiil have to do the 4-5 and 6 steps of the release.txt file.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,8 @@ astroid's ChangeLog
 
 What's New in astroid 2.5.0?
 ============================
-Release Date: 2020-12-27
+Release Date: TBA
+
 
 * Add `python 3.9` support.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,8 +5,7 @@ astroid's ChangeLog
 
 What's New in astroid 2.5.0?
 ============================
-Release Date: TBA
-
+Release Date: 2020-12-27
 
 * Add `python 3.9` support.
 
@@ -78,11 +77,6 @@ Release Date: TBA
 * Fix a crash in functools.partial inference when the arguments cannot be determined
 
   Close PyCQA/pylint#3776
-
-
-What's New in astroid 2.4.3?
-============================
-Release Date: TBA
 
 * Fix a crash caused by a lookup of a monkey-patched method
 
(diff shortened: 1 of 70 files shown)


hippo91:
@PCManticore have you got an idea on how to put a deprecation warning when `astroid` is used with `python3.5`?

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
@hippo91, regarding the deprecation warning I would use the following in `setup.py`:

```
import sys
import warnings

if sys.version_info.major == 3 and sys.version_info.minor <=5:
    warnings.warn("You will soon need to upgrade to python 3.6 in order to use the latest version of Astroid.", DeprecationWarning)
```

hippo91:
@Pierre-Sassoulas definitely a good idea! The message will be printed during install and not at runtime. It's probably sufficient.

ChangeLog
```diff
@@ -5,8 +5,7 @@ astroid's ChangeLog
 
 What's New in astroid 2.5.0?
 ============================
-Release Date: TBA
-
+Release Date: 2020-12-27
```
PCManticore, line 8: This is already obsolete. My suggestion is to simply backport the code changes but keep the `__pkginfo__` and `ChangeLog` for a separate change in the day of the release.

astroid/modutils.py
```diff
@@ -89,11 +91,23 @@
             pass
 
 if platform.python_implementation() == "PyPy":
+    #  The get_python_lib(standard_lib=True) function does not give valid
```
PCManticore, line 94: Fantastic!

PCManticore (approved):
Approved, but ideally we should remove or change the release date to today.

ChangeLog
hippo91: @PCManticore i will move back the Release Date in the ChangeLog. However i didn't change the `__pkginfo__`.
By the way, have you an idea concerning the release date? There are many people waiting for a release that support python 3.9.

PCManticore: Thanks. Not right now. I am still going through the latest updates, mails and issues and don't have a clear picture on what's needed for Python 3.9 compatibility. I did notice some Python 3.9 issues specific to `pylint` though.

FRidh:
Could there be a tag and PyPI release?

PR #883, opened 2021-01-19
Bug pylint 4001

hippo91:
## Steps

## Description
This PR adds a brain to mock the fact that, starting with `python3.9`, `type` object should behave as it had a `__class_getitem__` method so that it is possible to use such annotation: `type[int]`.
Note that `python` types, that inherit from `type` object, do not have such method, otherwise it would be possible to use `str[int]` in annotation. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/astroid/brain/brain_type.py
+++ b/astroid/brain/brain_type.py
@@ -29,7 +29,7 @@ def _looks_like_type_subscript(node):
     Try to figure out if a Name node is used inside a type related subscript
 
     :param node: node to check
-    :type node: nodes.Name
+    :type node: astroid.node_classes.NodeNG
     :return: true if the node is a Name node inside a type related subscript
     :rtype: bool
     """
--- a/astroid/brain/brain_type.py
+++ b/astroid/brain/brain_type.py
@@ -18,12 +18,7 @@ Thanks to Lukasz Langa for fruitful discussion.
 """
 import sys
 
-from astroid import (
-    MANAGER,
-    extract_node,
-    inference_tip,
-    nodes,
-)
+from astroid import MANAGER, extract_node, inference_tip, nodes
 
 
 def _looks_like_type_subscript(node):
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_type.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+import sys
+
+from astroid import MANAGER, extract_node, inference_tip, nodes
+
+
+PY39 = sys.version_info >= (3, 9)
+
+
+def _looks_like_type_subscript(node):
```
Pierre-Sassoulas, line 27: I think we can add typing directly in signature as we do not have python 3.5 to take into consideration anymore.
```suggestion
def _looks_like_type_subscript(node: nodes.Name) -> bool:
```

astroid/brain/brain_type.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Try to figure out if a Name node is used inside a type related subscript
+
+    :param node: node to check
+    :type node: nodes.Name
```
Pierre-Sassoulas, line 32: Seems to me it's not a class `nodes.Name` as we make a check to see if it is in the function ?

Pierre-Sassoulas (approved):
I learned something reading the comment, thanks.

astroid/brain/brain_type.py
hippo91: @Pierre-Sassoulas, sadly we did not yet throw away `python3.5`...
#864 has not been merged yet. I wanted to have @PCManticore opinion about it (i.e should the drop of `python3.5` be part of the next release or should it be delayed to the one after..)
I hope it will be done soon.

Pierre-Sassoulas: Ho ok, sorry :)

PCManticore: Hey sorry, don't hold PRs off because of me. I might not be able to respond in a timely manner. Feel free to drop Python 3.5 if the Python version is already EOLed.

PR #890, opened 2021-02-07
Fix regression property no member 2641

hippo91:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description
This PR adds the `attr_fset` method to the `PropertModel` class, so that `pylint` is able to deal with property setter.
The need for such a fix has been highlighted with #740 and is confirmed by the bug report PyCQA/pylint#3480.

## Type of Changes
|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes PyCQA/pylint#3480

PR #841, opened 2020-09-12
add support of 'six.with_metaclass' 

fmigneault:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description

Fixes issue #713 by providing corresponding transform and unit tests.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |
| ✓  | :sparkles: New feature |

## Related Issue

Closes #713

PR #893, opened 2021-02-11
Fix six with metaclass

hippo91:
## Steps

## Description
This PR corrects the brain dealing with `six.with_metaclass`.
The error came from the tweaking of the node bases. It has not to be done. 
For example, following code:
```
src = """
class A:
    pass

def get_parent():
    return A

class B(get_parent()):
    pass

B
"""
node=extract_node(src)
cls_node = node.inferred()[0]
print(list(cls_node.bases))
print(list(cls_node.ancestors()))
```
gives:
```
[<Call l.8 at 0x7f0ad5ecc250>]
[<ClassDef.A l.2 at 0x7f0ad5faf350>, <ClassDef.object l.0 at 0x7f0ad6ea45d0>]
```
The base is a CallNode, whereas the ancestors are ClassDef from the class hierarchy.
Thus in the `six.with_metaclass` case, the base has to be a call node.
Doing this, the `six` module is seen as really used and this avoids the emission of `unused-import` message.

This PR should correct the failing CI on `pylint` side.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Failing CI on `pylint`.

### DIFF
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -484,6 +484,34 @@ class SixBrainTest(unittest.TestCase):
         inferred = next(ast_node.infer())
         self.assertIsInstance(inferred, nodes.FunctionDef)
 
+    def test_with_metaclass_subclasses_inheritance(self):
+        ast_node = builder.extract_node(
+            """
+        class A(type):
+            def test(cls):
+                return cls
+
+        class C:
+            pass
+
+        import six
+        class B(six.with_metaclass(A, C)):
+            pass
+
+        B #@
+        """
+        )
+        inferred = next(ast_node.infer())
+        self.assertIsInstance(inferred, nodes.ClassDef)
+        self.assertEqual(inferred.name, "B")
+        bases = inferred.bases
+        self.assertIsInstance(bases[0], nodes.Call)
+        ancestors = tuple(inferred.ancestors())
+        self.assertIsInstance(ancestors[0], nodes.ClassDef)
+        self.assertEqual(ancestors[0].name, "C")
+        self.assertIsInstance(ancestors[1], nodes.ClassDef)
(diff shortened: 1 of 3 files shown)


hippo91:
@fmigneault are you ok with this change?
Related discussion:
#890 #841 PyCQA/pylint#4006

tests/unittest_brain.py
```diff
@@ -484,6 +484,34 @@ def test_from_submodule_imports(self):
...
+        inferred = next(ast_node.infer())
+        self.assertIsInstance(inferred, nodes.ClassDef)
+        self.assertEqual(inferred.name, "B")
+        bases = inferred.bases
+        self.assertIsInstance(bases[0], nodes.Call)
+        ancestors = tuple(inferred.ancestors())
+        self.assertIsInstance(ancestors[0], nodes.ClassDef)
+        self.assertEqual(ancestors[0].name, "C")
+        self.assertIsInstance(ancestors[1], nodes.ClassDef)
+        self.assertEqual(ancestors[1].name, "object")
```
Pierre-Sassoulas, line 513: Shouldn't this be `A` and not `object` here ? I don't know anything about `six.with_metaclass` :)

Pierre-Sassoulas (approved):

tests/unittest_brain.py
hippo91: @Pierre-Sassoulas first argument of six.with_metaclass is indeed the metaclass whereas the second argument is the base class. Thus i think that it is correct to not have class A in the ancestors.
What do you think about it?

Pierre-Sassoulas: OK, it makes sense, thank you for explaining.

Issue #899, opened 2021-02-17
False positive with Enum.__members__.items()


chiefnoah:
### Steps to reproduce
1. Use an enum's `__members__` field:
```python

from enum import Enum

class TestEnum(Enum):
    ONE = 1
    TWO = 2

print(x for x in TestEnum.__members__.items())
```
2. Run pylint with astroid 2.5 installed
```shell
$ pylint --version
pylint 2.6.2
astroid 2.5
Python 3.8.7 (default, Dec 26 2020, 08:45:55)
[GCC 10.2.1 20201203]
``` 

### Current behavior

$ pylint test.py
************* Module test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:3:0: C0115: Missing class docstring (missing-class-docstring)
test.py:7:17: E1101: Function '__members__' has no 'items' member (no-member)

--------------------------------------------------------------------
Your code has been rated at -4.00/10 (previous run: -4.00/10, +0.00)

### Expected behavior

Not complaining about `__members__.items()`

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output

2.5

See #895

hippo91:
@chiefnoah thanks for this report and the reproducer.
I can reproduce it.
The faulty commit is the same as the one of #895 (cc3bfc5dc94062a582b7b3226598f09d7ec7044e).
I'm looking for a fix.

hippo91:
@chiefnoah i reverted the faulty commit in #901. I am the author of this commit. It was a try to improve the inference of `numpy` objects.
Apologies for the regression you faced.

PR #884, opened 2021-01-24
Sum and multiply clean

hippo91:
Just the same PR as `sum_and_multiply` with a cleaner log history.
No need to review

PR #901, opened 2021-02-17
Bug astroid 895 899

### LINKED ISSUE #895: Unexpected error from pylint after upgrade to astroid 2.5.
### Steps to reproduce
astroid 2.5
pylint 2.6.0
elasticsearch 7.10.1
python 3.7.3
```
from elasticsearch import Elasticsearch
es = Elasticsearch(["test"])
es.indices.delete(index="test", ignore=[400, 404])
```
1.  run pylint

### Current behavior
line 3 >>> E1123: Unexpected keyword argument 'ignore' in method call (unexpected-keyword-arg)

### Expected behavior
astroid 2.4.2: no error from pylint

hippo91:
## Steps

## Description
This PR reverts #884. The goal of this PR was to ensure that `numpy` ufunc functions where inferred unconditionally as ufunc objects (i.e Uninferable was not part of the inferred values).
However doing this led to false positives such as #895 and #899. The modules used in those bug reports do not have anything to do with `astroid` brain. They use the classical astroid engine (when i say classical it means no use of brains).
IMHO it is not acceptable that trying to improve `numpy` inference (which already uses astroid brains) leds to downgrading inference of other modules. That's why i propose this PR.

In case `numpy` inference really has to be improved, i'll try to find another solution.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes #895 #899

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,15 @@
 astroid's ChangeLog
 ===================
 
+What's New in astroid 2.5.1?
+============================
+Release Date: TBA
+
+* The ``context.path`` is reverted to a set because otherwise it leds to false positives 
+  for non `numpy` functions.
+
+  Closes #895 #899
+
 What's New in astroid 2.5?
 ============================
 Release Date: 2021-02-15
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
:+1:

PR #908, opened 2021-02-22
Update pre-commit config

cdce8p:
## Description
Updated pre-commit hooks. Removed invalid config option for `black`.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

--

This should be safe to cherry-pick for `2.5.1`.


setup.py
```diff
@@ -22,8 +22,11 @@
 from setuptools.command import easy_install  # pylint: disable=unused-import
 from setuptools.command import install_lib  # pylint: disable=unused-import
 
-if sys.version_info.major == 3 and sys.version_info.minor <=5:
-    warnings.warn("You will soon need to upgrade to python 3.6 in order to use the latest version of Astroid.", DeprecationWarning)
+if sys.version_info.major == 3 and sys.version_info.minor <= 5:
```
Pierre-Sassoulas, line 25: Unrelated to the MR but I think we can now remove that ?

Pierre-Sassoulas (approved):

setup.py
cdce8p: Yeah, absolutely. Didn't even notice it while preparing the PR 😅

cdce8p:
@Pierre-Sassoulas I would suggest to cherry-pick these commits to the `2.5` branch. Since a lot of files have changed, this will prevent further conflicts.

Pierre-Sassoulas:
Yeah, maybe even all the master branch could be directly merged by the look of it.

cdce8p:
That would probably work as well. Could you take a look at #914 if you have some spare time? Might be something we also want to include.

PR #912, opened 2021-02-26
Add ImportAlias node as children of Import nodes

msuozzo:
## Steps

## Description

Since the name and alias metadata was already available as Import.names, the main addition here is of the location metadata associated with the alias. The intent is to generate line- and column-accurate lint locations in pylint for e.g. unused location.

For situations with compound imports (e.g. `from foo import bar, baz, qux`), addressing lints to a specific unused import instead of at the statement granularity as it is now would be a substantial quality-of-life improvement.

The `ImportAlias` node type shadows the `_ast.alias` type in the builtin ast module.

I've left the `Import.names` field to maintain backwards compatibility but it might be nice to change it to type `list(ImportAlias)` in the next major version.

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |


msuozzo:
Any movement here? Should I link to the accompanying pylint change?

hippo91:
@msuozzo thanks for this PR and sorry for the delay. I will try to have a look ASAP.

hippo91 (changes requested):
@msuozzo thanks for this clear development!
Overall it looks good. 
However i think adding tests would gives us more confidence.
Would you mind adding such a test in `unittest_builder.py` or `unittest_raw_building.py`?
By the way i would also have @cdce8p opinion on this.

cdce8p:
I agree, this change should definitely have tests!
As for the code itself, the addition seems reasonable. I don't know what your opinion is @hippo91, but I would prefer to use type annotations for variables (instead of type comments) going forward. Additionally it might be a good idea to define / declare all possible instance variables inside `__init__` and later reassign them, see pylint check for `attribute-defined-outside-init`.

cdce8p:
We will also need to check for unintended side effects with `pylint` before merging it.

Pierre-Sassoulas:
@msuozzo are you still interested in finishing this PR ?

DanielNoord:
Closing this as there no longer seems interest to finish this PR or add tests.

PR #918, opened 2021-03-05
Bug pylint 3974

hippo91:
## Steps

## Description
This PR fixes the way an Import node is detected as a valid `numpy` import. It concerns only the `numpy`'s brain.
The `_is_a_numpy_module` function was bugged because it returns True for `import numpy  as alias` but False for the simplest case `import numpy`.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes PyCQA/pylint#3974

### DIFF
--- a/tests/unittest_brain_numpy_core_multiarray.py
+++ b/tests/unittest_brain_numpy_core_multiarray.py
@@ -77,105 +77,122 @@ class BrainNumpyCoreMultiarrayTest(unittest.TestCase):
         )
         return node.infer()
 
+    def _inferred_numpy_no_alias_func_call(self, func_name, *func_args):
+        node = builder.extract_node(
+            """
+        import numpy
+        func = numpy.{:s}
+        func({:s})
+        """.format(
+                func_name, ",".join(func_args)
+            )
+        )
+        return node.infer()
+
     def test_numpy_function_calls_inferred_as_ndarray(self):
         """
         Test that calls to numpy functions are inferred as numpy.ndarray
         """
-        for func_ in self.numpy_functions_returning_array:
-            with self.subTest(typ=func_):
-                inferred_values = list(self._inferred_numpy_func_call(*func_))
-                self.assertTrue(
-                    len(inferred_values) == 1,
-                    msg="Too much inferred values ({}) for {:s}".format(
-                        inferred_values, func_[0]
-                    ),
-                )
-                self.assertTrue(
-                    inferred_values[-1].pytype() == ".ndarray",
-                    msg="Illicit type for {:s} ({})".format(
-                        func_[0], inferred_values[-1].pytype()
-                    ),
-                )
+        for infer_wrapper in (self._inferred_numpy_func_call, self._inferred_numpy_no_alias_func_call):
+            for func_ in self.numpy_functions_returning_array:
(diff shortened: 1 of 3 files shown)


hippo91:
@cdce8p for an unknown reason, i cannot add you as a reviewer but i would be happy to have your opinion on this PR.

tests/unittest_brain_numpy_core_multiarray.py
```diff
@@ -77,105 +77,137 @@ def _inferred_numpy_func_call(self, func_name, *func_args):
...
+                        ),
+                    )
+                    self.assertTrue(
+                        inferred_values[-1].pytype() == ".ndarray",
+                        msg="Illicit type for {:s} ({})".format(
+                            func_[0], inferred_values[-1].pytype()
+                        ),
+                    )
 
     def test_numpy_function_calls_inferred_as_bool(self):
```
Pierre-Sassoulas, line 116:
```suggestion
INFER_WRAPPERS = [self._inferred_numpy_func_call, self._inferred_numpy_no_alias_func_call]

    @pytest.parametrize("infer_wrapper", INFER_WRAPPERS))
    def test_numpy_function_calls_inferred_as_bool(self, infer_wrapper ):
```
I think using parametrize would make the diff and the code more terse.

tests/unittest_brain_numpy_core_multiarray.py
```diff
@@ -36,7 +36,7 @@ class BrainNumpyCoreMultiarrayTest(unittest.TestCase):
         ("is_busday", "['2011-07-01', '2011-07-02', '2011-07-18']"),
         ("lexsort", "(('toto', 'tutu'), ('riri', 'fifi'))"),
         ("packbits", "np.array([1, 2])"),
-        ("ravel_multi_index", "np.array([[1, 2], [2, 1]])", "(3, 4)"),
+        # ("ravel_multi_index", "np.array([[1, 2], [2, 1]])", "(3, 4)"),
```
Pierre-Sassoulas, line 39: We can remove dead code (I could add an auto-remover in pre-commit btw).

Pierre-Sassoulas (approved):
False positives with numpy are annoying, I like this !

tests/unittest_brain_numpy_core_multiarray.py
cdce8p: Usually I would agree. However this test module still uses `unittest.TestCase` and in particular `self.subTest` which doesn't exist in `pytest` (without an extension). So using `@pytest.mark.parametrize` would require a complete rewrite which probably isn't worth it at this point.

cdce8p (commented):
The change looks reasonable. Although I know too little about astroid, to evaluate it without debugging 😅

I'm not a member of `PyCQA` and don't have any permissions for either repo (astroid and pylint). That's probably why you couldn't request a review from me.

tests/unittest_brain_numpy_core_multiarray.py
hippo91: @Pierre-Sassoulas i didn't know about this `pytest` functionality. Thanks! I will remember it for my next devs.
@cdce8p thanks for your remark. I will let this as it is.

PR #916, opened 2021-02-28
Fix typing alias

cdce8p:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description
Replaces #913
CC: @hippo91 @Pierre-Sassoulas

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #905
Closes PyCQA/pylint#4093
Closes PyCQA/pylint#4131
Closes PyCQA/pylint#4145
Closes PyCQA/pylint#3247
Closes PyCQA/pylint#2717

PR #921, opened 2021-03-14
Bug pylint 4206

hippo91:
## Steps

## Description
This PR:

1. handles simple subscript inference dealing with `__class_getitem__` method. If a class implements this method then the class is subscriptable without the need that its metaclass implements `__getitem__` method. In this PR only FunctionDef nodes are taken into account. Maybe it should be enlarge to other node types such as AssignName but it should be made in another PR
1. For python39 and above, allows the `collections` objects implementing `__class_getitem__` to be detected as subscriptable. This is done in the `collections` brain because the implementation is a bit tricky and without the hack in the brain it would be necessary to improve the way `__class_getitem__` is inferred (see point 1).
1. For python versions lower than 3.9, allows the `typing` types that are alias of `collections` object to be detected as subscriptable. This is done without modifying inheritance tree and without hacking the metaclass. The way chosen here is to add `__class_getitem__` method on the types that are detected as subscriptable. Original way of adding `__getitem__` on the metaclass is not a good idea, because the same metaclass instance is shared between subscriptable and not subscriptable types.
1. Modifies the way `collections` brain hacks are registered so that `pylint`s acceptance tests are ok.

Points 1, 2, and 3 are necessary to fix the bug specified in PyCQA/pylint#4206 but it is not sufficient to make the acceptance test ok. That's why point 4 exists.

I would like to highlight the fact `deque`, `defaultdict` and `OrderedDict` are defined in the `_collections` module which is a C library. In order message dealing with this module (in fact its representation in the brain) to be displayed correctly (with no crash) i have specified the file path as `/tmp/unknown`. It is uggly but i dont know what to do otherwise. Moreover it wont impact many users, except those executing `pylint` against `typing` module (as the `pylint`s acceptance test does).

One last thing, for python versions lower thant 3.9, the case where both `typing` and `collections` modules are imported in the same module may lead to spurious false positive. This is because parsing the `typing` module while give subscript ability to some objects also defined in collection module that are not subscriptable. I don't think it matters because it is not wise, IMHO, to mix those modules. Instead the user should choose between object in collection that are not subscriptable or the same in typing that are subscriptable.

## Type of Changes
| ✓  | :bug: Bug fix  |

## Related Issue
First step towards: PyCQA/pylint#4206


hippo91:
@cdce8p if you could have a look to this PR it would be cool!

astroid/brain/brain_collections.py
```diff
@@ -85,3 +85,42 @@ def __class_getitem__(cls, item): return cls"""
...
+    node.locals["__class_getitem__"] = [func_to_add]
+
+if PY39:
+    # Starting with Python39 some objects of the collection module are subscriptable
+    # thanks to the __class_getitem__ method but the way it is implemented in
+    # _collection_abc makes it difficult to infer. (We would have to handle AssignName inference in the 
+    # getitem method of the ClassDef class) Instead we put here a mock of the __class_getitem__ method
+    astroid.MANAGER.register_transform(
+        astroid.nodes.ClassDef, easy_class_getitem_inference, _looks_like_subscriptable
+    )
```
Pierre-Sassoulas: You can install pre-commit (so black is applied automatically for each commit) with `pre-commit install`

tests/unittest_brain.py
```diff
@@ -1349,17 +1353,12 @@ class CustomTD(TypedDict):
...
-        assert isinstance(meta, nodes.ClassDef)
-        assert meta.name == "ABCMeta"
-
-    @test_utils.require_version("3.8")
+    @test_utils.require_version(minver="3.7")    
     def test_typing_alias_type(self):
         """
         Test that the type aliased thanks to typing._alias function are
         correctly inferred.
+        typing_alias function is introduced with python37
```
Pierre-Sassoulas, line 1364: 👍

Pierre-Sassoulas (approved):
I've made some surface comment, because to be fair I'm not understanding a lot of what is happening here 😄 At least I learned about MutableSet 😄 ! Fixing the typing acceptance fail is nice and I don't think we can avoid making special case for python 3.9. It would be the right time to release astroid 2.5.2 after we merge this, right ?

cdce8p:
@hippo91 I'll definitely take a look at it, hopefully later today. Thanks for taking the time to work on it!

cdce8p:
I had some time to take a closer look and do some tests.

The good news first, I really like the use of `__class_getitem__` instead of `__getitem__` and the change doesn't seem to cause any regression bugs.

However, it doesn't solve the problem completely. Since you add `__class_getitem__` directly to the `collections.abc` class (without an intermediate), those will be subscriptable as well (if the typing module is imported) although they aren't in 3.7 and 3.8. This seems to be more a less the issue I had with #916.

```py
# a.py
import typing
from collections.abc import Iterable

var: Iterable[int]
```

```bash
# python38
pylint a.py
************* Module a
a.py:1:0: W0611: Unused import typing (unused-import)

# python38, without the typing import 
pylint a.py
************* Module a
a.py:3:5: E1136: Value 'Iterable' is unsubscriptable (unsubscriptable-object)
```

**Regarding a solution**
1. Is is possible to repurpose a `Call` node as `ClassDef`? If that's the case, we could use the `_alias` node as an intermediate and add `__class_getitem__` to it. The issue I had was that I create a new `ClassDef` from nothing. That caused the acceptance test to fail.
2. If (**1**) is not possible, we could address this in `pylint`. Since this is limited to mostly the `collections` classes, it might be possible to add some special case handling.

hippo91:
@cdce8p thanks a lot for taking time to review this.
IMHO, the following code:
```
import typing
from collections.abc import Iterable
```
is a bit weird. I don't see any reason to use a container of the collections module if the typing module is already imported.
Why not use directly the corresponding container of the typing module?

hippo91:
@cdce8p you are right concerning PEP 585. But the deprecation should start with python3.9. Isn't it? If yes can we consider that mixing typing and collections with python3.8 and lower is bad practice?

cdce8p:
@hippo91 Sorry for not getting back to you earlier. I had a pretty busy week.

**Regarding your last comment (Mixed usage of `typing` and `collections`)**
I'm don't think we can assume this as bad practice. With postponed evaluation of type annotations (`from __future__ import annotations) is actually something that can happen rather often. As example: [generic_alias_mixed_py37](https://github.com/PyCQA/pylint/blob/c32b8f7152072d5d9dc42736ad01b327dd47a3d8/tests/functional/g/generic_alias/generic_alias_mixed_py37.py).

The example is part of a collection of tests I wrote a few days ago: https://github.com/PyCQA/pylint/pull/4239. After a while, I got so confused about what should be possible and what shouldn't that this was the only way left. I think they can help quite a lot when validating any astroid changes. I just needed to remind myself to also run the acceptance tests frequently. This is what I usually do
```
pytest tests/test_functional -k generic_alias
pytest -m acceptance -k "test_libmodule[typing.py]"
pytest -m acceptance -k "test_libmodule[collections]"
```

I mentioned that I was also looking into a possible solution. I didn't have the time to properly clean it up, but at least the idea is there. You can find it in this draft MR: #923. I would appreciate if you take a look at it. Maybe the idea is worth integrating here. One issue I still have is with the acceptance test. It checks the `ClassDef.list` with `tolineno = 0` which raises an AssertionError on `assert node.tolineno`. This also happens for `ClassDef.set` and `ClassDef.dict`. So basically all builtin types. Types from `collections` and `collections.abc` are fine. Maybe this is an issue with the `builtins` brain?

Lastly, I think we don't need to fix all issues with the tests I wrote, right now. The issue I just mentioned for example is a direct result of adding `nodes.Name` in the `isinstance` call here:
https://github.com/PyCQA/astroid/blob/98178861ad3de8d667e666b13308d9f216b2f804/astroid/brain/brain_typing.py#L137

The corresponding test case
```py
import typing
class DerivedList(typing.List[int]):
    pass
```

Without this change, it results in an `inherit-non-class` error, but maybe we can live with that for now and fix it later.

astroid/brain/brain_collections.py
```diff
@@ -77,7 +77,69 @@ def _ordered_dict_mock():
...
+        def __class_getitem__(cls, item): return cls"""
     return base_ordered_dict_class
 
 
-astroid.register_module_extender(astroid.MANAGER, "collections", _collections_transform)
+def _collections_module_properties(node, context=None):
+    """
+    Adds a path to a fictive file as the _collections module is a pure C lib.
+    """
+    node.file = "/tmp/unknown"
```
AWhetter, line 91: Minor: Could we import the collections module and use it's location to come up with a more realistic location. Maybe `.../_collections.pyi`. Or can we use a placeholder that definitely won't represent a real file. I think we've used `<?>` elsewhere.

astroid/brain/brain_collections.py
```diff
@@ -77,7 +77,69 @@ def _ordered_dict_mock():
...
+
+astroid.MANAGER.register_transform(
+    astroid.Module, _collections_module_properties, lambda n: n.name == "_collections"
+)
+astroid.register_module_extender(
+    astroid.MANAGER, "_collections", _collections_transform
+)
+
+
+PY39 = sys.version_info >= (3, 9)
```
AWhetter, line 103: Minor: This is already at the top of the file. Do we need to define it again?

astroid/brain/brain_collections.py
```diff
@@ -77,7 +77,69 @@ def _ordered_dict_mock():
...
+CLASS_GET_ITEM_TEMPLATE = """
+@classmethod
+def __class_getitem__(cls, item):
+    return cls
+"""
+
+
+def easy_class_getitem_inference(node, context=None):
+    #  Here __class_getitem__ exists but is quite a mess to infer thus
+    #  put instead an easy inference tip
```
AWhetter, line 133: Typo: Should this read "put inside an easy inference tip"?

astroid/scoped_nodes.py
```diff
@@ -2617,7 +2617,22 @@ def getitem(self, index, context=None):
...
-            raise exceptions.AstroidTypeError(node=self, context=context) from exc
+            if isinstance(self, ClassDef):
+                # subscripting a class definition may be
+                #  achieved thanks to __class_getitem__ method
+                #  which is a classmethod defined in the class
+                #  that supports subscript and not in the metaclass
+                try:
+                    methods = self.getattr("__class_getitem__")
+                    # Here it is assumed that the __class_getitem__ node is
+                    #  a FunctionDef. One possible improvment would be to deal
```
AWhetter, line 2628: Typo: "improvement"

AWhetter (approved):
Looks good to me! There's only a few minor suggestions.

astroid/brain/brain_collections.py
hippo91: @Pierre-Sassoulas thanks! I forgot to install it inside my virtual env

astroid/brain/brain_collections.py
hippo91: @AWhetter thanks for your review!
I choose to follow your first advice, so i import the collections module.

astroid/brain/brain_collections.py
hippo91: Definitely useless!

hippo91:
@cdce8p thanks a lot for your feedback and also for all the tests you added in pylint. 
I had not a lot of free time lastly so i did not have time to look at #923 but i will do it ASAP.

cdce8p:
@hippo91 It would be great if we could finish this MR this weekend. The next pylint release would solve a lot of issues and this is the only blocker item. Please let me know how I can help.

Just so you know, I updated the tests in https://github.com/PyCQA/pylint/pull/4239 today, after I found some more issues.

Pierre-Sassoulas (approved):
Huge work you did here ! I think we'll need to squash; but it seems ready to be merged !  Do you want to release the next version of astroid right after @hippo91 ?

AWhetter (approved):

hippo91:
@Pierre-Sassoulas @cdce8p @AWhetter IMHO it is way to soon to merge this in the master.
There are still bugs that need to be addressed, such as false negatives when typing module and collections module are imported together.
@cdce8p have you make progress on this last item?
I did not had time to.

cdce8p:
@hippo91 Unfortunately not really.  I still have some issues with the `builtins` module.

As for your last changes, I really like the idea to monkey-patch  `node.getattr`. That at least solves `typing.ByteString`.
This did not however improve anything with the collections types. I'm still wondering what the best approach here is. It might really make sense to insert an **alias** class for each type that gets the `__class_getitem__` method. That does solve most issues in my MR #923.

Would it help if we chat more about that? If you like, I could setup a Discord server. That could offer faster communication.
Maybe this is something that could even be used for general `astroid` and `pylint` dev communication, @Pierre-Sassoulas ?

Pierre-Sassoulas:
Good idea. There is a mailing list or something, but I never use it personally. We'll update the documentation later : https://discord.gg/q5VWQEaG

astroid/brain/brain_typing.py
```diff
@@ -165,6 +141,29 @@ def _looks_like_typing_alias(node: nodes.Call) -> bool:
...
+        # protocol defined in collections module) whereas the typing module consider it should not
+        origin_func: node.getattr
+        #  We do not want __class_getitem__ to be found in the classdef
+        def raiser(attr):
+            if attr == "__class_getitem__":
+                raise AttributeInferenceError("__class_getitem__ access is not allowed")
+            else:
+                return origin_func(attr)
+
+        node.getattr = raiser
```
cdce8p, line 162:
```suggestion
        #  We do not want __class_getitem__ to be found in the classdef
        def raiser(origin_func, attr, *args, **kwargs):
            if attr == "__class_getitem__":
                raise AttributeInferenceError("__class_getitem__ access is not allowed")
            else:
                return origin_func(attr, *args, **kwargs)

        node.getattr = partial(raiser, node.getattr)
```

I needed to use `partial` (from `functools`) here to get the acceptance test to pass.

hippo91: Totally agree! I came to the same conclusion. Thanks!

hippo91:
@Pierre-Sassoulas @cdce8p i think it is almost good now!
With last modifications:
- with `python3.9`, the generic alias test of PyCQA/pylint#4239 are all ok. There is just a weird thing. If i run:
   `pytest tests/test_functional.py -k generic_alias` there is a failure on collections tests. Whereas if i run:
   `pytest tests/test_functional.py -k generic_alias_collections`, i got no errors;
- with `python3.7`, the generic alias test are ok except for `Pattern` and `Match` which should need a specific brain.

@cdce8p in PyCQA/pylint#4239 i needed to change in `base.py` the `collections.deque` and `collections.defaultdict` in `_collections.deque` and `_collections.defaultdict`.

With this PR all pylint's acceptance tests are fines (at least on my machine :-) ). 

Moreover it seems that the problem with mixing `collections` and `typing` modules is gone away, probably thanks to the use of `inference_tip` function.

For example linting:
```
# pylint:disable=missing-module-docstring, invalid-name
import typing
import collections.abc


A = collections.abc.ByteString[int]
C = collections.abc.Iterable[int]
D = collections.abc.List[int]
B = typing.ByteString[int]
E = typing.Iterable[int]
D = typing.List[int]
H = typing.Dict[int]
peillexg@UN00308016:
```
gives, with `python3.7`:
```
test.py:6:4: E1136: Value 'collections.abc.ByteString' is unsubscriptable (unsubscriptable-object)
test.py:7:4: E1136: Value 'collections.abc.Iterable' is unsubscriptable (unsubscriptable-object)
test.py:8:4: E1101: Module 'collections.abc' has no 'List' member (no-member)
test.py:9:4: E1136: Value 'typing.ByteString' is unsubscriptable (unsubscriptable-object)
```
and with `python3.9`:
```
test.py:8:4: E1101: Module 'collections.abc' has no 'List' member (no-member)
test.py:9:4: E1136: Value 'typing.ByteString' is unsubscriptable (unsubscriptable-object)
```
I think it is correct.

hippo91:
> @hippo91 Unfortunately the latest change doesn't improve much on my part. I'll need more time to figure out what exactly the issue are, but most tests aren't passing. Maybe fetch the latest changes? As for the file you posted, for `py37`:
> 
> ```python
> D = collections.abc.List[int]  # this should be `unsubscriptable-object`, not `no-member`
> ```
> 
> and for `py39`:
> 
> ```python
> D = collections.abc.List[int]  # no issue, just like `collections.abc.Iterable`
> ```

@cdce8p it seems that there is not `List` object in `collections.abc`. 

```
ipython
Python 3.9.1 (default, Mar  5 2021, 20:11:01) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import collections.abc

In [2]: collections.abc.List
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-832f6e94cf81> in <module>
----> 1 collections.abc.List

AttributeError: module 'collections.abc' has no attribute 'List'
```

Thus i think that `no-member` is correct.

cdce8p:
True, my mistake. It's a builtin `list`.
I'll need to go through my test results, before I say anymore. It's just that I saw a lot of red

cdce8p:
@hippo91 I think I've found something now. The strange part was that when I ran the tests with
```bash
pytest tests/test_functional.py -k generic_alias
```
I saw a whole lot of errors that miraculously disappeared when I only ran a single file.

--

Try this example (Python 3.7 / 3.8)
```py
import typing

A = list[int]  # [unsubscriptable-object]
B = typing.List[int]
```
The result is
```bash
test2.py:3:4: E1136: Value 'list' is unsubscriptable (unsubscriptable-object)
```
If you now switch lines 3 and 4, the error will disappear.

--

I'd guess the root cause here might be caching on `pylints` part. Wouldn't it be easier to just insert intermediate classes?

astroid/brain/brain_typing.py
```diff
@@ -161,10 +137,43 @@ def _looks_like_typing_alias(node: nodes.Call) -> bool:
...
 
+def full_raiser(origin_func, attr, *args, **kwargs):
+    """
+    Raises an AttributeInferenceError in case of access to __class_getitem__ method.
+    Otherwise just call origin_func.
+    """
+    if attr == "__class_getitem__":
+        raise AttributeInferenceError("__class_getitem__ access is not allowed")
+    else:
+        return origin_func(attr, *args, **kwargs)
```
cdce8p, line 157: I would probably put this inside of `_forbid_class_getitem_access`. It's only ever used there.

astroid/brain/brain_typing.py
```diff
@@ -179,33 +188,48 @@ def infer_typing_alias(
     res = next(node.args[0].infer(context=ctx))
```
cdce8p, line 188: You didn't change it, but I noticed that the line above is redundant:
```py
if not isinstance(node, nodes.Call):
    return None
```
This will never match. `node` is always a `nodes.Call` instance.

astroid/brain/brain_typing.py
```diff
@@ -179,33 +188,48 @@ def infer_typing_alias(
...
+                # The typing object is subscriptable if the second argument of the _alias function
+                # is a TypeVar or a tuple of TypeVar. We could check the type of the second argument but
+                #  it appears that in the typing module the second argument is only TypeVar or a tuple of TypeVar or empty tuple.
+                # This last value means the type is not Generic and thus cannot be subscriptable
+                func_to_add = astroid.extract_node(CLASS_GETITEM_TEMPLATE)
+                res.locals["__class_getitem__"] = [func_to_add]
+            elif (
+                isinstance(maybe_type_var, node_classes.Tuple)
+                and not maybe_type_var.elts
+            ):
```
cdce8p, line 216: Can't this just be a simple `else`? Either we add `__class_getitem__` or we need to forbid the access.

astroid/scoped_nodes.py
```diff
@@ -2627,6 +2644,19 @@ def getitem(self, index, context=None):
...
+            #  However in such case the method is bound to an EmptyNode and
+            #  EmptyNode doesn't have infer_call_result method yielding to
+            #  AttributeError
+            if (
+                isinstance(method, node_classes.EmptyNode)
+                and self.name in ("list", "dict", "set", "tuple", "frozenset")
+                and PY39
+            ):
+                return self
+            raise
```
cdce8p, line 2659: I didn't need that for my solution. Maybe not needed here as well?

cdce8p (changes requested):
I really like the use of `inference_tip`! It seems this was the missing piece.

Nevertheless, I think `intermediate` classes solve a lot of the remaining issues ;)

astroid/scoped_nodes.py
cdce8p: **Update**: It is necessary. Please ignore my previous comment.

tests/unittest_brain.py
```diff
@@ -1266,60 +1439,111 @@ class Derived2(typing.OrderedDict[int, str]):
...
+            src = """
+            import typing
+            typing.{:s}[int]
+            """.format(
+                typename
+            )
+            right_node = builder.extract_node(src)
+            inferred = next(right_node.infer())
+            self.assertIsInstance(inferred, ClassDef)
+            self.assertIsInstance(inferred.getattr("__iter__")[0], FunctionDef)
```
cdce8p, line 1546: `ClassDef` and `FunctionDef` are also accessible through `nodes.ClassDef` and `nodes.FunctionDef`. That would be more consistent with your other tests.

tests/unittest_brain.py
```diff
@@ -1026,6 +1027,183 @@ def test_invalid_type_subscript(self):
...
+        for typename in ("tuple", "list", "dict", "set", "frozenset"):
+            src = """
+            {:s}[int]
+            """.format(
+                typename
+            )
+            right_node = builder.extract_node(src)
+            inferred = next(right_node.infer())
+            self.assertIsInstance(inferred, ClassDef)
+            self.assertIsInstance(inferred.getattr("__iter__")[0], FunctionDef)
```
cdce8p, line 1044: Same here: `nodes.ClassDef`, `nodes.FunctionDef`

hippo91:
@cdce8p i took into account your remarks. Thanks!
For the remaining problem of simultaneous import of `typing` and `collections` module, i would prefer a hack into `pylint` over introducing intermediate classes in the hierarchy. IMHO it is a non worthy complication to handle just what i think is a corner case.
For example in the `pylint`'s `utils` module i modified the `supports_getitem` method, to something along those lines:
```
def supports_getitem(
    value: astroid.node_classes.NodeNG, node: astroid.node_classes.NodeNG
) -> bool:
    def is_typing_imported(node: astroid.node_classes.NodeNG):
        """
        Returns True if the typing module is imported
        """
        rt = node.root()
        import_nodes = rt.nodes_of_class((astroid.Import,))
        for i_node in import_nodes:
            if 'typing' in i_node.names[0]:
                return True
        return False

    def is_name_imported_from_collections(node: astroid.node_classes.NodeNG):
        """
        Returns True if the Name node is imported from collections module
        """
        cls_name = node.value
        if isinstance(cls_name, astroid.Name):
            lookup_res = cls_name.lookup(cls_name.name)
            if lookup_res:
                for l_res in lookup_res[1]:
                    if l_res.modname.startswith("collections"):
                        return True
        return False

    def is_attribute_from_collections(node: astroid.node_classes.NodeNG):
        """
        Returns True if the Attribute node is linked to collections module
        """
        cls_name = node.value
        if isinstance(cls_name, astroid.Attribute):
            for lmod in cls_name.expr.inferred():
                if lmod.name.startswith('collections'):
                    return True
        return False

    if is_typing_imported(node) and PY39_LESS and isinstance(node, astroid.Subscript):
        if is_name_imported_from_collections(node) or is_attribute_from_collections(node):
            return False
    if isinstance(value, astroid.ClassDef):
        if _supports_protocol_method(value, CLASS_GETITEM_METHOD):
            return True
        if is_class_subscriptable_pep585_with_postponed_evaluation_enabled(value, node):
            return True
    return _supports_protocol(value, _supports_getitem_protocol)
```
And it seems to give good results even if it should probably be improved.
@cdce8p @AWhetter @Pierre-Sassoulas what do you think about it?

cdce8p:
@hippo91 Unfortunately your suggestion doesn't work in `pylint` and even introduces regression bugs. I also don't agree  with your assessment that it's only a small corner case. IMO it's much more than that. In PY37 and 38 the use of collection classes is only possible in some very limited circumstances. A user, who doesn't know about that, might attempt to just replace all typing imports which will cause difficult to find bugs. This is one of the key things I believe `pylint` should be able to detect.

As for the intermediate classes. The implementation is much simpler in `astroid` as you probably think. Take a look at https://github.com/cdce8p/astroid/compare/bug_squash_merged...cdce8p:type-alias-rebased. These are all changes that  would be necessary on top of your last commit to fully support the collection classes without the need to change anything in pylint. If you like, I can do this as a followup after your MR is merged.

It might seem like these intermediate classes are a hack, but honestly it's it more similar to what `cpython` actually does. It also inserts new classes in the hierarchy with `_GenericAlias`.

Lastly, it would also enable the possibility to do deprecation detection for the `typing` aliases (for PY39+), since we would be able to easily differentiate `typing` aliases from `collections` classes.

hippo91:
@cdce8p you convinced me! You are right concerning the fact that your implementation is not so complex and it mimics the cpython trick with `_GenericAlias`. My apologies for being so long to convince!
I would appreciate if you could add this as a followup as you proposed.

cdce8p (approved):

cdce8p:
@hippo91 That sound like a plan! I'm ready to get this finally done 🚀
FYI: With the last commit, I've pushed some small style changes to reduce the diff for the followup MR.

PR #923, opened 2021-03-19
Better handling of generic aliases

cdce8p:
## Steps

## Description
Followup to #921, see: https://github.com/PyCQA/astroid/pull/921#issuecomment-813937201
Improve handling of typing aliases by inserting intermediate classes in mro.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes: https://github.com/PyCQA/pylint/issues/4206
Closes: https://github.com/PyCQA/pylint/issues/4284

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -163,6 +163,11 @@ def _builtin_filter_predicate(node, builtin_name):
         and node.parent.targets[0].name in ("Pattern", "Match")
     ):
         # Handle re.Pattern and re.Match in brain_re
+        # Match these patterns from stdlib/re.py
+        # ```py
+        # Pattern = type(...)
+        # Match = type(...)
+        # ```
         return False
     if isinstance(node.func, nodes.Name) and node.func.name == builtin_name:
         return True
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -19,7 +19,6 @@
 """Astroid hooks for various builtins."""
 
 from functools import partial
-from textwrap import dedent
 
 from astroid import (
     MANAGER,
@@ -153,6 +152,18 @@ _extend_builtins(
 
 
 def _builtin_filter_predicate(node, builtin_name):
+    if (
+        builtin_name == "type"
+        and node.root().name == "re"
+        and isinstance(node.func, nodes.Name)
+        and node.func.name == "type"
+        and isinstance(node.parent, nodes.Assign)
+        and len(node.parent.targets) == 1
+        and isinstance(node.parent.targets[0], nodes.AssignName)
+        and node.parent.targets[0].name in ("Pattern", "Match")
+    ):
+        # Handle re.Pattern and re.Match in brain_re
+        return False
     if isinstance(node.func, nodes.Name) and node.func.name == builtin_name:
         return True
     if isinstance(node.func, nodes.Attribute):
--- a/astroid/brain/brain_re.py
+++ b/astroid/brain/brain_re.py
@@ -47,7 +47,14 @@ def __class_getitem__(cls, item):
 
 
 def _looks_like_pattern_or_match(node: nodes.Call) -> bool:
-    """Check for re.Pattern or re.Match call in stdlib."""
+    """Check for re.Pattern or re.Match call in stdlib.
+
+    Match these patterns from stdlib/re.py
+    ```py
+    Pattern = type(...)
+    Match = type(...)
+    ```
+    """
     return (
         node.root().name == "re"
         and isinstance(node.func, nodes.Name)
--- a/astroid/brain/brain_re.py
+++ b/astroid/brain/brain_re.py
@@ -2,8 +2,11 @@
 # For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
 import sys
 import astroid
+from astroid import MANAGER, inference_tip, nodes, context
 
 PY36 = sys.version_info >= (3, 6)
+PY37 = sys.version_info[:2] >= (3, 7)
+PY39 = sys.version_info[:2] >= (3, 9)
 
 if PY36:
     # Since Python 3.6 there is the RegexFlag enum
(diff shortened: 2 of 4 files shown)


astroid/brain/brain_typing.py
```diff
@@ -161,7 +179,7 @@ def _looks_like_typing_alias(node: nodes.Call) -> bool:
         isinstance(node, nodes.Call)
         and isinstance(node.func, nodes.Name)
         and node.func.name == "_alias"
-        and isinstance(node.args[0], nodes.Attribute)
+        and isinstance(node.args[0], (nodes.Attribute, nodes.Name))
```
cdce8p, line 182: This change is necessary, but causes many tests to fail (pylint acceptance "typing", astroid). Might be related to the brain for `builtins`.

hippo91: @cdce8p you are right it is necessary. I found a fix for this. It will be part of my next push on #921.

cdce8p: @hippo91 Sounds good! I'm looking forward to it.

cdce8p:
/CC: @hippo91, @Pierre-Sassoulas

As discussed, this is the followup to #921. It addresses almost all test cases defined in https://github.com/PyCQA/pylint/pull/4239 with the exception that it can't detect the wrong usage of `re.Pattern` and `re.Match` in PY37 and 38. See example below. This would require changes to the `brain_re.py` which is out of scope for this MR. (Probably not worth it to begin with)
```py
import re
re.Pattern[str]       # [unsubscriptable-object]
re.Match[str]         # [unsubscriptable-object]
var: re.Pattern[int]  # [unsubscriptable-object]
```

tests/unittest_brain.py
```diff
@@ -1398,12 +1420,12 @@ class Derived1(MutableSet[T]):
         """
         )
         inferred = next(node.infer())
-        check_metaclass_is_abc(inferred)
         assertEqualMro(
             inferred,
             [
                 "Derived1",
                 "MutableSet",
+                "MutableSet",
```
Pierre-Sassoulas, line 1428: Could this be a set with unique value ?

cdce8p: It basically is. If you resolve the ClassDef with `qname`, you would get
```py
typing.MutableSet
collections.abc.MutableSet
```
`assertEqualMro` only checks the `name` of the `ClassDef`. That's why they appear the same.

Pierre-Sassoulas: Ho, ok I get it now. We could do the test on `qname` so the test is more explicit then ?

cdce8p: I've updated the tests to use `qname()`

cdce8p:
**Small update**: I've updated the brain for the `re` module. With that, **all** test cases are passing now!

cdce8p:
If there are no objections, I would like to merge this by this weekend. After a new `astroid` release, we would finally be able to close this chapter.

Pierre-Sassoulas:
I think @hippo91 is the one that will really understand what he's reviewing :D I did not see anything wrong with the changes myself.

astroid/brain/brain_builtin_inference.py
```diff
@@ -153,6 +152,18 @@ def _extend_builtins(class_transforms):
...
+        builtin_name == "type"
+        and node.root().name == "re"
+        and isinstance(node.func, nodes.Name)
+        and node.func.name == "type"
+        and isinstance(node.parent, nodes.Assign)
+        and len(node.parent.targets) == 1
+        and isinstance(node.parent.targets[0], nodes.AssignName)
+        and node.parent.targets[0].name in ("Pattern", "Match")
+    ):
+        # Handle re.Pattern and re.Match in brain_re
```
hippo91, line 165: I think a comment with an example of what is filtered could help here.

astroid/brain/brain_re.py
```diff
@@ -34,3 +37,43 @@ def _re_transform():
...
+CLASS_GETITEM_TEMPLATE = """
+@classmethod
+def __class_getitem__(cls, item):
+    return cls
+"""
+
+
+def _looks_like_pattern_or_match(node: nodes.Call) -> bool:
+    """Check for re.Pattern or re.Match call in stdlib."""
+    return (
```
hippo91, line 51: Idem here, a simple example in the comment could help to understand

astroid/brain/brain_typing.py
```diff
@@ -88,7 +133,13 @@ def infer_typing_attr(node, context=None):
...
         raise UseInferenceDefault from exc
 
-    if not value.qname().startswith("typing."):
+    if (
+        not value.qname().startswith("typing.")
+        or PY37
+        and value.qname() in TYPING_ALIAS
+    ):
+        # If typing subscript belongs to an alias
+        # (PY37+) handle it separately.
```
hippo91, line 142: Well done. I do not know how i can have missed that.

astroid/brain/brain_typing.py
```diff
@@ -174,55 +223,54 @@ def full_raiser(origin_func, attr, *args, **kwargs):
...
+    ):
+        # If typing alias is subscriptable, add `__class_getitem__` to ClassDef
+        func_to_add = astroid.extract_node(CLASS_GETITEM_TEMPLATE)
+        class_def.locals["__class_getitem__"] = [func_to_add]
+    else:
+        # If not, make sure that `__class_getitem__` access is forbidden.
+        # This is an issue in cases where the aliased class implements it,
+        # but the typing alias isn't subscriptable. E.g. `typing.ByteString` for PY39+
+        _forbid_class_getitem_access(class_def)
+    return iter([class_def])
```
hippo91, line 273: I like very much this concise style! :+1: 
However you delete a lot of comments. It is not a pb for me if you think they where not helpful.

hippo91 (approved):
Well done @cdce8p . I am ok with this PR. Just left a bunch of comments. Feel free to address them or not!

astroid/brain/brain_typing.py
cdce8p: I tend to prefer less / only the necessary comments. Most information is contained in the code already. Comments also have the habit of becoming outdated pretty quickly at which point they are more confusion than actually helpful.

cdce8p:
@hippo91 Thanks for the quick review!

PR #928, opened 2021-04-01
Use inference_tip for typing.TypedDict brain

cdce8p:
## Steps

## Description
The typing brain should use `inference_tip` for `TypedDict` instead of a normal transform.
The pylint tests continue to pass with this change.

## Type of Changes

|   | Type |

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -104,17 +104,15 @@ def _looks_like_typedDict(  # pylint: disable=invalid-name
 
 def infer_typedDict(  # pylint: disable=invalid-name
     node: nodes.FunctionDef, ctx: context.InferenceContext = None
-) -> None:
+) -> typing.Iterator[nodes.ClassDef]:
     """Replace TypedDict FunctionDef with ClassDef."""
     class_def = nodes.ClassDef(
         name="TypedDict",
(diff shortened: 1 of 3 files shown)


cdce8p:
@hippo91 Would you mind taking a quick look at this one? After you used `inference_tip`, I realized that it's probably a better fit here too.

astroid/brain/brain_typing.py
```diff
@@ -107,13 +107,11 @@ def infer_typedDict(  # pylint: disable=invalid-name
     """Replace TypedDict FunctionDef with ClassDef."""
     class_def = nodes.ClassDef(
         name="TypedDict",
-        doc=node.doc,
         lineno=node.lineno,
```
hippo91, line 110: Why is it necessary to remove the doc?

hippo91 (approved):
@cdce8p it seems to be a better implementation. Well done.

astroid/brain/brain_typing.py
cdce8p: Tbh I don't really know why I added it in the first place. AFAIK it isn't necessary.

hippo91: Ok let's remove it so. :+1:

PR #931, opened 2021-04-07
Modify infernce tip for typing.Generic and typing.Annotated with ``__class_getitem__``

cdce8p:
## Steps

## Description
With Python 3.7, `typing.Generic` is now subscriptable through `__class_getitem__`. Same for `typing.Annotated` which was added in 3.9. As with the `__class_getitem__` methods for the `collections.abc` classes, astroid needs a bit of help inferring them correctly.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
This will address: https://github.com/PyCQA/pylint/issues/2822
I would like to add a test case to `pylint` before closing it, though.

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -160,6 +160,16 @@ def infer_typing_attr(
         # infer the native methods, replace them for an easy inference tip
         func_to_add = astroid.extract_node(CLASS_GETITEM_TEMPLATE)
         value.locals["__class_getitem__"] = [func_to_add]
+        if (
+            isinstance(node.parent, nodes.ClassDef)
+            and node in node.parent.bases
+            and getattr(node.parent, "__cache", None)
+        ):
+            # node.parent.slots is evaluated and cached before the inference tip
+            # is first applied. Remove the last result to allow a recalculation of slots
+            cache = getattr(node.parent, "__cache")
+            if cache and cache.get(node.parent.slots) is not None:
+                del cache[node.parent.slots]
         return iter([value])
 
     node = extract_node(TYPING_TYPE_TEMPLATE.format(value.qname().split(".")[-1]))
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -88,6 +88,12 @@ TYPING_ALIAS = frozenset(
     )
 )
 
+CLASS_GETITEM_TEMPLATE = """
+@classmethod
+def __class_getitem__(cls, item):
+    return cls
+"""
+
 
 def looks_like_typing_typevar_or_newtype(node):
     func = node.func
@@ -126,7 +132,9 @@ def _looks_like_typing_subscript(node):
     return False
 
 
-def infer_typing_attr(node, context=None):
+def infer_typing_attr(
+    node: nodes.Subscript, ctx: context.InferenceContext = None
+) -> typing.Iterator[nodes.ClassDef]:
     """Infer a typing.X[...] subscript"""
     try:
         value = next(node.value.infer())
@@ -142,8 +150,20 @@ def infer_typing_attr(node, context=None):
         # (PY37+) handle it separately.
         raise UseInferenceDefault
 
+    if (
+        PY37
+        and isinstance(value, nodes.ClassDef)
+        and value.qname() in ("typing.Generic", "typing.Annotated")
+    ):
+        # With PY37+ typing.Generic and typing.Annotated (PY39) are subscriptable
+        # through __class_getitem__. Since astroid can't easily
+        # infer the native methods, replace them for an easy inference tip
+        func_to_add = astroid.extract_node(CLASS_GETITEM_TEMPLATE)
+        value.locals["__class_getitem__"] = [func_to_add]
+        return iter([value])
+
     node = extract_node(TYPING_TYPE_TEMPLATE.format(value.qname().split(".")[-1]))
-    return node.infer(context=context)
+    return node.infer(context=ctx)
 
 
 def _looks_like_typedDict(  # pylint: disable=invalid-name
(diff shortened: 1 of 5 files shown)


cdce8p:
During testing, I noticed two related issues with `typing.Generic` which I've also addressed.
1. The `inferrence_tip` for `Generic` isn't loaded if the inheriting class is inferred, only after the subscript itself is. `pylint` already handles it. However, `astroid` uses caching, among other things for `slots`, which led to an incorrect result. To fix it, I deleted the cached result for `slots` inside the `inferrence_tip`.
2. The MRO for `Generic` wasn't correct. `Generic` is handled specially in Python as it's possible to inherit from it multiple times which isn't the case for normal classes. Previously `astroid` didn't account for that. Consider the following example:
```py
T = TypeVar('T')
class A:
    pass
class B(Generic[T]):
    pass
class C(A, B[T], Generic[T]):
    pass

print(C.__mro__)

# The result will be
(<class '__main__.C'>, <class '__main__.A'>, <class '__main__.B'>, <class 'typing.Generic'>, <class 'object'>)

# Previously, astroid would have resolved this to
['.C', '.A', '.B', '.Generic', '.Generic', 'builtins.object']

# Now
['.C', '.A', '.B', 'typing.Generic', 'builtins.object']
```

astroid/brain/brain_typing.py
```diff
@@ -142,8 +150,31 @@ def infer_typing_attr(node, context=None):
...
+    ):
+        # With PY37+ typing.Generic and typing.Annotated (PY39) are subscriptable
+        # through __class_getitem__. Since astroid can't easily
+        # infer the native methods, replace them for an easy inference tip
+        func_to_add = astroid.extract_node(CLASS_GETITEM_TEMPLATE)
+        value.locals["__class_getitem__"] = [func_to_add]
+        if (
+            isinstance(node.parent, nodes.ClassDef)
+            and node in node.parent.bases
+            and getattr(node.parent, "__cache", None)
```
hippo91, line 167: Is it not redundant with the test in line 172 `if cache` ?

hippo91: By the way it is a nice catch. How do you find this?

astroid/scoped_nodes.py
```diff
@@ -103,6 +103,39 @@ def _c3_merge(sequences, cls, context):
     return None
 
 
+def clean_typing_generic_mro(sequences: List[List["ClassDef"]]) -> None:
+    """typing.Generic is allowed to appear multiple times in the initial mro.
+    The final one however, MUST only contain ONE.
```
hippo91, line 108: The terms initial and final are a bit confusing. What are their meanings?

astroid/scoped_nodes.py
```diff
@@ -103,6 +103,39 @@ def _c3_merge(sequences, cls, context):
...
+
+    This method will check if Generic is in inferred_bases, but also
+    part of bases_mro. If true, remove it from inferred_bases
+    as well as its entry the bases_mro.
+
+    Format sequences: [[self]] + bases_mro + [inferred_bases]
```
hippo91, line 114: I wonder if it could not be clearer to pass 3 arguments instead of one (i.e self, bases_mro and inferred_bases).
What do you think about it?

astroid/scoped_nodes.py
```diff
@@ -103,6 +103,39 @@ def _c3_merge(sequences, cls, context):
...
+    """
+    pos_generic_in_main_bases = -1
+    # Check if part of inferred_bases
+    for i, base in enumerate(sequences[-1]):
```
hippo91, line 118: I think it would be definitely clearer to manipulate self (or an alias), bases_mro and inferred bases instead of dealing with only one sequence and multiples indexes.

hippo91 (approved):
Great job @cdce8p . I just left a couple of comments. The most important IMHO is the one dealing with `clean_typing_generic_mro`.

astroid/brain/brain_typing.py
cdce8p: > Is it not redundant with the test in line 172 `if cache` ?

👍🏻 

> By the way it is a nice catch. How do you find this?

A bit by surprise tbh. I usually disable the cache for debugging and noticed that the test was passing after I did that. I wasn't expecting that, that's for sure.

astroid/scoped_nodes.py
cdce8p: I would like to leave `list(clean_duplicates_mro(unmerged_mro, self, context))` in `_compute_mro` untouched. This should be run before the MRO is cleaned from Generics. However, that also means we only have access to the whole sequence. Creating references inside `clean_typing_generic_mro` might work though.

cdce8p:
@hippo91 I've pushed a new commit to address your comments.

**clean_typing_generic_mro**
Unfortunately, I couldn't pass `self`, `bases_mro`, and `inferred_bases` individually, since (as explained earlier) I would like `clean_duplicates_mro` to run first. It was however possible to define aliases inside the function, so it should now be at least a bit easier to understand.

I've also redone the comment. As you said, `initial` and `final` weren't as clear as they could have been. `initial` should have referred to the MRO before the c3_merge and `final` after.

cdce8p:
@hippo91 Do you had time to take it a short look at the changes yet? I was thinking that we could release a new patch version of `astroid` after this MR is merged.

/CC: @Pierre-Sassoulas What is your opinion?

Pierre-Sassoulas:
Yes, I think we should release as soon as possible because the packaging changes in pylint could cause problems we did not anticipate and the less other bugs we get handle for the next release the better.

hippo91 (approved):
Thanks @cdce8p ! It is clearer! :+1:

PR #933, opened 2021-04-07
Fix inference of instance attributes defined in base class

### LINKED ISSUE #932: Instance attributes defined in parent classes can raise InferenceError
### Steps to reproduce

Inferring an attribute on a class instance where the instance attribute is defined in a base class will fail, but
inferring that same attribute via `igetattr` on an `astroid.Instance` will succeed.

```python
import astroid
attr = astroid.extract_node("""
class Parent:
    def __init__(self):
        self.data = 123

class Test(Parent):
    pass
t = Test()
t.data
""")
# next(attr.infer())
# raises astroid.exceptions.InferenceError: Inference failed for <Attribute.data l.9 at 0x...>.
instance = next(attr.expr.infer())
print(next(instance.igetattr('data')))
# Const.int(value=123)
```

Discovered while debugging #904. I couldn't find an exact match for this issue before, but probably https://github.com/PyCQA/pylint/issues/960 is the same issue.

### Current behavior

As above

### Expected behavior

Inferring attribute should return `Const.int(value=123)`

2.6.0-dev0

nelfin:
## Steps

## Description

Ref #932. The cause of this appears to be failing to specify the
`lookupname` on the inference context in `BaseInstance.igetattr`.
Additionally, the context path key was not set in a manner that matched
other usages. This appears to date back to at least 2015 when context
path keys were changed. The usage as a tuple instead of a node with
implict `lookupname` was changed last in 3d342e85 but the lookup key was
preserved throughout various changes since then.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #932

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -4050,6 +4050,42 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         assert not isinstance(inferred, nodes.ClassDef)  # was inferred as builtins.type
         assert inferred is util.Uninferable
 
+    def test_infer_subclass_attr_instance_attr_indirect(self):
+        node = extract_node(
+            """
+        class Parent:
+            def __init__(self):
+                self.data = 123
+
+        class Test(Parent):
+            pass
+        t = Test()
+        t
+        """
+        )
+        inferred = next(node.infer())
+        assert isinstance(inferred, Instance)
+        const = next(inferred.igetattr("data"))
+        assert isinstance(const, nodes.Const)
+        assert const.value == 123
+
+    def test_infer_subclass_attr_instance_attr(self):
+        node = extract_node(
+            """
+        class Parent:
+            def __init__(self):
+                self.data = 123
+
+        class Test(Parent):
+            pass
+        t = Test()
+        t.data
+        """
+        )
+        inferred = next(node.infer())
+        assert isinstance(inferred, nodes.Const)
+        assert inferred.value == 123
+
 
 class GetattrTest(unittest.TestCase):
     def test_yes_when_unknown(self):
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,10 @@ What's New in astroid 2.6.0?
 ============================
 Release Date: TBA
 
+* Fix inference of instance attributes defined in base classes
+
+  Closes #932
+
 * Do not set instance attributes on builtin object()
 
  Closes #945
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -207,8 +207,9 @@ class BaseInstance(Proxy):
         if not context:
             context = contextmod.InferenceContext()
         try:
+            context.lookupname = name
             # avoid recursively inferring the same attr on the same class
-            if context.push((self._proxied, name)):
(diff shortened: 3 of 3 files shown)


cdce8p:
@nelfin Would you mind resolving the merge conflicts? This PR seems like one I can review within the next few days.

tests/unittest_inference.py
```diff
@@ -4050,6 +4050,42 @@ def inner():
...
+                self.data = 123
+
+        class Test(Parent):
+            pass
+        t = Test()
+        t
+        """
+        )
+        inferred = next(node.infer())
+        # self.assertIsInstance(inferred, bases.Instance)
```
cdce8p, line 4067: Why comment it out? Either test or remove it. The test should be fine

tests/unittest_inference.py
```diff
@@ -4050,6 +4050,42 @@ def inner():
...
+        const = next(inferred.igetattr("data"))
+        self.assertIsInstance(const, nodes.Const)
+        self.assertEqual(const.value, 123)
```
cdce8p, line 4070: Can you use `assert isinstance` and `assert .. == ..` instead? We try to move more towards `pytest` and away from `unittest`. After you changed it, you probably need to add `@staticmethod` as well.

Same for the test case below.

cdce8p (changes requested):
The change itself LGTM. Just some minor comments regarding the tests.

tests/unittest_inference.py
nelfin: Sure, I wasn't sure what the prevailing style was at the time so I just tried to match the surrounding tests

tests/unittest_inference.py
nelfin: Just cruft I overlooked

cdce8p (approved):
Thanks 🐬

nelfin:
I know there's no perfect way to manage pull requests on GitHub, but I don't like how it disregarded all of the commit message and instead tells people to refer to this PR link. I'd rather that those messages stayed in the history.

cdce8p:
> I know there's no perfect way to manage pull requests on GitHub, but I don't like how it disregarded all of the commit message and instead tells people to refer to this PR link. I'd rather that those messages stayed in the history.

Not completely sure about that. Up until now, when I merged something I decided on the go what made sense to keep. For this one, as it's just a small bugfix, I chose to remove it. I almost never look at a commit message outside of Github history where the PR is also available so I'm not the best one to judge. If you like, I can try to keep yours if I happen to merge any other of your PRs.

nelfin:
> Not completely sure about that. Up until now, when I merged something I decided on the go what made sense to keep. For this one, as it's just a small bugfix, I chose to remove it.

That's fine, it's not likely to cause too much confusion. To me, the real value in a small bugfix is not the lines itself, but the explanation as to why things broke.

> I almost never look at a commit message outside of Github history where the PR is also available so I'm not the best one to judge.

This repository is actually older than GitHub itself so it's probably prudent to keep a "backup".

> If you like, I can try to keep yours if I happen to merge any other of your PRs.

If it's not too much trouble.

Issue #925, opened 2021-03-29
Problem with partial and pickle


renefritze:
I'm not using astroid directly, but discovered this issue [downstream](https://github.com/readthedocs/sphinx-autoapi/issues/264) using sphinx-autoapi.
It seems I've run into an edge case with `partial` and the `pickle` module. I've added a new test case to `astroid` to showcase my problem: https://github.com/PyCQA/astroid/compare/master...renefritze:conditional_test_pickle

With the last commit on that branch the test suite succeeds: https://travis-ci.com/github/renefritze/astroid/jobs/487475875
Dropping the last commit, the suite fails: https://travis-ci.com/github/renefritze/astroid/jobs/487452325
The difference being https://github.com/PyCQA/astroid/commit/307f3acbfaa8bac3e7e4f1f0d0f38a965e54c53d

hippo91:
@renefritze thanks for your report. Would you mind opening a PR on astroid please?

renefritze:
Sure thing.

Pierre-Sassoulas:
Reopening because I did not understand that the merge request was only illustrative

renefritze:
Is there anything more I can do to help that doesn't require in-depth astroid knowledge?

DanielNoord:
I have investigated this a little.

```python
 def test_conditional_definition_partial_pickle(self) -> None:
        """Regression test for a conditional defintion of a partial function wrapping pickle.

        Reported in https://github.com/PyCQA/astroid/issues/925
        """
        module = resources.build_file(
            "data/conditional_pickle_import/conditional_importer.py"
        )

        dump_nodes = module.lookup("dump")[1]
        assert len(dump_nodes) == 2

        # Check the function defintion node
        func_def_nodes = list(dump_nodes[0].infer())
        assert len(func_def_nodes) == 1
        assert isinstance(func_def_nodes[0], FunctionDef)

        # Check the partial node
        partial_node = list(dump_nodes[1].infer())
        assert len(partial_node) == 1
        assert isinstance(partial_node[0], objects.PartialFunction)
```

This should pass with:
```python
# conditional_importer
import pickle

if False:

    def dump(obj, file, protocol=None):
        pass


else:
    from functools import partial

    dump = partial(pickle.dump, protocol=0)
```


The problem is that `pickle.dump` can be either `pickle.dump` or `_pickle.dump`. 
[...]

PR #935, opened 2021-04-09
Conditional test pickle

renefritze:
This is the code as described in #925


cdce8p (commented):
@renefritze Thanks for opening the PR. I'm not sure I fully understand it though. I assume this shouldn't just be a regression test for #925? In that case it would help to also have a test case which is currently failing.

renefritze:
> @renefritze Thanks for opening the PR. I'm not sure I fully understand it though. I assume this shouldn't just be a regression test for #925? In that case it would help to also have a test case which is currently failing.

I've tried to outline this in https://github.com/PyCQA/astroid/issues/925#issue-843129695
If you drop the last commit on my branch here, you get the failure. The idea being that the last commit documents that the issue is somehow related to which function is input into partial. This really isn't meant to be merged at all from my end. I just did as asked to open a PR with the example code.

renefritze:
Can I do anything to make the problem clearer?

Pierre-Sassoulas (approved):
Sorry for the review time, I was waiting for hippo to review but he's not available right now and this seems pretty safe to merge. Thank you for the merge request :)

renefritze:
Uhm. 

> This really isn't meant to be merged at all from my end. I just did as asked to open a PR with the example code.

renefritze:
I mean, it's safe to merge because it doesn't do anything useful with the last commit included? As I mentioned before, only without the last commit you'll see the potential underlying problem.

Pierre-Sassoulas:
My bad ! If I understand correctly the point of this merge request was to illustrate the problem with pickle ?

renefritze:
> My bad ! If I understand correctly the point of this merge request was to illustrate the problem with pickle ?

Indeed. Maybe I should set the PR branch to the commit with failing tests and add the "fix" just as a comment here? Maybe that would be clearer for everyone?

Pierre-Sassoulas:
:+1: A branch with the latest commit reverted seems like a good idea. (So we can see the error in the CI for each interpreter)

renefritze:
> > My bad ! If I understand correctly the point of this merge request was to illustrate the problem with pickle ?
> 

OK, so just like at the top of #925 I now have two branches again: [supposed to fail](https://github.com/renefritze/astroid/tree/conditional_test_pickle_failing)
 and [supposedly not failing](https://github.com/renefritze/astroid/tree/conditional_test_pickle_no_error)

Then I had to enable GitHub actions for those branches. 
Builds for both branches are now failing everywhere though: https://github.com/renefritze/astroid/actions

IMO the situation as described in #925 with the travis build logs was much clearer :disappointed: 

BTW: don't you want to remove the squashed commit on top in main for this PR?

Pierre-Sassoulas:
Could you open a pull request with the failing branch, please ? It can stay broken as an illustration.

PR #937, opened 2021-04-10
Upgrade packaging to setuptools

Pierre-Sassoulas:
This was pretty similar to pylint upgrade. I think we could release 2.5.4 directly after merging this.

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |


setup.cfg
```diff
@@ -1,3 +1,51 @@
...
+author = Python Code Quality Authority
+author_email = code-quality@python.org
+license = LGPL-2.1
+license_file = LICENSE
+classifiers =
+    Development Status :: 6 - Mature
+    Environment :: Console
+    Intended Audience :: Developers
+    License :: OSI Approved :: GNU General Public License (GPL)
+    License :: OSI Approved :: GNU Lesser General Public License v2 (LGPLv2)
```
cdce8p, line 17: We should only add the classifier for `LGPLv2` here

setup.cfg
```diff
@@ -1,3 +1,51 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+    Topic :: Software Development :: Libraries :: Python Modules
+    Topic :: Software Development :: Quality Assurance
+    Topic :: Software Development :: Testing
+keywords = static code analysis python abstract syntax tree
```
cdce8p, line 31: These should be `,` separated. https://packaging.python.org/specifications/core-metadata/#keywords

cdce8p (changes requested):
I like this one. Thanks for doing the work!
Regarding the whole license situation. Would you mind creating an extra MR for it? I like that you renamed `COPYING.LESSER` to `LICENSE`, but doing it separately would make it easier to follow. With that I also believe all headers would need updating.

Aside from that, there is also the `MANIFEST.in` which can be removed and `pytest.ini`.

setup.cfg
```diff
@@ -1,3 +1,51 @@
...
+name = astroid
+version = attr: astroid.__pkginfo__.__version__
+description = An abstract syntax tree for Python with inference support.
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+url = https://github.com/PyCQA/astroid
+author = Python Code Quality Authority
+author_email = code-quality@python.org
+license = LGPL-2.1
+license_file = LICENSE
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Actually it's the pre-commit hook that did it. It also detected the exact license in COPYING.LESSER and changed the value itself 😄

setup.cfg
Pierre-Sassoulas: Same problem exists in pylint btw

setup.cfg
```diff
@@ -1,5 +1,54 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+    Topic :: Software Development :: Libraries :: Python Modules
+    Topic :: Software Development :: Quality Assurance
+    Topic :: Software Development :: Testing
+keywords = static code analysis, python, abstract syntax tree
```
cdce8p, line 30:
```suggestion
keywords = static code analysis,python,abstract syntax tree
```

setup.cfg
```diff
@@ -1,5 +1,54 @@
...
+    "Bug tracker" = https://github.com/PyCQA/astroid/issues
+    "Discord server" = https://discord.gg/kFebW799
+    "Mailing list" = mailto:code-quality@python.org
+
+[options]
+packages = find:
+install_requires =
+    lazy_object_proxy>=1.4.0
+    wrapt>=1.11,<1.13
+    typed-ast>=1.4.0,<1.5;implementation_name== "cpython" and python_version<"3.8"
```
cdce8p, line 41:
```suggestion
    typed-ast>=1.4.0,<1.5;implementation_name=="cpython" and python_version<"3.8"
```
This changed again

setup.cfg
```diff
@@ -1,5 +1,54 @@
...
+name = astroid
+version = attr: astroid.__pkginfo__.__version__
+description = An abstract syntax tree for Python with inference support.
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+url = https://github.com/PyCQA/astroid
+author = Python Code Quality Authority
+author_email = code-quality@python.org
+license = LGPL-2.1
+license_file = LICENSE
```
cdce8p, line 11:
```suggestion
license_files = LICENSE
```
I would suggest you remove the `pre-commit` hook. The config is unlikely to change any time soon anyway.

cdce8p (changes requested):
I'm not yet comfortable with the changes you did in the `astroid` package. Can you explain your motivation? Maybe we can come up with a better plan.

setup.cfg
Pierre-Sassoulas: Both license_file and license_files are valid (and deprecated in favor of just giving the license in long_descripton) see https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#metadata

Pierre-Sassoulas:
The packaging is taking the ``__version__`` from astroid.__pkginfo__`` the problem is that in order to load that, we need to have wrapt and enum installed. But we don't have wrapt installed when we want to install the package. I think ``__init__.py`` should only be an API and not define function relying on dependencies. Initially I added ``__version__`` to ``__all__`` before importing wrapt or enum.

setup.cfg
cdce8p: Not entirely correct ;) https://setuptools.readthedocs.io/en/latest/history.html#v56-0-0
It's also mentioned here: https://setuptools.readthedocs.io/en/latest/references/keywords.html

`license_files` works better with `wheel` and also now supports glob pattern matching and defaults.

Pierre-Sassoulas: Ok, as we have a LICENSE we can remove this part entirely then ?

cdce8p: That should work. If not, I would have done something wrong. Although it's a bit more explicit with it being there.

cdce8p:
> The packaging is taking the `__version__` from `astroid.__pkginfo__` the problem is that in order to load that, we need to have wrapt and enum installed. But we don't have wrapt installed when we want to install the package.

I see. As far as I can tell this only effects the sdist though which isn't used normally. The build distribution works fine.

> I think `__init__.py` should only be an API and not define function relying on dependencies. Initially I added `__version__` to `__all__` before importing wrapt or enum.

Long term I agree, but we should think about it more before making a decision there. For now it might be best to use the previous method in `setup.py`.

```py
with open(pkginfo, "rb") as fobj:
    exec(compile(fobj.read(), pkginfo, "exec"), locals())

setup(
    version=__version__
)
```
Not the most elegant, but it should work.

setup.cfg
Pierre-Sassoulas: Yeah I like the expliciteness too. I'm not used to that level of magic happening in setuptools 😄 Also, apparently the license_file deprecation is a little too fresh for setup-cfg-fmt to act on it, so I made a fork.

Pierre-Sassoulas:
Yeah that circular import is not going to solve itself but let's do it in another MR.

cdce8p:
Looks good so far, but the tests are still failing.

**As for travis - pylint**: I believe the reason is there is that the `astroid.brain` module isn't included yet. It's missing the `__init__.py` file that's why it isn't recognized as package.

**For AppVeyor**: The test script in `appveyor.yml` just calls `python -m tox` which still uses Python 2.7. Since this is also used for to execute `setup.py`, it fails. `pathlib` was added in 3.4. It's probably enough to change all references to `python` in `appveyor.yml` - `install` and `test_script` - to `python3.6`.

cdce8p:
We should probably start to move to Github Actions after this MR is done 🤔

Pierre-Sassoulas:
I'll have a look soon thank you for analyzing the problem. You're right for github actions, having that in pylint was really night and day. Astroid see a lot less traffic so it's not as painful but still it would be a very welcome change.

setup.cfg
```diff
@@ -1,5 +1,53 @@
...
+    Topic :: Software Development :: Quality Assurance
+    Topic :: Software Development :: Testing
+keywords = static code analysis,python,abstract syntax tree
+project_urls =
+    "Bug tracker" = https://github.com/PyCQA/astroid/issues
+    "Discord server" = https://discord.gg/kFebW799
+    "Mailing list" = mailto:code-quality@python.org
+
+[options]
+packages = find_namespace:
```
cdce8p, line 36: `find:` was the right directive. You just need to add `astroid/brain/__init__.py`

Pierre-Sassoulas: ``__init__.py`` is not mandatory since python 3.3 I think ?

cdce8p: Not exactly sure what you mean. `setuptools` just looks for the `__init__.py` file to determine if it's a package that should be included.

Pierre-Sassoulas: https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3-3 Seeing the second answer you're right.

Pierre-Sassoulas: The pylint bug that prevent pylint from analysing directory without ``__init__.py`` is bitting us there. I don't have the time to fix all of those legacy warnings.

Pierre-Sassoulas:
Need #947 to be merged and rebased on first.

Pierre-Sassoulas:
Thanks for the review, the two useless job you caught on travis will save us a lot of time waiting for pipeline to run !

cdce8p (changes requested):
The change itself is fine, but `astroid.brain` is not included currently. #969 should fix that

cdce8p (approved):
It seems like everything works now! I just checked the source and build distributions and the `brain` files are included. IMO this MR is finally finished.

@Pierre-Sassoulas Thanks for all the work you have done here. Who would have thought that so many different things would need to be changed / fixed. 🚀

Issue #898, opened 2021-02-16
Add kind field to Const


federicobond:
Python 3.8 added the optional `kind` field to `ast.Constant`. This is used to store the string prefix, like `u` for `u"foo"`. `astroid.Const` should include this field too.

hippo91:
@federicobond thanks for this suggestion!

PR #943, opened 2021-04-12
Add kind field to Const nodes

federicobond:
## Description

Adds a kind field to Const nodes.

Note: this field is only available in Python's 3.8+ AST. Should `astroid` backport this for older supported versions of Python or are we fine as is?

## Type of Changes

## Related Issue

Closes https://github.com/PyCQA/astroid/issues/898

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,11 @@ Release Date: TBA
 
   Closes #926
 
+* Add ``kind`` field to ``Const`` nodes, matching the structure of the built-in ast Const.
+  The kind field is "u" if the literal is a u-prefixed string, and ``None`` otherwise.
+
+  Closes #898
+
 
 What's New in astroid 2.5.6?
 ============================
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -2557,7 +2557,7 @@ class Const(mixins.NoChildrenMixin, NodeNG, bases.Instance):
 
     _other_fields = ("value",)
 
-    def __init__(self, value, lineno=None, col_offset=None, parent=None):
+    def __init__(self, value, lineno=None, col_offset=None, parent=None, kind=None):
         """
         :param value: The value that the constant represents.
         :type value: object
@@ -2571,12 +2571,12 @@ class Const(mixins.NoChildrenMixin, NodeNG, bases.Instance):
 
         :param parent: The parent node in the syntax tree.
         :type parent: NodeNG or None
+
+        :param kind: The string prefix. "u" for u-prefixed strings and ``None`` otherwise. Python 3.8+ only.
+        :type kind: str or None
         """
         self.value = value
-        """The value that the constant represents.
-
-        :type: object
-        """
+        self.kind = kind
 
         super().__init__(lineno, col_offset, parent)
 
--- a/tests/unittest_nodes.py
+++ b/tests/unittest_nodes.py
@@ -542,6 +542,19 @@ class ConstNodeTest(unittest.TestCase):
     def test_unicode(self):
         self._test("a")
 
+    @pytest.mark.skipif(
+        not PY38, reason="kind attribute for ast.Constant was added in 3.8"
+    )
+    def test_str_kind(self):
+        node = builder.extract_node(
+            """
+            const = u"foo"
+        """
+        )
+        assert isinstance(node.value, nodes.Const)
+        assert node.value.value == "foo"
+        assert node.value.kind, "u"
+
     def test_copy(self):
         """
         Make sure copying a Const object doesn't result in infinite recursion
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -425,6 +425,7 @@ class TreeRebuilder:
             getattr(node, "lineno", None),
             getattr(node, "col_offset", None),
             parent,
+            getattr(node, "kind", None),
         )
 
     def visit_continue(self, node, parent):
@@ -814,6 +815,7 @@ class TreeRebuilder:
             getattr(node, "lineno", None),
             getattr(node, "col_offset", None),
             parent,
+            getattr(node, "kind", None),
         )
 
     # Not used in Python 3.8+.


astroid/node_classes.py
```diff
@@ -2567,11 +2567,14 @@ class Const(mixins.NoChildrenMixin, NodeNG, bases.Instance):
 
     _other_fields = ("value",)
 
-    def __init__(self, value, lineno=None, col_offset=None, parent=None):
+    def __init__(self, value, kind=None, lineno=None, col_offset=None, parent=None):
```
cdce8p, line 2570: Technically this is the right position (after `value`). I was just thinking that this could break calls which don't use keyword arguments. Would it be better to add `kind` as last argument? @hippo91

In the long term it's probably a good idea to add `*` and change to keyword only arguments.

astroid/node_classes.py
```diff
@@ -2583,6 +2586,7 @@ def __init__(self, value, lineno=None, col_offset=None, parent=None):
         :type parent: NodeNG or None
         """
         self.value = value
+        self.kind = kind
         """The value that the constant represents.
 
         :type: object
```
cdce8p, line 2592: Can you remove this comment? It seems to belong to `self.value` but that is already mentioned in the function docstring.

ChangeLog
```diff
@@ -6,6 +6,7 @@ What's New in astroid 2.6.0?
 ============================
 Release Date: TBA
 
+* Add `kind` field to `Const` nodes, matching the structure of the built-in `ast` Const.
```
cdce8p, line 9:
```suggestion
* Add `kind` field to `Const` nodes, matching the structure of the built-in `ast` Const.

  Closes #898
```

astroid/node_classes.py
```diff
@@ -2567,11 +2567,14 @@ class Const(mixins.NoChildrenMixin, NodeNG, bases.Instance):
 
     _other_fields = ("value",)
 
-    def __init__(self, value, lineno=None, col_offset=None, parent=None):
+    def __init__(self, value, kind=None, lineno=None, col_offset=None, parent=None):
         """
         :param value: The value that the constant represents.
         :type value: object
 
+        :param kind: The kind of value that the constant represents.
```
cdce8p, line 2575: Can you add a small comment that this is `3.8+` only?

cdce8p (changes requested):
Thanks for working on this! I left a few comments. Additional, should we a a simple test case for it?
```py
node = extract_node("""
const = u""
""")
assert node.value.kind == "u"
```

There is also a small merge conflict with the `ChangeLog`.

federicobond:
Thanks for the review @cdce8p! I addressed all the feedback except the constructor argument order. Awaiting comments from @hippo91.

tests/unittest_nodes.py
```diff
@@ -542,6 +542,19 @@ def test_str(self):
     def test_unicode(self):
         self._test("a")
 
+    @pytest.mark.skipif(
+        not PY38, reason="needs to be able to extract kind from ast.Constant"
```
cdce8p, line 546: What do you think about: "kind attribute for ast.Constant was added in 3.8"

tests/unittest_nodes.py
```diff
@@ -542,6 +542,19 @@ def test_str(self):
...
+    )
+    def test_str_kind(self):
+        node = builder.extract_node(
+            """
+            const = u"foo"
+        """
+        )
+        self.assertIsInstance(node.value, nodes.Const)
```
cdce8p, line 554: Can you use `assert isinstance(...)`? We try to move to pytest (at least for the newly added tests). The same for assertEqual below.

ChangeLog
```diff
@@ -23,6 +23,10 @@ Release Date: TBA
 
   Closes #926
 
+* Add `kind` field to `Const` nodes, matching the structure of the built-in `ast` Const.
+
+	Closes #898
```
cdce8p, line 28: Really small detail, but every other indent is only two spaces (I believe).

ChangeLog
```diff
@@ -23,6 +23,10 @@ Release Date: TBA
 
   Closes #926
 
+* Add `kind` field to `Const` nodes, matching the structure of the built-in `ast` Const.
```
cdce8p, line 26: Just notice it, the Changelog ist in a rst format so "kind" and "const" need to be escaped twice (take a look at other entries how it's done there). I would also recommend to remove the escaping from "ast"

cdce8p (changes requested):

tests/unittest_nodes.py
federicobond: yes, much better!

tests/unittest_nodes.py
federicobond: sure!

federicobond:
Fixed!

astroid/node_classes.py
hippo91: I agree with you @cdce8p . Putting the new parameter at the end is safer.

cdce8p: @federicobond Would you mind changing it then? Since `kind` is an optional argument, you should be able to get rid of some `None` arguments with that.

federicobond: Should be fixed now.

astroid/node_classes.py
```diff
@@ -2571,12 +2571,12 @@ def __init__(self, value, lineno=None, col_offset=None, parent=None):
 
         :param parent: The parent node in the syntax tree.
         :type parent: NodeNG or None
-        """
-        self.value = value
-        """The value that the constant represents.
 
-        :type: object
+        :param kind: The kind of value that the constant represents. Python 3.8+ only.
```
hippo91, line 2575: Idem here. A better picture of what `kind` represent could be nice.

astroid/node_classes.py
hippo91: Thanks @federicobond!

hippo91 (approved):
@federicobond thanks a lot for this MR!
I just left a couple of comments.

ChangeLog
```diff
@@ -23,6 +23,10 @@ Release Date: TBA
 
   Closes #926
 
+* Add ``kind`` field to ``Const`` nodes, matching the structure of the built-in ast Const.
```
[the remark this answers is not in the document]
federicobond: Unfortunately there is no PEP or release notes referencing this. I added further explanation though which should help clarify the meaning of the field.

federicobond:
Updated!

ChangeLog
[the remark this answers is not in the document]
hippo91: Nice! Thanks @federicobond!

hippo91:
@federicobond it is all good for me. I will just wait for @cdce8p acceptance to merge it!
Thanks again!

cdce8p (approved):
LGTM! Thanks 🚀

PR #944, opened 2021-04-12
Introduce speedups implemented in cython

matusvalo:
## Steps

## Description

This PR introduces PoC of implementing speed ups in Cython. This PR is increasing speed of pylint by 10-15% (based on my measurements). 

The optimization in cython is achieved annotation of code using 3 ways:
1. introducing `.pxd` header providing details about types and C structures
2. introducing `@cython` decorators to annotate the code (when executed via python this decorators are empty)
3. annotating the code via PEP 3107 and PEP 526 annotations.

### Motivation
Cython is well established compiler compiling python code to C where programmer can partially optimize code to avoid slow python API calls. Cython is wide used in multiple projects including Numpy [1], aiohttp [2] and also cython itself [3] (cython uses pure python mode),  Speedups written in this PR are written using *pure python mode* of Cython [4]. This ensures that

* the python code is kept almost untouched (the python code contains refactor changes only to write in an a way to be easily cythonized)
* running code does not require compling to binary

The main benefit of cython is that
* there is no need of rewriting huge code bases to different language
* speed ups can be implemented by small steps, introducing only simple changes in the beginning and after more invasive changes can be written.

Current PoC shows around **10-15%** speed increase comparing to python version.

### Drawbacks

* To improve performance via cython, the programmer needs to understand C language and basics of python API. 
* Using Cython requires learning Cython language. Cython language can challenging to understand can in some places - e.g. exception propagation in c methods/functions [5] (`except -1` vs. `except? -1` vs. `except *`).
* Cython contains 3 major issues which complicates optimization of astroid code base:
  -  generators using `yield` keyword cannot be cythonized to C method/function [6]
  - There is a bug which causes creas of cython for syntax of providing generator as parameter - e.g.: `any(... for ... in ...)` [7]
  - Cython C classes does not play well with Multiple Inheritance [8]

### Possible improvements
This PR servers as PoC showing that Cython can be viable platform to improve pylint/astroid performance. But there are more additional improvements which varies in complexity:
1. PR cythonize only portion of code. Whole astroid/pylint code base can be cythonized which is simple brute force approach which can saves some small  execution time for free.
2. based on my measurements the most expensive part of astroid is `scoped_nodes.py` module. This module can be rewritten to enable compiling more code in C mode (avoiding python API calls)
3. `AstroidManager` class should be easily compiled to cdef class.
4. shared global variables can be migrated to C variables after some small changes in python code.
5. `InferenceContext` class should be further optimized when some changes in python code are changed - see [9]
6. pylint codebase can be cythonized and C declarations from astroid can be imported in pylint to access directly C symbols of astroid

### Compilation and building binaries

Cython directly integrates to setup.py [10], but the goal of this PR is just to show how cython can improve performance in astroid. For building extention the following command needs to be executed in `astroid/astroid` directory:
```
cythonize -3ib arguments.py  bases.py  context.py  decorators.py  helpers.py  inference.py  node_classes.py  protocols.py  scoped_nodes.py  transforms.py
```

This command produces .so files (on linux platform which are used during import instead of .py files). For investigating the optimization the following command:
```
cython -3 -a <FILE>.py
```
produces a html page showing the optimizations of cython - see [11] and 

[1] https://github.com/numpy/numpy/tree/main/numpy/random
[2] https://github.com/aio-libs/aiohttp/tree/master/aiohttp
[3] https://github.com/cython/cython/blob/master/Cython/Compiler/FlowControl.pxd
[4] https://cython.readthedocs.io/en/latest/src/tutorial/pure.html
[5] https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values
[6] https://github.com/cython/cython/issues/2597
[7] https://github.com/cython/cython/issues/3477
[8] https://cython.readthedocs.io/en/latest/src/tutorial/cdef_classes.html#extension-types-aka-cdef-classes
[9] https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#fast-instantiation
[10] https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#basic-setup-py
[11] https://cython.readthedocs.io/en/latest/src/tutorial/cython_tutorial.html#primes

## Type of Changes

### DIFF
--- a/astroid/arguments.pxd
+++ b/astroid/arguments.pxd
@@ -1,0 +1,21 @@
+import cython
+from .context cimport InferenceContext, CallContext
+
+@cython.final
+cdef class CallSite:
+    cdef public dict argument_context_map
+    cdef public set duplicated_keywords
+    cdef list _unpacked_args
+    cdef dict _unpacked_kwargs
+    cdef public list positional_arguments
+    cdef public dict keyword_arguments
+
+
+    cpdef int has_invalid_arguments(self) except -1
+    cpdef has_invalid_keywords(self)
+
+
+    cdef dict _unpack_keywords(self, list keywords, InferenceContext context=*)
+    cdef list _unpack_args(self, list args, InferenceContext context=*)
+
+    cpdef infer_argument(self, funcnode, name, InferenceContext context)
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -8,12 +8,13 @@
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/master/COPYING.LESSER
 
+import cython
 
 from astroid import bases
-from astroid import context as contextmod
 from astroid import exceptions
 from astroid import nodes
 from astroid import util
+from .context import InferenceContext, CallContext
 
 
 class CallSite:
@@ -33,6 +34,7 @@ class CallSite:
         An instance of :class:`astroid.context.Context`.
     """
 
+    @cython.locals(callcontext=CallContext, context=InferenceContext)
     def __init__(self, callcontext, argument_context_map=None, context=None):
         if argument_context_map is None:
             argument_context_map = {}
@@ -62,8 +64,8 @@ class CallSite:
         """
 
         # Determine the callcontext from the given `context` object if any.
-        context = context or contextmod.InferenceContext()
-        callcontext = contextmod.CallContext(call_node.args, call_node.keywords)
+        context = context or InferenceContext()
+        callcontext = CallContext(call_node.args, call_node.keywords)
         return cls(callcontext, context=context)
 
     def has_invalid_arguments(self):
@@ -88,7 +90,7 @@ class CallSite:
 
     def _unpack_keywords(self, keywords, context=None):
         values = {}
-        context = context or contextmod.InferenceContext()
+        context = context or InferenceContext()
         context.extra_context = self.argument_context_map
         for name, value in keywords:
             if name is None:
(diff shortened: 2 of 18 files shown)


hippo91:
@matusvalo sorry for the delay and thanks a lot for this PR. 
I'am not very comfortable with such a strong modification of the code base. I'm afraid it involves too much modifications of the source and requires to learn the `cython` pseudo language. While it could be very interesting, i think it will not encourage new programmers to submit pull request and it will maybe causes troubles to maintain the existing sources.
As i'm afraid that my writings are too much opinionated i will require other maintainers opinions on this subject.

Pierre-Sassoulas (commented):
The performance improvement sounds great but I'm afraid it will be very hard to find contributors that understands cython. I think it would be a mistake to cytonize pylint. Having said that it's probably possible to cythonize small well chosen part of astroid, in order to get the greater benefit with the smallest possible amount of hard to maintain code. Maybe even more ? I have the feeling that astroid contributors are a lot more technical than pylint's contributors.

I also think doing cython first is kinda brute forcing. Importing astroid takes a lot of time (0.35s) and I think it's possible to do better than that while staying in pure python.

cdce8p (commented):
I'm with @hippo91 on this one. Performance improvements are nice, but using `cython` has some real drawbacks that we shouldn't ignore.
- It's hard enough as it is to find people how understand the code and are willing to contribute. Going forward with it will only increase that threshold.
- @Pierre-Sassoulas Already mentioned that there are other areas we should probably explore first, that can give as at least some improvements.
- If I'm not mistaken, `cython` would require a large change to the way we build the distribution package. It would need to be compile for multiple platforms and/or user would need to do it themselves. That would probably be a large ask, especially for beginners (and on Windows).
- Aside from `pylint`, I think quite a few plugins also use `astroid` for inference and to customize pylint. Would they need to modify their code?

All in all, I don't think we should go there with `astroid` itself. The only option I see would be if it's possible to make this change fully optional and opt in. But even then, the added complexity and maintenance burden might not be worth it.

Pierre-Sassoulas:
I'm closing as there is a clear consensus among the three most active maintainers against using cython right now. It does not mean that we won't re-consider cython later on if the circumstances change.

PR #947, opened 2021-04-17
Add same github actions than pylint but withtout benchmark/spelling

Pierre-Sassoulas:
## Description

Introduce github action because appveyor is hell and travis is changing his commercial proposal. Also python 3.10 seems to work for astroid so we might as well test it.

## Type of Changes


.travis.yml
```diff
@@ -47,7 +29,7 @@ deploy:
...
+    secure: lAlz/mySOEOqIMp9vYb6WVvd4YP/XmnP1XmDJWAziit4+ydSB52H0wUprBZjMHenChtflANIKXggiaVO6sw6EqU8mxMEMz+6ixs9ZA0robYy9CgYdMrXSAYgr8NHbf3WPTiD65ajP5bpQ/v6i5YhVXhTgotORBmhnMyn5LA/OvbQGWZqHsdtdXZpsflXuzEDD9SL/MgrvfOEBINJzHuXyKDqwOzqjNL9VeUoUHbubBk/haJtbXHPvAQR9SOtS1hBeq9sVAQghdxQTs39XNPAnzukgEwW0UNmmuW6bQ6UWbxztHHQYgXBni5cfhGE7B5GO2L0Cneuiwz99HGyDvdOSNgxNahLcIlAWCWzp71T7KSRnPhAFMVbw7/65eb5VIJKyrO9rwZi5zCo4+c9Wi0er7+l1PVLcEw9O+ouEYs1+1iY7JFyP4cHAPGd6h0POG/IE3UJZ/5yhOSBR6sYwRbR4Qc2zPflnZrjSgBCpaJ37Y+FZwg7BzPvElGteTmqm3PsdqWWJshYs/l5QaRuzUOalPlxJHDrau9JPm3KAlosJde7cUD5zooiy08GHfd8fle2zAbGjgk9p7VAFf/2BFJj261h9eAmFHwIgBW7jje3eBCYUbBuzl+uzGGQNdfyoNzrbRcnuVWr/Is9PefVf0OmLDPNTgJy0gevsMZgfoCCuiQ=
```
cdce8p, line 32: Are you sure this should be changed?

requirements_test.txt
```diff
@@ -0,0 +1,9 @@
+-r requirements_test_pre_commit.txt
+-r requirements_test_min.txt
+coveralls~=3.0
+coverage~=5.5
+pre-commit~=2.12
+pyenchant~=3.2
+pytest-cov~=2.11
+pytest-profiling~=1.7
+pytest-xdist~=2.2
```
cdce8p, line 9: We probably only need `pytest-cov` from those

requirements_test_pre_commit.txt
```diff
@@ -0,0 +1,7 @@
+autoflake==1.4
+black==20.8b1
+flake8==3.9.0
+isort==5.8.0
+mypy==0.812
+pyupgrade==2.11.0
+black-disable-checker==1.0.0
```
cdce8p, line 7: Currently, `astroid` only uses `pyupgrade` (in v2.12.0) and `black` of those here. Unless I'm missing something

requirements_test_min.txt
```diff
@@ -0,0 +1,2 @@
+-e .
+pytest~=6.2
```
cdce8p, line 2: You might need to add `lazy-object-proxy`, `nose` and `wrapt` here as well. Take a look at the `tox.ini` file for that.

cdce8p (changes requested):
Thanks for starting the work here! I never really got to it.

Some general comments:
- Should we add the classifier for `py310` if the tests already pass?
- Travis does run `pylint` as well. This isn't covered here
- The requirement files probably need some work. From looking at the `tox.ini` file, we probably need to add a `requirements_pypy.txt` and a `requirements_six.txt` file as well. https://github.com/PyCQA/astroid/blob/f2b197a4f8af0ceeddf435747a5c937c8632872a/tox.ini#L13-L28
- All tests (linux and window) should be run **with** and **without** `six`. Maybe this could work with `strategy.matrix`?

.travis.yml
Pierre-Sassoulas: The deployment on tag is not working for astroid, but its working for pylint. This is a bet where I lose nothing if it does not work 😄

requirements_test_pre_commit.txt
Pierre-Sassoulas: No you're right. It can't use pylint either because once you start analyzing everything there's a hundred of warnings.

requirements_test_min.txt
Pierre-Sassoulas: Thank you for pointing that out, lazy object and wrapt are installed from the ``-e .`` but apparently we need nose and numpy.

.travis.yml
cdce8p: Didn't know that. So you uploaded all releases to pypi manually in the past?

requirements_test_pre_commit.txt
cdce8p: Are you sure with `pylint`? I just ran it locally and the `astroid` directory doesn't have any issues. That's also the only one tox checked.
```
commands = pylint -rn --rcfile={toxinidir}/pylintrc {toxinidir}/astroid
```

.travis.yml
Pierre-Sassoulas: Only for astroid :)

Pierre-Sassoulas: https://github.com/PyCQA/astroid/blob/master/doc/release.md

requirements_test_pre_commit.txt
Pierre-Sassoulas: Yeap pylint does not work on package without ``__init__.py``. So if you add an init in ``astroid/brain`` everything starts to be analysed. Or use pre-commit because it launch pylint file by file.

```yaml
  - repo: local
    hooks:
      - id: pylint
        name: pylint
        entry: pylint
        language: system
        types: [python]
        args:
          [
            "-rn",
            "-sn",
            "--rcfile=pylintrc",
          ]
        exclude: tests/testdata
```

Pierre-Sassoulas:
I left the ``x-six`` and ``pylint`` jobs in the travis.yml.

cdce8p:
I opened #964 and #965 which should simplify this MR, once they are merged.
Can you rebase it as well? It still includes the `backports.functools_lru_cache` dependency and the issue with `requirements_pre_commit.txt` which are already fixed.

After those things are done, I believe we only need to check that all the right requirements are installed and the cache key is generated correctly and it should be good to go. Maybe we should also pin the requirements in `requirements_test_brain.txt`?

If you like, let me know once you're done with the rebase and I can take another look regarding the dependencies.

Pierre-Sassoulas:
#965 is not mergeable right now, I tried to do it too but we need to think about why six or not six is required for tests (I don't know much about six). So I think this is reviewable now.

appveyor.yml
```diff
@@ -1,31 +1,5 @@
 version: "{branch}-{build}"
 build: off
-cache:
-  - 'C:\\tmp'
-environment:
-  matrix:
-    - PYTHON: "C:\\Python36"
-      TOXENV: "py36,py36-six"
```
cdce8p, line 8: What do you think? If we keep the `six` test on linux / travis for now, should we also keep those here? IMO the travis ones are probably enough.

.travis.yml
```diff
@@ -7,18 +7,16 @@ jobs:
     - stage: prechecks
       python: 3.6
       env: TOXENV=pylint
```
cdce8p, line 9: Isn't `pylint` done with the `formatting` checks from pre-commit?

.travis.yml
```diff
@@ -7,18 +7,16 @@ jobs:
...
-    - python: 3.6
-      env: TOXENV=formatting
     - python: pypy3
       env: TOXENV=pypy
```
cdce8p, line 11: `pypy3` will be checked by Github actions

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+      - name: Set up Python ${{ env.DEFAULT_PYTHON }}
+        id: python
+        uses: actions/setup-python@v2.2.1
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+      - name: Generate partial Python venv restore key
+        id: generate-python-key
+        run: >-
+          echo "::set-output name=key::base-venv-${{ env.CACHE_VERSION }}-${{
+            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_min.txt')
```
cdce8p, line 34:
```suggestion
            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_pre_commit.txt',
                      'requirements_test_min.txt', 'requirements_test_brain.txt')
```
Besides `requirements_test_brain.txt`, we probably need to add `requirements_test_pre_commit.txt` as well -> for `pylint`.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+            steps.generate-python-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-base-venv-${{ env.CACHE_VERSION }}-
+      - name: Create Python virtual environment
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt
```
cdce8p, line 52:
```suggestion
          pip install -U -r requirements_test.txt -r requirements_test_brain.txt
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        id: python
+        uses: actions/setup-python@v2.2.1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Generate partial Python venv restore key
+        id: generate-python-key
+        run: >-
+          echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
+            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_min.txt')
```
cdce8p, line 135:
```suggestion
            hashFiles('setup.cfg', 'requirements_test.txt',
                      'requirements_test_min.txt', 'requirements_test_brain.txt')
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+            steps.generate-python-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
+      - name: Create Python virtual environment
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt
```
cdce8p, line 153:
```suggestion
          pip install -U -r requirements_test.txt -r requirements_test_brain.txt
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        id: python
+        uses: actions/setup-python@v2.2.1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Generate partial Python venv restore key
+        id: generate-python-key
+        run: >-
+          echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
+            hashFiles('setup.cfg', 'requirements_test_min.txt')
```
cdce8p, line 258:
```suggestion
            hashFiles('setup.cfg', 'requirements_test_min.txt', 'requirements_test_brain.txt')
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,388 @@
...
+            steps.generate-python-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
+      - name: Create Python virtual environment
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv\\Scripts\\activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test_min.txt
```
cdce8p, line 276:
```suggestion
          pip install -U -r requirements_test_min.txt -r requirements_test_brain.txt
```

cdce8p (changes requested):
Only a few small things, mainly `requirements_test_brain.txt`. That file isn't loaded inside `requirements_test.txt`, which is the right decision, so it needs to be installed manually.

appveyor.yml
Pierre-Sassoulas: Yeah, basically I want to get rid of appveyor :)

.travis.yml
Pierre-Sassoulas: You're right we can actually optimize that !

appveyor.yml
Pierre-Sassoulas: Maybe we can also get rid of non six tests, we'll see in #965

.travis.yml
Pierre-Sassoulas: Nice catch !

.github/workflows/ci.yaml
cdce8p: Any specific reason why you didn't add this one here?

Pierre-Sassoulas: No my bad !

cdce8p: Do you what do fix it or shall I? :)

Pierre-Sassoulas: I'll fix in a MR for fixing pylint's warning again. (pylint's master changed)

cdce8p: We should probably pin the `pylint` version for the formatting checks

.travis.yml
```diff
@@ -4,21 +4,14 @@ stages:
...
   include:
-    - stage: prechecks
-      python: 3.6
-      env: TOXENV=pylint
     - python: 3.6
-      env: TOXENV=formatting
-    - python: pypy3
-      env: TOXENV=pypy
-    - python: 3.6
-      env: TOXENV=py36,py36-six
```
cdce8p, line 15: Just noticed, we probably broke the deploy stage by removing `py36` here.

.github/workflows/ci.yaml
Pierre-Sassoulas: Yeah I thought that as well, we don't want to have to fix unrelated pylint warning in a MR just because pylint changed.

See : https://github.com/PyCQA/astroid/pull/968/files

.travis.yml
Pierre-Sassoulas: It was already broken but not the same way 😄 If you want to fix it feel free.. I'm aiming to fix pylint's warning and finish #937 before releasing 2.5.4

PR #948, opened 2021-04-17
Add inference tip for ``typing.Tuple``

cdce8p:
## Description
The alias for `typing.Tuple` is handled differently form all other typing aliases. Instead of a call to the `_alias` function, it's actually a class instantiation of `_VariadicGenericAlias` / `_TupleType`.

## Type of Changes

## Related Issue
--

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -297,6 +297,44 @@ def infer_typing_alias(
     return iter([class_def])
 
 
+def _looks_like_tuple_alias(node: nodes.Call) -> bool:
+    """Return True if call is for Tuple alias.
+
+    In PY37 and PY38 the call is to '_VariadicGenericAlias' with 'tuple' as
+    first argument. In PY39+ it is replaced by a call to '_TupleType'.
+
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_typing.py
```diff
@@ -297,6 +297,44 @@ def infer_typing_alias(
...
+    return (
+        isinstance(node, nodes.Call)
+        and isinstance(node.func, nodes.Name)
+        and (
+            not PY39
+            and node.func.name == "_VariadicGenericAlias"
+            and isinstance(node.args[0], nodes.Name)
+            and node.args[0].name == "tuple"
+            or PY39
+            and node.func.name == "_TupleType"
```
hippo91, line 318: Should we also test the first argument being equal to "tuple"?

hippo91 (approved):
Thanks @cdce8p ! It looks good to me. I just left a comment but nothing serious.

astroid/brain/brain_typing.py
cdce8p: 👍🏻

PR #952, opened 2021-04-19
Add isort pylint and autoflake to pre-commit

Pierre-Sassoulas:
## Description

Follow up to #950 

## Type of Changes

## Related Issue

#937

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,12 @@ repos:
     hooks:
       - id: prettier
         args: [--prose-wrap=always, --print-width=88]
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args: ["-rn", "-sn", "--rcfile=pylintrc"]
+        exclude: tests/testdata|setup.py|conf.py
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -34,15 +34,15 @@
 :var BUILTIN_MODULES: dictionary with builtin module names has key
 """
 
-# We don't want distutils to be a requirement
-# pylint: disable=import-error, no-name-in-module,useless-suppression
+# We disable the import-error so pylint can work without distutils installed.
+# pylint: disable=no-name-in-module,useless-suppression
 
 import importlib.util
 import itertools
 import os
 import platform
 import sys
-from distutils.errors import DistutilsPlatformError
+from distutils.errors import DistutilsPlatformError  # pylint: disable=import-error
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
 
 from astroid.interpreter._import import spec, util
--- a/astroid/nodes.py
+++ b/astroid/nodes.py
@@ -20,7 +20,8 @@ All nodes inherit from :class:`~astroid.node_classes.NodeNG`.
 """
 # pylint: disable=unused-import,redefined-builtin
 
-from astroid.node_classes import (  # Node not present in the builtin ast module.
+# Node not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
+from astroid.node_classes import (
     AnnAssign,
     Arguments,
     Assert,
@@ -120,6 +121,7 @@ ALL_NODE_CLASSES = (
     Compare,
     Comprehension,
     Const,
+    const_factory,
     Continue,
     Decorators,
     DelAttr,
@@ -131,6 +133,7 @@ ALL_NODE_CLASSES = (
     Expr,
     Ellipsis,
     EmptyNode,
+    EvaluatedObject,
     ExceptHandler,
     Exec,
     ExtSlice,
@@ -165,6 +168,7 @@ ALL_NODE_CLASSES = (
     TryFinally,
     Tuple,
     UnaryOp,
+    Unknown,
     While,
     With,
     Yield,
--- a/astroid/nodes.py
+++ b/astroid/nodes.py
@@ -20,7 +20,7 @@ All nodes inherit from :class:`~astroid.node_classes.NodeNG`.
 """
 # pylint: disable=unused-import,redefined-builtin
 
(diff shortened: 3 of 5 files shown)


setup.cfg
```diff
@@ -3,3 +3,11 @@ test = pytest
...
 [tool:pytest]
 testpaths = tests
+
+[isort]
+multi_line_output = 3
+line_length = 88
+known_third_party = sphinx, isort, pytest, mccabe, six, toml
+include_trailing_comma = True
+skip_glob = tests/testdata
+src_paths = astroid
```
cdce8p, line 13: This might also not be necessary

astroid/modutils.py
```diff
@@ -33,22 +33,24 @@
 :type BUILTIN_MODULES: dict
 :var BUILTIN_MODULES: dictionary with builtin module names has key
 """
+
+# We don't want distutils to be a requirement
```
cdce8p, line 37:
```suggestion
# TODO: We don't want distutils to be a requirement
```
?

astroid/modutils.py
```diff
@@ -33,22 +33,24 @@
...
+# pylint: disable=import-error, no-name-in-module,useless-suppression
```
cdce8p, line 38:
```suggestion
# pylint: disable=import-error,no-name-in-module,useless-suppression
```

astroid/modutils.py
```diff
@@ -33,22 +33,24 @@
...
+
 import importlib.util
+import itertools
 import os
 import platform
 import sys
-import itertools
+from distutils.errors import DistutilsPlatformError
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
```
cdce8p, line 46: `pylint` disable already on L38 for whole file

.pre-commit-config.yaml
```diff
@@ -10,6 +10,11 @@ repos:
           - --expand-star-imports
           - --remove-duplicate-keys
           - --remove-unused-variables
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.8.0
+    hooks:
+      - id: isort
+        exclude: tests/testdata|astroid/__init__.py
```
cdce8p, line 17: That should probably be fully covered by / move to the `setup.cfg`

tests/unittest_inference.py
```diff
@@ -37,25 +37,20 @@
...
-from astroid.bases import Instance, BoundMethod, UnboundMethod, BUILTINS
-from astroid import arguments
+from astroid import InferenceError, Slice, arguments, builder
 from astroid import decorators as decoratorsmod
-from astroid import exceptions
-from astroid import helpers
-from astroid import objects
-from astroid import test_utils
-from astroid import util
+from astroid import exceptions, helpers, nodes, objects, test_utils, util
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Do you think we could open a bug in astroid ? I think the second imports are all namespaces.

astroid/nodes.py
```diff
@@ -20,15 +20,18 @@
 """
 # pylint: disable=unused-import,redefined-builtin
 
-from astroid.node_classes import (
+from astroid.node_classes import (  # Node not present in the builtin ast module.
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Updated comment.

tests/unittest_inference.py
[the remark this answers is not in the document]
cdce8p: Just double checked it. Yeah, you're right. In that case, we can probably just leave it. It's a test file after all.

setup.cfg
```diff
@@ -3,3 +3,11 @@ test = pytest
 
 [tool:pytest]
 testpaths = tests
+
+[isort]
+multi_line_output = 3
+line_length = 88
+known_third_party = sphinx, isort, pytest, mccabe, six, toml
```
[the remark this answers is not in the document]
Pierre-Sassoulas: We have six, nose, numpy, attr, (others) in brains. (So in fact I should add some)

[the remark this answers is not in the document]
cdce8p: But aren't they grouped anyway if you don't add `known_third_party`?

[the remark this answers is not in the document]
Pierre-Sassoulas: It depends if they're installed locally or not, so with this setting isort can work the same if numpy is installed or not.

astroid/modutils.py
Pierre-Sassoulas: (We disable the import-error so pylint can work without distutil installed.) I updated the comment :)

.pre-commit-config.yaml
Pierre-Sassoulas: I don't understand ? We need this because there is a circular import and if isort is applied on ``astroid/__init__.py`` it breaks everything. We'll still have it after we move to setup.cfg. This will probably be a lot of work to fix.

cdce8p: 🤦🏻 I mixed that on up with the config for `flake8`. There you have a setting `extend_exclude` for the `setup.cfg` so you don't need to specify the excludes in the `.pre-commit-config`.

astroid/nodes.py
```diff
@@ -20,7 +20,8 @@
 """
 # pylint: disable=unused-import,redefined-builtin
 
-from astroid.node_classes import (  # Node not present in the builtin ast module.
+# Node not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
```
cdce8p, line 23:
```suggestion
# Nodes not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
```

cdce8p:
`pypy` doesn't seem to handle `astroid/brain/__init__.py` that well. Maybe that should be addressed later?

Pierre-Sassoulas:
Yeah you're right let's do it in another MR. Although adding that init was the whole point of the operation. But having the pylint's warning be checked by pre-commit is important so we don't have to do them all at once again.

cdce8p (approved):

PR #955, opened 2021-04-19
Enable useless supression

Pierre-Sassoulas:
## Description

Following #950 we enable useless-suppression

## Type of Changes

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -29,10 +29,7 @@
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/master/LICENSE
 
-# pylint: disable=too-many-lines; https://github.com/PyCQA/astroid/issues/465
-
-"""Module for some node classes. More nodes in scoped_nodes.py
-"""
+"""Module for some node classes. More nodes in scoped_nodes.py"""
 
 import abc
 import builtins as builtins_mod
@@ -791,7 +788,7 @@ class NodeNG:
         indent="   ",
         max_depth=0,
         max_width=80,
-    ):
+    ) -> str:
         """Get a string representation of the AST from this node.
 
         :param ids: If true, includes the ids with the node type names.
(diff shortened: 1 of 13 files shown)


astroid/node_classes.py
```diff
@@ -791,7 +788,7 @@ def repr_tree(
         indent="   ",
         max_depth=0,
         max_width=80,
-    ):
+    ) -> str:
```
cdce8p, line 791: I wouldn't start with type annotations just yet. Chances are you're not going to finish for a while :D

cdce8p (approved):
LGTM 👍🏻

tests/unittest_nodes.py
```diff
@@ -155,8 +155,13 @@ class Language(metaclass=Natural):
...
         code_annotations = textwrap.dedent(code)
-        # pylint: disable=line-too-long
-        expected = 'def function(var: int):\n    nonlocal counter\n\n\nclass Language(metaclass=Natural):\n    """natural language"""'
+        expected = '''\
+def function(var: int):
+    nonlocal counter
+
+
+class Language(metaclass=Natural):
+    """natural language"""'''
```
cdce8p, line 164: I would indent these lines here. For reference, see L149-L55.

Pierre-Sassoulas: We're testing ``textwrap.dedent(code)`` so we're actually testing that the code line 149-155 become like the one here.

```
>       self.assertEqual(ast.as_string().strip(), expected)
E       AssertionError: 'def function(var: int):\n    nonlocal cou[68 chars]e"""' != '        def function(var: int):\n        [100 chars]e"""'
E       - def function(var: int):
E       +         def function(var: int):
E       ? ++++++++
E       -     nonlocal counter
E       +             nonlocal counter
E       ? ++++++++
E         
E         
E       - class Language(metaclass=Natural):
E       +         class Language(metaclass=Natural):
E       ? ++++++++
E       -     """natural language"""+             """natural language"""? ++++++++
```

cdce8p: 👍🏻 Got it

PR #956, opened 2021-04-20
Fix crash when evaluating ``typing.NamedTuple``

cdce8p:
## Steps

## Description
Small MR to fix a crash when evaluating ``typing.NamedTuple``

## Type of Changes

## Related Issue
Closes https://github.com/PyCQA/pylint/issues/4383

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,10 @@ Release Date: TBA
 
 * Added inference tip for ``typing.Tuple`` alias
 
+* Fix crash when evaluating ``typing.NamedTuple``
+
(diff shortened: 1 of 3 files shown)


bdraco:
Thanks for the quick fix @cdce8p   🎉

hippo91 (approved):
Well done @cdce8p ! For me it is ok. I just wonder if we should add a unittest covering this bug. 
What do you think about it?

cdce8p:
> Well done @cdce8p ! For me it is ok. I just wonder if we should add a unittest covering this bug.
> What do you think about it?

I've added two test cases for it

PR #965, opened 2021-04-23
Consolidate test cases

cdce8p:
## Steps

## Description
`six` is only used to test `brain_six.py`. Besides that, it isn't used anywhere. There is no reason why we can't include it with the other third party brain dependencies. Similar story for `python-dateutil`.

The tests have guards to check if `six` is installed, so there is no need for further action besides this MR.
https://github.com/PyCQA/astroid/blob/54e6efeb3989bbb1c1127a73e8e0cbc2aef78912/tests/unittest_brain.py#L63-L82

/CC: @Pierre-Sassoulas, @hippo91 

Preparation for #947

## Type of Changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,7 @@ language: python
 jobs:
   include:
     - python: 3.6
-      env: TOXENV=py36,py36-six
-    - python: 3.7
-      env: TOXENV=py37-six
-    - python: 3.8
-      env: TOXENV=py38-six
-    - python: 3.9
-      env: TOXENV=py39-six
+      env: TOXENV=py36
 before_install:
   - python --version
   - uname -a
--- a/requirements_test_brain.txt
+++ b/requirements_test_brain.txt
@@ -1,3 +1,5 @@
+attrs
 nose
 numpy
-attrs
+python-dateutil
+six
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -33,7 +33,6 @@
 
 """Tests for basic functionality in astroid.brain."""
 import io
-import os
 import queue
 import re
 import sys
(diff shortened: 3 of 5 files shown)


Pierre-Sassoulas:
Yeah I tried that locally too but there is a check inside the tests that six is not installed. The question then is, can we remove this check ?

Pierre-Sassoulas (changes requested):
My hypothesis is that test with six and without six are required because we'd break code that do not use six. Apparently it's compatibility lib that could stay a very long time in codebase. I don't know if we can remove the tests without six. @hippo91 do you know why it was tested like this ?

cdce8p:
> Yeah I tried that locally too but there is a check inside the tests that six is not installed. The question then is, can we remove this check ?

I would thing it should work, but I'll take a closer look.

hippo91:
> My hypothesis is that test with six and without six are required because we'd break code that do not use six. Apparently it's compatibility lib that could stay a very long time in codebase. I don't know if we can remove the tests without six. @hippo91 do you know why it was tested like this ?

Sorry guys but have got no idea about why it has been tested like this.

cdce8p:
I did look at this one again. At the moment, we basically have 8 tests that are run only in `six` environments. 7 of them are guarded by `HAS_SIX` and one by `HAS_DATEUTIL`. The one for `dateutil` is probably a mistake. `tox.ini` has this line here: `!py{36,37,38,39}-six: python-dateutil`. It seems as the `!` is not recognized as a logical `not` and instead just ignored.

As for the special guard in `SixBrainTest`: I believe it isn't necessary and was just a precautionary measure. There are tests in `unittest_scoped_nodes` that just check for `HAS_SIX` and they are fine.

To summarize, it should be fine to remove the `six` specific tests and just include it in `requirements_test_brain.txt`.

Pierre-Sassoulas (approved):

PR #968, opened 2021-04-23
Fix pylint warning from master branch

Pierre-Sassoulas:
## Description

Pin pylint version for the CI, and fix the current warning in the codebase.

## Type of Changes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -150,7 +150,7 @@ jobs:
           python -m venv venv
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
-          pip install -U -r requirements_test.txt
+          pip install -U -r requirements_test.txt -r requirements_test_brain.txt
 
   pytest-linux:
     name: Run tests Python ${{ matrix.python-version }} (Linux)
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -50,6 +50,7 @@ MANAGER = manager.AstroidManager()
 
 
 def open_source_file(filename):
+    # pylint: disable=consider-using-with
     with open(filename, "rb") as byte_stream:
         encoding = detect_encoding(byte_stream.readline)[0]
     stream = open(filename, newline=None, encoding=encoding)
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -109,6 +109,7 @@ jobs:
       - name: Run formatting check
         run: |
           . venv/bin/activate
+          pip install -r requirements_test_pre_commit.txt # Pylint is a system hook we need it
```
cdce8p, line 112: I haven't looked at the action logs, but we do restore the Python venv here which includes `pylint` already.

.github/workflows/ci.yaml
```diff
@@ -334,7 +335,8 @@ jobs:
         run: >-
           echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
             hashFiles('setup.cfg', 'requirements_test_min.txt',
-          'requirements_test_brain.txt', 'requirements_test_pre_commit.txt') }}"
+           'requirements_test_brain.txt')
+          }}"
```
cdce8p, line 339: Previously, we didn't run the extra tests for `pypy` which is quite reasonable I believe.

cdce8p (changes requested):
Would you mind adding the "fix" for travis as well? https://github.com/PyCQA/astroid/commit/f157130d665ffb271e426964dc9aa105b395cb73

.github/workflows/ci.yaml
Pierre-Sassoulas: I removed that

.github/workflows/ci.yaml
Pierre-Sassoulas: Also removed

Pierre-Sassoulas:
> Would you mind adding the "fix" for travis as well? [f157130](https://github.com/PyCQA/astroid/commit/f157130d665ffb271e426964dc9aa105b395cb73)

I pushed it directly on master.

Pierre-Sassoulas:
I think we should also do https://github.com/PyCQA/pylint/issues/4390

cdce8p (approved):
👍🏻

PR #969, opened 2021-04-24
Add __init__ to astroid.brain package

cdce8p:
## Steps

## Description
For #937 to work, `astroid.brain` needs to be a package, i.e. have an `__init__.py` file.
With this change, `astroid.brain` is no longer added to `sys.path`, instead I use `importlib.import_module` and the full path `astroid.brain.brain_xxx.py`. A good side effect is that we can now use relative imports inside the brain modules.

/CC: @Pierre-Sassoulas @hippo91 

## Type of Changes

### DIFF
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -41,7 +41,8 @@ Main modules are:
 import enum
 import itertools
 import os
-import sys
+from importlib import import_module
+from pathlib import Path
 
 import wrapt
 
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
Thank you for fixing this. We had to fix all pylint warning in astroid.brain before being able to do that, that's why it was not done in #937 yet :) Quite a lot of unexpected work in order to do this !

PR #971, opened 2021-04-24
Add flake8 in pre-commit configuration

Pierre-Sassoulas:
## Description

Fix the legacy problem with flake8, some of them quite ancient (python 2.6 deprecation!) and add an automated checker in pre-commit configuration.

## Type of Changes

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+extend-ignore = E203, E266, E501, C901, F401
+max-complexity = 20
+select = B,C,E,F,W,T4,B9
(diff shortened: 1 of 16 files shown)


Pierre-Sassoulas:
Ready, but need #972 to be merged first.

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+extend-ignore = E203, E266, E501, C901, F401
```
cdce8p, line 2:
```suggestion
extend-ignore = E203,E266,E501,C901,F401
```

Don't really know if that would make a difference

cdce8p (approved):
LGTM. One small comment

.flake8
Pierre-Sassoulas: Yeah I think so, I had to make the max_complexity higher and C901 was supposed to be ignored.

PR #972, opened 2021-04-24
Move a function in order to remove a circular import

Pierre-Sassoulas:
## Description

Fix a TODO and a suspicious code construct, one can wonder why it was this way. But the tests seems to still work.

## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -316,9 +316,19 @@ class Instance(BaseInstance):
                 return True
         return result
 
-    # This is set in inference.py.
     def getitem(self, index, context=None):
-        pass
+        # TODO: Rewrap index to Const for this case
+        new_context = contextmod.bind_context_to_node(context, self)
+        if not context:
+            context = new_context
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
@cdce8p could you review this one please ? :)

astroid/bases.py
```diff
@@ -283,6 +283,23 @@ def callable(self):
         except exceptions.AttributeInferenceError:
             return False
 
+    def getitem(self, index, context=None):
```
cdce8p, line 286: Why not put it at its original position (L319)?

cdce8p (approved):
One small comment. Besides that, the change looks fine. Although I also can't really tell why it was done that way to begin with and thus am not 100% sure it's safe to do. Anyway pylint tests pass, so I guess the circular dependency isn't present anymore.

astroid/bases.py
Pierre-Sassoulas: Just realized the function already existed after moving it :) Fixed.

PR #979, opened 2021-04-29
Add inference of Compare nodes

### LINKED ISSUE #846: Compare node inferring
This might be obvious, but I can't find any issue on compare node inferring. I maybe wrong here. but I thought that it will be a straight forward implementation, similar to `binop` implementation. Perhaps because of multiple comparators complicate things?

### Steps to reproduce
from astroid.builder import parse
module = parse(
    """
    x = 1 > 2
    """
)
inferred = next(module["x"].infer())
print(inferred)

### Current behavior
Uninferable

### Expected behavior
`False` const

2.5.0

nelfin:
## Steps

## Description

This change adds a first attempt at inferring the boolean value of `Compare` nodes.

## Type of Changes

## Related Issue

Closes #846

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -857,7 +857,7 @@ def _do_compare(
     return retval  # it was all the same value
 
 
-def _infer_compare(self: nodes.Compare, context: contextmod.InferenceContext) -> Any:
+def _infer_compare(self: nodes.Compare, context: InferenceContext) -> Any:
     """Chained comparison inference logic."""
     retval = True
 
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -26,9 +26,11 @@
 """this module contains a set of functions to handle inference on astroid trees
 """
 
+import ast
 import functools
 import itertools
 import operator
+from typing import Any, Iterable
 
 import wrapt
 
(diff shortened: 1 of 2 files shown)


nelfin:
Updated to reflect latest changes in astroid API and to catch an edge case I found in later testing. Since this is a new feature (even if it's small) this might be better as part of a minor release instead of a point release so I'm happy to wait for 2.8.x before merging.

astroid/inference.py
```diff
@@ -798,6 +800,98 @@ def infer_binop(self, context=None):
...
+        if retval is not True:
+            break  # short-circuit
+        lhs = rhs  # continue
+    if retval is util.Uninferable:
+        yield retval
+    else:
+        yield nodes.Const(retval)
+
+
+nodes.Compare._infer = _infer_compare
```
Pierre-Sassoulas, line 893: I suppose we're forced to do that or we get a circular import ? Is this due to the typing for ``_infer_compare`` maybe ? What happen if we use ``def _infer_compare(self: "nodes.Compare", context: contextmod.InferenceContext) -> Any:``  ?

Pierre-Sassoulas (approved):
Looks good, the tests are on point and very thorough.

astroid/inference.py
Pierre-Sassoulas: I tried, it's not worth it 😄

nelfin: I'm not sure I understand your question, was it about the `nodes.Compare._infer = _infer_compare` line specifically?

Pierre-Sassoulas: Yes, I realized since that trying to put the infer function directly in the nodes.Compare class was not feasible.

astroid/inference.py
```diff
@@ -790,6 +792,98 @@ def infer_binop(self, context=None):
...
+        if retval is None:
+            retval = expr
+        elif retval != expr:
+            return util.Uninferable
+            # (or both, but "True | False" is basically the same)
+
+    return retval  # it was all the same value
+
+
+def _infer_compare(self: nodes.Compare, context: InferenceContext) -> Any:
```
cdce8p, line 860: In almost all `_infer_xxx` methods `context` is annotated as `Optional[InferenceContext]` with `None` as default. It works either way, but it might be nice to use the current default.

cdce8p (commented):
Great work @nelfin 🚀
I now, I'm a bit late for the review. Notice some small things that I wanted to mention nevertheless.

PR #966, opened 2021-04-23
Support python 3.10

Pierre-Sassoulas:
## Description

Merge request for supporting python 3.10 that will be valid once #947 is merged.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |
| ✓   | :scroll: Docs          |

PR #986, opened 2021-05-04
Add initial support for pattern matching py310

cdce8p:
## Description
Python 3.10 will at pattern matching which introduces a lot of new AST nodes. This MR adds new node classes and extends the `TreeRebuilder` to parse `match` statements correctly.

Although not yet tested, I expect the impact for pylint itself to be fairly contained. Only three node types have some kind of `name` field that we need to account for: `MatchMapping ('rest')`, `MatchStar ('name')`, and `MatchAs ('name')`. Just this MR alone is enough so that pylint doesn't crash when parsing `match` statements.

**Some useful references**
[typeshed/_ast.pyi](https://github.com/python/typeshed/blob/59390469d60f3b05c6bcbb4e1441756b84ede607/stdlib/_ast.pyi#L379)
https://docs.python.org/3.10/library/ast.html#pattern-matching

## Type of Changes


Pierre-Sassoulas:
I don't know about merging this without tests. At the same time we currently have crashes in pylint on match statements.  What do you think @cdce8p ? Do you want my help adding test cases ? It would make this part of astroid compatible with DO178-E, that's always nice 😛

cdce8p:
> I don't know about merging this without tests. At the same time we currently have crashes in pylint on match statements. What do you think @cdce8p ? Do you want my help adding test cases ? It would make this part of astroid compatible with DO178-E, that's always nice 😛

It was my plan to add at least some test cases for it. I just haven't gotten around to doing it. Let's see if I can change that now

Pierre-Sassoulas:
Hey I cherry-picked #966, here so I can close the other one.

cdce8p:
@Pierre-Sassoulas I added some tests. They should cover most cases, although probably not all. Let me know if I should add any more or if they are too detailed.

Btw: It was a good idea not to merge it without tests. I did find one small issue. Although given the large scope, just one error is quite good I would say.

cdce8p:
@hippo91 I you have some time to take a look, I would appreciate the feedback from you as well. Since the change is fully typed and covered mostly by tests, I'm quite confident but maybe you find something I missed. Especially since it's a larger change.

Pierre-Sassoulas:
> Although given the large scope, just one error is quite good I would say.

Haha, yeah you can say that ! Surprisingly good even.

Pierre-Sassoulas (approved):
Thank you that's a lot of tests :) It's hard to tell if we handle all the case of the match pattern, but I saw all the use cases I know about.

astroid/node_classes.py
```diff
@@ -36,13 +36,20 @@
 import itertools
 import pprint
 import sys
+import typing
```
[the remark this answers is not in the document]
cdce8p: I needed to avoid name conflicts as `astroid` itself has a `List` class (in `node_classes.py`). It would have been possible to use `Optional`, but I figured to it's easier to understand if I stick to `typing.xxx` naming. Once we drop support for `3.6` we can simplify it with `from __future__ import annotations` and `list`.

tests/unittest_nodes.py
```diff
@@ -1371,5 +1371,181 @@ def test():
...
+            case None:
+                pass
+            case _:
+                pass
+        """
+        )
+        assert isinstance(node, nodes.Match)
+        assert isinstance(node.subject, nodes.Name)
+        assert isinstance(node.cases, list) and len(node.cases) == 4
+        assert node.subject.name == "status"
```
hippo91, line 1396: Maybe this line would make more sense if it is written right after `assert isinstance(node.subject, nodes.Name)` because both lines are dealing with `subject`.

tests/unittest_nodes.py
```diff
@@ -1371,5 +1371,181 @@ def test():
...
+        case0, case1, case2, case3 = node.cases
+
+        assert isinstance(case0.pattern, nodes.MatchValue)
+        assert case0.guard is None
+        assert isinstance(case0.body[0], astroid.Pass)
+        assert (
+            isinstance(case0.pattern.value, astroid.Const)
+            and case0.pattern.value.value == 200
+        )
```
hippo91, line 1405: Idem here. I think we can group the lines dealing with `pattern`. IMHO it would make more sense but maybe i'm nitpicking.

hippo91 (approved):
Amazing work @cdce8p ! It is really a good start for python 3.10 support! Congrats.
I just have two comments but i'm probably nitpicking so feel free to just ignore them!

cdce8p:
@hippo91 Thanks for taking the time to review it!

PR #998, opened 2021-05-24
Update requirements

cdce8p:
## Description
Followup to #997
/CC: @Pierre-Sassoulas

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,5 +1,6 @@
 -r requirements_test_min.txt
+-r requirements_test_pre_commit.txt
 coveralls~=3.0
 coverage~=5.5
-pre-commit~=2.12
+pre-commit~=2.13
 pytest-cov~=2.11
--- a/requirements_test_pre_commit.txt
+++ b/requirements_test_pre_commit.txt
@@ -1,6 +1,6 @@
 autoflake==1.4
 black==21.5b1
-pyupgrade==2.16.0
+pyupgrade==2.18.1
 black-disable-checker==1.0.1
 pylint==2.8.2
 isort==5.8.0
--- a/requirements_test_pre_commit.txt
+++ b/requirements_test_pre_commit.txt
@@ -1,7 +1,4 @@
-autoflake==1.4
 black==21.5b1
-pyupgrade==2.18.1
-black-disable-checker==1.0.1
 pylint==2.8.2
 isort==5.8.0
 flake8==3.9.2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 1
+  CACHE_VERSION: 3
   DEFAULT_PYTHON: 3.6
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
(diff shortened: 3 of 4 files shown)


requirements_test.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements_test_min.txt
+-r requirements_test_pre_commit.txt
```
Pierre-Sassoulas, line 2: I just removed that in #978 because it makes a circular import where astroid need pylint 2.8.2 that need astroid 2.5.6 and break the tests

cdce8p, line 2: You mentioned here https://github.com/PyCQA/astroid/commit/664555ac1f26113c78397297cbf340fb64736a24 that this caused an error. I tried to setup a new venv and everything worked out fine. They seem to be resolved as a whole.
```bash
python -m venv venv
source venv/bin/activate
pip install -r requirements_test.txt
```

cdce8p: Interesting... The pylint requirements should be fine in theory as it allows `>=2.5.6,<2.7` [pylint - setup.cfg](https://github.com/PyCQA/pylint/blob/091cb92314dc701f10390136da78fbbb362e892e/setup.cfg#L44)

requirements_test_pre_commit.txt
```diff
@@ -1,6 +1,6 @@
 autoflake==1.4
```
Pierre-Sassoulas, line 1: I'd really like to have the pre-commit configuration file as the only source of truth, every week for 4 weeks we had to accept 4/5 dependabot upgrades on pylint. We could have only one upgrade by the pre-commit bot. Maybe we could just add pylint/flake8 (because these tool do not have online autofix) and pre-commit in the test requirements ?

cdce8p: I would be fine with removing `pyupgrade`, `autoflake`, and `black-disable-checker` here as they aren't integrated into IDEs and only called with `pre-commit`. Everything else has some use

requirements_test.txt
cdce8p: Found the issue: It seems to be an issue with `setuptools_scm` as `pip install -e .` would install version `0.1.dev1+...` https://github.com/PyCQA/astroid/pull/998/checks?check_run_id=2659259745#step:6:97

Pierre-Sassoulas: The fix is to remove the pre-commit requirement from ``requirements_test.txt`` I don't know if it's a circular import but I think we cannot have both the local astroid and the dependency from pylint.

cdce8p: The issue isn't a circular dependency. Somehow `pip install -e .` will install version `0.1.dev1` of astroid which is incompatible with `pylint`. I think that `setuptools_scm` doesn't get the correct version here.

cdce8p: Found the issue: https://github.com/pypa/setuptools_scm/issues/414
Will look at solutions tomorrow.

cdce8p:
@Pierre-Sassoulas It works now. The solution was to add `fetch-depth: 0` to the Github checkout action. That way all refs and tags are fetched and `setuptools_scm` works correctly.

Pierre-Sassoulas:
Nice thank you. I guess we can fix https://github.com/PyCQA/pylint/issues/4505 with something similar.

Issue #1003, opened 2021-05-30
Performance regressions with 2.5.7


nelfin:
@cdce8p, @Pierre-Sassoulas, @hippo91: I meant to highlight this before there was a tagged release, but additional copies in the `InferenceContext.clone` cause some performance slowdown, especially in modules with deep inference chains. Maybe there should be additional benchmarking before `pylint` bumps it's minimum required version of `astroid`?

benfred:
I've also run into a major performance regression with astroid 2.5.7 using pylint against https://github.com/nvidia/nvtabular . 

With astroid 2.5.6 I can run ```pylint nvtabular/workflow.py``` in about 10 seconds, but with astroid 2.5.7 pylint has been running for 40+ minutes without finishing.  I've posted a [flamegraph here](https://gist.githubusercontent.com/benfred/458b0783ff97bd15225acf26c1d2e8c5/raw/9c66cde5f5a6664aa39118d4a68e7fcfc78517e2/astroid.svg) showing the callstack

nelfin:
@benfred: I don't have a conda install so I wasn't able to work with `nvtabular`, but I followed its dependency chain up one level to `cudf` and I was able to get some pathological slowdown with files like `cudf/python/cudf/cudf/core/series.py`

benfred:
thanks for looking into this @nelfin! 

I peaked a bit at the slowdown I was seeing with py-spy, and I can also verify that the problem seems to be in the cudf dependency. Running ```py-spy dump --locals -p PID``` shows slow processing of cudf.core.index at one point and cudf.core.dataframe at another - with frames like this on the stack trace: 

```
.. 
    ast_from_file (astroid/manager.py:100)
        Arguments::
            self: <AstroidManager at 0x7fe28efa9400>
            filepath: "/home/ben/miniconda3/envs/nvtabular_cudf19/lib/python3.8/site-packages/cudf/core/index.py"
            modname: "cudf.core.index"
            fallback: False
            source: True
        Locals::
            AstroidBuilder: <type at 0x5636e352d8c0>
    ast_from_module_name (astroid/manager.py:192)
        Arguments::
            self: <AstroidManager at 0x7fe28efa9400>
            modname: "cudf.core.index"
            context_file: None
        Locals::
            found_spec: ("index", <ModuleType at 0x7fe28f8f0f10>, ...)
...
```

Pierre-Sassoulas:
Probably related I encountered a RecursionError on code I can't share:

```
[...]

PR #1009, opened 2021-06-03
Performance improvements to counter context.clone slowdown

nelfin:
## Steps

## Description

This adds a couple of changes to `InferenceContext` to eliminate copies of the `inferred` nodes cache and limit the total number of inferred nodes in a context and all of its clones (preventing deep inference chains).

## Type of Changes

## Related Issue

Ref #1003

### DIFF
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -30,9 +30,16 @@ class InferenceContext:
         "boundnode",
         "inferred",
         "extra_context",
+        "_nodes_inferred",
     )
 
-    def __init__(self, path=None, inferred=None):
+    max_inferred = 100
+
+    def __init__(self, path=None, inferred=None, nodes_inferred=None):
+        if nodes_inferred is None:
+            self._nodes_inferred = [0]
+        else:
+            self._nodes_inferred = nodes_inferred
         self.path = path or set()
         """
         :type: set(tuple(NodeNG, optional(str)))
@@ -81,6 +88,20 @@ class InferenceContext:
         for call arguments
         """
 
+    @property
+    def nodes_inferred(self):
+        """
+        Number of nodes inferred in this context and all its clones/decendents
+
+        Wrap inner value in a mutable cell to allow for mutating a class
+        variable in the presence of __slots__
+        """
+        return self._nodes_inferred[0]
+
+    @nodes_inferred.setter
+    def nodes_inferred(self, value):
+        self._nodes_inferred[0] = value
+
     def push(self, node):
         """Push node into inference path
 
@@ -103,7 +124,11 @@ class InferenceContext:
         starts with the same context but diverge as each side is inferred
         so the InferenceContext will need be cloned"""
         # XXX copy lookupname/callcontext ?
-        clone = InferenceContext(self.path.copy(), inferred=self.inferred.copy())
+        clone = InferenceContext(
+            self.path.copy(),
+            inferred=self.inferred.copy(),
+            nodes_inferred=self._nodes_inferred,
+        )
         clone.callcontext = self.callcontext
         clone.boundnode = self.boundnode
         clone.extra_context = self.extra_context
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -13,7 +13,17 @@
 """Various context related utilities, including inference and call contexts."""
 import contextlib
 import pprint
-from typing import Optional
+from typing import TYPE_CHECKING, MutableMapping, Optional, Sequence, Tuple
+
+if TYPE_CHECKING:
+    from astroid.node_classes import NodeNG
+
+
+_INFERENCE_CACHE = {}
+
+
+def _invalidate_cache():
+    _INFERENCE_CACHE.clear()
 
 
 class InferenceContext:
@@ -28,14 +38,13 @@ class InferenceContext:
         "lookupname",
         "callcontext",
         "boundnode",
-        "inferred",
         "extra_context",
         "_nodes_inferred",
     )
 
     max_inferred = 100
 
-    def __init__(self, path=None, inferred=None, nodes_inferred=None):
+    def __init__(self, path=None, nodes_inferred=None):
         if nodes_inferred is None:
             self._nodes_inferred = [0]
         else:
@@ -72,14 +81,6 @@ class InferenceContext:
 
         e.g. the bound node of object.__new__(cls) is the object node
         """
-        self.inferred = inferred or {}
-        """
-        :type: dict(seq, seq)
-
-        Inferred node contexts to their mapped results
-        Currently the key is ``(node, lookupname, callcontext, boundnode)``
-        and the value is tuple of the inferred results
-        """
         self.extra_context = {}
         """
         :type: dict(NodeNG, Context)
@@ -101,6 +102,20 @@ class InferenceContext:
     @nodes_inferred.setter
     def nodes_inferred(self, value):
         self._nodes_inferred[0] = value
+
+    @property
+    def inferred(
+        self,
+    ) -> MutableMapping[
+        Tuple["NodeNG", Optional[str], Optional[str], Optional[str]], Sequence["NodeNG"]
+    ]:
+        """
+        Inferred node contexts to their mapped results
+
+        Currently the key is ``(node, lookupname, callcontext, boundnode)``
+        and the value is tuple of the inferred results
+        """
+        return _INFERENCE_CACHE
 
     def push(self, node):
         """Push node into inference path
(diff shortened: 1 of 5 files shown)


astroid/context.py
```diff
@@ -81,6 +85,31 @@ def __init__(self, path=None, inferred=None):
...
+    @property
+    def inferred(self):
+        """
+        :type: optional[NodeNG]
+
+        The bound node of the given context
+
+        e.g. the bound node of object.__new__(cls) is the object node
+        """
+        return _INFERENCE_CACHE
```
hippo91, line 111: If i understand well, here we will store every inferred node in a global variable that will exists throughout the life of the program. 
It will definitely allow cpu time sparing but what about memory consumption?

tests/unittest_brain.py
```diff
@@ -1469,6 +1469,19 @@ def test_typing_types(self):
             inferred = next(node.infer())
             self.assertIsInstance(inferred, nodes.ClassDef, node.as_string())
 
+    @test_utils.require_version(minver="3.7")
+    def test_plain_tuple_type(self):
```
hippo91, line 1473: I can't find a relation between this test and your modification. Can you explain a little bit please?

hippo91 (approved):
Thanks @nelfin for this smart MR!
It seems to be ok. I just have a question regarding memory consumption.
By the way i would really appreciate the review of @cdce8p. 
I think it is also necessary to test this MR against pylint but maybe have you already done it.

astroid/context.py
```diff
@@ -81,6 +85,31 @@ def __init__(self, path=None, inferred=None):
...
+    @nodes_inferred.setter
+    def nodes_inferred(self, value):
+        self._nodes_inferred[0] = value
+
+    @property
+    def inferred(self):
+        """
+        :type: optional[NodeNG]
+
+        The bound node of the given context
```
Pierre-Sassoulas, line 107:
```suggestion
    def inferred(self) -> Optional[NodeNG]:
        """The bound node of the given context
```
We can do that since we're not supporting python 3.5 anymore :)

Pierre-Sassoulas (approved):
👍

astroid/context.py
Pierre-Sassoulas: Is the number of inferred node limit sufficient for the speed up ? There are really big code bases using pylint and having a cache can affects them to the point it's not possible to run pylint on them at all.

cdce8p:
I'll take a look at it later today

cdce8p (changes requested):
I can confirm the speed improvements. However this change causes some `StopIteration` exceptions when testing against HomeAssisant: https://github.com/cdce8p/ha-core/runs/2753897777?check_suite_focus=true#step:7:55

astroid/context.py
nelfin: > If i understand well, here we will store every inferred node in a global variable that will exists throughout the life of the program.
> It will definitely allow cpu time sparing but what about memory consumption?

Your understanding is correct. It could theoretically grow without bound, but I suppose many other parts of astroid (such as `astroid.MANAGER.astroid_cache`) can as well. I regularly see pylint/astroid runs in the hundreds of MiB, but without a benchmark, I'm not sure there's a meaningful way to answer this question.

I tried running pylint (commit 9361fa0630ce920081b57865be7882a39e9da1e3) on itself with and without these astroid changes.

With `wip/performance` (0d7d7989458326a863f7fa942e662c5446d170e1):

```
26.77user 0.25system 0:27.08elapsed 99%CPU (0avgtext+0avgdata 231516maxresident)k
3312inputs+248outputs (3major+81935minor)pagefaults 0swaps
```

Without (d3fea8fee322200884920620cd9f9cb339ee9158):

```
45.15user 0.24system 0:45.43elapsed 99%CPU (0avgtext+0avgdata 235892maxresident)k
0inputs+336outputs (0major+83907minor)pagefaults 0swaps
```

So memory usage (max RSS) went down from 235MB -> 231MB


> Is the number of inferred node limit sufficient for the speed up ? There are really big code bases using pylint and having a cache can affects them to the point it's not possible to run pylint on them at all.

They address different problems: the inferred node limit is to stop pathological cases, like really large files. The inferred node cache helps with re-use, like when scanning a package with a lot of files. The main problem I saw with the inference cache was that it was re-built from scratch for every node so the hit-rate was very low, and cloning it introduced a large number of small copies which was inefficient time-wise. In fact, it was so inefficient that just removing it altogether produces a small but noticeable (~2%) speed-up on astroid 2.5.6 (middle of 3 runs):

astroid-2.5.6 (36dda3fc8a5826b19a33a0ff29402b61d6a64fc2):
```
38.57user 0.26system 0:38.86elapsed 99%CPU (0avgtext+0avgdata 231020maxresident)k
0inputs+72outputs (0major+81740minor)pagefaults 0swaps
```

With no-cache.patch applied:
```
37.77user 0.25system 0:38.05elapsed 99%CPU (0avgtext+0avgdata 230788maxresident)k
0inputs+72outputs (0major+81551minor)pagefaults 0swaps
```

`no-cache.patch`:
```patch
diff --git a/astroid/context.py b/astroid/context.py
index 18220ec..73f1f4d 100644
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -27,11 +27,10 @@ class InferenceContext:
         "lookupname",
         "callcontext",
         "boundnode",
-        "inferred",
         "extra_context",
     )
 
-    def __init__(self, path=None, inferred=None):
+    def __init__(self, path=None):
         self.path = path or set()
         """
         :type: set(tuple(NodeNG, optional(str)))
@@ -64,14 +63,6 @@ class InferenceContext:
 
         e.g. the bound node of object.__new__(cls) is the object node
         """
-        self.inferred = inferred or {}
-        """
-        :type: dict(seq, seq)
-
-        Inferred node contexts to their mapped results
-        Currently the key is ``(node, lookupname, callcontext, boundnode)``
-        and the value is tuple of the inferred results
-        """
         self.extra_context = {}
         """
         :type: dict(NodeNG, Context)
@@ -102,7 +93,7 @@ class InferenceContext:
         starts with the same context but diverge as each side is inferred
         so the InferenceContext will need be cloned"""
         # XXX copy lookupname/callcontext ?
-        clone = InferenceContext(self.path, inferred=self.inferred)
+        clone = InferenceContext(self.path)
         clone.callcontext = self.callcontext
         clone.boundnode = self.boundnode
         clone.extra_context = self.extra_context
diff --git a/astroid/node_classes.py b/astroid/node_classes.py
index 7faf681..089b579 100644
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -358,11 +358,6 @@ class NodeNG:
             yield from self._infer(context, **kwargs)
             return
 
-        key = (self, context.lookupname, context.callcontext, context.boundnode)
-        if key in context.inferred:
-            yield from context.inferred[key]
-            return
-
         generator = self._infer(context, **kwargs)
         results = []
 
@@ -378,7 +373,6 @@ class NodeNG:
 
         # Cache generated results for subsequent inferences of the
         # same node using the same context
-        context.inferred[key] = tuple(results)
         return
 
     def _repr_name(self):
```

tests/unittest_brain.py
nelfin: Before I added the cache invalidation logic in `TransformVisitor` the `test_tuple_type` test was the only test that failed. I added this test while I was debugging that, but I guess it doesn't make any sense without an explanation.

astroid/context.py
nelfin: It's actually the wrong doc comment any way >_>

nelfin:
@cdce8p, I was able to fix the `StopIteration` bubbling up through `safe_infer` by just catching it, but I haven't been able to come up with a good way to write a regression test for it. Re-running this on your `ha-core` fork, I got some new messages on `homeassistant/components/mysensors/light.py`:

```
************* Module homeassistant.components.mysensors.light
homeassistant/components/mysensors/light.py:157:23: E1120: No value for argument 'iG' in function call (no-value-for-parameter)
homeassistant/components/mysensors/light.py:157:23: E1120: No value for argument 'iB' in function call (no-value-for-parameter)
```

but these don't seem to be related. I've raised PyCQA/pylint#4546 since I was able to replicate a false-positive with astroid 2.5.6.

cdce8p (approved):

cdce8p:
@nelfin Thanks for fixing the issue! Indeed the `no-value-for-parameter` error seems to be unrelated.

Pierre-Sassoulas (approved):
I think this is mergeable, as the caching also got better. Do you agree @hippo91 ?

hippo91:
Yes it seems to be ok.

Pierre-Sassoulas:
Thank you a lot for this, @nelfin , great performance improvement, and the information you gave were super detailed and informative :fire:

cdce8p:
I can only agree, thanks @nelfin 🚀

PR #1010, opened 2021-06-06
Improve support for Pattern Matching

cdce8p:
## Description
1. Fix pylint `undefined-variable` and `unused-variable` errors for Pattern Matching (Python 3.10).
At the moment, the following nodes use strings for new variable names instead of the `ast.Name` node: `MatchAs` (for `name`), `MatchStar` (also for `name`), and `MatchMapping` (for `rest`). https://bugs.python.org/issue43994

```py
var = 42
match var:
    case _ as new_var:
        print(new_var)

var2 = [1, 2, 3]
match var2:
    case [*rest1]:
        print(rest1)

var3 = {1: "Hallo", 2: "World"}
match var3:
    case {**rest2}:
        print(rest2)
```

2. Fix `lineno` and `col_offset` for `MatchAs` and `MatchOr` nodes.
3. Add string representations for match nodes
4. Add example nodes in docstrings

--
**Useful links**
https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching
https://docs.python.org/3.10/library/ast.html#pattern-matching

## Type of Changes

### DIFF
--- a/astroid/as_string.py
+++ b/astroid/as_string.py
@@ -27,6 +27,21 @@
 * :func:`dump` function return an internal representation of nodes found
   in the tree, useful for debugging or understanding the tree structure
 """
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .node_classes import (
+        Match,
+        MatchAs,
+        MatchCase,
+        MatchClass,
+        MatchMapping,
+        MatchOr,
+        MatchSequence,
+        MatchSingleton,
+        MatchStar,
+        MatchValue,
+    )
 
 # pylint: disable=unused-argument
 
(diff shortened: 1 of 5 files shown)


astroid/node_classes.py
```diff
@@ -4910,21 +4910,20 @@ def get_children(self) -> typing.Generator["PatternTypes", None, None]:
             yield from self.patterns
 
 
-class MatchMapping(NodeNG):
+class MatchMapping(mixins.AssignTypeMixin, NodeNG):
```
hippo91, line 4913: I was wondering if it could be interesting to add an example of such node in the docstring. What do you think about it?

hippo91 (approved):
@cdce8p it is (as ever) a nice piece of work! I do not have very interesting comments just one regarding typing hint and another regarding docstring.
By the way i would like to thank you very much for your involvement in astroid and the quality of your devs! :+1:

astroid/as_string.py
```diff
@@ -591,6 +606,80 @@ def visit_starred(self, node):
         """return Starred node as string"""
         return "*" + node.value.accept(self)
 
+    def visit_match(self, node: "Match") -> str:
```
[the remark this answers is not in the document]
cdce8p: The `Match` nodes are only imported with `if TYPE_CHECKING` to prevent a circular dependency. Since we can't use postponed evaluation of annotations just yet (requires 3.7), these need to be strings for now.

astroid/node_classes.py
cdce8p: You got me 😄 I've been putting off doing that.
Added a new commit with examples for each `Match` node.

The link to the docs preview:
https://pylint--1010.org.readthedocs.build/projects/astroid/en/1010/api/astroid.nodes.html#astroid.nodes.Match

cdce8p:
@hippo91 Thanks for the quick review!

astroid/as_string.py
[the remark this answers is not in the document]
hippo91: Yes of course. I forgot the 'if' statement!

astroid/node_classes.py
hippo91: Nice job! At least i have been a bit useful! :wink:

cdce8p: A review is always useful!

PR #1011, opened 2021-06-06
Add lineno and col_offset for `Keyword` nodes

cdce8p:
## Description
Python 3.9 added line and column information to the `ast.keyword` node. Besides `lineno` and `col_offset`, `end_lineno` and `end_col_offset` have been added as well. I'm not sure if it would be valuable to us if I add them as well, since astroid uses `tolineno` only with regards to blocks, i.e. `body`.

https://bugs.python.org/issue40141

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,7 @@ What's New in astroid 2.5.8?
 ============================
 Release Date: TBA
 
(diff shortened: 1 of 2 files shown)


hippo91 (approved):
Thanks @cdce8p . It is ok for me. Regarding `end_lineno` and `end_col_offset` i don't think it is useful to add them.

PR #1019, opened 2021-06-09
Rebuilder improvements and cleanup

cdce8p:
## Description
* Removed `_visit_or_none`. The function is difficult to type and prone to errors since we must pass strings. It's better to extend the existing `visit` method to simply handle `None` as well.
* Replaced `getattr` calls as they are not needed anymore. All currently supported versions do have these attributes.

## Type of Changes

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -54,18 +54,6 @@ PY38 = sys.version_info >= (3, 8)
 PY39 = sys.version_info >= (3, 9)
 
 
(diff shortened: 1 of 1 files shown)


cdce8p:
@Pierre-Sassoulas The last one before I add type annotations

Pierre-Sassoulas (approved):
Definitely a great simplification 🔥 It does not make sense to handle ``Optional[]`` everywhere if we can avoid it.

PR #1020, opened 2021-06-10
Improve inference of Enum members called "name" and "value"

nelfin:
Improve inference of Enum members called "name" and "value"

## Steps

## Description

The attributes .name and .value on the enum.Enum are defined as data-descriptors (but not plain `@property` methods) that behave differently when the attribute access is on the class or an instance. This change updates the enum brain to provide better guesses for values (and appropriate changes if the enum happens to define a member named "name" or "value").

## Type of Changes

## Related Issue

PyCQA/pylint#1932
PyCQA/pylint#2062


Pierre-Sassoulas (approved):
The explanation in the comment is great, I needed that :)

nelfin:
> The explanation in the comment is great, I needed that :)

Thanks. I was thinking that it'd probably need explaining, and instead of sticking it in the PR, I'll just stick it in a comment straight away.

Re: the test failure on 3.10, it looks like the enum module has undergone a fairly extensive re-write. I'll update the tests since I don't think we really care what the value of the ".value" attribute is in a class context

PR #1022, opened 2021-06-12
Add exclude_lines coverage

cdce8p:
## Description
Add `exclude_lines` setting to `coveragerc` file. Some lines will never be executed. We shouldn't use them to calculate coverage.

https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html#advanced-exclusion
Example Home-Assistant: [.coveragerc](https://github.com/home-assistant/core/blob/c242e56b8c9eb2b4de51aac24406e638cfa843de/.coveragerc#L1249)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,3 +7,13 @@ include =
     astroid/*
 omit =
     */tests/*
+exclude_lines =
+    # Re-enable default pragma
+    pragma: no cover
+
+    # Debug-only code
+    def __repr__
+
+    # Type checking code not executed during pytest runs
+    if TYPE_CHECKING:
+    @overload


cdce8p:
@Pierre-Sassoulas Anything else that you can think of what is worth adding?

Pierre-Sassoulas (approved):
Testing that a ``NotImplemented`` is raised do not take a lot of effort so I'd test it. What is ``@overload`` though ?

cdce8p:
> Testing that a `NotImplemented` is raised do not take a lot of effort so I'd test it. What is `@overload` though ?

It's a typing decorator. I'll use open a MR with it soon. It should be a lot clear then.
https://docs.python.org/3/library/typing.html#typing.overload

Pierre-Sassoulas:
Neat, I didn't know about this one. Makes sense to not compute coverage for it too.

PR #1023, opened 2021-06-12
Remove py2 nodes

cdce8p:
## Description
`Repr`, `Exec`, and `Print` are no longer needed as these were used for Python 2 only.
Some additional `rebuilder` changes:
* **visit_call**: `starargs` and `kwargs` are also Python 2 only. These are now included in `args` and `keywords`.
* **visit_name**: In Python 2 `id` it was possible to reassign `True`, `False`, and `None`. With Python 3 this is a syntax error.

https://docs.python.org/3.6/library/ast.html#abstract-grammar

## Type of Changes


Pierre-Sassoulas (approved):
Nice cleanup ! I guess you're doing that after checking the coverage ? 😄  The integration in github MR is not that great and having more visible coverage in the interface would permit to clean up a lot of code. (What do you think about adding something like https://about.codecov.io/blog/announcing-line-by-line-coverage-via-github-checks/ ?)

cdce8p:
> Nice cleanup ! I guess you're doing that after checking the coverage ? 😄

Not  quite. I actually noticed it while working on typing. I could find the `ast` nodes and thus discovered that they aren't used anymore.

> The integration in github MR is not that great and having more visible coverage in the interface would permit to clean up a lot of code.

I agree. I can't even see  which lines are actually covered. `Source not available` when selecting a specific file: https://coveralls.io/jobs/81875682 `Codecov´ seems to handle it better.

> What do you think about adding something like https://about.codecov.io/blog/announcing-line-by-line-coverage-via-github-checks/ ?

I'm +0 on that one. If you like go for it. As mentioned before, just seeing which lines are actually covered would go a long way. That applies to `pylint` as well, btw.

cdce8p:
One point I didn't mention yet. I wouldn't consider this MR a breaking change as I only removed nodes that aren't used anymore. In theory it could still break some plugin which hasn't removed all Python 2 code. We'll have to see about that.

Pierre-Sassoulas:
> That applies to pylint as well, btw.

Sure :) I don't like codecov particularly, just checking what exists. Do you have experience on something like this in home assistant ?

Pierre-Sassoulas:
Using gitlab CI there is an indicator on the left side of the code to show if the code is covered or not... It's very handy.

cdce8p:
> Do you have experience on something like this in home assistant ?

I had to check. Seems like they are using codecov 😅 https://app.codecov.io/gh/home-assistant/core
If you find something else, that would work too.

Pierre-Sassoulas:
> it could still break some plugin which hasn't removed all Python 2 code. We'll have to see about that.

I think we can put this in the changelog and release as a minor version, python 2 was already removed from pylint a long time ago, in 2.0.

cdce8p:
> > it could still break some plugin which hasn't removed all Python 2 code. We'll have to see about that.
> 
> I think we can put this in the changelog and release as a minor version, python 2 was already removed from pylint a long time ago, in 2.0.

That's what I was thinking as well!

PR #1024, opened 2021-06-13
Add type annotations to rebuilder

cdce8p:
## Description

Now for the payoff of the previous MRs: Adding type annotations to the rebuilder module. With that it will be much easier to begin adding type hints to the node classes as well.

Some details on the implementation:
* The main part of the `TreeBuilder` class is obviously the `visit` method as it delegates most incoming function calls. Therefore I believe it is important to type it as best as possible. Here I made extensive use of the `@overload` decorator. As long as for any new `visit_xxx` method a new `overload` call is added, it should work.
* Some `visit_xxx` methods aren't natively called as there are no corresponding `ast` nodes. Thus the `visit` method doesn't have an `overload` for them. It might make sense to rename in the future. For now I've added note in their docstrings. Those are:
  * `visit_module` (called as a starting point directly)
  * `visit_assignname`
  * `visit_decorators`
* The `overload` for `visit_ellipsis` is currently incompatible with the remain ones, since `nodes.Ellipsis` doesn't inherit from `nodes.Const`. It would make sense to directly replace it with `nodes.Const` as that is already happing for Python 3.9+. Some other methods already do that: `visit_str`, `visit_num`, ... I plan an opening a MR for that after this one is done.
* `visit_const` and `visit_emptynode` seems to be never called. For `visit_const` there is `visit_constant`. Not sure but it might be that the `ast` node was named `Const` at one point . However, since at least 2.7 it's called `Constant`. `visit_emptynode` is not called as it is a node type specially create for `astroid`. I'll investigate if both can be removed safely.

In summary, typing isn't perfect just yet, but it's definitely a good starting point. The remaining issues can be addressed later. We first need to get to a point where we can even enable `mypy` 😅

## Type of Changes

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -30,17 +30,38 @@ order to get a single Astroid representation
 """
 
 import sys
-from typing import TYPE_CHECKING, Optional
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
+
+try:
+    from typing import Final
+except ImportError:
+    # typing.Final was added in Python 3.8
+    from typing_extensions import Final
 
 import astroid
-from astroid import nodes
+from astroid import node_classes, nodes
 from astroid._ast import ParserModule, get_parser_module, parse_function_type_comment
 from astroid.node_classes import NodeNG
 
 if TYPE_CHECKING:
     import ast
 
-REDIRECT = {
+    from astroid.manager import AstroidManager
+
+REDIRECT: Final[Dict[str, str]] = {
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Wow, nice job. This must have taken some time ! Maybe we can even start adding the `py.typed` now (from pep 561) ? Being able to activate mypy would be a huge step indeed.

cdce8p:
> Maybe we can even start adding the `py.typed` now (from pep 561) ? Being able to activate mypy would be a huge step indeed.

That is probably a bit early. We still have 187 mypy errors. Most of them in `node_classes`. Additionally the change only fully works in `mypy` with Python 3.10 as the Match nodes aren't included before then. We probably need to add version guards to these code parts, i.e. `if sys.version_info >= (3, 10): ...` Never the less, it's already a great help for additional additional types.

PR #1025, opened 2021-06-13
Deprecate Ellipsis node

cdce8p:
## Description
The `ast.Ellipsis` node was deprecated in Python 3.8 and fully replaced by `ast.Constant`. Similar to `ast.Str`, which was also replaced in Python 3.8, the `visit_ellipsis` method will now return a `nodes.Const` node. That will simplify development as the node for `Ellipsis` will no longer depend on the Python version.

#### Breaking Change
Although technically a breaking change, this only effects checkers that haven't been updated for Python 3.8. Thus I imagine the impact will be fairly limited. `pylint` tests pass without any changes. End user code will not be impacted. It will still be possible to check Python 3.6 only code.

#### Required Changes
After the next release of `astroid`, all references to the astroid `Ellipsis` node need to be removed from `pylint`. That's quite easy to do: https://github.com/PyCQA/pylint/pull/4567

#### Removal
I suggest to remove the `Ellipsis` node with the release of the next minor version, i.e. `2.7.0`.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,8 +10,8 @@ Release Date: TBA
   they represented have been removed with the change to Python 3
 
 * Deprecate ``Ellipsis`` node. It will be removed with the next minor release.
-  Checker that already support Python 3.8+ work without issues. It's only
-  necessary to remove all references to the astroid ``Ellipsis`` node.
+  Checkers that already support Python 3.8+ work without issues. It's only
+  necessary to remove all references to the ``astroid.Ellipsis`` node.
   This changes will make development of checkers easier as the resulting tree for Ellipsis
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
I like that we use astroid as a wrapper so we can handle old versions as if they were the latest. We could do that for everything and remove the interpreter version check in pylint entirely !

cdce8p:
> We could do that for everything and remove the interpreter version check in pylint entirely !

I'm working on `Subscript` with `Index` and `ExtSlice` at the moment. Are there any other that I'm missing?

Pierre-Sassoulas:
I don't think of anything else right now.

PR #1032, opened 2021-06-15
Small fix + improvements to Match nodes

cdce8p:
## Description
* Small fix for `MatchClass`: `kwd_attrs` isn't an `_astroid_field`. Instead it should belong to `_other_fields`.
* Added additional tests for Match `get_children` methods.
* Added type annotations for `_astroid_fields` and `_other_fields` for Match nodes.
* Removed `get_children` methods from Match nodes. As long as `_astroid_fields` is set correctly, `NodeNG.get_children` handles it perfectly fine. My focus here was just on the Match nodes, but it can probably be removed from others as well.
* Added `yield  from ()` as default for `NodeNG.get_children`, just in case a node doesn't have any items for `_astroid_fields`. That should make `mixins.NoChildrenMixin` redundant.

## Type of Changes

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -4851,7 +4851,8 @@ class MatchClass(NodeNG):
     <MatchClass l.5 at 0x10ca80880>
     """
 
-    _astroid_fields = ("cls", "patterns", "kwd_attrs", "kwd_patterns")
+    _astroid_fields = ("cls", "patterns", "kwd_patterns")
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
Removing the ``NoChildrenMixin`` look like a pretty nice simplification. The mixins are such a plague in pylint/astroid imho. Also the added typing 👍 Anyway good job !

PR #1033, opened 2021-06-15
Update Match nodes to be internally consistent

cdce8p:
## Description
* Define instance variables in `__init__` even if they are `None` at first
* At `ClassVar` to class variables: `_astroid_fields`, `_other_fields`
* Remove default parameter from `postinit`. Those should just always be called with all necessary parameters. Reduces the possibility of unknowingly missing one. If you really what to set just one, do it through direct assignment. At the end `postinit` is just a helper method anyway.
* Removed `Optional` annotation from `postinit` arguments where instance variable can't actually be `None`. Unfortunately `Optional` is still needed as most instance variables are initialized with `None`.
* Replaced `None` with empty list for initialization (where possible)
* Require keyword arguments for init of `MatchClass` as it accepts neither `lineno` nor `col_offset`
* Reorder arguments for `MatchSingleton.__init__`. `value` is require and thus should come first, `lineno`, `col_offset`, and `parent` can be optional. Change to require keyword arguments here as well.
* I was thinking about requiring keyword arguments for all `__init__` methods, but decided that the default (just `lineno`, `col_offset`, and `parent`) is probably fine. Only if that changes should we require keywords.

#### Breaking change
Technically this is a breaking change for all `Match` nodes. However, as it was just added recently, I think it's probably fine to do it. Better now than later, anyway.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,7 @@ Release Date: TBA
   depend on the python version. **Background**: With Python 3.9 ``ast.Index``
   and ``ast.ExtSlice`` were merged into the ``ast.Subscript`` node.
 
+* Updated all Match nodes to be internally consistent.
 
 
 What's New in astroid 2.5.8?
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -40,6 +40,7 @@ import pprint
 import typing
 from functools import lru_cache
 from functools import singledispatch as _singledispatch
+from typing import Optional
 
 from astroid import as_string, bases
 from astroid import context as contextmod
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):
Thank you for the typing work ! Would the refactor from ``Optional[x]`` to simply ``x`` would be big ? It seems that forcing to have the real value at instantiation when possible would simplify the code long term ?

cdce8p:
> It seems that forcing to have the real value at instantiation when possible would simplify the code long term ?

Indeed, that would be preferred. Mypy doesn't know that `postinit` is called immediately afterwards. We can't even assume that as some plugins might miss that when they use `transform`, so `mypy` is right here. The only really solution would be to switch the way instance variables and `parent` are assigned, i.e. parents need to be assigned retroactively. Tbh that would be a huge undertaking as other parts depend on `parent` being assigned first. Additionally, it would definitely break **ALL** custom plugins that use `transform` or set custom `inference_tips`. It might be possible, but I don't believe we should do it. Some things you have to live with, unfortunately.

> Would the refactor from `Optional[x]` to simply `x` would be big ?

That is something else that we can / should pursue. Just moving the initialization of instance variables to the `__init__` method is cumbersome, but doable, even without a breaking change! With that all list types can be set to `[]`. That doesn't work in a class context as all instance would work with the same list.

Changing the signature for the remaining `postinit` methods would also be possible. Although I would only combine it with a bump to `v3.0` as it would be a major breaking change. However that would also allow us to remove some old code. So not entirely a bad idea. There is also the MR for the `importFrom` node and I was planing on refactoring the `arguments` and `try` nodes to better match the current ast. Maybe we should open a tracking issue for `v3.0`?

PR #1034, opened 2021-06-15
Create a file for inference tip

Pierre-Sassoulas:
## Description

This permit to not have to import wrapt directly in ``__init__.py``

It's a dependency and we want to be able to use ``attr: astroid.__version__`` for packaging without installing dependencies first.

## Type of Changes

## Related Issue

#1030

### DIFF
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -40,12 +40,11 @@ Main modules are:
 """
 
 import enum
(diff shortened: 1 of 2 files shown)


cdce8p (approved):
Looks good. I just ran the pylint tests and everything seems to work as expected.

PR #1035, opened 2021-06-15
Reduce cyclic import by always importing exceptions from astroid.exceptions internally

Pierre-Sassoulas:
## Description

This normalize the way the exceptions are imported. Previously there was a wildcard import of exceptions in ``astroid.__init__.py`` and then:
```python
from astroid import exceptions

exceptions.InferenceError
```

```python
import astroid

astroid.InferenceError
```
This created circular imports in all the file where it was used. The one true way **internally** is:

```python
from astroid.exceptions import InferenceError

InferenceError
```

The import should be done from astroid like before for external code (in pylint for example).

## Type of Changes

### DIFF
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -60,7 +60,34 @@ del _Context
 # pylint: disable=wrong-import-order,wrong-import-position,redefined-builtin
 
 # make all exception classes accessible from astroid package
-from astroid.exceptions import *
+from astroid.exceptions import (
+    AstroidBuildingError,
+    AstroidBuildingException,
+    AstroidError,
+    AstroidImportError,
+    AstroidIndexError,
+    AstroidSyntaxError,
+    AstroidTypeError,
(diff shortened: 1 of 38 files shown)


astroid/__init__.py
```diff
@@ -60,7 +60,34 @@
 # pylint: disable=wrong-import-order,wrong-import-position,redefined-builtin
 
 # make all exception classes accessible from astroid package
-from astroid.exceptions import *
+from astroid.exceptions import (
```
cdce8p, line 63: Instead of importing all exceptions individually here, what do you think about adding `__all__` to `astroid.exceptions`?

cdce8p (commented):
I like this change. In general, it makes much more sense to import the individual exceptions.

astroid/__init__.py
Pierre-Sassoulas: Make sense, I'm going to do it.

PR #1036, opened 2021-06-15
Remove unused rebuilder methods

cdce8p:
## Description
Followup to #1024
It seems `TreeRebuilder.visit_const` and `TreeRebuilder.visit_emptynode` can safely be removed as they don't have corresponding ast nodes.

**Notes**
- Constants are handled by `visit_constant`
- `EmptyNode` is only an astroid node class

## Type of Changes

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -776,16 +776,6 @@ class TreeRebuilder:
         )
         return newnode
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Astroid is getting nicer little by little 😄

PR #1040, opened 2021-06-17
Start typing node classes 01

cdce8p:
## Description

Start adding type annotations to some node classes.

* For: `NodeNG`, `Statement`, and `_BaseContainer`
* Only `ClassVar`, `__init__`, `postinit`, and instance variables
* Moved instance variable initialization to `__init__` (`NodeNG`)
* Added `autoclass_content = "both"` to properly document `__init__`
* `:type ...:` docstring can now be removed as it's handled by type annotation

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#configuration

The plan is to continue with smaller increments as doing all nodes at once would be difficult to review and keep track of.

## Type of Changes

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -274,77 +274,56 @@ class NodeNG:
     This is the base class for all Astroid node classes.
     """
 
-    is_statement = False
-    """Whether this node indicates a statement.
-
-    :type: bool
-    """
-    optional_assign = False  # True for For (and for Comprehension if py <3.0)
+    is_statement: ClassVar[bool] = False
+    """Whether this node indicates a statement."""
+    optional_assign: ClassVar[
+        bool
+    ] = False  # True for For (and for Comprehension if py <3.0)
     """Whether this node optionally assigns a variable.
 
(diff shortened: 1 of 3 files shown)


astroid/node_classes.py
```diff
@@ -274,77 +274,56 @@ class NodeNG:
...
 
     This is for loop assignments because loop won't necessarily perform an
     assignment if the loop has no iterations.
     This is also the case from comprehensions in Python 2.
-
-    :type: bool
     """
-    is_function = False  # True for FunctionDef nodes
-    """Whether this node indicates a function.
+    is_function: ClassVar[bool] = False  # True for FunctionDef nodes
```
Pierre-Sassoulas, line 288: I don't understand why it's not `bool` here ?

Pierre-Sassoulas (approved):
Thank you for this, I juste have a small question.

astroid/node_classes.py
cdce8p: It is. Just additionally marked as ClassVar

PR #1043, opened 2021-06-17
Create a astroid.constants.py and put the python interpreter check in it

Pierre-Sassoulas:
## Description

Little clearer code, the constants.py will be useful to avoid circular imports, and this is a performance improvment as the version checks will be done once instead of multiple time during parsing.

## Type of Changes

### DIFF
--- a/astroid/constants.py
+++ b/astroid/constants.py
@@ -1,6 +1,5 @@
 import sys
 
-PY36 = sys.version_info >= (3, 6)
 PY37 = sys.version_info >= (3, 7)
 PY38 = sys.version_info >= (3, 8)
 PY39 = sys.version_info >= (3, 9)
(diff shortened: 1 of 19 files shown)


astroid/constants.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+PY36 = sys.version_info >= (3, 6)
```
cdce8p, line 3: Isn't that always `True`? Python 3.6 is the min required version

cdce8p (commented):
I guess that will work

astroid/constants.py
Pierre-Sassoulas: You're right, I removed the check for less than py36 but I can remove that too.

PR #1045, opened 2021-06-17
Break avoidable cyclic import, apply isort on ``astroid.__init__.py``

Pierre-Sassoulas:
## Description

This breaks a number of existing cyclic import. Enough to sort ``__init__.py``, not enough to enable the cyclic-import warning in pylint.

## Type of Changes

### DIFF
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -6,9 +6,9 @@ Astroid hook for the attrs library
 Without this hook pylint reports unsupported-assignment-operation
 for attrs classes
 """
-
-import astroid
 from astroid import MANAGER
+from astroid.node_classes import AnnAssign, Assign, Call, Unknown
+from astroid.scoped_nodes import ClassDef
 
 ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib"))
 ATTRS_NAMES = frozenset(("attr.s", "attrs", "attr.attrs", "attr.attributes"))
@@ -20,7 +20,7 @@ def is_decorated_with_attrs(node, decorator_names=ATTRS_NAMES):
     if not node.decorators:
         return False
     for decorator_attribute in node.decorators.nodes:
-        if isinstance(decorator_attribute, astroid.Call):  # decorator with arguments
+        if isinstance(decorator_attribute, Call):  # decorator with arguments
             decorator_attribute = decorator_attribute.func
         if decorator_attribute.as_string() in decorator_names:
             return True
@@ -33,12 +33,12 @@ def attr_attributes_transform(node):
     """
     # Astroid can't infer this attribute properly
     # Prevents https://github.com/PyCQA/pylint/issues/1884
-    node.locals["__attrs_attrs__"] = [astroid.Unknown(parent=node)]
+    node.locals["__attrs_attrs__"] = [Unknown(parent=node)]
 
     for cdefbodynode in node.body:
-        if not isinstance(cdefbodynode, (astroid.Assign, astroid.AnnAssign)):
+        if not isinstance(cdefbodynode, (Assign, AnnAssign)):
             continue
-        if isinstance(cdefbodynode.value, astroid.Call):
+        if isinstance(cdefbodynode.value, Call):
             if cdefbodynode.value.func.as_string() not in ATTRIB_NAMES:
                 continue
         else:
@@ -50,7 +50,7 @@ def attr_attributes_transform(node):
         )
         for target in targets:
 
-            rhs_node = astroid.Unknown(
+            rhs_node = Unknown(
                 lineno=cdefbodynode.lineno,
                 col_offset=cdefbodynode.col_offset,
                 parent=cdefbodynode,
@@ -59,6 +59,4 @@ def attr_attributes_transform(node):
             node.instance_attrs[target.name] = [rhs_node]
 
 
-MANAGER.register_transform(
-    astroid.ClassDef, attr_attributes_transform, is_decorated_with_attrs
-)
+MANAGER.register_transform(ClassDef, attr_attributes_transform, is_decorated_with_attrs)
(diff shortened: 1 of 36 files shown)


astroid/context.py
```diff
@@ -12,12 +12,34 @@
...
+
+class Context(enum.Enum):
+    Load = 1
+    Store = 2
+    Del = 3
+
+
+Load: Context = Context.Load
+Store: Context = Context.Store
+Del: Context = Context.Del
```
cdce8p, line 42: I don't think `astroid/context.py` is the right file for it. True `Context` and `context`, but as I understand it the file mostly contains stuff regarding the `InferenceContext`. `const.py` would probably be better for it. Btw. what do you think about renaming `constants.py` to `const.py`?

astroid/nodes.py
```diff
@@ -20,10 +20,94 @@
...
 All nodes inherit from :class:`~astroid.node_classes.NodeNG`.
 """
-# pylint: disable=redefined-builtin
 
 # Nodes not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
-from astroid.node_classes import (
+
+__all__ = (
+    "AsyncFunctionDef",
+    "AsyncFor",
```
cdce8p, line 28: Not sure `__all__` is necessary here. Don't we already only export those names?

astroid/rebuilder.py
```diff
@@ -135,7 +136,7 @@ def _get_context(
             "ast.Tuple",
         ],
     ):  # TODO return type needs change to _Context enum
-        return self._parser_module.context_classes.get(type(node.ctx), astroid.Load)
+        return self._parser_module.context_classes.get(type(node.ctx), Load)
```
cdce8p, line 139: I would just import `Context` and then use `Context.Load`. That would also make it a bit clearer.

astroid/brain/__init__.py
```diff
@@ -0,0 +1,3 @@
+__all__ = ["register_module_extender"]
+
+from astroid.brain.register_module_extender import register_module_extender
```
cdce8p, line 3: This isn't necessary. Nobody should use `from astroid.brain import *` in the first place. There isn't a need for it.

astroid/brain/register_module_extender.py
```diff
@@ -0,0 +1,13 @@
+from astroid.scoped_nodes import Module
+
+
+def register_module_extender(manager, module_name, get_extension_mod):
```
cdce8p, line 4: I feel like a general `astroid/brain/helper[s].py` file would be a better fit.

astroid/inference.py
```diff
@@ -45,11 +54,62 @@
...
+    is_subtype,
+    is_supertype,
+    object_type,
+    safe_infer,
+)
 from astroid.interpreter import dunder_lookup
+from astroid.manager import AstroidManager
+from astroid.node_classes import (
+    Arguments,
+    AssignAttr,
```
cdce8p, line 68: I'm not sure this is really an improvement here. Readability wise I would prefer the old version of `astroid/inference.py`. Especially for `nodes`, `decorators`, `helpers`, `protocols`, and `utils`.

cdce8p (commented):
Left a few comments

astroid/context.py
Pierre-Sassoulas: I put that in contants originally so yes I agree 😄

astroid/rebuilder.py
Pierre-Sassoulas: Yes I was trying to not change the way it was done before but it's hard to justify doing it like this.

astroid/inference.py
Pierre-Sassoulas: The thing is import by entire module look nicer but they also hide circular imports. Files are enormous so we want to use import by module so it feels reasonable, but I think we also need to burst those enormous 5000+ lines file importing 50+ classes from a single file, so we have manageable import in the results and we can see where the circular imports are. There are global values hidden because of that (Thing like that are not a one time thing: https://github.com/PyCQA/astroid/pull/972/files). Maybe we could start by splitting file first ?

cdce8p: Let's do everything except `nodes`. It does have `__all__` so I think it should be fine (for now).

Pierre-Sassoulas: I'm going to split this MR into multiple one, there is too much in it right now.

astroid/nodes.py
Pierre-Sassoulas: Right

astroid/inference.py
Pierre-Sassoulas: I removed this part, it did not help for the cyclic import at all in facT.

astroid/brain/brain_attrs.py
```diff
@@ -6,9 +6,9 @@
 Without this hook pylint reports unsupported-assignment-operation
 for attrs classes
 """
-
-import astroid
 from astroid import MANAGER
+from astroid.node_classes import AnnAssign, Assign, Call, Unknown
+from astroid.scoped_nodes import ClassDef
```
cdce8p, line 11: I still think it would make sense to import those from `astroid.nodes` instead. Even more so if we want to reorganize `node_classes` in the future. Might make sense to split those 5000 lines into multiple files inside a sub-package (called nodes?).

Pierre-Sassoulas: >  Might make sense to split those 5000 lines into multiple files inside a sub-package (called nodes?).

Yes, I'm working on it ! 

But right now there are cyclic import that prevent to import directly from nodes without hacks (and will for a long time seeing how it goes right now). I want to do this in multiple small steps :) So yes the end goal would be to import from ``astroid.nodes``

Pierre-Sassoulas: Original idea here : https://github.com/PyCQA/astroid/pull/587#issuecomment-403326307

cdce8p: Sounds good. I would just suggest to do it in smaller pieces. That will make resolving conflicts easier.

PR #1051, opened 2021-06-18
Remove Optional annotation if parameter is required

cdce8p:
## Description
Currently, some instance variables are typed with `Optional` just because they are initialized with `None`. They are later set during the `postinit` call, but that doesn't allow us to change the initial type. The consequence is that `pylint` has to assume `None` is possible although it technically will never be (at least not in correct code).

If `postinit` requires the correct type and doesn't allow the use of `None`, it is possible to safely remove the initialization and with that the `Optional` annotation.

## Type of Changes

### DIFF
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -4805,7 +4805,7 @@ class Match(Statement):
         col_offset: Optional[int] = None,
         parent: Optional[NodeNG] = None,
     ) -> None:
-        self.subject: Optional[NodeNG] = None
+        self.subject: NodeNG
         self.cases: typing.List["MatchCase"] = []
         super().__init__(lineno=lineno, col_offset=col_offset, parent=parent)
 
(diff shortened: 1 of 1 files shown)


cdce8p:
@Pierre-Sassoulas What do you think? This will improve typing. The cost however will be that we could encounter errors if `postinit` isn't called. I think the tradeoff might be worth it, especially since developers (of custom checkers / extensions) should notice the error immediately.

Pierre-Sassoulas:
I think it's reasonable, there's a lot of work to do so that typing is useful, I tried to add mypy and there's 1000+ warnings right now. This is a step in the right direction, I'm sure in some places the code do not handle the case for None because ``postinit`` is always called.

Pierre-Sassoulas (approved):

cdce8p:
> I'm sure in some places the code do not handle the case for None because postinit is always called.

I would think so too. It wouldn't make much sense otherwise. We would need to add special case handling for half the variables which aren't / shouldn't ever be `None`. At least with this change IDE type hints will recognize them correctly.

It was pretty easy for the Match nodes as all arguments for `postinit` are required. That isn't the case for the remaining once. I feel like all of this is leading up to a breaking change in the future.

PR #1052, opened 2021-06-18
Create a ``Context`` enum to fix a circular import in ``astroid.__init__.py`` and improve typing

Pierre-Sassoulas:
## Description

First step toward #1045 

## Type of Changes


astroid/__init__.py
```diff
@@ -69,6 +62,8 @@
 
 # more stuff available
 from astroid import raw_building
+
+from astroid.const import Context, Load, Store, Del
```
cdce8p, line 66: Do we want to add `Context` here? `Load`, `Store`, and `Del` are required so we don't break stuff but I feel like it might make sense to not add any more. We could even think about removing all here for the next major release. It shouldn't be a big issue to import it directly from `astroid.const`. What do you think?

The spirit of `astroid` so far has been to export everything from `__init__`, but that's obviously causing issues now.

astroid/const.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Context(enum.Enum):
+    Load = 1
+    Store = 2
+    Del = 3
+
+
+Load: Context = Context.Load
+Store: Context = Context.Store
+Del: Context = Context.Del
```
cdce8p, line 18:
```suggestion
Load = Context.Load
Store = Context.Store
Del = Context.Del
```
That should be enough

astroid/rebuilder.py
```diff
@@ -135,7 +135,7 @@ def _get_context(
             "ast.Tuple",
         ],
     ):  # TODO return type needs change to _Context enum
-        return self._parser_module.context_classes.get(type(node.ctx), astroid.Load)
```
cdce8p, line 138: Would you mind fixing the `TODO` here as well? I put it there for exactly the time we refactor `Context` 😅 It should probably be: `-> Context`

cdce8p (changes requested):

astroid/__init__.py
Pierre-Sassoulas: I think the import from astroid init directly are not a problem as long as    it's in pylint. It actually permits to change the structure in astroid without creating breaking changes. In pylint an import directly from a module must even be avoided as a change in structure would break it.

 But in astroid itself we can't do that as it's creating cyclic import and more problem than what an elegant import solves. (Runtime definition of variable or function in some class). And if we're refactoring a module in astroid we're aware of it so changing the import is not costly at all (my ide handles everything).

astroid/rebuilder.py
Pierre-Sassoulas: Nice catch. Yeah I thought that deleting the temporary enum would prevent from using it as Typing; Very strange design decision (probably taken in python 2's days).

astroid/__init__.py
Pierre-Sassoulas: So yes I think we should create an API with everything in ``astroid.__init__.py`` so we can refactor astroid without affecting the tools using it.

cdce8p: > So yes I think we should create an API with everything in `astroid.__init__.py` so we can refactor astroid without affecting the tools using it.

Ok. That does make sense.

PR #1054, opened 2021-06-19
Create a file for the global manager to prevent cyclic import

Pierre-Sassoulas:
## Description

This remove 6 cyclic imports. There was a global ``MANAGER`` that was directly in ``astroid.__init__.py`` that could not be imported another way than by using directly ``from astroid import MANAGER``. I also realized that ``AstroidManager`` is a singleton so the intermediate variables do not matter.

## Type of Changes

## Related Issue

Need #1045 to be merged first.

### DIFF
--- a/astroid/brain/brain_argparse.py
+++ b/astroid/brain/brain_argparse.py
@@ -1,4 +1,5 @@
-from astroid import MANAGER, arguments, inference_tip, nodes
+from astroid import arguments, inference_tip, nodes
+from astroid.astroid_manager import MANAGER
 from astroid.exceptions import UseInferenceDefault
 
 
--- a/astroid/brain/brain_argparse.py
+++ b/astroid/brain/brain_argparse.py
@@ -1,6 +1,6 @@
 from astroid import arguments, inference_tip, nodes
-from astroid.astroid_manager import MANAGER
 from astroid.exceptions import UseInferenceDefault
+from astroid.manager import AstroidManager
 
 
 def infer_namespace(node, context=None):
@@ -30,6 +30,6 @@ def _looks_like_namespace(node):
     return False
 
 
-MANAGER.register_transform(
+AstroidManager().register_transform(
     nodes.Call, inference_tip(infer_namespace), _looks_like_namespace
 )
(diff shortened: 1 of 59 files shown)


Pierre-Sassoulas:
I'm going to wait for a review of this one because rebasing on it will be painful.

astroid/brain/brain_argparse.py
```diff
@@ -29,6 +30,6 @@ def _looks_like_namespace(node):
     return False
 
 
-MANAGER.register_transform(
+AstroidManager().register_transform(
```
cdce8p, line 33: I'm probably missing something, but isn't the idea to use the `MANAGER` singleton here?

Pierre-Sassoulas: As it's a singleton/borg we can also instantiate.

cdce8p: But isn't `MANGER` the singleton. `AstroidManger()` will create a new instance of `AstroidManager`.

Pierre-Sassoulas: I think that the principle of the singleton is to get the same unique instance when you instantiate wherever you want, so you don't need to keep track of a global. I don't think this take much more time this way as ``self.__dict__`` will already be set. How do you use the singleton pattern ?

cdce8p: > I don't think this take much more time this way as self.__dict__ will already be set.

I didn't saw that. You're right, that it will create a shared state which would work.

> How do you use the singleton pattern ?

I expected something like this
```py
class Manager:
    _instance = None

    def __new__(cls):
        if cls._instance is not None:
            cls._instance = super().__new__(cls)
        return cls._instance

Manager() == Manager()  # true
```

Pierre-Sassoulas: Ha, yes it's a "borg" not a singleton, so the implementation is a little different. I had to check what a borg is in fact because I did not know. I don't think it was clear for other astroid's dev either because there was a MANAGER constants in various module. Originally I just wanted to change the import. Apparently a borg is better for multiprocessing (?)

cdce8p: > Apparently a borg is better for multiprocessing (?)

Don't know about that. Had to look up what a `borg` is myself. We can keep it like this then

PR #1058, opened 2021-06-19
Create a statement file from node_classes.

Pierre-Sassoulas:
## Description

Small step toward bursting node_classes. There is a return Statement that could go in this file too, but right not a typing would create a circular import.

Require #1057 to be merged first.

## Type of Changes

### DIFF
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -68,7 +68,6 @@ from astroid.inference_tip import _inference_tip_cached, inference_tip
 # importing with a wildcard would clash with astroid/nodes/scoped_nodes
 # and astroid/nodes/node_classes.
 from astroid.nodes import (  # pylint: disable=redefined-builtin (Ellipsis)
-    CONST_CLS,
     AnnAssign,
     Arguments,
     Assert,
--- a/astroid/node_classes.py
+++ b/astroid/node_classes.py
@@ -2,7 +2,7 @@
 
 import warnings
 
-from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (Ellipsis)
+from astroid.nodes import (  # pylint: disable=redefined-builtin (Ellipsis)
     CONST_CLS,
     AnnAssign,
     Arguments,
@@ -46,6 +46,7 @@ from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (E
     JoinedStr,
     Keyword,
     List,
+    LocalsDictNodeNG,
     Match,
     MatchAs,
     MatchCase,
--- a/astroid/nodes/__init__.py
+++ b/astroid/nodes/__init__.py
@@ -27,7 +27,6 @@ from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (E
     CONST_CLS,
     AnnAssign,
     Arguments,
-    Assert,
     Assign,
     AssignAttr,
     AssignName,
@@ -54,7 +53,6 @@ from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (E
     EmptyNode,
     EvaluatedObject,
     ExceptHandler,
-    Expr,
     ExtSlice,
     For,
     FormattedValue,
@@ -83,7 +81,6 @@ from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (E
     Nonlocal,
     Pass,
     Pattern,
-    Raise,
     Return,
     Set,
     Slice,
@@ -98,6 +95,7 @@ from astroid.nodes.node_classes import (  # pylint: disable=redefined-builtin (E
     With,
     Yield,
     YieldFrom,
+    _BaseContainer,
     are_exclusive,
     const_factory,
     unpack_infer,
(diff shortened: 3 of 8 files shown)


Pierre-Sassoulas:
@cdce8p I've tried to remove that TODO but this is a trap. Ended up creating a file from base containers and then the circular import become apparent and very hard to fix. Clearly those circular imports are going to take some serious time to fix. It was a lucky guess to try to do statements first after NodeNG because the whole file is boobytrapped 😄 Maybe let's start with not moving too much mud.

cdce8p (commented):
You can fix the circular import by keeping `Statement` in `node_classes` for now. That way you are able to import `Tuple` in `statement.py`. That does especially make sense considering that there are still a lot of other classes that also inherit from `Statement`. An alternative would be to create a separate file for the statement base class.

I'm curious why you decided to move these selected nodes in particular as there are quite a few more: https://docs.python.org/3.10/library/ast.html. (That doesn't mean that I would support moving all to `statement.py` as we wouldn't gain anything by that.)

--
Of the top of my head some groupings that might make sense:
- Exception handeling: `TryExcept`, `TryFinally`, `ExceptHandler`, `Raise`
- Assignment and delete: `Assign`, `AssignAttr`, `AssignName`, `DelAttr`, `Delete`, `AugAssign`, `AnnAssign`
- Container: `List`, `Set`, `Tuple`, `Starred`, (`_BaseContainer`)
- Loops: `For`, `AsyncFor`, `While`
- Imports: `Import`, `ImportFrom`
- If: `If`, `IfExp`
- Compare: `Compare`, `BoolOp`, `BinOp`, `UnaryOp`
- Subscript: `Subscript`, `Slice`, `ExtSlice`, `Index`
- Call: `Call`, `Argument`, `Keyword`
- Constant: `Const`, `Ellipsis`, 
- Pattern matching: `Match`, `Match**`, `Pattern`

Some might make more sense than others but you might find it useful nevertheless.

Pierre-Sassoulas:
Thank you for thinking about the association in detail, let's use that to burst it. 

I tried to do ``Container: List, Set, Tuple, Starred, (_BaseContainer)`` but this is a very long spaghetti to un-knit. There's a specific mechanism to initialize see ``def _two_step_initialization(cls, value):`` with ``postinit``. To be fair I don't understand well enough what it does for now, but I can see it's probably the reason why the startup time of pylint is bad and why this import cycle will be very hard to break.

So I'm just trying to merge the one group I identified and chose to begin with originally. I took only the class that are pure ``Statement`` without multi-inheritance. I don't know if it makes a lot of sense but the reasoning was that we would not depend on another class so ``statements.py`` could be a base dependency without any circular import. If we keep ``Statement`` in ``node_classes `` this will create a circular import between child class in ``statement.py`` and ``Statement`` itself in ``node_classes.py``, right ?

cdce8p:
> If we keep `Statement` in `node_classes` this will create a circular import between child class in `statement.py` and `Statement` itself in `node_classes.py`, right ?

It shouldn't, or at least didn't when I tired it. As long as you don't put classes in `statement.py` that you than need in `node_classes.py`.

Pierre-Sassoulas:
Ok I understand, we're using a different approach, I'm trying to create base dependencies for ``node_classes.py`` (like NodeNG which was easy) and you're trying to create new file that depends on ``node_classes.py``. Seems like since we already separated the grand mother from the family, now separating the children from their mother is easier than separating the mothers from their children, let's do that 😄 .

cdce8p:
> Seems like since we already separated the grand mother from the family, now separating the children from their mother is easier than separating the mothers from their children, let's do that 😄 .

Took me a moment to understand 😂

> I'm trying to create base dependencies for `node_classes.py` (like NodeNG which was easy)

If that's your goal, you probably need to isolate `class Statement` itself in a separate file. Then it might work.

Pierre-Sassoulas:
> If that's your goal, you probably need to isolate class Statement itself in a separate file. Then it might work.

Agreed ! But in order to do that we need to separate Containers in their own file first, and this looked like a week or more of work. Let's start small and tidy up what was already done. I'm going to use https://github.com/PyCQA/pylint/issues/4814#issuecomment-896831816 to check that something is happening because it would be a shame if performance got worst instead of better.

cdce8p:
> Agreed ! But in order to do that we need to separate Containers in their own file first, and this looked like a week or more of work. Let's start small and tidy up what was already done.

It's completely fine to leave the Containers as is (for now). I just keep mentioning them as they are the most prominent examples. But indeed the better approach would be to start with the small (and hopefully easy) stuff.

> I'm going to use [PyCQA/pylint#4814 (comment)](https://github.com/PyCQA/pylint/issues/4814#issuecomment-896831816) to check that something is happening because it would be a shame if performance got worst instead of better.

I would be surprised if you were to find any performance difference (better or worse) just from moving classes. That shouldn't change anything.

Pierre-Sassoulas:
> I would be surprised if you were to find any performance difference (better or worse) just from moving classes. That shouldn't change anything.

You're probably right 😄 I hope we get a perf gain with the removal of the two step initialization. Hopefully we'll understand better why it's required at this point.

Pierre-Sassoulas:
Well LookupMixin depends on are_exclusive, are_exclusive depends on some statement Node, some statement Node depends on AssignName, AssignName inherit from LookupMixin. This is not something I can do without planning and a couple dedicated hours. Let's put this in draft again. Unless we can merge the current statement.py without much of the child classes in ?

cdce8p:
> Unless we can merge the current statement.py without much of the child classes in ?

Not completely sure what you mean. The way it is now, I wouldn't recommend to merge it. There simply isn't much value in it. Furthermore, it will make anything else more difficult as both `node_classes` and `statement` now depend more or less on each other (just short of having a circular dependency).

At the very least, I would move the `Statement` class back into `node_classes`.

Pierre-Sassoulas:
I'm going to close completely along with #1060, this is just some copy paste and we're going to need to tackle some vicious circular import that will be a lot more complex than that.

PR #1060, opened 2021-06-19
Refactor saved from removing circular import in ``as_string.py``

Pierre-Sassoulas:
## Description

Move ``as_string`` in astroid.nodes and remove the circular import it contains.

## Type of Changes

## Related Issue

Need #1058 to be merged first.


astroid/nodes/as_string.py
```diff
@@ -20,34 +20,17 @@
...
-        MatchOr,
-        MatchSequence,
-        MatchSingleton,
-        MatchStar,
-        MatchValue,
-    )
-
-# pylint: disable=unused-argument
+# Do not add typing as it will create circular import and the function cannot
+# be used with the wrong type anyway because of the visitor pattern.
```
cdce8p, line 24: That's not correct. Adding types will make sure the method contents itself are checked, i.e. that all instance variables for `node` actually exist for that node type. Additionally `if TYPE_CHECKING` should not create circular imports, does it?

astroid/nodes/node_ng.py
```diff
@@ -509,7 +510,7 @@ def as_string(self):
         :returns: The source code.
         :rtype: str
         """
-        return as_string.to_code(self)
+        return AsStringVisitor("    ")(self)
```
cdce8p, line 513: Maybe we could add 4 spaces as default for `AsStringVisitor.indent` in `__init__`?

astroid/nodes/as_string.py
Pierre-Sassoulas: In my view the function are always accessed through the visitor pattern so if ``visit_classdef``` is called, the instance will always be a ``ClassDef`` and the typing does not bring new information. Maybe we need to make the visit function private so it really cannot be accessed any other way ? Also a circular import  (according to pylint, not something that create a python error) is created by an import of Node when we also need to import the visitor in ``NodeNG.as_string``.

cdce8p: Try the following with and without the argument annotation:
```py
class Custom:
    var: int

def visit(node: Custom) -> None:
    print(node.var)
    print(node.does_not_exist)
```

The point isn't that it's always `ClassDef`. That I know, it's to help check that all accessed attributes of node actually exist for the node type.

> Maybe we need to make the visit function private so it really cannot be accessed any other way ?

I don't really see a need for it.

> Also a circular import (according to pylint, not something that create a python error) is created by an import of Node when we also need to import the visitor in NodeNG.as_string.

It works fine the way it currently is. Just `visit_matchas` isn't ideal as the import is needed at runtime for the `isinstance` check.

Pierre-Sassoulas: > Just visit_matchas isn't ideal as the import is needed at runtime for the isinstance check.

Well this MR is supposed to fix that, it's just on top of #1057 and #1058 https://github.com/PyCQA/astroid/pull/1060/commits/8df6cc4cd8c41230f0a8e516e11c55a154c79122

Also I think we're using the first answer on stackoverflow to deal with circular import right now, but the second one that follow PEP484 also make sense : https://stackoverflow.com/a/57005931/2519059, what do you think ?

cdce8p: > Well this MR is supposed to fix that, it's just on top of #1057 and #1058 8df6cc4

I don't think that is worth it. Using imports has the clear advantage that you don't need test to catch errors if the class name changes. Every linter will tell you that. In addition, a string comparison prevents type checkers from doing any kind of type narrowing for the indented block.

> Also I think we're using the first answer on stackoverflow to deal with circular import right now, but the second one that follow PEP484 also make sense : https://stackoverflow.com/a/57005931/2519059, what do you think ?

If you really what to, go for it. Keep in mind however that you would need to create protocol / mixin classes for each and ever one of them. Remember also to always update them if the source changes. I guess that might be a lot of work. Using `if TYPE_CHECKING: ...` is certainly easier.

astroid/nodes/as_string.py
```diff
@@ -624,11 +620,11 @@ def visit_matchstar(self, node: "MatchStar") -> str:
...
-        # pylint: disable=import-outside-toplevel
-        # Prevent circular dependency
-        from astroid.nodes.node_classes import MatchClass, MatchMapping, MatchSequence
-
-        if isinstance(node.parent, (MatchSequence, MatchMapping, MatchClass)):
+        if node.parent.__class__.__name__ in (
+            "MatchSequence",
+            "MatchMapping",
+            "MatchClass",
+        ):
```
cdce8p, line 627: I don't think this should be change. Although the current code isn't ideal either, it's still better than comparing the class name with predefined strings.

astroid/nodes/node_ng.py
cdce8p: This comment is still open

astroid/nodes/as_string.py
Pierre-Sassoulas: We have to find something else then, it means redesigning the code in depth, right ?

cdce8p: The proper way would probably be to resolve the circular dependency. To do that `NodeNG` can't import `as_string`. Thus `NodeNG.as_string()` should be set later.

Pierre-Sassoulas: When you say that it should be set later do you mean using a postinit like in other node classes ? I fear this is the reason that importing astroid takes a long time and that pylint startup is slow now matter what.

cdce8p: What I was thinking about was something similar to how we deal with the inference methods at the moment. I.e.
```py
# astroid/nodes/__init__.py

from astroid.nodes.node_ng import NodeNG
...
from astroid.as_string import AsStringVisitor

def as_string(self):
    return AsStringVisitor("    ")(self)

NodeNG.as_string = as_string
```

Tbh I like the current solution a bit better but that would be an alternative.

astroid/nodes/node_ng.py
Pierre-Sassoulas: Done.

astroid/nodes/as_string.py
Pierre-Sassoulas: I removed that so the only thing to merge now are the refactor that were done around it.

cdce8p (approved):
LGTM!

PR #1062, opened 2021-06-19
Misc typing improvements

cdce8p:
## Description
Fix issues found while working on adding basic type annotations to nodes.

## Type of Changes

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -29,6 +29,7 @@
 order to get a single Astroid representation
 """
 
+import sys
 from typing import (
     TYPE_CHECKING,
     Callable,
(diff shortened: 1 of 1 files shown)


astroid/rebuilder.py
```diff
@@ -675,13 +671,27 @@ def visit_annassign(self, node: "ast.AnnAssign", parent: NodeNG) -> nodes.AnnAss
...
     ) -> nodes.AssignName:
+        ...
+
+    @overload
+    def visit_assignname(
+        self, node: "ast.AST", parent: NodeNG, node_name: None
+    ) -> None:
+        ...
+
+    def visit_assignname(
```
Pierre-Sassoulas, line 686: For me a visit_x function is accessed with:
```python
    def accept(self, visitor):
        """Visit this node using the given visitor."""
        func = getattr(visitor, "visit_" + self.__class__.__name__.lower())
        return func(self)
```
And if it's named ``visit_assignname`` the node is going to be an ``AssignName`` not a NodeNG ? What am I missing ?

Pierre-Sassoulas (commented):
I think there's a problem with the overload functions, when I'm running mypy on it I get a lot of those kind of errors:
```
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 66's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 67's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 68's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 69's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 70's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 71's parameter type(s) are the same or broader
astroid/rebuilder.py:471: error: Overloaded function signature 76 will never be matched: signature 72's parameter type(s) are the same or broader
```

astroid/rebuilder.py
cdce8p: Actually `accept` isn't ever used for the `rebuilder`, as I can see it, it's just for `as_string`. The rebuilder uses `def visit` instead. However, `visit_assignname` is also never called by `visit` as there is no corresponding `ast` node. It is more a helper function, probably with a not ideal name. If already added a note to the docstring to explain that it's never called by `visit`. (Same is true for `visit_decorators` and `visit_module`)

cdce8p:
> I think there's a problem with the overload functions, when I'm running mypy on it I get a lot of those kind of errors

I guess you used Python < 3.10 for the mypy run? `ast.Match` and the others are only defined for
`sys.version_info >= (3, 10)`: [typeshed -> stdlib/_ast.pyi](https://github.com/python/typeshed/blob/9ec2f8712480c57353cea097a65d75a2c4ec1846/stdlib/_ast.pyi#L380-L383). In theory something like this should be possible, but `mypy` doesn't seem to like it. Probably a bug, as it works fine with `pyright`. Have to investigate it.
```py
@overload
def visit(node: ast.Assert) -> None:
    ...

@overload
def visit(node: ast.Add) -> None:
    ...

if sys.version_info >= (3, 10):
    @overload
    def visit(node: ast.Match) -> int:
        ...

def visit(node: ast.AST) -> None | int:
    pass
```

astroid/rebuilder.py
Pierre-Sassoulas: But ``Rebuilder`` look like a ``Visitor`` though, maybe we could refactor it so it's used like one ?

cdce8p:
> I think there's a problem with the overload functions, when I'm running mypy on it I get a lot of those kind of errors:

@Pierre-Sassoulas Does it work for you to run `mypy` with `3.10` for now? If not, I would need to remove / comment the overloads for `ast.Match`, ... together with removing the annotation from `visit_match`, ...

astroid/rebuilder.py
Pierre-Sassoulas: OK then.

Pierre-Sassoulas:
I'm on 3.8 right now, I'm getting:
```
astroid/rebuilder.py:459: error: Name 'ast.MatchAs' is not defined
astroid/rebuilder.py:459: error: Overloaded function signature 73 will never be matched: signature 65's parameter type(s) are the same or broader
```
Is it supposed to work only for python 3.10 ?

Pierre-Sassoulas:
This MR still remove 9 errors compared to the main branch so it could be merged as is imo.

cdce8p:
I pushed a new commit with the changes that are required so that most mypy errors for `astroid/rebuilder.py` are now fixed. Even for Python < 3.10

Pierre-Sassoulas (approved):
From 1000+ errors to ``Found 117 errors in 22 files`` along with #1068, it's not looking too bad now ! 🎉

cdce8p:
I don't really like the duplication for the overloads, but for now that's the only way `mypy` will accept. Let's keep it then.

Pierre-Sassoulas:
I forgot something during the review I opened a follow-up here: https://github.com/PyCQA/astroid/pull/1069.

PR #1064, opened 2021-06-20
Fix some invalid name for pep8

Pierre-Sassoulas:
## Description

Function names should be in snake case. Add a TODO on variable we should remove in 3.0.

## Type of Changes

### DIFF
--- a/astroid/const.py
+++ b/astroid/const.py
@@ -13,6 +13,7 @@ class Context(enum.Enum):
     Del = 3
 
 
-Load = Context.Load
-Store = Context.Store
-Del = Context.Del
+# TODO Remove in 3.0 in favor of Context
+Load = Context.Load  # pylint: disable=invalid-name
+Store = Context.Store  # pylint: disable=invalid-name
+Del = Context.Del  # pylint: disable=invalid-name
--- a/pylintrc
+++ b/pylintrc
@@ -123,7 +123,7 @@ enable=useless-suppression
 bad-functions=
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_
+good-names=i,j,k,e,f,m,cm,Run,_,n,op,it
 
 # Bad variable names which should always be refused, separated by a comma
 bad-names=foo,bar,baz,toto,tutu,tata
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -462,7 +462,7 @@ class DefaultDictTest(unittest.TestCase):
 
 
 class ModuleExtenderTest(unittest.TestCase):
-    def testExtensionModules(self):
+    def test_extension_modules(self):
         transformer = MANAGER._transform
         for extender, _ in transformer.transforms[nodes.Module]:
             n = nodes.Module("__main__", None)
@@ -1612,7 +1612,7 @@ class TypingBrain(unittest.TestCase):
         self.assertIsInstance(inferred, astroid.Instance)
 
     @test_utils.require_version("3.8")
-    def test_typedDict(self):
+    def test_typed_dict(self):
         node = builder.extract_node(
             """
         from typing import TypedDict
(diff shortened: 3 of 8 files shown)


pylintrc
```diff
@@ -123,7 +123,7 @@ enable=useless-suppression
 bad-functions=
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_
+good-names=i,j,k,e,f,m,cm,Run,_,n,op,it
```
cdce8p, line 126: Might be just my personal preference, but I normally use `ex` over just `e`.

astroid/const.py
```diff
@@ -13,6 +13,7 @@ class Context(enum.Enum):
     Del = 3
 
 
-Load = Context.Load
-Store = Context.Store
-Del = Context.Del
+# TODO Remove in 3.0 in favor of Context
```
cdce8p, line 16: Maybe we should add a Changelog entry for it?

tests/unittest_nodes.py
```diff
@@ -570,7 +570,7 @@ def test_copy(self):
 
 
 class NameNodeTest(unittest.TestCase):
-    def test_assign_to_True(self):
+    def test_assign_to_rrue(self):
```
cdce8p, line 573: `true`

tests/unittest_brain.py
```diff
@@ -1612,7 +1612,7 @@ def test_typing_namedtuple_dont_crash_on_no_fields(self):
         self.assertIsInstance(inferred, astroid.Instance)
 
     @test_utils.require_version("3.8")
-    def test_typedDict(self):
+    def test_typed_dict(self):
```
cdce8p, line 1615: `typedDict` would be more fitting

tests/unittest_inference.py
```diff
@@ -901,12 +901,12 @@ class D(C):
...
             "module",
         )
-        should_be_C = list(ast[0].infer())
-        should_be_D = list(ast[1].infer())
-        self.assertEqual(1, len(should_be_C))
-        self.assertEqual(1, len(should_be_D))
-        self.assertEqual("module.C", should_be_C[0].qname())
-        self.assertEqual("module.D", should_be_D[0].qname())
+        should_be_c = list(ast[0].infer())
+        should_be_d = list(ast[1].infer())
```
cdce8p, line 905: Was there really a reason to change `C` and `D` to lowercase?

cdce8p (commented):
I don't think these changes are an improvement. There is absolutely no need to follow PEP 8 naming for tests, especially since the old style was often times more expressive of the actual test case.

tests/unittest_inference.py
Pierre-Sassoulas: pep8 says variable name should be in snake_case.

pylintrc
Pierre-Sassoulas: Let's add both then :)

astroid/const.py
Pierre-Sassoulas: I think we can add it in the breaking changes for 3.0 directly. Except if you know how to raise a deprecation warning if a constant is imported ? I don't think it's possible

tests/unittest_brain.py
Pierre-Sassoulas: I think pep8 says that method name should be in snake case.

cdce8p:
The point I was trying to make was that I don't think PEP8 is a useful guidance for tests. It's completely ok IMO to keep what makes the most sense, even if it would violate, for example, `snake_case` naming.

Pierre-Sassoulas:
I agree that tests should be coded differently as they need to be more understandable and less generic or maintainable than the code itself. But regarding naming, what make the most sense is different for everyone, if we make exception we'll have to think each time we type a variable if there is an exception or if this part of the code is following pep8 (or worse discuss capitalization during review which bring no value). This negate the advantages of following a convention, so I don't think the naming style should be relaxed for tests.

PR #1065, opened 2021-06-20
Remove default init values Arguments + fix raw_building

cdce8p:
## Description
`args`, `default`, `kwonlyargs`, `kw_defaults`, and `annotations` are required parameter for `Arguments.postinit`. With that it's safe to remove the default initialization values for them as long as `postinit` is always called.

## Type of Changes

### DIFF
--- a/tests/unittest_raw_building.py
+++ b/tests/unittest_raw_building.py
@@ -66,6 +66,10 @@ class RawBuildingTC(unittest.TestCase):
         node = build_function(name="MyFunction", posonlyargs=["a", "b"])
         self.assertEqual(2, len(node.args.posonlyargs))
 
+    def test_build_function_kwonlyargs(self):
+        node = build_function(name="MyFunction", kwonlyargs=["a", "b"])
+        assert len(node.args.kwonlyargs) == 2
+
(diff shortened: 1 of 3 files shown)


tests/unittest_raw_building.py
```diff
@@ -66,6 +66,10 @@ def test_build_function_posonlyargs(self):
         node = build_function(name="MyFunction", posonlyargs=["a", "b"])
         self.assertEqual(2, len(node.args.posonlyargs))
 
+    def test_build_function_kwonlyargs(self):
+        node = build_function(name="MyFunction", kwonlyargs=["a", "b"])
+        assert len(node.args.kwonlyargs) == 2
```
Pierre-Sassoulas, line 71:
```suggestion
        assert len(node.args.kwonlyargs) == 2
        assert "a" in node.args.kwonlyargs
        assert "b" in node.args.kwonlyargs
```

Pierre-Sassoulas (approved):

tests/unittest_raw_building.py
cdce8p: 👍🏻 I've added it to the test.

PR #1067, opened 2021-06-21
Add ``NodeNg`` to ``astroid.nodes`` as it can be used for typing

Pierre-Sassoulas:
## Description

``NodeNg`` can be used for typing in library using astroid as a dependency.

## Type of Changes

## Related Issue

#1057


astroid/nodes.py
```diff
@@ -24,6 +24,7 @@
 # Nodes not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
 
 from astroid.node_classes import (  # pylint: disable=redefined-builtin (Ellipsis)
+    CONST_CLS,
```
cdce8p, line 27: I believe `CONST_CLS` is just used internally for astroid. It therefore can be accessed directly.

astroid/nodes.py
```diff
@@ -189,6 +194,8 @@
     With,
     Yield,
     YieldFrom,
-    FormattedValue,
-    JoinedStr,
 )
+
+# Can't create a proper __all__ with string because of a cyclic import for ClassDef
+__all__ = ["CONST_CLS"]
+__all__ += [c.__name__ for c in ALL_NODE_CLASSES]
```
cdce8p, line 201: I'm still not convinced this changes anything

astroid/nodes.py
```diff
@@ -78,6 +79,7 @@
     MatchValue,
     Name,
     NamedExpr,
+    NodeNG,
```
cdce8p, line 82: A Changelog entry explaining that `NodeNG` can now be accessed through `from astroid import NodeNG` would be helpful. It could also mention that the previous import is now deprecated.

cdce8p (changes requested):

astroid/nodes.py
Pierre-Sassoulas: This clarify the API and also let linter see that the import is actually used (autoflake8 remove unused import)

astroid/nodes.py
Pierre-Sassoulas: You're right, I removed it.

astroid/nodes.py
Pierre-Sassoulas: I added a message in changelog. The previous import will be deprecated at the same time than all the other in #1057

astroid/nodes.py
Pierre-Sassoulas: Also it's necessary if we want to be able to keep the old API and deprecate it in #1057 :
 ```
 +# It's impossible to import from astroid.nodes with a wildcard, because
 +# there is a cyclic import that prevent creating an __all__ in astroid/nodes
 +# and we need astroid/scoped_nodes and astroid/node_classes to work. So
 +# importing with a wildcard would clash with astroid/nodes/scoped_nodes
 +# and astroid/nodes/node_classes.
```

cdce8p (approved):
Ok. Let's do it then.

PR #1069, opened 2021-06-21
Follow up to merge request 1062 and 1068

Pierre-Sassoulas:
## Description

Just two things forgotten during review.

## Type of Changes

## Related Issue

#1062

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -5,5 +5,5 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
-types-typed-ast
+types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.2
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -5,3 +5,5 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
+types-typed-ast
+types-pkg_resources==0.1.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,7 +71,14 @@ repos:
         types: [python]
         args: []
         require_serial: true
-        additional_dependencies: ["types-pkg_resources==0.1.2"]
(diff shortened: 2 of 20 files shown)


cdce8p (changes requested):
It should be `sys.version_info >= (3, 10)`. That's the preferred version for type checkers.

Pierre-Sassoulas:
I think this constant is misleading as in pylint it's named PY310_PLUS instead but it's the same one: https://github.com/PyCQA/astroid/blob/master/astroid/const.py#L7 Do you want to change them all to PY3X_PLUS ?

cdce8p:
> Do you want to change them all to PY3X_PLUS ?

IMO we should remove them completely.
https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

cdce8p:
Regarding the 930+ errors: I think that's because of `PY310`

Pierre-Sassoulas:
> IMO we should remove them completely.

It's not used only for mypy though, they are used in a lot of places.

cdce8p:
> It's not used only for mypy though, they are used in a lot of places.

Let's change the existing ones to `PYXX_PLUS` then and keep `sys.version_info` where it's required for type checking.

Pierre-Sassoulas:
I think that everything is ready after that one, but I'm going to release 2.6.0 tomorrow (and after checking in pylint first), I've coded too much today 😄

cdce8p:
Can you add the additional typing dependencies to `requirements_test_brain.txt` / `requirement_test.txt`?

requirements_test.txt
```diff
@@ -5,3 +5,5 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
+types-typed-ast
```
cdce8p, line 8:
```suggestion
types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
```
Must sure if it's necessary as this is just a types package. Shouldn't harm though

cdce8p (approved):
Looks good overall. Thanks for doing it 🚀

PR #1070, opened 2021-06-21
Create a builtins constant in ``astroid.const``

Pierre-Sassoulas:
## Description

Cleanup of the BUILTINS constant used and redefined in a lot of places.

## Type of Changes

### DIFF
--- a/astroid/const.py
+++ b/astroid/const.py
@@ -1,3 +1,4 @@
+import builtins
 import enum
 import sys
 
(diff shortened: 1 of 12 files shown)


astroid/const.py
```diff
@@ -17,3 +18,5 @@ class Context(enum.Enum):
 Load = Context.Load  # pylint: disable=invalid-name
 Store = Context.Store  # pylint: disable=invalid-name
 Del = Context.Del  # pylint: disable=invalid-name
+
+BUILTINS = builtins.__name__  # Could be just 'builtins' ?
```
cdce8p, line 22: In theory it shouldn't matter, but I kinda understand why `builtins.__name__` was chosen. That way you know at first glance it's the name of the builtins module and don't need an additional comment explaining it.

cdce8p (approved):
Looks good overall. As explained earlier, I would just wait a bit before merging.

PR #1076, opened 2021-06-26
Infer the type of the result of calling typing.cast()

timmartin:
## Description

Added an inference rule in `brain_typing.py` to infer the return type of `typing.cast`. So that:

```py
from typing import cast

class A:
    pass

b = cast(A, x)  # Can infer that b is an instance of A
```

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,7 +11,8 @@ What's New in astroid 2.6.6?
 ============================
 Release date: TBA
 
-* Added support to infer return type of typing.cast()
+* Added support to infer return type of ``typing.cast()``
+
 
 What's New in astroid 2.6.5?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,7 +11,7 @@ What's New in astroid 2.6.6?
 ============================
 Release date: TBA
 
-
+* Added support to infer return type of typing.cast()
 
 What's New in astroid 2.6.5?
 ============================
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -28,6 +28,7 @@ from astroid.node_classes import (
     Call,
     Const,
     Name,
+    NodeNG,
     Subscript,
 )
 from astroid.scoped_nodes import ClassDef, FunctionDef
@@ -356,6 +357,29 @@ def infer_tuple_alias(
     return iter([class_def])
 
 
+def _looks_like_typing_cast(node: Call) -> bool:
+    return isinstance(node, Call) and (
+        isinstance(node.func, Name)
+        and node.func.name == "cast"
+        or isinstance(node.func, Attribute)
+        and node.func.attrname == "cast"
+    )
+
+
+def infer_typing_cast(
+    node: Call, ctx: context.InferenceContext = None
+) -> typing.Iterator[NodeNG]:
+    """Infer call to cast() returning same type as casted-from var"""
+    try:
+        func = next(node.func.infer(context=ctx))
+    except InferenceError as exc:
+        raise UseInferenceDefault from exc
+    if func.qname() != "typing.cast" or len(node.args) != 2:
+        raise UseInferenceDefault
+
+    return node.args[1].infer(context=ctx)
+
+
 AstroidManager().register_transform(
     Call,
     inference_tip(infer_typing_typevar_or_newtype),
@@ -363,6 +387,9 @@ AstroidManager().register_transform(
 )
 AstroidManager().register_transform(
     Subscript, inference_tip(infer_typing_attr), _looks_like_typing_subscript
+)
+AstroidManager().register_transform(
+    Call, inference_tip(infer_typing_cast), _looks_like_typing_cast
 )
 
 if PY39_PLUS:
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -370,11 +370,18 @@ def infer_typing_cast(
     node: Call, ctx: context.InferenceContext = None
 ) -> typing.Iterator[NodeNG]:
     """Infer call to cast() returning same type as casted-from var"""
+    if not isinstance(node.func, (Name, Attribute)):
+        raise UseInferenceDefault
+
     try:
         func = next(node.func.infer(context=ctx))
     except InferenceError as exc:
         raise UseInferenceDefault from exc
-    if func.qname() != "typing.cast" or len(node.args) != 2:
+    if (
+        not isinstance(func, FunctionDef)
+        or func.qname() != "typing.cast"
+        or len(node.args) != 2
+    ):
         raise UseInferenceDefault
 
     return node.args[1].infer(context=ctx)
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -1810,6 +1810,38 @@ class TypingBrain(unittest.TestCase):
             self.assertIsInstance(inferred, nodes.ClassDef)
             self.assertIsInstance(inferred.getattr("__iter__")[0], nodes.FunctionDef)
 
+    def test_typing_cast(self):
+        node = builder.extract_node(
+            """
+        from typing import cast
+        class A:
+            pass
+
+        b = list()
+        a = cast(A, b)
+        a
+        """
+        )
+        inferred = next(node.infer())
+        assert isinstance(inferred, bases.Instance)
+        assert inferred.name == "list"
+
+    def test_typing_cast_attribute(self):
+        node = builder.extract_node(
+            """
+        import typing
+        class A:
+            pass
+
+        b = list()
+        a = typing.cast(A, b)
+        a
+        """
+        )
+        inferred = next(node.infer())
+        assert isinstance(inferred, bases.Instance)
+        assert inferred.name == "list"
+
 
 class ReBrainTest(unittest.TestCase):
     def test_regex_flags(self):
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -1817,14 +1817,14 @@ class TypingBrain(unittest.TestCase):
         class A:
             pass
 
-        b = list()
+        b = 42
         a = cast(A, b)
         a
         """
         )
         inferred = next(node.infer())
-        assert isinstance(inferred, bases.Instance)
-        assert inferred.name == "list"
+        assert isinstance(inferred, nodes.Const)
+        assert inferred.value == 42
 
     def test_typing_cast_attribute(self):
         node = builder.extract_node(
@@ -1833,14 +1833,14 @@ class TypingBrain(unittest.TestCase):
         class A:
             pass
 
-        b = list()
+        b = 42
         a = typing.cast(A, b)
         a
         """
         )
         inferred = next(node.infer())
-        assert isinstance(inferred, bases.Instance)
-        assert inferred.name == "list"
+        assert isinstance(inferred, nodes.Const)
+        assert inferred.value == 42
 
 
 class ReBrainTest(unittest.TestCase):


Pierre-Sassoulas (approved):
Thank you for the pull request ! It's pretty straightforward :)

ChangeLog
```diff
@@ -6,7 +6,7 @@ What's New in astroid 2.7.0?
 ============================
 Release Date: TBA
 
-
+* Added support to infer return type of typing.cast()
```
cdce8p, line 9: @Pierre-Sassoulas Do you want to include it in `2.6.1`? In that case it would need to be moved to the section below.

astroid/brain/brain_typing.py
```diff
@@ -344,6 +345,29 @@ def infer_tuple_alias(
     return iter([class_def])
 
 
+def _looks_like_typing_cast(node: Call) -> bool:
+    return (
+        isinstance(node, Call)
+        and isinstance(node.func, Name)
+        and node.func.name == "cast"
```
cdce8p, line 352:
```suggestion
        and (
            isinstance(node.func, Name)
            and node.func.name == "cast"
            or isinstance(node.func, Attribute)
            and node.func.attrname == "cast"
        )
```
`cast` can also be called with `typing.cast` in which case it's an `Attribute`.

astroid/brain/brain_typing.py
```diff
@@ -344,6 +345,29 @@ def infer_tuple_alias(
...
+
+def infer_typing_cast(
+    node: Call, ctx: context.InferenceContext = None
+) -> typing.Iterator[Instance]:
+    """Infer call to cast() returning an instance of the cast-to type"""
+    try:
+        func = next(node.func.infer(context=ctx))
+    except InferenceError as exc:
+        raise UseInferenceDefault from exc
+    if func.qname() != "typing.cast":
```
cdce8p, line 364:
```suggestion
    if func.qname() != "typing.cast" and len(node.args) == 2:
```
Just to prevent errors if the cast statement syntax isn't correct.

astroid/brain/brain_typing.py
```diff
@@ -344,6 +345,29 @@ def infer_tuple_alias(
...
+        raise UseInferenceDefault
+
+    type_node = next(node.args[0].infer(context=ctx))
+    return iter([Instance(type_node)])
```
cdce8p, line 368:
```suggestion
    type_node = next(node.args[1].infer(context=ctx))
    return iter([type_node])
```

tests/unittest_brain.py
```diff
@@ -1801,6 +1801,22 @@ def test_typing_object_builtin_subscriptable(self):
...
 
+    def test_typing_cast(self):
+        node = builder.extract_node(
+            """
+        from typing import cast
+        class A:
+            pass
+
+        b = deserialize()
+        a = cast(A, b)
```
cdce8p, line 1812: Could you add a second case for the use of `typing.cast`?

cdce8p (changes requested):
I'm not sure we should infer `b` as instance of `A` for `b = cast(A, x)`. `cast` is used just for type hints and we probably shouldn't mix astroid inference and type hints. IMO the logical step would be to infer `b` as `x` since that's what `cast` does. @Pierre-Sassoulas What do you think?

astroid/brain/brain_typing.py
timmartin: Good point, but I think this should be `func.qname() != "typing.cast" or len(node.args) != 2`, unless I've misunderstood?

cdce8p: My mistake. Yours is correct.

timmartin:
Thanks for the helpful review comments. In terms of what the inferred type of the result should be, maybe I'm fixing this in the wrong place. The original motivation was [this report](https://github.com/PyCQA/pylint/issues/4534) in Pylint; if we infer that `typing.cast(A, x)` has the same type as `x` then we won't fix this issue.

The way I look at it, `typing.cast` exists to provide information to make static checking better, so in an ideal world a linter could use the knowledge of the casted-to type to give better feedback. But does this mean that Pylint is the right place to apply this logic and not Astroid?

cdce8p:
@timmartin I'm a bit hesitant with this one. Who makes sure `A` is truly the correct type? In an ideal world, `x` would be inferred correctly and we can just say that the result of `cast` equals `x`. Although I haven't looked at the issue in detail, it seems that there are two distinct issues:
1. Currently, the result of `cast` is always `Uninferrable`
2. `wave.open` might not be inferred correctly. This should be fixed separately by adding a `brain` module or writing a custom plugin.

One last thing, it's should be possible to use abstract classes for `A` just so the type checker is satisfied. However, that might prevent `pylint` from analyzing it further although it sometimes could.

timmartin:
The developer who adds the cast is responsible for checking that `A` is the right type. If they put the wrong annotation on, they'll get the wrong results from the type checker / linter. Garbage in, garbage out.

To me, the issue is that `typing.cast` is specifically designed to be used in the case where the correct type isn't / can't reasonably be inferred. It has no runtime semantics, it's purely a hint to the type checker. If `x` was being correctly inferred, there'd be no point in applying a cast.

I think you're right in theory that any problem that can be solved by a cast can be solved by sufficiently good type declarations / better inference. I think casting exists for the cases where that doesn't yet exist.

cdce8p:
The issue I have with it is that `cast` is designed for type checking and not for inference. You could for example say that the type should be an abstract or protocol class. That will work perfectly fine for `mypy`, but won't offer any additional insides for us.

It might be an idea to add something like `astroid.cast` / `pylint.cast` that can be used explicitly for inference tips.

Pierre-Sassoulas:
> It might be an idea to add something like astroid.cast / pylint.cast that can be used explicitly for inference tips.

It would add work for pylint users, what is the meaningful difference between typing for inference and typing for type/mypy ? I think ``cast`` does the work and have the advantages of being official and widespread. Also as @timmartin said, garbage in, garbage out, if you mislead mypy or pylint you get the error you deserve 😄

cdce8p:
@Pierre-Sassoulas The point I'm trying to make here is that with inference we can already do better then what type checkers can. So using `cast` doesn't add anything for us. In contrast it might even prevent us from offering better suggestions.

Pierre-Sassoulas:
> using cast doesn't add anything for us. In contrast it might even prevent us from offering better suggestions.

Well it's true that it might prevent us from offering better suggestions but only when the user made an explicit typing mistake. I think it add something for us if the inference fail. So maybe we should rely on ``cast`` if the inference failed and provide an option to be able to check ``cast`` ?

timmartin:
AIUI type checkers like mypy already do type inference, they can't rely on every variable having its type statically declared. Is there a class of inference that Astroid does that wouldn't be available in a sufficiently smart type checker?

I can certainly see how there will be cases in the real world where the cast may be needed for mypy and not for Pylint, say. That would make it annoying if adding the needed cast helped mypy but broke Pylint. On the other hand, there will be cases where Astroid isn't (yet) making the inference and it would be useful to have the option of a cast.

It should be easy enough to make it infer the type of `cast(A, x)` as the type of `x` if it's inferable, otherwise an instance of `A`. I think that would avoid making inference worse in any case, although it still might miss the opportunity to do (say) a downcast when the variable is known to be an instance of a parent class. Is that a good solution here?

TBH I don't have a strong opinion here, I just picked this one up because I thought it would be an easy way to get used to the codebase. If it's controversial, I'm happy to leave this to someone else.

Pierre-Sassoulas:
@hippo91 do you have an opinion on this ?

hippo91:
Well that is not an easy question. Thanks all for giving your point of view. IMHO the best option here is to infer the result of the cast as 'A' only if the inference of 'x' failed. Otherwise I am afraid we lose precious informations as @cdce8p mentioned it. But I am afraid there is not a real clear solution for this.

cdce8p:
Thanks for your input @hippo91!
I agree with your point and like the suggestion. So +1 from me

ChangeLog
cdce8p, line 9: Can you move this to the `2.6.3` section? You probably need to fetch the latest changes first.

astroid/brain/brain_typing.py
```diff
@@ -344,6 +346,34 @@ def infer_tuple_alias(
...
+        isinstance(node.func, Name)
+        and node.func.name == "cast"
+        or isinstance(node.func, Attribute)
+        and node.func.attrname == "cast"
+    )
+
+
+def infer_typing_cast(
+    node: Call, ctx: context.InferenceContext = None
+) -> typing.Iterator[Instance]:
```
cdce8p, line 360:
```suggestion
) -> typing.Union[typing.Iterator[NodeNG], typing.Iterator[Instance]]:
```
Not completely sure about this as `infer` isn't yet typed.

astroid/brain/brain_typing.py
```diff
@@ -344,6 +346,34 @@ def infer_tuple_alias(
...
+    """Infer call to cast() returning an instance of the cast-to type"""
+    try:
+        func = next(node.func.infer(context=ctx))
+    except InferenceError as exc:
+        raise UseInferenceDefault from exc
+    if func.qname() != "typing.cast" or len(node.args) != 2:
+        raise UseInferenceDefault
+
+    peek_value_node, value_node = itertools.tee(node.args[1].infer(context=ctx))
```
cdce8p, line 369: The use of `itertools.tee` isn't really necessary. We are only really interested in the first return value.


This should be enough
```py
    inferred = next(node.args[1].infer(context=ctx)
```

astroid/brain/brain_typing.py
```diff
@@ -344,6 +346,34 @@ def infer_tuple_alias(
...
+    if next(peek_value_node) is Uninferable:
+        type_node = next(node.args[0].infer(context=ctx))
+        return iter([Instance(type_node)])
+    else:
+        return value_node
```
cdce8p, line 374: `else` is redundant and can be removed since the `if` clause ends with `return`. That's the `pylint` error.

```py
    return iter([inferred])
```

cdce8p (changes requested):
Looks good overall! Just a few last comments

astroid/brain/brain_typing.py
timmartin: My first attempt was roughly what you describe here, but it didn't work, because if the first element of the iterable isn't uninferable then we want to return the whole iterator (so all values inferred for `x` can be inferred as the return type of `cast()`. AIUI by having called `next()` I've already consumed the first element of the iterator at this point, so I'm not able to return the full iterator.

Am I understanding this right?

cdce8p: Though about this again. Yeah, you are right. I didn't account for the case that more than one inference result would be returned (I believe in most situations it's only one).

To still avoid the overhead of `itertools.tee`, you could use one of the following options:
1. Store results of `infer` in list and check against first value / return `iter(inferred_lst)`.
2. Just call `node.args[1].infer(context=ctx)` two times. Inference results are already cached with `context`, so I suspect there isn't much / if any performance hit.
https://github.com/PyCQA/astroid/blob/c9c498348174b38ce35bfe001353c8ebea262802/astroid/node_classes.py#L357-L360

astroid/brain/brain_typing.py
```diff
@@ -351,6 +353,34 @@ def infer_tuple_alias(
...
+        isinstance(node.func, Name)
+        and node.func.name == "cast"
+        or isinstance(node.func, Attribute)
+        and node.func.attrname == "cast"
+    )
+
+
+def infer_typing_cast(
+    node: Call, ctx: context.InferenceContext = None
+) -> typing.Union[typing.Iterator[NodeNG], typing.Iterator[Instance]]:
```
cdce8p, line 367: To fix the tests, you need to add the import for `NodeNG` (from `astroid.node_classes`)

astroid/brain/brain_typing.py
timmartin: I believe the `tee()` solution has less overhead than either of your suggestions, not more. It guarantees that the original generator will be executed at most once, and has only trivial temporary storage in this case (where only one element is being stored in a temporary FIFO, which will presumably be consumed the moment we return our result to the caller).

Never the less, if there's no significant overhead to running the inference twice then that has the advantage of making the intent of the code much clearer, so I've done that. Thanks for the suggestion.

timmartin: I don't fully understand what's going on here. If I call the `node.args[1].infer(context=ctx)` twice then it doesn't work the second time. Trying to consume it gives:

    astroid.exceptions.InferenceError: StopIteration raised without any error information.

If I don't pass in the context the second time then it works as expected, so presumably there's some state being stored in the context that is making the second inference depend on the first. I don't really understand what's stored in the context and if this is expected behaviour, though.

cdce8p: Found the issue: The inference result is only cached in the context if the first iterator is run till completion, i.e. `tuple(node.args[1].infer(context=ctx))[0] is Uninferable`. Without that the context only knows that it has seen the name for `args[1]` already and thus thinks it can safely abort by raising a `StopIteration`.

You made a good point regarding `itertools.tee()`, so let's use it then. Thank you for trying it though.

cdce8p:
@Pierre-Sassoulas I would like to include it in `2.6.3` if that's fine with you. It will improve inference and thus help solve some issue.

cdce8p:
@timmartin I tested the change against some code and noticed two errors:

1. The type for `cast` can be provided as string in which case it will raise a `TypeError`
```py
from typing import cast

my_dict = func_returns_dict()  # not inferable
my_dict = cast("dict[str, int]", my_dict)
dict2 = my_dict.copy()
```

2. Especially if `TypeVars` are used, the change will result in many false-positive `no-member` errors.
```py
from typing import cast, TypeVar

class SomeClass:
    def some_method():
        pass

ClassT = TypeVar("ClassT", bound=SomeClass)

var = func_returns_SomeClass()  # not inferable
var = cast(ClassT, var)
var.some_method()  # 'no-member' error
```

astroid/brain/brain_typing.py
timmartin: Thanks for the clear explanation of what's happening. Saved me a few hours' debugging!

timmartin:
@cdce8p Good catch. Is it reasonable to report uninferable in both cases? I can see how both cases could in principle be handled, but I don't know what infrastructure we already have for parsing type expression strings and working with typevars.

cdce8p:
> @cdce8p Good catch. Is it reasonable to report uninferable in both cases? I can see how both cases could in principle be handled, but I don't know what infrastructure we already have for parsing type expression strings and working with typevars.

@timmartin Either that or raising `UseInferenceDefault` (which should ultimately also result in `Uninferable`).

timmartin:
I think my latest change deals with both cases. I'm not totally sure in the typevar case; in my testing it appears that the typevar just gets inferred to `Uninferable`, which we can handle easily. I don't know if this will suffice in general though, if we improve typevar inference presumably we could regress this functionality.

ChangeLog
```diff
@@ -21,6 +20,8 @@ Release date: TBA
 
   Closes PyCQA/pylint#4671
 
+* Added support to infer return type of typing.cast()
```
cdce8p, line 23:
```suggestion
* Added support to infer return type of ``typing.cast()``
```
The ChangeLog ist `rst` formatted.

tests/unittest_brain.py
```diff
@@ -1801,6 +1801,90 @@ def test_typing_object_builtin_subscriptable(self):
...
+        """
+        )
+        inferred = next(node.infer())
+        self.assertIs(util.Uninferable, inferred)
+
+    def test_typing_cast_typevar(self):
+        """Casting to a typevar is not currently inferred"""
+        node = builder.extract_node(
+            """
+        from typing import cast
```
cdce8p, line 1872:
```suggestion
        from typing import cast, TypeVar
```

astroid/brain/brain_typing.py
```diff
@@ -351,6 +354,43 @@ def infer_tuple_alias(
...
+        raise UseInferenceDefault from exc
+    if func.qname() != "typing.cast" or len(node.args) != 2:
+        raise UseInferenceDefault
+
+    peek_value_node, value_node = itertools.tee(node.args[1].infer(context=ctx))
+    if next(peek_value_node) is Uninferable:
+        type_node = next(node.args[0].infer(context=ctx))
+
+        if type_node is Uninferable:
+            raise UseInferenceDefault
```
cdce8p, line 382: This is good to have, but it doesn't solve the `TypeVar` issue.

cdce8p (changes requested):
Astroid does have some (albeit limited) inference for `TypeVar` (see below). The issue for us here is that it basically returns an empty class with the bare minimum and doesn't account for and constraints or bounds on the original TypeVar.

Just from a quick view, I fear we won't be able to easily find out if a type is a `TypeVar` or not. That unfortunately will bring us back to the point I made at the beginning: It's probably best to only infer `cast` as the variable itself even if it's `Uninferable`.

If you have another idea, I would be open to testing that out.

https://github.com/PyCQA/astroid/blob/a7e55b4caefd532d404ac9429f490edcc5ad7d5a/astroid/brain/brain_typing.py#L114-L128

timmartin:
> Just from a quick view, I fear we won't be able to easily find out if a type is a `TypeVar` or not. That unfortunately will bring us back to the point I made at the beginning: It's probably best to only infer `cast` as the variable itself even if it's `Uninferable`.

That's a shame. I'm a little puzzled though, because I was assuming that astroid would either successfully determine that the argument was a TypeVar, or else it would be uninferable. Are you saying that TypeVars are handled in a way that makes it impossible to distinguish a class from a typevar?

cdce8p:
> That's a shame. I'm a little puzzled though, because I was assuming that astroid would either successfully determine that the argument was a TypeVar, or else it would be uninferable. Are you saying that TypeVars are handled in a way that makes it impossible to distinguish a class from a typevar?

At the moment that is the case. Keep in mind though that this behavior is completely fine since `astroid` doesn't use type hints for inference tips. We just need to make sure they have some basic properties. There are probably some more cases that will cause issues: `NewType`, `Protocol`, `Generic`.

timmartin:
If that's the case then it's probably best to do as you suggest and just infer from the second argument.

Will astroid differentiate between classes and typevars in the future? It seems like it would be necessary to catch some errors e.g. that typevars can't be constructed.

Looking at the typevar code, is there any way to distinguish based on metaclass? It looks like typevars are given a custom metaclass; if we could only infer `cast` for classes with a default metaclass that might be a small win. I fiddled around for a while but I couldn't come up with anything that didn't feel like a kludge, though.

cdce8p:
> If that's the case then it's probably best to do as you suggest and just infer from the second argument.

I would probably prefer that.

> Will astroid differentiate between classes and typevars in the future? It seems like it would be necessary to catch some errors e.g. that typevars can't be constructed.

Maybe, there are multiple issues though. First you need someone to add it, then to review it, and most important someone to maintain it. (1) might be possible, but the other two unlikely. As it currently is, I too struggle sometimes to understand a change completely. That's the cost of good, but really old legacy code. Regarding `typing` in particular, I'm happy that most major issues are solved now. It has taken us moths to do that at the start of the year. (Also a reason why I prefer the safe route here.)

> Looking at the typevar code, is there any way to distinguish based on metaclass? It looks like typevars are given a custom metaclass; if we could only infer `cast` for classes with a default metaclass that might be a small win. I fiddled around for a while but I couldn't come up with anything that didn't feel like a kludge, though.

You could try `next(node.args[0].infer()).metaclass()`, but again I would probably suggest to leave it be even it might technically be possible. I found the two issues after only a short test run. Don't know what other will come up with once it is released. Usually you have to decide if the feature you are adding is worth the additional bugs it might create. I'm just really skeptical about that.

I understand, that's probably not the answer you were hoping for but that's how I feel about it. I hope you can understand it.

timmartin:
I think my last change does what's required here. Should I change the tags on this PR to indicate that it's ready for re-review?

Pierre-Sassoulas:
@timmartin there's a little "reload" button in the reviewers so you can ask for a new review that way. :)

Pierre-Sassoulas:
Capture d’écran de 2021-07-29 18-03-41

astroid/brain/brain_typing.py
```diff
@@ -370,11 +370,18 @@ def infer_typing_cast(
     node: Call, ctx: context.InferenceContext = None
 ) -> typing.Iterator[NodeNG]:
     """Infer call to cast() returning same type as casted-from var"""
+    if not isinstance(node.func, (Name, Attribute)):
+        raise UseInferenceDefault
```
cdce8p, line 374: Technically this is covered by `_looks_like_typing_cast` but type checker don't know about that.

tests/unittest_brain.py
```diff
@@ -1817,14 +1817,14 @@ def test_typing_cast(self):
...
+        b = 42
         a = cast(A, b)
         a
         """
         )
         inferred = next(node.infer())
-        assert isinstance(inferred, bases.Instance)
-        assert inferred.name == "list"
+        assert isinstance(inferred, nodes.Const)
+        assert inferred.value == 42
```
cdce8p, line 1827: Usually it's a bit easier to test for a const value.

cdce8p (approved):
Thanks for the ping. I must have missed your last push.
As for the PR, I committed some last changes (I've explained the reasoning behind them below).

It looks good now.
Thank you @timmartin 🐬

Pierre-Sassoulas (approved):
👍 Thank you for being so persistent 👏 This will go in astroid 2.6.6 then pylint 2.10

PR #1081, opened 2021-06-30
Catch the StopIteration that result from PEP-479 or add default value in ``next``

### LINKED ISSUE #1080: Unhandled `StopIteration` during node inference
### Steps to reproduce

a.py
```
from b import getSourceFilePath

class A:
    languageCodes = ['en', 'fr', 'it', 'de', 'cn', 'nl']
    def readExcerpts(self):
        for languageCode in self.languageCodes:
            # does not need to be an existing path
            thisExcerpt = getSourceFilePath() / (languageCode + '.txt')
            with thisExcerpt.open(encoding='utf-8') as f:
                pass
```

b.py
```
import inspect
import pathlib

def getSourceFilePath() -> pathlib.Path:
    fpThis = pathlib.Path(inspect.getfile(getSourceFilePath)).resolve()
    return fpThis.parent.parent
```

### Current behavior

```
Exception on node <AssignName.f l.9 at 0x7fd0582e4f10> in file ...
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/astroid/protocols.py", line 463, in _infer_context_manager
    inferred = next(mgr.infer(context=context))
StopIteration
```

2.6.1

Appears to be a regression in cf6528cbc158097c4903f0cab68242ff14bb591b.

Pierre-Sassoulas:
StopIteration exceptions raised directly or indirectly in coroutines
and generators are transformed into RuntimeError exceptions for python
3.7+.

## Description

Handle all the StopIteration and not just the one that caused https://github.com/PyCQA/pylint/issues/4631

## Type of Changes

## Related Issue

Closes https://github.com/PyCQA/pylint/issues/4631
Closes #1080

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,12 @@ Release date: TBA
 
   Closes PyCQA/pylint#4633
 
+* Fix unhandled StopIteration during inference, following the implementation
+  of PEP479 in python 3.7+
+
+  Closes PyCQA/pylint#4631
+  Closes #1080
+
 What's New in astroid 2.6.1?
 ============================
 Release date: 2021-06-29
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -101,6 +101,8 @@ class CallSite:
                 except InferenceError:
                     values[name] = util.Uninferable
                     continue
+                except StopIteration:
+                    continue
 
                 if not isinstance(inferred, nodes.Dict):
                     # Not something we can work with.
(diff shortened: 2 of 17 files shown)


Pierre-Sassoulas:
I'm going to merge as it's preventing crash and all pylint tests still work, but I'm still interested in a review if you think there's a more elegant way to do it :)

cdce8p:
I'm not too familiar with the `infer` logic unfortunately. The only other way I can think of at the moment would be to make sure each `infer` method always returns an iterator and doesn't raise `StopIteration` on the first call of `next`. However that might be difficult to do, as we would need to understand and check every inference method.

hippo91:
@Pierre-Sassoulas i think you did the right thing to do.

PR #1083, opened 2021-06-30
Use explicit `__all__` for `astroid.nodes`

cdce8p:
## Description
A computed `__all__` for `astroid.nodes` doesn't work for `pylance` as it can't infer the names.

## Type of Changes

## Related Issue
https://github.com/PyCQA/astroid/pull/1067#discussion_r655412771


cdce8p:
@Pierre-Sassoulas Would you mind taking a look at this one? I know you mentioned an issue with `cyclic-imports` but from my testing so far, everything seems to be fine. Even with this change applied to #1057. Am I missing something?

Pierre-Sassoulas:
The cyclic import is not a real one, it's detected by pylint due to [this problem](https://github.com/PyCQA/pylint/issues/3285)

Pierre-Sassoulas (approved):

cdce8p:
> The cyclic import is not a real one, it's detected by pylint due to [this problem](https://github.com/PyCQA/pylint/issues/3285)

Strange, I'm not getting any error. Besides neither `scoped_nodes` nor `node_classes` uses `TYPE_CHECKING`

Pierre-Sassoulas:
There is a pragma to prevent isort from mixing the import order in ``astroid.__init__.py``, so that may be why it does not appear ? I'll check that when pylint 2.9.0 do not have crashes anymore :)

PR #1095, opened 2021-07-10
Update pylint config and remove useless-suppressions

cdce8p:
## Description

Coincidentally, I was also working on the `pylint` update when you opened and merged #1094

Besides removing some `useless-suppression` messages, I would like to set `report=no`. That is also the way it's in the pylint repo: [pylint -> pylintrc](https://github.com/PyCQA/pylint/blob/73089b5a02d11d4a2bb9513d3764aec40b873a4a/pylintrc#L85).

## Type of Changes

### DIFF
--- a/astroid/const.py
+++ b/astroid/const.py
@@ -15,8 +15,8 @@ class Context(enum.Enum):
 
 
 # TODO Remove in 3.0 in favor of Context
-Load = Context.Load  # pylint: disable=invalid-name
-Store = Context.Store  # pylint: disable=invalid-name
-Del = Context.Del  # pylint: disable=invalid-name
+Load = Context.Load
+Store = Context.Store
+Del = Context.Del
 
 BUILTINS = builtins.__name__  # Could be just 'builtins' ?
--- a/pylintrc
+++ b/pylintrc
@@ -50,7 +50,7 @@ output-format=text
 files-output=no
 
 # Tells whether to display a full report or only the messages
-reports=yes
+reports=no
 
 # Python expression which should return a note less than 10 (10 is the highest
 # note). You have access to the variables errors warning, statement which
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas:
Some useless-suppression for invalid-name on const are actually false negative in pylint (can't link the issue right now but there is one). Also useless suppression seems to go unreported when there is no other warning.

pylintrc
```diff
@@ -50,7 +50,7 @@ output-format=text
 files-output=no
 
 # Tells whether to display a full report or only the messages
-reports=yes
+reports=no
```
Pierre-Sassoulas, line 53: I dislike this report too but we should add more tests for it in pylint if we remove it from astroid (we already broke it and realized only because it was used in astroid)

astroid/const.py
```diff
@@ -15,8 +15,8 @@ class Context(enum.Enum):
 
 
 # TODO Remove in 3.0 in favor of Context
-Load = Context.Load  # pylint: disable=invalid-name
-Store = Context.Store  # pylint: disable=invalid-name
-Del = Context.Del  # pylint: disable=invalid-name
+Load = Context.Load
+Store = Context.Store
+Del = Context.Del
```
Pierre-Sassoulas, line 20: I was hoping we could fix the false negative in pylint first but I don't have much time right now. Can we separate those three removal in a single commit so we can revert it later ? (I'm on mobile so maybe it's already the case)

Pierre-Sassoulas (changes requested):

astroid/const.py
Pierre-Sassoulas: This is already the case, sorry I was on mobile.

pylintrc
Pierre-Sassoulas: On the other hand, I never look at the report anyway.

Pierre-Sassoulas (approved):

astroid/const.py
cdce8p: Don't think about it. I rebased and force pushed it after your comment 😄

PR #1097, opened 2021-07-12
Fix incorrect scope for functools partials

Alphadelta14:
## Steps

## Description

While working on https://github.com/PyCQA/pylint/pull/4357, I happened to notice an issue with partial functions where a given scope had both the original function and some remote partial usage returned by the same inference call.

I tracked the underlying issue to partials having their parents set the original function parent.

So in the [abbreviated] code from pylint/checkers/utils.py below, `func` (`partial(error_of_type, ...)`) gets Module as the parent (and since it has the same name, PartialFunction(name=error_of_type)), both of these values are locals in Module with the same name. This lead to the extremely strange case of `error_of_type(handler, exception)` inferred as both `error_of_type` and `partial(error_of_type, ...)`.
The proper parent for `partial(error_of_type, ...)` should just be the normal assignment to `func`.

```py
def error_of_type(handler: astroid.ExceptHandler, error_type) -> bool:
    pass

def _except_handlers_ignores_exception(
    handlers: astroid.ExceptHandler, exception
) -> bool:
    func = partial(error_of_type, error_type=(exception,))
    return any(func(handler) for handler in handlers)

def get_exception_handlers(
    node: astroid.node_classes.NodeNG, exception=Exception
) -> Optional[List[astroid.ExceptHandler]]:
    context = find_try_except_wrapper_node(node)
    if isinstance(context, astroid.TryExcept):
        return [
            handler for handler in context.handlers if error_of_type(handler, exception)
        ]
    return []
```

This also fixes the underlying issue seen in https://github.com/PyCQA/pylint/issues/2588 / https://github.com/PyCQA/astroid/commit/87b55a6e without a special case.

## Type of Changes

## Related Issue

https://github.com/PyCQA/pylint/pull/4357

### DIFF
--- a/astroid/objects.py
+++ b/astroid/objects.py
@@ -260,7 +260,10 @@ class PartialFunction(scoped_nodes.FunctionDef):
     def __init__(
         self, call, name=None, doc=None, lineno=None, col_offset=None, parent=None
     ):
-        super().__init__(name, doc, lineno, col_offset, parent)
+        super().__init__(name, doc, lineno, col_offset, parent=None)
+        # A typical FunctionDef automatically adds its name to the parent scope,
+        # but a partial should not, so defer setting parent until after init
+        self.parent = parent
         self.filled_positionals = len(call.positional_arguments[1:])
         self.filled_args = call.positional_arguments[1:]
         self.filled_keywords = call.keyword_arguments
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -2742,6 +2742,24 @@ class TestFunctoolsPartial:
         scope = partial_func3.parent.scope()
         assert scope.name == "test3_scope", "parented by closure"
 
+    def test_partial_does_not_affect_scope(self):
+        """Make sure partials are not automatically assigned."""
+        ast_nodes = astroid.extract_node(
+            """
+        from functools import partial
+        def test(a, b):
+            return a + b
+        def scope():
+            test2 = partial(test, 1)
+            test2 #@
+        """
+        )
+        test2 = next(ast_nodes.infer())
+        mod_scope = test2.root()
+        scope = test2.parent.scope()
+        assert set(mod_scope) == {"test", "scope", "partial"}
+        assert set(scope) == {"test2"}
+
 
 def test_http_client_brain():
     node = astroid.extract_node(
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -2720,6 +2720,28 @@ class TestFunctoolsPartial:
             assert isinstance(inferred, astroid.Const)
             assert inferred.value == expected_value
 
+    def test_partial_assignment(self):
+        """Make sure partials are not assigned to original scope."""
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas:
I tested in pylint and I got:
```
___________________________________________________________________________________ test_functional[too_many_arguments] ___________________________________________________________________________________

self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x7f37c98c7730>

    def runTest(self):
>       self._runTest()
E       AssertionError: Wrong results for file "too_many_arguments":
E       
E         32: too-many-function-args
```

On :
```python
def func_call():
    """Test we don't emit a FP for https://github.com/PyCQA/pylint/issues/2588"""
    partial_func = partial(root_function, 1, 2, 3)
    partial_func()
    return root_function(1, 2, 3) # false positive too-many-function-args
```

This is the test case for #2588 (#632 )

Alphadelta14:
okay. I've figured out how to reproduce that test locally. Will make sure that is addressed!

astroid/objects.py
```diff
@@ -260,7 +260,10 @@ class PartialFunction(scoped_nodes.FunctionDef):
     def __init__(
         self, call, name=None, doc=None, lineno=None, col_offset=None, parent=None
     ):
-        super().__init__(name, doc, lineno, col_offset, parent)
+        super().__init__(name, doc, lineno, col_offset, parent=None)
```
Alphadelta14, line 263: A typical FunctionDef will define its name in the parent scope, so in the example below, inner_func should get added to outer_func.locals, but `partial(inner_func, ...)` should not be added until an Assign node adds it.
```
def outer_func():
    def inner_func(a, b):
        return a + b
   inner_partial = partial(inner_func, 1, 2)
```

I would probably prefer if there was an interface to add self to the parent frame for LocalsDictNodeNG for ClassDef and FunctionDef that could be disabled for Partial and Lambda.

tests/unittest_brain.py
```diff
@@ -2720,6 +2720,46 @@ def other_test(a, b, *, c=1):
...
+        def scope():
+            test2 = partial(test, 1)
+            test2 #@
+        """
+        )
+        test2 = next(ast_nodes.infer())
+        mod_scope = test2.root()
+        scope = test2.parent.scope()
+        assert set(mod_scope) == {"test", "scope", "partial"}
+        assert set(scope) == {"test2"}
```
Alphadelta14, line 2761: before, there was an extra item in locals: `{"test2": PartialFunction("test"), "test": PartialFunction("test")}`

Pierre-Sassoulas (approved):
It's working in pylint now 👍

astroid/objects.py
Pierre-Sassoulas: Do you think it's worth a refactor ? :)

Alphadelta14: Maybe; I don't quite think I'm qualified here yet to say. LocalsDictNodeNG <- Lambda <- FunctionDef <- PartialFunction is a pretty weird hierarchy.

PR #1099, opened 2021-07-13
Add new If guard helper methods

cdce8p:
## Description
Add `If.is_sys_guard` and `If.is_typing_guard` helper methods. Most useful to check if analysis of the `if` body should be skipped, e.g. because module are not available for the checked python version.

## Type of Changes

## Related Issue
https://github.com/PyCQA/pylint/pull/4702
https://github.com/PyCQA/pylint/pull/4703

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,12 +6,12 @@ What's New in astroid 2.7.0?
 ============================
 Release date: TBA
 
+* Added ``If.is_sys_guard`` and ``If.is_typing_guard`` helper methods
 
 
 What's New in astroid 2.6.3?
 ============================
 Release date: TBA
-
 
 * Fix a bad inferenece type for yield values inside of a derived class.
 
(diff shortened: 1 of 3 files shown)


cdce8p:
As this is technically a new feature, should it go in `2.7.0` or can we include it in `2.6.3`?

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
This is useful enough that I want it asap in pylint but it's true it's a new feature. I would not mind having it in 2.6.3 though.

Pierre-Sassoulas:
I think it would force us to release astroid 2.7.0 very fast, so we can put https://github.com/PyCQA/pylint/pull/4702 in pylint 2.10.

Pierre-Sassoulas:
Moving this to 2.6.3 and merging because we need it for https://github.com/PyCQA/pylint/pull/4703 that is in pylint 2.9.4.

PR #1118, opened 2021-08-06
Remove use of deprecated ``node_classes`` and ``scoped_node`` API in astroid

Pierre-Sassoulas:
## Description

This removes uses of deprecated API in astroid following #1057 

## Type of Changes


astroid/nodes/__init__.py
```diff
@@ -204,13 +204,89 @@
...
+    "AsyncFunctionDef",
+    "AsyncWith",
+    "Attribute",
+    "AugAssign",
+    "Await",
+    "BinOp",
+    "BoolOp",
+    "Break",
     "builtin_lookup",
-    "are_exclusive",
```
cdce8p, line 211: Not sure if we still need it in `pylint`

cdce8p (commented):
The change looks good. I was just wondering if the intention should be different: Shouldn't the imports all be from `astroid.nodes` now? With the current solution we would need to come back to all of them once we split up `node_classes` and `scoped_nodes`.

astroid/nodes/__init__.py
Pierre-Sassoulas: We need it internally in astroid :D

Pierre-Sassoulas:
> Shouldn't the imports all be from astroid.nodes now?

Totally agree with that. And this is the goal. The MR is a small step toward it because we need to remove the circular import first, then change the imports.

astroid/nodes/__init__.py
cdce8p: The comment was referring to the fact that you removed it from `__all__`. I haven't checked if it's used in `pylint`. If it is, it might make sense to keep it in `__all__`. Similar with `unpack_infer`.

Pierre-Sassoulas: Ho my mistake, nice catch !

Issue #1115, opened 2021-08-02
slow pylint performance with a hot spot in astroid


robin-wayve:
tl;dr I have been trying to figure out why `pylint` is slow for us and profiling points to `astroid`. I'd appreciate some pointers on how I could get a clearer signal on what is going on.

### Versions

```
astroid 2.6.5
pylint 2.9.6
python 3.7.9
```

### Description

We run `pylint` against individual packages/modules in our codebase and the performance is highly variable. I've been trying to figure out why some runs are slow with `yappi`, similar to how it was done in #610.

One such module of ours has about 1000 lines of code and has become infeasibly slow on the versions listed above (I gave up after waiting 1 hour with the profiler on). I managed to get it to complete a profile in ~30 minutes by limiting the run to a single 200 line file. It does complete in under 5 minutes on our current versions: `pylint` 2.7.4 with `astroid` 2.5.2 -- I haven't captured a profile on these versions yet, but I could if that would be helpful.

I'm not able to share the code since it is proprietary, and the profile may contain sensitive information. I may attempt to scrub the profile and share the dot graph if that would be of interest.

### Profling result

The hottest node from the `callgrind` was this by quite some margin:
```
Name.infer astroid/node_classes.py:325 57.62% (16.12%) 82185683×
```
I believe that refers to this code: https://github.com/PyCQA/astroid/blob/v2.6.5/astroid/node_classes.py#L325-L379

This node has quite a lot of edges making it hard to see exactly what is going on but I hope this an interesting enough starting point for us to figure out what the cause of the bottleneck could be?

robin-wayve:
There's quite a few other details I'm not sure whether to include: 

- to get a useful profile, this was done with `--jobs=1`: it does run faster with `--jobs=0` but it's still quite slow.
- I believe our `rcfile` is relatively boring apart from disabling quite a lot of pylint's checkers
- Other options we're using
```
--persistent=n --unsafe-load-any-extension=y
```

Pierre-Sassoulas:
Thank you for opening the issue. It makes sense that the inference would be what is costing a lot on astroid. We'd need to dig deeper and analyse what inferences exactly are taking more time than they should and probably in what pylint checker.

Pierre-Sassoulas:
Coule you try to single out deep recursion call, if possible ?

robin-wayve:
I tidied up the graph, hope this will help:
[...]

PR #1120, opened 2021-08-07
Cache ancestors

dwo:
## Steps

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -43,7 +43,7 @@ import builtins
 import io
 import itertools
 import typing
-from collections import OrderedDict
+from collections import defaultdict, OrderedDict
 from typing import List, Optional
 
 from astroid import bases
@@ -1963,8 +1963,8 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
     # a dictionary of class instances attributes
     _astroid_fields = ("decorators", "bases", "keywords", "body")  # name
 
-    all_ancestors = None
-    direct_ancestors = None
+    all_ancestors = {}
+    direct_ancestors = {}
 
     decorators = None
     """The decorators that are applied to this class.
@@ -2335,10 +2335,10 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
         :returns: The base classes
         :rtype: iterable(NodeNG)
         """
-        if recurs and self.all_ancestors is not None:
-            yield from self.all_ancestors.keys()
-        elif not recurs and self.direct_ancestors is not None:
-            yield from self.direct_ancestors.keys()
+        if recurs and context in self.all_ancestors:
+            yield from self.all_ancestors[context].keys()
+        elif not recurs and context in self.direct_ancestors:
+            yield from self.direct_ancestors[context].keys()
 
         # FIXME: should be possible to choose the resolution order
         # FIXME: inference make infinite loops possible here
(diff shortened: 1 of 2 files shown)


tests/unittest_scoped_nodes.py
```diff
@@ -1040,7 +1040,7 @@ class Past(Present):
         astroid = builder.parse(data)
         past = astroid["Past"]
         attr = past.getattr("attr")
-        self.assertEqual(len(attr), 1)
+        self.assertEqual(len(attr), 1, attr)
```
Pierre-Sassoulas, line 1043:
```suggestion
        assert len(attr) == 1
```
I think when using pytest this is possible (?)

Pierre-Sassoulas (commented):
Thank you for that, if we believe the graph given in https://github.com/PyCQA/astroid/issues/1115#issuecomment-890961827 this is gonna be a huge performance boost. Could we check that this is really a performance boost, please ?

astroid/nodes/scoped_nodes.py
```diff
@@ -2331,13 +2335,26 @@ def ancestors(self, recurs=True, context=None):
...
         """
+        if recurs and context in self.all_ancestors:
+            yield from self.all_ancestors[context].keys()
+        elif not recurs and context in self.direct_ancestors:
+            yield from self.direct_ancestors[context].keys()
+
         # FIXME: should be possible to choose the resolution order
         # FIXME: inference make infinite loops possible here
-        yielded = {self}
+        yielded = OrderedDict()
```
dwo, line 2345: I think this is necessary for Python 3.6 to preserve the order of yielded ancestors: a plain `dict` achieves the same thing from Python 3.7+

tests/unittest_scoped_nodes.py
dwo: Any Ideas why this might have started failing though? I had a read through `getattr` but since it doesn't call `ancestors` directly I'm going to need to dig a little more.

dwo:
> Thank you for that, if we believe the graph given in [#1115 (comment)](https://github.com/PyCQA/astroid/issues/1115#issuecomment-890961827) this is gonna be a huge performance boost. Could we check that this is really a performance boost, please ?

Once I get the tests to pass I want to try this against the example given in https://github.com/PyCQA/pylint/issues/4079 and then perhaps retry #1115 too.

Are there some codebases to run `pylint` against that make a good benchmark?

I also wondered if I should try and measure what the extra memory cost will be: I know there's prior art for this from https://rtpg.co/2020/10/12/pylint-usage.html

tests/unittest_scoped_nodes.py
dwo: on `main`, `attr` is...
```
[<AssignName.attr l.3 at 0x[...]>]
```
with my changes this becomes...
```
[<AssignName.attr l.3 at 0x[...]>, <Const.int l.3 at 0x[...]>]
```

dwo: Oh I was looking at the wrong `getattr` before... https://github.com/PyCQA/astroid/blob/f86d6e9f67e449209fe53a6e9cef4f7b55adf52a/astroid/nodes/scoped_nodes.py#L2514-L2517

dwo: I suspect that using `ClassDef`s and `InferenceContext`s as keys was a bit naive: they aren't Singletons and don't implement `__hash__` and `__eq__` so I am only caching specific instances by object id instead of detecting when the `ancestor`s or `context`s are equivalent.

DanielNoord:
@dwo Are you still interested in picking this PR back up or should we close it?

dwo:
Damn, I forgot about this and would have liked to spend more time on it... will close for now.

DanielNoord:
I still think this would indeed be very valuable. `ancestors` is one of the most expensive calls in regular `pylint` calls.

PR #1135, opened 2021-08-17
Fix variable collision in _infer_stmts

nickdrozd:
Was this causing any bugs? Hard to say.


nickdrozd:
This really seems like something that Pylint ought to have flagged.

Pierre-Sassoulas (approved):
I think if we infer something in the loop it was implicitly True, and we were fixing inferred to true right after the loop anyway so the value was always coherent.

cdce8p:
Didn't notice any changes during a short test. Seems to be ok

PR #1137, opened 2021-08-19
 Replace the constant ``BUILTINS`` by the string 'builtins'

Pierre-Sassoulas:
## Description

 This make for clearer and also slightly faster code (means time
 seems to decrease by 0.68% with this change alone (astroid/pylint)
 in the pylint tests benchmarks). Done because we were using an
 import from astroid from astroid.bases for one of those, which is
 kinda messy.

## Type of Changes

## Related Issue

Require https://github.com/PyCQA/pylint/pull/4868 to be merged in pylint, pylint 2.10.0 being released and the pylint version used in pre-commit to be updated in astroid to work.

It's a little one, but it's a breaking change, I don't know if we can put that in 2.7.2 or should wait for 2.8 or even 3.0.  (Edit : as we need pylint 2.10 to be released it will go in 2.8.0 most likely, maybe astroid 2.7.x)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,7 +6,8 @@ What's New in astroid 2.8.0?
 ============================
 Release date: TBA
 
-
+* ``astroid.const.BUILTINS`` and ``astroid.bases.BUILTINS`` have been removed,
+  simply replace this by the string 'builtins' for better performances.
 
 What's New in astroid 2.7.2?
 ============================
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -29,7 +29,7 @@ import collections
 
 from astroid import context as contextmod
 from astroid import decorators, util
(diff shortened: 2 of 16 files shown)


cdce8p:
I think `BUILTINS` should probably never have been imported from `astroid` in the first place.
What do you think of this?
* Keep `BUILTINS = "builtins"` in `astroid/const.py`, and `from astroid.const import BUILTINS` in `astroid/bases` for backwards compatibility with `pylint`. That way the PR could be merged now.
* Remove it with `2.8.0`

Pierre-Sassoulas:
You're right let's keep this compatible with older pylint for now.

cdce8p:
> You're right let's keep this compatible with older pylint for now.

You can still move forward with this change if you like. Just keep the two things I mentioned before and it should be fine.

PR #1138, opened 2021-08-19
Remove constraint on numpy for 3.10

cdce8p:
## Description
The numpy wheels for `3.10` have been uploaded to PyPI. Thus it should be fine to remove the constraint now.


cdce8p:
Seems like the current `pylint` version is blocking `pip install`. Delaying it until `2.10` is released

Pierre-Sassoulas (approved):

cdce8p:
Rebased this one onto #1142. Should be fine as long as the pylint update is merged first.

Issue #792, opened 2020-05-22
Continuously increasing memory usage when pylint is run via its API


char101:
### Steps to reproduce

When using pylint in-process to lint a file (in this case in vim), the memory usage of the editor is continually increasing. The increase is even higher when the cache is cleared. I believe that this is caused because the module is still referenced by `lru_cache` even though it has been deleted from the manager cache and each new parsing adds a new cache entry.

```python
import gc

import astroid
import psutil
from pylint import lint
from pylint.reporters.base_reporter import BaseReporter


class Reporter(BaseReporter):
    def _display(self, layout):
        pass


def run1():
    print('run without clearing cache')
    for _ in range(0, 5):
        lint.Run(['--reports', 'n', 'pyparsing1.py'], reporter=Reporter(), exit=False)
        print(psutil.Process().memory_info().rss)


def run2():
    print('run with cache clear')
    for _ in range(0, 5):
        lint.Run(['--reports', 'n', 'pyparsing1.py'], reporter=Reporter(), exit=False)
        astroid.builder.MANAGER.astroid_cache.clear()
        gc.collect()
        print(psutil.Process().memory_info().rss)


def run3():
    print('run with clearning only checked module cache')
    for _ in range(0, 5):
        lint.Run(['--reports', 'n', 'pyparsing1.py'], reporter=Reporter(), exit=False)
        del astroid.builder.MANAGER.astroid_cache['pyparsing1']
        gc.collect()
        print(psutil.Process().memory_info().rss)


def main():
    run1()
    run2()
    run3()


if __name__ == '__main__':
    main()
```

The linted file is `253KB` `pyparsing.py` renamed to `pyparsing1.py`.

```
run without clearing cache
118296576
130756608
143630336
156053504
168882176
run with cache clear
181768192
263176192
345382912
427147264
507686912
run with clearning only checked module cache
589205504
615079936
641482752
665550848
689803264
```

### Current behavior
Increasing memory usage

### Expected behavior
Stable memory usage

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output

`2.4.1`

char101:
`LookupMixIn` uses `lru_cache` on its `lookup` method with unlimited size, that means any object of the classes that uses `LookupMixIn` will be cached indefinitely.

wzorgdrager:
Is there any workaround for this?

richtier:
perhaps this: https://stackoverflow.com/a/50699209
```py
import functools
[...]

PR #1139, opened 2021-08-19
Removed mutable default value in _inference_tip_cache

superbobry:
## Description

The mutable default is problematic when astroid is used as a library,
because it effectively becomes a memory leak, see #792.

This commit moves the cache to the global namespace and adds a public
API entry point to clear it.

## Type of Changes

|   | :bug: Bug fix          |
| ✓    | :hammer: Refactoring   |

## Related Issue

This PR partially addresses #792. More changes are needed to make memory consumption stable. See discussion in the issue.

### DIFF
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -8,13 +8,21 @@ import typing
 
 import wrapt
 
-# pylint: disable=dangerous-default-value
 from astroid.exceptions import InferenceOverwriteError
 from astroid.nodes import NodeNG
 
+InferFn = typing.Callable[..., typing.Any]
+
+_cache: typing.Dict[typing.Tuple[InferFn, NodeNG], typing.Any] = {}
+
+
+def clear_inference_tip_cache():
+    """Clear the inference tips cache."""
+    _cache.clear()
+
 
 @wrapt.decorator
-def _inference_tip_cached(func, instance, args, kwargs, _cache={}):  # noqa:B006
+def _inference_tip_cached(func, instance, args, kwargs):
     """Cache decorator used for inference tips"""
     node = args[0]
     try:
@@ -27,12 +35,10 @@ def _inference_tip_cached(func, instance, args, kwargs, _cache={}):  # noqa:B006
         return original
 
 
-# pylint: enable=dangerous-default-value
-
 
 def inference_tip(
-    infer_function: typing.Callable, raise_on_overwrite: bool = False
-) -> typing.Callable:
+    infer_function: InferFn, raise_on_overwrite: bool = False
+) -> InferFn:
     """Given an instance specific inference function, return a function to be
     given to AstroidManager().register_transform to set this inference function.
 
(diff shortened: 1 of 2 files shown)


astroid/inference_tip.py
```diff
@@ -8,13 +8,21 @@
...
-# pylint: disable=dangerous-default-value
 from astroid.exceptions import InferenceOverwriteError
 from astroid.nodes import NodeNG
 
+InferFn = typing.Callable[..., typing.Any]
+
+_cache: typing.Dict[typing.Tuple[InferFn, NodeNG], typing.Any] = {}
+
+
+def clear_inference_tip_cache():
```
superbobry, line 19: Alternatively, I can add a single `clear_caches()` entry point which also clears LRU caches on various methods in `astroid` (as discussed in #792). Wdyt?

Pierre-Sassoulas: What do you think about using the `@lru_cache(maxsize=?)` decorator to handle that ? I think it would remove the need for the API. But maybe we'd need a mechanism to set the max size instead ?

superbobry: We'd need some extra logic to account for the return value being a generator, right?

Otherwise, `lru_cache` would do the job. I'd still prefer for us to have an explicit way to clear the cache, because regardless of capacity, references to nodes will keep (potentially expensive) object around in memory for longer.

Pierre-Sassoulas: I don't know lru_cache well I did not know of the potential problem with generator. Just thought that we might as well use a specialized library to have a sane limit without any human intervention (Maybe 8 Go ?). (We can also have the API on top of it by the way).

superbobry: Sorry if I was unclear. I was referring to the bit of logic in the existing implementation, where the cached function does `itertools.tee` to make the cached generator "re-usable". We could, of course, eagerly materialize the generator into a `list`/`tuple` and cached that, instead. Does that sg?

Pierre-Sassoulas: I don't understand the ``itertools.tee`` implementation well enough to guess if it will be better one way or the other. A way to test the speed of an implementation is to use the benchmark test in pylint by doing a local install of astroid and comparing the result. But I think as long as we set a limit to the cache we store it will probably be better than the current situation.

superbobry: I think `itertools.tee` has to cache the elements somewhere, and logically if one of the resulting iterators is fully consumed, `itertools.tee` will have to keep a full copy of the original iterable around internally. With this in mind, it seems we can just drop the `itertools.tee` and always return a list iterator.

As for `lru_cache`, I'm afraid we cannot use it here because the current implementation only uses the first argument as key (because the other one is `context`?), and `lru_cache` does not allow that.

Pierre-Sassoulas (approved):
Thank you for refactoring this and permitting to clear the cache. Do you have an opinion on what need to be done to make lru_cache clear automatic ? Maybe a decorator that take a string of all the argument given to ``_inference_tip_cached`` and is cached itself ?

Pierre-Sassoulas:
Hey sorry for now answering this sooner. I thought I might as well partially answer: The inference tip always have two arguments (a node and a context). I asked other active maintainers about constant context during astroid's execution and it look like no one knows precisely. So I would have to look at the only place of truth (The CODE 😄) to answer that, but did not find the time.

PR #1140, opened 2021-08-20
Generate synthetic __init__ method for dataclasses

david-yz-liu:
## Steps

## Description

For a dataclass, an initializer is automatically generated based on its attributes. Currently astroid's `brain_dataclasses.py` doesn't infer this generated init, which makes pylint unable to detect errors in how the initializer is called, or report false positives if a different initializer is incorrectly inherited (e.g., https://github.com/PyCQA/pylint/issues/3201).

I added code to the brain the generate this initializer. In addition to basic attributes, the generated method:

- Ignores `ClassVar`s
- Includes `InitVar`s as parameters but not as instance attributes
- Includes parameter type annotations and default values
- Includes attributes that were inherited from superclass dataclasses
- Handles `field` default values, including using a sentinel for default factories. The latter is a bit tricky: I used a similar approach to the one used in Python itself, see https://github.com/python/cpython/blob/d26dbba9297adb59bc49a6583c647804ef56fc58/Lib/dataclasses.py#L504-L520.
- The method body initializes all instance attributes, though this isn't currently used for any inference.

Also, in the case that dataclass fields are in an invalid order (field with defaults precede those without defaults), no initializer is generated. In principle there should be a Pylint check for this, I can make an issue if you agree.

## Type of Changes

## Related Issue

Closes https://github.com/PyCQA/pylint/issues/3201

Also, I believe https://github.com/PyCQA/pylint/issues/3754 was closed in #1126, but I missed it, sorry about that. I can't reproduce it on main.

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -3,12 +3,12 @@
 """
 Astroid hook for the dataclasses library
 """
-from typing import Generator, Tuple, Union
+from typing import Generator, List, Optional, Tuple
 
 from astroid import context, inference_tip
 from astroid.builder import parse
 from astroid.const import PY37_PLUS, PY39_PLUS
-from astroid.exceptions import InferenceError
+from astroid.exceptions import AstroidSyntaxError, InferenceError, MroError
 from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import (
     AnnAssign,
@@ -26,6 +26,7 @@ from astroid.util import Uninferable
 DATACLASSES_DECORATORS = frozenset(("dataclass",))
 FIELD_NAME = "field"
 DATACLASS_MODULE = "dataclasses"
+DEFAULT_FACTORY = "_HAS_DEFAULT_FACTORY"  # based on typing.py
 
 
 def is_decorated_with_dataclass(node, decorator_names=DATACLASSES_DECORATORS):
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_dataclasses.py
```diff
@@ -57,17 +58,7 @@ def is_decorated_with_dataclass(node, decorator_names=DATACLASSES_DECORATORS):
...
-            assign_node.target, AssignName
-        ):
-            continue
-
-        if _is_class_var(assign_node.annotation) or _is_init_var(
-            assign_node.annotation
-        ):
-            continue
-
+    for assign_node in _get_dataclass_attributes(node):
```
Pierre-Sassoulas, line 61: 👌 I like that change, very elegant

tests/unittest_brain_dataclasses.py
```diff
@@ -295,3 +295,281 @@ class A:
...
+
+    B.__init__  #@
+    """
+    )
+    init = next(node.infer())
+    assert [a.name for a in init.args.args] == ["self", "arg0", "arg2", "arg1"]
+    assert [a.as_string() if a else None for a in init.args.annotations] == [
+        None,
+        "float",
+        "list",  # not str
```
Pierre-Sassoulas, line 555: Very nice !

Pierre-Sassoulas (approved):
Let's include that in 2.10 :)

PR #1141, opened 2021-08-20
Use specific imports instead of importing whole module

Pierre-Sassoulas:
## Description

In order to fight circular import more effectively and speed up astroid import time, we need be specific in what we need to import so we can realize a new architecture can be made.

## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -436,7 +436,7 @@ class BoundMethod(UnboundMethod):
         needs to be a tuple of classes
         """
         # pylint: disable=import-outside-toplevel; circular import
-        from astroid.nodes import node_classes
+        from astroid.nodes.node_classes import Pass
 
         # Verify the metaclass
         try:
@@ -507,7 +507,7 @@ class BoundMethod(UnboundMethod):
             col_offset=caller.col_offset,
             parent=caller,
         )
-        empty = node_classes.Pass()
+        empty = Pass()
         cls.postinit(
             bases=bases.elts,
             body=[empty],
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -27,19 +27,25 @@ inference utils.
 
 import collections
 
-from astroid import context as contextmod
-from astroid import decorators, util
+from astroid import decorators
 from astroid.const import PY310_PLUS
+from astroid.context import (
+    CallContext,
+    InferenceContext,
+    bind_context_to_node,
+    copy_context,
+)
 from astroid.exceptions import (
     AstroidTypeError,
     AttributeInferenceError,
     InferenceError,
     NameInferenceError,
 )
+from astroid.util import Uninferable, lazy_descriptor, lazy_import
 
-objectmodel = util.lazy_import("interpreter.objectmodel")
-helpers = util.lazy_import("helpers")
-manager = util.lazy_import("manager")
+objectmodel = lazy_import("interpreter.objectmodel")
+helpers = lazy_import("helpers")
+manager = lazy_import("manager")
 
 
 # TODO: check if needs special treatment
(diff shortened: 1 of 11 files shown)


astroid/bases.py
```diff
@@ -87,8 +91,13 @@ def _is_property(meth, context=None):
     if not meth.decorators:
         return False
     for decorator in meth.decorators.nodes or ():
-        inferred = helpers.safe_infer(decorator, context=context)
-        if inferred is None or inferred is util.Uninferable:
+        # Circular import
+        from astroid.helpers import (  # pylint: disable=import-outside-toplevel
+            safe_infer,
+        )
```
cdce8p, line 97: We should avoid putting an import inside a `for` loop.

astroid/interpreter/objectmodel.py
```diff
@@ -799,7 +808,9 @@ def attr_fset(self):
 
         func = self._instance
 
-        def find_setter(func: objects.Property) -> Optional[astroid.FunctionDef]:
+        def find_setter(
+            func: Property,  # pylint: disable=used-before-assignment
+        ) -> Optional[astroid.FunctionDef]:
```
cdce8p, line 813:
```suggestion
        def find_setter(func: "Property") -> Optional[astroid.FunctionDef]:
```
This current code causes a runtime exception as `Property` isn't defined. Use `"Property"` instead.

cdce8p (commented):
Haven't looked at it in detail yet, will do so later. Left a few comments about things I noticed so far.

astroid/bases.py
Pierre-Sassoulas: I did that change too fast, you're damn right.

Pierre-Sassoulas:
Yes I didn't know lazy import well when I started that : the import time actually increase with this change which is why I did not ask for a review on this one. Let's keep or extend the lazy import if required. 

I don't like the third option though, because it's very fragile and if you can easily start to have an actual circular dependency and then you suddenly need to do a huge refactor because the problem was not really addressed. I feel that if we stop using module import (``from astroid import a_module`` and use ``from astroid.a_module import WhatWeReallyUse``) its easier to understand where exactly there are circular dependencies. Because you just have to look at the import at the top of the file instead of scanning every imported module use in the whole file to get a mental map of the import "web". 

I agree that trying to touch ``bases -> helpers`` and ``objectmodel -> objects`` is really demoralizing. I tried too, it felt bad 😄 

But I think we have other tools than the three you talked about though. 
- We can burst the ginormous files that are creating circular dependencies 
- Change the architecture when classes that should be together in a module are not
- Create new file for things that should be a dependency of multiple file instead of being in one and being imported from the other files. 
- Use ``TYPE_CHECKING`` guard when the circular import is due to typing. 
- In last resort, we can also change the design

But I feel that we need to be able to understand the structure first and for that having to look at the import instead of having to look at the full file would help a lot.

cdce8p:
> Let's keep or extend the lazy import if required.

Sounds good to me! Do you want to update this PR then?

> I feel that if we stop using module import (`from astroid import a_module` and use `from astroid.a_module import WhatWeReallyUse`) its easier to understand where exactly there are circular dependencies.

👌🏻

> But I think we have other tools than the three you talked about though.

Yeah, there are more options. The three I mentioned are just the once to I found to resolve the immediate issues.

Just a hunch, but I fear that to truly fix the dependency mess, we'll need to unlink the inference functions from the nodes itself. That would be a lot of work. Might be easier to continue using lazy imports instead.

Pierre-Sassoulas:
Yes I'll update this one later :)

Pierre-Sassoulas:
Last MR for 2.7.3, then I'll release it :) I asked for a review but it's the same that you already reviewed only it's rebased and without the lazy import removal.

cdce8p (approved):
Looks good! Just a tiny comment

astroid/bases.py
```diff
@@ -436,7 +442,7 @@ def _infer_type_new_call(self, caller, context):
         needs to be a tuple of classes
         """
         # pylint: disable=import-outside-toplevel; circular import
-        from astroid.nodes import node_classes
+        from astroid.nodes.node_classes import Pass
```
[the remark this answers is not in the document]
Pierre-Sassoulas: It's a gamble as long as the circular import are not really fixed, some file would work some won't, (and some will break if we change something else later on) which is why I don't try right now.

PR #1143, opened 2021-08-22
Add filter for abc.collections in dataclass inference

david-yz-liu:
## Steps

## Description

Okay, so the filter introduced in #1130 correctly excluded imports from `typing`, but didn't exclude imports from `collections.abc`. Fixed that by adding the module's name to the existing filter. *Note*: the module defines its `__name__` as `'collections.abc'`, but the `.name` node attribute is generated from the file name, `_collections_abc.py`.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#4895

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,10 @@ What's New in astroid 2.7.3?
 ============================
 Release date: TBA
 
+* When processing dataclass attributes, exclude the same type hints from abc.collections
+  as from typing.
+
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
👌 Thanks for the fast and straightforward fix !

david-yz-liu:
No problem, I'm trying to keep an eye out for dataclass type-related issues 😅

PR #1144, opened 2021-08-23
Relax dataclass filter for Uninferable nodes.

david-yz-liu:
## Steps

## Description

The filtering introduced in #1130 excluded pydantic's [dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/), which are a drop-in replacement for built-in dataclasses.

One twist when fixing this was that inference doesn't seem to work for pydantic's code:

```python
import astroid
node = astroid.extract_node(
    """
from pydantic.dataclasses import dataclass

dataclass
"""
)

print(node.inferred())  # prints [Uninferable]
```

So I modified the filter so that any Uninferable node that looks like "dataclass" or "<expr>.dataclass" will be treated as a dataclass decorator. Probably should have done this to begin with, as it better aligns with the astroid/pylint philosophy of treating `Uninferable` as a permissive wildcard (e.g., `no-member` isn't raised for `Uninferable` values).

If inference improves and `pydantic.dataclasses` can be inferred, we'll need to add `pydantic.dataclasses` to the `DATACLASS_MODULE` variable. I didn't do that in this PR since it doesn't currently fix the problem, but I did parametrize the tests to use both `dataclasses` and `pydantic.dataclasses`, so we should be able to catch a regression if `pydantic.dataclasses` becomes inferable.

*Comment*: the code change is bigger than the actual logical change, since I refactored the two places that the change is relevant into a single helper, `_looks_like_dataclass_decorator`.

## Type of Changes

## Related Issue

Closes https://github.com/PyCQA/pylint/issues/4899

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,10 @@ Release date: TBA
 
   Closes PyCQA/pylint#4895
 
+* Apply dataclass inference to pydantic's dataclasses.
+
+  Closes PyCQA/pylint#4899
+
 
 What's New in astroid 2.7.2?
 ============================
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -2,6 +2,10 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 """
 Astroid hook for the dataclasses library
+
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):
Look good, I like the parametrization in the tests.

PR #1147, opened 2021-08-24
Rewrite while-loop

nickdrozd:

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -91,26 +91,19 @@ def path_wrapper(func):
         if context is None:
             context = contextmod.InferenceContext()
         if context.push(node):
-            return None
+            return
 
         yielded = set()
-        generator = _func(node, context, **kwargs)
 
-        try:
-            for res in generator:
-                # unproxy only true instance, not const, tuple, dict...
-                if res.__class__.__name__ == "Instance":
-                    ares = res._proxied
-                else:
-                    ares = res
-                if ares not in yielded:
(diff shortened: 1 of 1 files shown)


astroid/decorators.py
```diff
@@ -95,22 +95,22 @@ def wrapped(node, context=None, _func=func, **kwargs):
...
+            for res in generator:
+                # unproxy only true instance, not const, tuple, dict...
+                if res.__class__.__name__ == "Instance":
+                    ares = res._proxied
+                else:
+                    ares = res
+                if ares not in yielded:
+                    yield res
+                    yielded.add(ares)
+        except StopIteration as error:
```
Pierre-Sassoulas, line 109: Can the stop iteration happen anymore while using a for loop ? Shouldn't this be a for else instead ?

astroid/decorators.py
```diff
@@ -95,22 +95,22 @@ def wrapped(node, context=None, _func=func, **kwargs):
...
-            if res.__class__.__name__ == "Instance":
-                ares = res._proxied
-            else:
-                ares = res
-            if ares not in yielded:
-                yield res
-                yielded.add(ares)
+
+        try:
+            for res in generator:
```
Pierre-Sassoulas, line 100:
```suggestion
            for res in _func(node, context, **kwargs):
```

Pierre-Sassoulas (commented):
Thanks for the cleanup.

PR #1148, opened 2021-08-25
Bug pylint 3342

hippo91:
## Steps

## Description

I noticed that adding some transforms to a node through brains is done even if the module in which the node is defined is authorized to be dynamically imported. IMHO it is not a good idea.
It leads to situation where false positive are emitted for module that have been authorized to be imported. 
I remarked that situation arises especially for the `numpy` module.

This PR deactivates the transforms of a node if the module is inside the `extension_package_whitelist` set of the `Manager` instance. 

It is maybe a too strong decision and it is also possible to avoid this phenomenon only for `numpy` brains by playing with the predicate function. 

## Type of Changes

## Related Issue
Closes PyCQA/pylint#3342
Closes PyCQA/pylint#4326
if the `--extension-pkg-allow-list=numpy` option is activated.


astroid/builder.py
```diff
@@ -157,6 +160,10 @@ def _post_build(self, module, encoding):
 
         # Visit the transforms
         if self._apply_transforms:
+            if modutils.is_module_name_part_of_extension_package_whitelist(
+                module.name, self._manager.extension_package_whitelist
+            ):
+                return module
```
Pierre-Sassoulas, line 166:
```suggestion
            if module.is_in_extension_package_whitelist(
                self._manager.extension_package_whitelist
            ):
                return module
```
If we create a function in the `nodes.Module` class it permits to have auto-completion later when we'll have typing and it put the function where it's expected to be.

Pierre-Sassoulas (approved):
This makes a lot of sense, thank you for fixing this. There's probably also pandas that will have similar issue.

astroid/builder.py
cdce8p: I'm not sure this function should be part of `nodes.Module`. I see it more as a helper function for which `modutils` is actually a pretty good place already.

cdce8p (commented):
I'm not too familiar with this part of the code base. Didn't notice any regressions, so I think  this is ok

astroid/builder.py
Pierre-Sassoulas: If a function only works for a specific class, why create  a standalone function that will evade autocomplete on instance of the class and be hard to find ?

cdce8p: It doesn't depend on an instance of `nodes.Module`. The first argument is just the module name as string.

I would think of it differently: In case you have a `Module` node in pylint, are you likely to need to call this function? I think no, it's just a helper for astroid.

Pierre-Sassoulas: OK this make sense.

PR #1151, opened 2021-08-27
Some fixes for function call inference and inference contexts

### LINKED ISSUE #1008: Incorrect inference of `self` in methods defined in superclasses in unbound methods on a subclass
### Steps to reproduce
```python
import astroid
node = astroid.extract_node("""
class Base:
    def replace(self):
        return self

class Derived(Base):
    def method(self):
        return 123

this = Derived()
that = Derived.replace(this)
that
""")
inferred = next(node.infer())
inferred.getattr('method')
```

### Current behavior

`that` is inferred as an instance of `.Base` not `.Derived`, causing this exception:

```python
  File "astroid/bases.py", line 182, in getattr
    values = self._proxied.instance_attr(name, context)
  File "astroid/scoped_nodes.py", line 2433, in instance_attr
    raise exceptions.AttributeInferenceError(
astroid.exceptions.AttributeInferenceError: 'method' not found on <ClassDef.Base l.2 at 0x...>.
```

which leads to `no-member` warnings on `pylint`

### Expected behavior

`that` is inferred as an instance of `.Derived`.

2.5.7

### Related issues

Appears to be the root-cause of PyCQA/pylint#4377.
Ref PyCQA/pylint#4487

david-yz-liu:
## Steps

## Description

This PR addresses three issues for the inference of function calls. Again I went deep on this segment of the code and put the changes all in one branch, but I've split up the changes into three logically independent commits, and would be happy to split up into separate PRs if you want. (Honestly I think I got a bit carried away.)

I added tests for each change; as is my style, there are some new tests that pass on main, but I included them anyway to be thorough. I also put the Changelog entries under 2.8.0 since it affects some core algorithms.

1. https://github.com/PyCQA/pylint/issues/400. The issue here is an interesting interaction between `infer_call`, which sets a `CallContext` for inference, and calling `igetattr` in [`infer_attribute`](https://github.com/PyCQA/astroid/blob/main/astroid/inference.py#L300-L338), which statically scans all class methods for assignments to an instance attribute. The issue was that the `CallContext` was passed to `igetattr`, which caused attributes to be incorrectly inferred based on argument values for a completely different call! Here's an example from one of the test cases I added (`test_method_dynamic_instance_attr_3`):

    ```python
        node = builder.extract_node(
        """
    class A:
        def get_x(self, x):  # x is unused
            return self.x

        def set_x(self, x):
            self.x = x

    A().get_x(10)  #@
    """
    )
    inferred = node.inferred()
    assert len(inferred) == 1
    assert inferred[0] is Uninferable  # not 10!
    """
    ```

    Currently astroid uses the argument `10` to infer the `x` in `set_x`, which isn't even called at all in this simple program.

    The straightforward solution would be to clear the call context before calling `igetattr`, but this had the unintended consequence of removing some inference astroid currently does support, when an instance attribute is set in the method being called, e.g.

    ```python
    class A:
        def get_x(self, x):
            self.x = x
            return self.x

    A().get_x(10)  # Should infer to 10
    ```
    
    Without keeping the call context in this case, we'd get `Uninferable` for the call result. So to try to keep this behaviour but remove the invalid inference, I added a new `callee` attribute to the call context to store what function is actually being called in the context, and modified the way we infer argument `AssignName`s to only use the call context if the argument is for the callee. The other changes in the commit are just setting this new attribute and tidying up slightly.

2. https://github.com/PyCQA/astroid/issues/1008. This one is pretty straightforward: when we call an instance method from the class rather than an instance, e.g. `MyClass.method(obj, 10)`, the `MyClass.method` is inferred as an `UnboundMethod`, and inference context's `boundnode` attribute is `None`. I think this is the correct and intended behaviour. The problem is that when `self` is inferred in `CallSite.infer_arguments`, the current code tries to "fill in" a `boundnode` that's `None`.

    https://github.com/PyCQA/astroid/blob/0a697736138703c37571ac92d274e4ab0ee7adc3/astroid/arguments.py#L218-L241

    I added a special case to this method for when dealing with an instance method where `boundnode is None`, and at least one positional argument is passed. Since `BoundMethod.infer_call_result` [sets boundnode already](https://github.com/PyCQA/astroid/blob/0a697736138703c37571ac92d274e4ab0ee7adc3/astroid/bases.py#L522-L523), I think this is a safe change, as I don't think there are cases where `boundnode` should be `None`.

3. https://github.com/PyCQA/pylint/issues/4220. This one's a bit strange. Basically, given these classes:

    ```python
    
    class A:
        def f(self):
            return 42

    class B(A):
        def __init__(self):
            self.a = A()
            result = self.a.f()
        
        def f(self):
            pass
    ```

    In the expression `B().a.f()`, the `B().a` is an instance of `A`, but the `boundnode` of the context is set to `B`, which causes an inference error on the call to `f()`, since the `B.f` version is used rather than `A.f`. It seems the root cause is https://github.com/PyCQA/astroid/commit/d272438213e6235f642d864085b10768b501cade, which added a check in `infer_attribute` that essentially converts an inferred superclass instance (`Instance of A`) into a subclass instance, that that subclass instance is already the boundnode.

    I'm actually not sure if `boundnode` should be set in `infer_attribute` at all, since it's set appropriately in `infer_call_result` already, and I think only used in resulving method calls (`infer_attribute` itself just uses `igetattr`). But anyway I removed this check and verified that the tests that the commit added still pass, probably because `igetattr` has gotten better at dealing with inheritance since that commit was made.

## Type of Changes

## Related Issue

First commit:

Closes https://github.com/PyCQA/pylint/issues/400

Second commit:

Closes #1008
Closes https://github.com/PyCQA/pylint/issues/4377

Third commit:

Closes https://github.com/PyCQA/pylint/issues/4220

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,11 @@ Release date: TBA
     Closes #1008
     Closes PyCQA/pylint#4377
 
+* Fixed bug in inference of chained attributes where a subclass
+  had an attribute that was an instance of its superclass.
+
+    Closes PyCQA/pylint#4220
+
 
 What's New in astroid 2.7.3?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,9 @@ What's New in astroid 2.8.0?
 ============================
 Release date: TBA
 
+* Fixed bug in attribute inference from inside method calls.
+
+    Closes PyCQA/pylint#400
 
 
 What's New in astroid 2.7.3?
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,12 @@ Release date: TBA
 
     Closes PyCQA/pylint#400
 
+* Fixed bug in inference for superclass instance methods called
+  from the class rather than an instance.
+
+    Closes #1008
+    Closes PyCQA/pylint#4377
+
 
 What's New in astroid 2.7.3?
 ============================
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -216,11 +216,15 @@ class CallSite:
                     positional.append(arg)
 
         if argindex is not None:
+            boundnode = getattr(context, "boundnode", None)
             # 2. first argument of instance/class method
             if argindex == 0 and funcnode.type in ("method", "classmethod"):
-                if context.boundnode is not None:
-                    boundnode = context.boundnode
-                else:
+                # context.boundnode is None when an instance method is called with
+                # the class, e.g. MyClass.method(obj, ...). In this case, self
+                # is the first argument.
+                if boundnode is None and funcnode.type == "method" and positional:
+                    return positional[0].infer(context=context)
+                if boundnode is None:
                     # XXX can do better ?
                     boundnode = funcnode.parent.frame()
 
@@ -242,7 +246,7 @@ class CallSite:
             # if we have a method, extract one position
             # from the index, so we'll take in account
             # the extra parameter represented by `self` or `cls`
-            if funcnode.type in ("method", "classmethod"):
+            if funcnode.type in ("method", "classmethod") and boundnode:
                 argindex -= 1
             # 2. search arg index
             try:
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -169,6 +169,7 @@ def _infer_method_result_truth(instance, method_name, context):
         if not meth.callable():
             return util.Uninferable
         try:
+            context.callcontext = contextmod.CallContext(args=[], callee=meth)
             for value in meth.infer_call_result(instance, context=context):
                 if value is util.Uninferable:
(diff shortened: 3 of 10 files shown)


Pierre-Sassoulas (approved):
👍

hippo91:
@david-yz-liu congrats! It is a nice work and with very clear explanations! Thanks a lot!

david-yz-liu:
Thanks @hippo91!

@Pierre-Sassoulas I see you marked this for 2.7.3, but I put the Changelog entries under 2.8.0. Are you going to change that before the next release, or do you want me to submit a PR to do it?

Pierre-Sassoulas:
Don't worry I'll fix it before releasing if required.

PR #1154, opened 2021-08-28
Bug pylint 4896

hippo91:
## Steps

## Description
This PR adds a brain module to allow the inference of the `value` attribute of `ctypes` redefined types.
In a first try, i tried to implement a method based on an inference tip that was focused on the inference of the `value` attribute.
While the inference was correct, i have been unable to build a predicate that targeted precisely the `value` attribute of `c_types` instances. Thus the inference tip was also applied for too many classes and leads to inference problem (infinite recursions) in the unittest dedicated to the `enum` brain.
Instead this PR proposes an extension of the `ctypes` modules in which every `ctypes` redefined types is overloaded with its two members (`value` and `_type_`) redefined.
 

## Type of Changes

## Related Issue
Closes PyCQA/pylint#4896


Pierre-Sassoulas (approved):
LGTM :)

cdce8p (approved):

cdce8p:
@Pierre-Sassoulas Do you want to release it with `2.7.3` or do `2.8.0` next?

Pierre-Sassoulas:
I don't think there were major changes so I was going to uput everything in 2.7.3, but I don't have strong opinion about that.  I can put all the new change in 2.7.3 or in 2.8.0. What do you suggest ?

cdce8p:
The majority of changes would probably qualify as bugfix, so `2.7.3` should be fine. Just wanted to ask after I saw changelog entries for both `2.7.3` and `2.8.0`.

PR #1157, opened 2021-08-31
Deprecate some default values for node init calls

cdce8p:
## Description
Some arguments that should always have a value are currently `Optional` with default `None`. This complicates typing them correctly and necessitates additional checks in `pylint`. This PR adds `DeprecationWarnings` if any of these args are `None`.

* `name` (`str`) for `Name`, `AssignName`, `DelName`
* `attrname` (`str`) for `Attribute`, `AssignAttr`, `DelAttr`
* `op` (`str`) for `AugAssign`, `BinOp`, `BoolOp`, `UnaryOp`
* `names` (`list[tuple[str, str | None]]`) for `Import`

--
I plan on adding a few `DeprecationWarnings` in `2.8.0`. Especially for `postinit` methods.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,7 +6,7 @@ What's New in astroid 2.8.0?
 ============================
 Release date: TBA
 
-* Add additional deprecation warnings in preparation for astroid 3.0.0
+* Add additional deprecation warnings in preparation for astroid 3.0
 
   * Require attributes for some node classes with ``__init__`` call.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,15 @@ What's New in astroid 2.8.0?
 ============================
 Release date: TBA
 
+* Add additional deprecation warnings in preparation for astroid 3.0.0
+
+  * Require attributes for some node classes with ``__init__`` call.
+
+    * ``name`` (``str``) for ``Name``, ``AssignName``, ``DelName``
+    * ``attrname`` (``str``) for ``Attribute``, ``AssignAttr``, ``DelAttr``
+    * ``op`` (``str``) for ``AugAssign``, ``BinOp``, ``BoolOp``, ``UnaryOp``
+    * ``names`` (``list[tuple[str, str | None]]``) for ``Import``
+
 
 What's New in astroid 2.7.4?
 ============================
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -18,12 +18,24 @@
 """ A few useful function/method decorators."""
 
 import functools
+import inspect
+import sys
+import warnings
+from typing import Callable, TypeVar
 
 import wrapt
 
 from astroid import util
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceError
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
+R = TypeVar("R")
+P = ParamSpec("P")
 
 
 @wrapt.decorator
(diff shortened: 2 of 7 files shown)


Pierre-Sassoulas (approved):
This will be very helpful if we want typing to be useful :ok_hand:

cdce8p:
@Pierre-Sassoulas Is it ok to merge? Ideally this should be part of `2.8.0`. Don't know if you are planing a `2.7.4` release before that.

Pierre-Sassoulas:
I'm planning on releasing 2.11 quickly we might have to release new astroid version if there are new problem in it, but we can release 2.8 for astroid with just this deprecation I think.

cdce8p:
Found a more elegant solution using a decorator. This should make it much easier to reuse for the `postinit` methods as well.

Example deprecation message:
```txt
tests/unittest_transforms.py::TestTransforms::test_transform_patches_locals
  /Users/marc/Develop/astroid/astroid/decorators.py:198: DeprecationWarning: 'name' will be a required attribute for 'AssignName.__init__' in astroid 3.0 ('name' should be of type: 'str')
    warnings.warn(
```

Pierre-Sassoulas (approved):
Awesome, I like this decorator, very clean !

DanielNoord:
@cdce8p @Pierre-Sassoulas I did a little performance analysis and for `pylint pylint` 15% of the times is spent in the functions added in this commit. See:
         79452892 function calls (71616899 primitive calls) in 53.587 seconds

   Ordered by: internal time

   187363    2.431    0.000    8.151    0.000 inspect.py:2276(_signature_from_function)
17167051/17154349    1.928    0.000    2.094    0.000 {built-in method builtins.isinstance}
     1232    1.874    0.002    1.874    0.002 {built-in method builtins.compile}
  1311574    1.683    0.000    3.166    0.000 inspect.py:2630(__init__)
1039362/407980    1.576    0.000   25.164    0.000 node_ng.py:121(infer)
1279086/135393    1.327    0.000    1.868    0.000 node_ng.py:525(nodes_of_class)
350640/334705    1.188    0.000    5.589    0.000 transforms.py:33(_transform)

And the following `.png`. (Scale is horrible I know):
out

I think it would be very worthwhile to see if we can remove the `inspect.signature`  call here. I can open an issue if wanted!

Pierre-Sassoulas:
Great analysis ! I hope some day we'll have a warning that will tell use "hey, everything now takes 15% longer" in github actions 😄 

On the top of my head I see three solutions:
- Less elegant solution for deprecation without inspection. (hard coded / copy pasted messages ?)
- Semi hard coded where we remove the need to call inspect but keep the decorators
- We declare the deprecation period over, apply the changes, and release 3.0 (maybe a bit extreme, 4 months isn't that long a time for a deprecation.)

cdce8p:
Yeah, doing a `inspect.signature` call is expensive. I had hoped that it wouldn't be that bad since it's limited to `__init__` calls. Guess that wasn't the case 🙈

Removing the introspection is unfortunately not an option. It would create so many false positives that the message would become useless. As for moving ahead with `3.0`, although tempting, the issue is that not all attributes that need to are actually deprecated. I didn't ever got around to finishing it as it was good enough.

There might be one idea that could work though. @DanielNoord I mentioned it here https://github.com/PyCQA/astroid/pull/1372#issuecomment-1025240741 already but in a different context. What if we guard the decorator itself?

I don't have the time to check how it would be possible, but if we would know that deprecation warnings are hidden, we wouldn't need to run the expensive parts and could use a passthrough decorator instead. I.e. something like this
```py
if deprecation_warning_enabled:
    def deprecate_default_argument_values(
        astroid_version: str = "3.0", **arguments: str
    ) -> Callable[[Callable[P, R]], Callable[P, R]]:
        ...
        # with inspect.signature
else:
    def deprecate_default_argument_values(
        astroid_version: str = "3.0", **arguments: str
    ) -> Callable[[Callable[P, R]], Callable[P, R]]:
        def deco(func: Callable[P, R]) -> Callable[P, R]:

            @functools.wraps(func)
            def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
                return func(*args, **kwargs)

            return wrapper
        return deco
```

The obvious requirement would probably be that they should be enabled for `pytest` runs.

Pierre-Sassoulas:
I created #1383 so that this discussion is not lost in a closed MR

PR #1161, opened 2021-09-02
Support pyz imports

grayjk:
## Description

Adds pyz and pyzw (PEP 441) to the list of zip import extensions

## Type of Changes

## Related Issue

Closes PyCQA/pylint#3887

### DIFF
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -220,6 +220,17 @@ class AstroidManagerTest(
                 os.path.sep.join(["data", os.path.normcase("MyPyPa-0.1.0-py2.5.zip")])
             )
 
+    def test_ast_from_module_name_pyz(self):
+        try:
+            linked_file_name = os.path.join(resources.RESOURCE_PATH, "MyPyPa-0.1.0-py2.5.pyz")
+            os.symlink(os.path.join(resources.RESOURCE_PATH, "MyPyPa-0.1.0-py2.5.zip"),
+                       linked_file_name)
+
+            with self._restore_package_cache():
+                self._test_ast_from_zip(linked_file_name)
+        finally:
+            os.remove(linked_file_name)
+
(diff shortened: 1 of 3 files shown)


tests/unittest_manager.py
```diff
@@ -220,6 +220,21 @@ def test_ast_from_module_name_zip(self):
...
+            )
+            os.symlink(
+                os.path.join(resources.RESOURCE_PATH, "MyPyPa-0.1.0-py2.5.zip"),
+                linked_file_name,
+            )
+
+            with self._restore_package_cache():
+                self._test_ast_from_zip(linked_file_name)
+        finally:
+            os.remove(linked_file_name)
```
Pierre-Sassoulas, line 236: We should use the ``tmp_path`` fixture from pytest.

Pierre-Sassoulas (commented):
Are those tests sufficient to check that astroid support pyz ? Shouldn't we try to parse a 'real' pyz package ? This seems... a little too easy. (Disclaimer: I did not know about pyz before reading the issue and pep 441)

tests/unittest_manager.py
grayjk: I think this test is sufficient. A pyz is just a Python package and `__main__.py` in a zip file. See https://bugs.python.org/issue1739468. This was surprisingly easy because astroid already supported zips via astroid.manager.zip_import_data. I just had to expand the file extension filter to allow a pyz through.

tmp_path from pytest doesn't appear to be used in any of the other tests. I tried to follow the style/pattern of other tests. See temporary symlinks created in unittest_modutils.py.

Pierre-Sassoulas: It's true the style of other code is not using ``tmp_path`` but it's because it's coming directly from 15 years ago and tmp_path was not available then. You have a point though we should probably refactor the whole code at once.

PR #1163, opened 2021-09-04
Add typing in tests directory

Pierre-Sassoulas:
## Description

Adding typing in tests, we're not ready for using mypy yet. The design of ``extract_nodes`` that return a Node or a list of Node is peculiar and will probably cause a lot of typing issues.

## Type of Changes

### DIFF
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -186,6 +186,7 @@ class NamedTupleTest(unittest.TestCase):
            pass
         """
         )
+        assert isinstance(klass, nodes.ClassDef)
         self.assertEqual(
             [anc.name for anc in klass.ancestors()], ["X", "tuple", "object"]
         )
@@ -203,6 +204,7 @@ class NamedTupleTest(unittest.TestCase):
            pass
         """
         )
+        assert isinstance(klass, nodes.ClassDef)
         base = next(base for base in klass.ancestors() if base.name == "X")
         self.assertSetEqual({"a", "b", "c"}, set(base.instance_attrs))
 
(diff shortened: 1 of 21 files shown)


tests/unittest_brain.py
```diff
@@ -106,14 +110,14 @@ def _assert_hashlib_class(self, class_obj):
         self.assertEqual(len(class_obj["digest"].args.args), 1)
         self.assertEqual(len(class_obj["hexdigest"].args.args), 1)
 
-    def test_hashlib(self):
+    def test_hashlib(self) -> None:
```
cdce8p, line 113: I'm not really sure we should strictly type all tests. There isn't really much of a benefit in doing so. Those are just tests after all. Especially adding `-> None` to every test method seems unnecessary. IMO we shouldn't do that.

--
I do however agree that typing test helper methods will be quite useful. (Mostly for improved intellisense.)

--
Just FYI: Home Assistant only runs `flake` and `black` over the test files. `mypy` and `pylint` only for the production code.

Pierre-Sassoulas: I think once it's done having it is better than not having it. Also useful to have mypy on the whole codebase (it's easier to not have a specific config for tests).

cdce8p: > I think once it's done having it is better than not having it. 

Don't believe you can convince me on that one 😄

> Also useful to have mypy on the whole codebase (it's easier to not have a specific config for tests)

Excluding the `tests` folder is pretty easy.

--
Anyway, the topic is not worth the arguments. Just do what you think.

PR #1165, opened 2021-09-05
Fix crash on datafields

DanielNoord:
## Steps

## Type of Changes

## Related Issue

Fixes the crash reported on https://github.com/PyCQA/pylint/issues/4963
Tests added there.

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -12,7 +12,7 @@ from typing import Generator, List, Optional, Tuple
 from astroid import context, inference_tip
 from astroid.builder import parse
 from astroid.const import PY37_PLUS, PY39_PLUS
-from astroid.exceptions import AstroidSyntaxError, InferenceError, MroError
+from astroid.exceptions import AstroidSyntaxError, InferenceError, MroError, UseInferenceDefault
 from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import (
     AnnAssign,
@@ -233,10 +233,10 @@ def infer_dataclass_attribute(
 
 def infer_dataclass_field_call(
     node: Call, ctx: Optional[context.InferenceContext] = None
-) -> Optional[Generator]:
+) -> Generator:
     """Inference tip for dataclass field calls."""
     if not isinstance(node.parent, (AnnAssign, Assign)):
-        return
+        raise UseInferenceDefault
     field_call = node.parent.value
     default_type, default = _get_field_default(field_call)
     if not default_type:
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -233,19 +233,20 @@ def infer_dataclass_attribute(
 
 def infer_dataclass_field_call(
     node: Call, ctx: context.InferenceContext = None
-) -> Generator:
+) -> Optional[Generator]:
     """Inference tip for dataclass field calls."""
-    if isinstance(node.parent, (AnnAssign, Assign)):
-        field_call = node.parent.value
-        default_type, default = _get_field_default(field_call)
-        if not default_type:
-            yield Uninferable
-        elif default_type == "default":
-            yield from default.infer(context=ctx)
-        else:
-            new_call = parse(default.as_string()).body[0].value
-            new_call.parent = field_call.parent
-            yield from new_call.infer(context=ctx)
+    if not isinstance(node.parent, (AnnAssign, Assign)):
+        return
+    field_call = node.parent.value
+    default_type, default = _get_field_default(field_call)
+    if not default_type:
+        yield Uninferable
+    elif default_type == "default":
+        yield from default.infer(context=ctx)
+    else:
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_dataclasses.py
```diff
@@ -231,19 +232,20 @@ def infer_dataclass_attribute(
...
-    default_type, default = _get_field_default(field_call)
-    if not default_type:
-        yield Uninferable
-    elif default_type == "default":
-        yield from default.infer(context=ctx)
-    else:
-        new_call = parse(default.as_string()).body[0].value
-        new_call.parent = field_call.parent
-        yield from new_call.infer(context=ctx)
+    if isinstance(node.parent, (AnnAssign, Assign)):
```
Pierre-Sassoulas, line 238:
```suggestion
    if not isinstance(node.parent, (AnnAssign, Assign)):
        return
```
Unless it does not works with yield ? It permits to simplify the number of things to keep in mind when reading the code and reduce the diff and future conflict.

Pierre-Sassoulas (commented):
👍

DanielNoord:
Also changed the typing of the return!

astroid/brain/brain_dataclasses.py
```diff
@@ -231,9 +232,11 @@ def infer_dataclass_attribute(
 
 
 def infer_dataclass_field_call(
-    node: AssignName, ctx: context.InferenceContext = None
-) -> Generator:
+    node: Call, ctx: context.InferenceContext = None
```
DanielNoord, line 235:
```suggestion
    node: Call, ctx: Optional[context.InferenceContext] = None
```

astroid/brain/brain_dataclasses.py
```diff
@@ -231,9 +232,11 @@ def infer_dataclass_attribute(
 
 
 def infer_dataclass_field_call(
-    node: AssignName, ctx: context.InferenceContext = None
-) -> Generator:
+    node: Call, ctx: Optional[context.InferenceContext] = None
+) -> Optional[Generator]:
     """Inference tip for dataclass field calls."""
+    if not isinstance(node.parent, (AnnAssign, Assign)):
+        return
```
cdce8p, line 239: I think this should be `raise UseInferenceDefault` instead. Please test it before changing it as I'm not 100% certain.

astroid/brain/brain_dataclasses.py
```diff
@@ -231,9 +232,11 @@ def infer_dataclass_attribute(
 
 
 def infer_dataclass_field_call(
-    node: AssignName, ctx: context.InferenceContext = None
-) -> Generator:
+    node: Call, ctx: Optional[context.InferenceContext] = None
+) -> Optional[Generator]:
```
cdce8p, line 236:
```suggestion
) -> Generator:
```
The return type will then only be `Generator`.

cdce8p (changes requested):
I saw that you opened a PR in pylint with regression tests. It would be great if you could add a test for astroid as well.

DanielNoord:
Not quite sure what the tests should be. I don't really know how `astroid` works, but saw the issue in `pylint` and saw that it was quite easily fixable.

Please let me know what the tests should be and then I will see if `raise UseInferenceDefault` also works.

tests/unittest_brain_dataclasses.py
```diff
@@ -643,3 +643,39 @@ class A:
...
+    from typing import cast
+
+    @dc.dataclass
+    class A:
+        attribute: int = cast(int, field(default_factory=dict))
+
+    """
+    )
+    inferred = node.inferred()
+    assert len(inferred) == 1
```
cdce8p, line 663:
```suggestion
    assert len(inferred) == 1 and isinstance(inferred[0], nodes.ClassDef)
    assert "attribute" in inferred[0].instance_attrs
```
Just testing that `attribute` is indeed an `instance_attr` should be enough.

tests/unittest_brain_dataclasses.py
```diff
@@ -643,3 +643,39 @@ class A:
...
+def test_enclosed_field_call(module: str):
+    """Test inference of dataclass attribute with a non-annotated field call in another function call"""
+    node = astroid.extract_node(
+        f"""
+    from {module} import dataclass, field
+    from typing import cast
+
+    @dc.dataclass
+    class A:
+        attribute = cast(int, field(default_factory=dict))
```
cdce8p, line 676: Why not combine both tests by adding this as `attribute2` in the first test. Not really necessary to have two separate once for it.

cdce8p (commented):
Had a bit of time to test it. `raise UseInferenceDefault` will work.

tests/unittest_brain_dataclasses.py
DanielNoord: Good idea! Going to do this now!

DanielNoord:
Turns out I did not know enough about `dataclasses`. The second test was non-sensical and did not cover anything not already covered by the other tests so I removed it.

cdce8p (approved):

Issue #1015, opened 2021-06-08
false not-an-iterable for class method with @overload (2.5.7 regression)


belm0:
### Steps to reproduce

Starting from astroid 2.5.7, I'm seeing false not-an-iterable when generator class methods are annotated with `@overload`.

```python
from typing import overload, Iterator

class MyClass:
    @overload
    def transitions(self, foo: int, bar: int) -> Iterator[int]: ...
    @overload
    def transitions(self, baz: str) -> Iterator[str]: ...
    def transitions(self, foo_or_baz, bar=None):
        yield

for _ in MyClass().transitions('hello'):
    pass
```

If `@overload` is removed, or the function is moved to the module level, or I switch to astroid 2.5.6, the problem goes away.

It happens with pylint-2.8.3 or pylint-3.0.0a3.

### Current behavior

E1133: Non-iterable value MyClass().transitions('hello') is used in an iterating context (not-an-iterable)

### Expected behavior

no error

belm0:
(in this example, I can work around it by replacing `...` with `yield`, but in my actual code this doesn't work for some reason...)

hippo91:
@belm0 thanks for the report.

cdce8p:
The issue seems to have started with #934, the previous commit is fine: https://github.com/PyCQA/astroid/commit/2e8417ffc2285e798ccdef86f743abb75958e2c6. (Tested with pylint `2.8.3`.)
@nelfin Would you like to take a look at this?

nelfin:
Sure, happy to take this one

nelfin:
This doesn't seem to have anything to do with `@overload`. Here's a minimum reproducing example:
```python
class A:
    def foo(self): ...
    def foo(self):
        yield

for _ in A().foo():
    pass
```

I think this has to do with the name resolution order, `Attribute.foo` is inferred as the `foo(): ...` instead of `foo(): yield` and the lack of a function body means `infer_call_result` infers `None` (arguably correct, just given incorrect inputs). I'm chasing down the root cause and a fix.

Pierre-Sassoulas:
https://github.com/PyCQA/pylint/issues/7624 have a use case related to this in pylint

mscuthbert:
THANKS @jacobtylerwalls -- this was a major bug for pre-typing code that returned different things depending on settings (would not write new code for that today) and needed `@overload` to specify the proper return values.

PR #1168, opened 2021-09-11
Add typing to ``NodeNG.nodes_of_class``

DanielNoord:
## Steps

## Description

This uses overloads to type the ``nodes_of_class`` function.

For some reason `pylint` doesn't pick up on the iterable context correctly. Some help would be appreciated! This is not shown in the CI but does show up locally. 

## Type of Changes


DanielNoord:
As said in the description: would like some help figuring out why `pylint` throws an error. I made a basic test case with `overload` and `pylint` did seem to understand them. Not sure why it doesn't pick up that the function will always return an iterable.

cdce8p (changes requested):
Thanks for starting the work on this one! It was long overdue.
I left a few comments. Basically, all current overloads are redundant and can be removed.

--
One area we still need to take a look at is the annotation for `klass`. As this can be a tuple as well, it complicates things. An idea:
```py
T_Nodes = TypeVar("T_Nodes", bound="NodeNG")
T_Nodes2 = TypeVar("T_Nodes2", bound="NodeNG")
T_Nodes3 = TypeVar("T_Nodes3", bound="NodeNG")
SkipKlassT = Union[None, Type["NodeNG"], Tuple[Type["NodeNG"], ...]]


class NodeNG:
    @overload
    def nodes_of_class(
        self,
        klass: Type[T_Nodes],
        skip_klass: SkipKlassT = None,
    ) -> Iterator[T_Nodes]:
        ...

    @overload
    def nodes_of_class(
        self,
        klass: Tuple[Type[T_Nodes], Type[T_Nodes2]],
        skip_klass: SkipKlassT = None,
    ) -> Union[Iterator[T_Nodes], Iterator[T_Nodes2]]:
        ...

    @overload
    def nodes_of_class(
        self,
        klass: Tuple[Type[T_Nodes], Type[T_Nodes2], Type[T_Nodes3]],
        skip_klass: SkipKlassT = None,
    ) -> Union[Iterator[T_Nodes], Iterator[T_Nodes2], Iterator[T_Nodes3]]:
        ...

    # Technically any number of additional overloads with a different number of
    # items for 'klass' could be added. However, 3 should cover most cases
    # Use the last overload as fallback.
    # Unfortunately, this doesn't work in 'mypy'. Probably because of a bug.

    @overload
    def nodes_of_class(
        self,
        klass: Tuple[Type[T_Nodes], ...],
        skip_klass: SkipKlassT = None,
    ) -> Iterator[T_Nodes]:
        ...

    def nodes_of_class(
        self,
        klass: Union[
            Type[T_Nodes],
            Tuple[Type[T_Nodes], Type[T_Nodes2]],
            Tuple[Type[T_Nodes], Type[T_Nodes2], Type[T_Nodes3]],
            Tuple[Type[T_Nodes], ...],  # related to the 'mypy' bug. Needs to be excluded as well
        ],
        skip_klass: SkipKlassT = None,
    ) -> Union[Iterator[T_Nodes], Iterator[T_Nodes2], Iterator[T_Nodes3]]:
        # implementation
        pass
```

--
Regarding the `pylint` issue. I would think this is a bug. Need to take a look at it sometime. For now, I would recommend to use `# pylint: ignore` and link the issue.

DanielNoord:
Wow, this is a much cleaner solution! I did not know `bound` worked like this. Hopefully it didn't take you you too much time to come up with this.
I do think this will be really helpful in tackling some of the `mypy` issues for `pylint` as we get much more precise typing now :)
Glad to have put this on your radar!

astroid/nodes/scoped_nodes.py
```diff
@@ -1734,7 +1734,11 @@ def infer_yield_result(self, context=None):
         :returns: What the function yields
         :rtype: iterable(NodeNG or Uninferable) or None
         """
-        for yield_ in self.nodes_of_class(node_classes.Yield):
+        for (
+            yield_
+        ) in self.nodes_of_class(  # pylint: disable=not-an-iterable # See https://github.com/PyCQA/astroid/issues/1015
```
cdce8p, line 1739: Just realized that we would need to add this to every call in `pylint` as well 😕
The issue should probably be fixed before we can merge it.

DanielNoord: I did a `pip instal -e local_astroid_folder` and succeeded in running the pre-commit to test for this.

The problem comes from `overload`s where the `...` is on a separate line (I think).
We don't have those in pylint (yet), so I think for the time-being we are fine.

You might want to check this though by doing `pip -e` as well, just to be sure!

cdce8p:
> Hopefully it didn't take you you too much time to come up with this.

I spend most of the time dealing with type checker issues to be honest. The one I mentioned for `mypy` and also opened `two` for `pyright`.

> do think this will be really helpful in tackling some of the `mypy` issues for `pylint` as we get much more precise typing now :)

We should be able to remove a few `typing.cast` as well.

astroid/nodes/scoped_nodes.py
cdce8p: Just tested it as well and you're right. At the moment `pylint` doesn't complain. The reason however isn't that we don't use `overloads`. It's actually that `node` in `visit_xxx(self, node)` cannot be inferred and thus no error is emitted. In contrast, just doing something like this will easily produce one:
```py
def func():
    node = nodes.For()
    for n in node.nodes_of_class(nodes.Name):  # not-an-iterable
        pass
```

astroid/nodes/scoped_nodes.py
```diff
@@ -1734,7 +1734,11 @@ def infer_yield_result(self, context=None):
...
+            node_classes.Yield
+        ):
```
cdce8p, line 1741:
```suggestion
        # pylint: disable=not-an-iterable
        # https://github.com/PyCQA/astroid/issues/1015
        for yield_ in self.nodes_of_class(node_classes.Yield):
```

astroid/nodes/scoped_nodes.py
DanielNoord: Didn't we just add typing to all `visit_xxx(self, node)` calls in `pylint`? Or is that not enough to help `mypy`?

I saw you just assigned somebody to #1015. I think we should block this until that issue has been fixed, right?

cdce8p: Except from some corner cases, type annotations **aren't** used to infer values in pylint. That is different in `mypy`.

>  I think we should block this until that issue has been fixed, right?

Since this doesn't actually cause any issues with `pylint`, I would say we can merge it now. It doesn't necessarily need to wait for #1015.

cdce8p:
@DanielNoord Would you like to work on a follow up PR for `pylint` to remove now redundant `typing.cast` calls? Searching for `= cast(nodes.` should probably do the trick.

astroid/nodes/node_ng.py
```diff
@@ -1,14 +1,20 @@
...
-from typing import ClassVar, Optional
+from typing import ClassVar, Iterator, Optional, Tuple, Type, TypeVar, Union, overload
 
 from astroid import decorators, util
 from astroid.exceptions import AstroidError, InferenceError, UseInferenceDefault
 from astroid.manager import AstroidManager
 from astroid.nodes.as_string import AsStringVisitor
 from astroid.nodes.const import OP_PRECEDENCE
 
+# These TypeVar's are used to type NodeNG.nodes_of_class()
```
cdce8p, line 12:
```suggestion
# Types for 'NodeNG.nodes_of_class()'
```

cdce8p (approved):
Looks good 🚀
Thanks @DanielNoord!

DanielNoord:
> @DanielNoord Would you like to work on a follow up PR for `pylint` to remove now redundant `typing.cast` calls? Searching for `= cast(nodes.` should probably do the trick.

Yes! You can assign that (imaginary) issue to me 😄

PR #1170, opened 2021-09-13
Add typing to `scope()`

DanielNoord:
## Steps

## Description

After stumbling across this in https://github.com/PyCQA/pylint/pull/4995 I thought I might as well make a start with this.

One of the problems I foresee is that in some cases we return `None` in `node_ng.py`. I am not sure if this ever happens in a real scenario, but this will trip up `mypy` for `pylint` as it will want us to assert that any call to `NodeNG.scope()` did not return `None`.
Based on the docstring of that function that might be (legacy) code which is never triggered, but I am not sure. Perhaps a maintainer knows more about this. 
In any case, if we can remove the `Optional` from the return typing this would be much more useful for `pylint`.

## Type of Changes

## Related Issue


astroid/nodes/node_classes.py
```diff
@@ -2020,7 +2023,7 @@ def postinit(self, nodes: typing.List[NodeNG]) -> None:
         """
         self.nodes = nodes
 
-    def scope(self):
+    def scope(self) -> Optional["LocalsDictNodeNG"]:
```
cdce8p, line 2026: I was debating with myself what to do here. Technically, `Optional` is correct as nodes can be created without parents. In practice however, only `Module` nodes won't have a `parent`.

We could remove the `Optional` and add `type: ignore` to the function body. The drawback is that we risk missing some obscure corner case where it's actually `None`. The safest course of action would likely be to leave it and modify `pylint` with `node.scrope() is not None` checks.

cdce8p, line 2026: The function body here isn't actually safe either. It should be
```py
if self.parent and self.parent.parent:
    return self.parent.parent.scope()
return None
```

astroid/nodes/node_ng.py
```diff
@@ -246,7 +259,7 @@ def frame(self):
         """
         return self.parent.frame()
 
-    def scope(self):
+    def scope(self) -> Optional["LocalsDictNodeNG"]:
```
cdce8p, line 262: Regarding `LocalsDictNodeNG`: I fear that at some point it will just create another circular dependency. Maybe we should start looking into `Protocol` classes for that 🤔 Haven't really done so myself, thus don't know for certain if it would work.

https://docs.python.org/3/library/typing.html#typing.Protocol

--
Another issue: If at some point some scoped node inherits from another mixin class (not `LocalsDictNodeNG`, this will break.

cdce8p (changes requested):
Added a few comments. Unfortunately, they will probably result in more work than you expected, sorry about that.

astroid/nodes/node_classes.py
DanielNoord: Just thinking out loud here, but what do we currently do when `scope()` returns `None`? As in, what is the use case of that option? I can imagine in those cases we want to go back a level and get the scope of the module. Or are there are checks that depend on the fact that `scope()` is `None`?
If the latter is false we might do something were if `scope()` would return `None` and `isinstance(self, nodes.Module)` we return self. And if it's not a Module do some error-handling. But this is completely dependent on the question whether there are any actual use-cases for returning `None`

cdce8p: > Just thinking out loud here, but what do we currently do when `scope()` returns `None`? As in, what is the use case of that option?

We can't really do anything. It's only important to make sure the it doesn't create an error. E.g. by trying to access an attribute of it.

> But this is completely dependent on the question whether there are any actual use-cases for returning `None`

As `None` shouldn't be returned unless someone didn't properly assign `parent`, the only thing for us to do in these cases is to ignore the node in the current check (i.e. `return` early) and move on.

--
**OR** we could decide that we actually want to fail in these cases (as they are errors anyway). In which case we could remove `Optional` and add `type: ignore` as explained earlier.

DanielNoord: I would be in favour of the latter. I feel adding checks to the `pylint` codebase for `scope() != None` should not be the preferred solution as it delegates an `astroid` problem to `pylint`. IMO a solution within `astroid` is preferable.

If I understand you correctly you want to leave the function body as is and just add a `type: ignore`. You probably have good reasons for this (as you always do 😄 ), but why wouldn't you want to check for `self.parent` and raise an `AttributeError` when it is not assigned?

cdce8p: > If I understand you correctly you want to leave the function body as is and just add a `type: ignore`. You probably have good reasons for this (as you always do 😄 )

Actually I don't have a reason for that specifically 😅

> but why wouldn't you want to check for `self.parent` and raise an `AttributeError` when it is not assigned?

Raising an exception is a good idea! Maybe not `AttributeError`, but a custom one inheriting from `AstroidError`. We could include the name of the class that doesn't have a parent in the error message. That should speed up debugging.

DanielNoord: After discussion above, this is no longer relevant.

astroid/nodes/node_ng.py
DanielNoord: Are you sure? Doesn't using `"Class"` prevent circular dependency?

astroid/exceptions.py
```diff
@@ -256,6 +256,19 @@ class InferenceOverwriteError(AstroidError):
...
+class ParentMissingError(AstroidError):
+    """Raised when a node which is expected to have a parent attribute is missing one
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    target = None
+
+    def __init__(self, message="Parent not found on {target!r}.", **kws):
```
cdce8p, line 268: Does `message` need to be an argument? Should be enough to define it in the `__init__` body.

--
Furthermore, typing `**kws` is quite difficult. `target` is the only expected (and required) argument. It can also be assigned to `self.target` in the function body instead of using `setattr` from `AstroidError`.

astroid/nodes/node_ng.py
cdce8p: The current code is fine. As you said, using a string as type annotation does prevent the circular dependency.
It's probably just that I'm not really a fan of the class name itself (`LocalsDictNodeNG`). Maybe that should be changed at some point 🤔

Anyway, let's leave it for now.

astroid/exceptions.py
DanielNoord: Something like this?
I did not actually set self.target. Passing the formatted message to the `super().__init__` call seemed sufficient, right?

astroid/exceptions.py
```diff
@@ -100,7 +100,7 @@ class TooManyLevelsError(AstroidImportError):
     def __init__(
         self,
         message="Relative import with too many levels " "({level}) for module {name!r}",
-        **kws
+        **kws,
```
DanielNoord, line 103: Not really sure why the pre-commit changed this?

astroid/exceptions.py
cdce8p: It got set here: https://github.com/PyCQA/astroid/blob/2d3f8ed45e7091df7968ab10d0de366431d6a00b/astroid/exceptions.py#L63-L67

I would recommend to assign `self.target` as it would make debugging a bit easier.

astroid/exceptions.py
```diff
@@ -256,6 +256,17 @@ class InferenceOverwriteError(AstroidError):
...
 
 
+class ParentMissingError(AstroidError):
+    """Raised when a node which is expected to have a parent attribute is missing one
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    def __init__(self, target):
```
cdce8p, line 266: Can you add type annotations here?

astroid/exceptions.py
cdce8p: Don't know either. Maybe a small bug with black?
Doesn't really matter though tbh.

DanielNoord:
Is this failing because of the recent update to `pre-commit`? 😕

cdce8p:
> Is this failing because of the recent update to `pre-commit`? 😕

No. `main` is fine. It seems to be something in this PR.

DanielNoord:
It has something to do with the import of `nodes`. Removing it fixes the issue.
Will investigate tomorrow!

cdce8p:
> It has something to do with the import of `nodes`. Removing it fixes the issue.
> Will investigate tomorrow!

I would suspect some kind of caching issue. Might not be that easy to track down.

astroid/exceptions.py
Pierre-Sassoulas: Black normalize the formatting, if there are multiple line it ends with a comma (and vice versa, add a comma and black will use multiline)

DanielNoord: Yeah, but nothing changed in the previous commit for `black` to suddenly want to add it (I think).

Pierre-Sassoulas: Ho, right. There's a change in the file. though. Maybe due to a new version of black tha thandle kwargs better and apply new formatting only if there is a change in the file ?

DanielNoord: That is indeed a very good theory!

DanielNoord:
I'm not sure if this solution works for you, but adding `disable` to the new error messages and removing on the `unused` places seems to solve the issue.
`no-member` has numerous false positives reported so I presumed this is likely a `pylint` issue rather than an issue with the actual code. Please correct if I'm wrong or if you would like another solution!

astroid/nodes/node_classes.py
```diff
@@ -61,6 +62,9 @@
 else:
     from typing_extensions import Literal
 
+if TYPE_CHECKING:
+    from astroid.scoped_nodes import LocalsDictNodeNG
```
Pierre-Sassoulas, line 66:
```suggestion
    from astroid.nodes import LocalsDictNodeNG
```
Or if really necessary:
```suggestion
    from astroid.nodes.scoped_nodes import LocalsDictNodeNG
```
Because ``astroid.scoped_nodes`` is deprecated and will be removed in 3.0

astroid/nodes/node_classes.py
```diff
@@ -2016,13 +2020,17 @@ def postinit(self, nodes: typing.List[NodeNG]) -> None:
         """
         self.nodes = nodes
 
-    def scope(self):
+    def scope(self) -> "LocalsDictNodeNG":
         """The first parent node defining a new scope.
 
         :returns: The first parent scope node.
         :rtype: Module or FunctionDef or ClassDef or Lambda or GenExpr
```
Pierre-Sassoulas, line 2027: Is a ``LocalsDictNodeNG`` the same as ``Module or FunctionDef or ClassDef or Lambda or GenExpr``? 
```suggestion
```

astroid/nodes/scoped_nodes.py
```diff
@@ -77,6 +77,8 @@
     {"classmethod", "staticmethod", "builtins.classmethod", "builtins.staticmethod"}
 )
 
+T = TypeVar("T")
```
Pierre-Sassoulas, line 80: I don't understand this part, what does ``T`` means ?

astroid/nodes/node_classes.py
DanielNoord: Yes! All classes which can considered `scopes` inherit from this base class!

astroid/nodes/node_classes.py
DanielNoord: `LocalsDictNodeNG` was not imported in `astroid.nodes`. I added it there and added an import from `astroid.nodes`.

astroid/nodes/scoped_nodes.py
DanielNoord: I don't know how it works under the hood. But for
```
def scope(self: T) -> T:
        """The first parent node defining a new scope.

        :returns: The first parent scope node.
        :rtype: Module or FunctionDef or ClassDef or Lambda or GenExpr
        """
        return self
```
It makes the function return with the correct typing. It sets the return type to the type of `self`.

astroid/nodes/node_classes.py
DanielNoord, line 2027:
```suggestion
        :rtype: LocalsDictNodeNG
```

astroid/nodes/scoped_nodes.py
Pierre-Sassoulas: Ok I've read the documentation and I think ``T = TypeVar("T")`` is the example for ``TypeVar`` from the python doc, but I guess we can change the name to something else ? Maybe ``Scope`` ?

astroid/nodes/node_classes.py
DanielNoord: I have added it to the docstring

Pierre-Sassoulas: I've never really checked but I guess sphinx should be able to create the rtype from the typing information in the signature ?

DanielNoord: I have no idea. I never really used `sphinx`. We might want to test this at some point and remove some of the redundant doc strings.

astroid/nodes/scoped_nodes.py
Pierre-Sassoulas: @cdce8p what do you think, isn't it better to have an explicit name here ?

Pierre-Sassoulas: Sorry for the discrepancies @DanielNoord I started to not review some MR that Marc is already reviewing because there is a lot to review :)

astroid/nodes/node_classes.py
Pierre-Sassoulas: Thank you for fixing the underlying problem, it will prevent problems during 3.0 migrations !

astroid/nodes/scoped_nodes.py
DanielNoord: No worries! With the typing PRs it never hurts to have an additional set of eyes.

cdce8p: It would be possible to rename the `TypeVar`, I just wouldn't recommend it in this case. `T` is the usual convention. Additionally, it doesn't have any constraints or bounds. Thus it could be reused in another function later.

That is, for example, different in the `TypeVar` we used in https://github.com/PyCQA/pylint/pull/4978.

astroid/nodes/node_classes.py
```diff
@@ -2016,13 +2020,17 @@ def postinit(self, nodes: typing.List[NodeNG]) -> None:
         """
         self.nodes = nodes
 
-    def scope(self):
+    def scope(self) -> "LocalsDictNodeNG":
         """The first parent node defining a new scope.
 
         :returns: The first parent scope node.
-        :rtype: Module or FunctionDef or ClassDef or Lambda or GenExpr
+        :rtype: LocalsDictNodeNG
```
cdce8p, line 2027: You could remove it all together. `Sphinx` should pick up the return type annotation. With each CI run, a documentation preview is also generated. You can check it here: https://pylint--1170.org.readthedocs.build/projects/astroid/en/1170/api/astroid.nodes.html#id174

astroid/nodes/node_ng.py
```diff
@@ -251,15 +269,15 @@ def frame(self):
         """
         return self.parent.frame()
 
-    def scope(self):
+    def scope(self) -> "LocalsDictNodeNG":
         """The first parent node defining a new scope.
```
cdce8p, line 273: I would name the possible scopes here

astroid/nodes/node_ng.py
```diff
@@ -251,15 +269,15 @@ def frame(self):
...
 
         :returns: The first parent scope node.
         :rtype: Module or FunctionDef or ClassDef or Lambda or GenExpr
```
cdce8p, line 276: This can then be removed

astroid/nodes/scoped_nodes.py
```diff
@@ -1606,6 +1608,7 @@ def blockstart_tolineno(self):
 
         :type: int
         """
+        # pylint: disable-next=no-member
```
cdce8p, line 1611: I'm a bit hesitant about these `pylint: disable` statements. Maybe it would be better to remove the annotations in `astroid/exceptions.py` for now.
They don't add much as the error is only used in two locations anyway.

astroid/exceptions.py
```diff
@@ -256,6 +261,18 @@ class InferenceOverwriteError(AstroidError):
     """
 
 
+class ParentMissingError(AstroidError):
```
cdce8p, line 264: Can you add it here as well, just so it's documented?
https://github.com/PyCQA/astroid/blob/main/doc/api/astroid.exceptions.rst

astroid/nodes/scoped_nodes.py
DanielNoord: I'm going to fix this in another PR. I'll tag you!

DanielNoord: See https://github.com/PyCQA/astroid/pull/1174

DanielNoord:
- [ ] Note to self after merging of #1175, the linkt to `LocalsDictNodeNG` needs to be changed in `doc/api/astroid.nodes.rst`

cdce8p:
> Note to self after merging of [Fix documentation after changes to astroid.nodes #1175](https://github.com/PyCQA/astroid/pull/1175), the linkt to `LocalsDictNodeNG` needs to be changed in `doc/api/astroid.nodes.rst`

You don't need to wait for it. That merge conflict should be easy to resolve.

astroid/nodes/scoped_nodes.py
cdce8p: I see now! These are actual bugs, only `Arguments` should be assigned to `self.args` not a `list`.
I was under the impression that these are `pylint` bugs.

In that case, these are ok.

astroid/nodes/node_classes.py
```diff
@@ -2016,13 +2020,17 @@ def postinit(self, nodes: typing.List[NodeNG]) -> None:
         """
         self.nodes = nodes
 
-    def scope(self):
+    def scope(self) -> "LocalsDictNodeNG":
         """The first parent node defining a new scope.
+        These can be  Module, FunctionDef, ClassDef, Lambda or GeneratorExp nodes
```
cdce8p, line 2025:
```suggestion
        These can be Module, FunctionDef, ClassDef, Lambda, or GeneratorExp nodes.
```

astroid/nodes/node_ng.py
```diff
@@ -251,15 +269,15 @@ def frame(self):
         """
         return self.parent.frame()
 
-    def scope(self):
+    def scope(self) -> "LocalsDictNodeNG":
         """The first parent node defining a new scope.
+        These can be  Module, FunctionDef, ClassDef, Lambda or GeneratorExp nodes
```
cdce8p, line 274:
```suggestion
        These can be Module, FunctionDef, ClassDef, Lambda, or GeneratorExp nodes.
```

cdce8p (approved):
Some final comments. Besides that, this LGTM!
Thanks @DanielNoord 🐬

Can be merged if @Pierre-Sassoulas is ok with using `T` for the generic `TypeVar`.

Pierre-Sassoulas (approved):

cdce8p:
@DanielNoord I've just merged the #1174 and #1175. Can you update this one so it can be merged as well? Let me know if I can help.

DanielNoord:
@cdce8p Should be good to go after these tests pass!

cdce8p (approved):
Thanks @DanielNoord 🚀

DanielNoord:
@cdce8p The check on this latest commit failed although `main` seems to be fine. Might be because I removed the branch mid-check, but thought I would notify you about this.

cdce8p:
@DanielNoord Thanks. Yeah, this isn't an issue. `main` is fine.

PR #1172, opened 2021-09-14
Feature / Bug Fix: Brain for the stdlib signal module

SupImDos:
## Steps

## Description

I've come across https://github.com/PyCQA/pylint/issues/2804 a few times in my projects, and as the fix appears to be relatively simple I'm hoping to help get it fixed upstream here in Astroid.

As per the referenced issue, the `signal` module generates the `Signals`, `Handlers` and `Sigmasks` `IntEnum`s dynamically using the `IntEnum._convert()` classmethod. This modifies the module globals and `Pylint` is unable to handle this type of code.

Without these hooks, the following are triggered by `Pylint`:
- E1101: Module 'signal' has no 'Signals' member (no-member)
- E1101: Module 'signal' has no 'Handlers' member (no-member)
- E1101: Module 'signal' has no 'Sigmasks' member (no-member)

In this PR I have added a `brain` for the stdlib `signal` module, which generates the ASTs for the `Signals`, `Handlers` and `Sigmasks` `IntEnum`s. This is done by simply importing the `signal` library and extracting the enums that are generated correctly at runtime. The existence of `Sigmasks` is checked first, as it appears it is only generated by the `signal` library if `pthread_sigmask()` is supported by the OS.

## Notes
I haven't added a `ChangeLog` entry yet, as was unsure what milestone this would be added to. Happy to add once advised if needed.

## Type of Changes

## Related Issue

Closes: https://github.com/PyCQA/pylint/issues/2804


hippo91 (changes requested):
@SupImDos by declaring `Signals = signal.Signals` you just add a node in the ast whose name is `Signals` but `astroid`cannot deduce anything about it. It is `uninferable`. You can show that with running the following snippet:
```
src = """
import signal
signal.Signals
"""
node=extract_node(src)
cls_node = node.inferred()[0]
print(cls_node)
```
It will return `Uninferable`.

I think the solution here would be something along the lines:
```
    return parse(
        """
        class Signals(enum.IntEnum):
            SIGABRT
            SIGFPE
            SIGKILL
            SIGRTMAX
            SIGTERM
            SIGURG
            SIGXCPU
            SIGALRM
            SIGHUP
            SIGPIPE
            SIGRTMIN
            SIGTRAP
            SIGUSR1
            SIGXFSZ
            SIGBUS
            SIGILL
            SIGPROF 
            SIGSEGV
            SIGTSTP
            SIGUSR2    
            SIGCHLD
            SIGINT
            SIGPWR
            SIGSTOP
            SIGTTIN
            SIGVTALRM  
            SIGCONT
            SIGIO
            SIGQUIT
            SIGSYS
            SIGTTOU
            SIGWINCH 
        """
    )
```
But with assigning the correct integer for each enum.

hippo91:
@SupImDos thanks a lot for this PR! Just a bit of work and it will be ok!

SupImDos:
@hippo91 Thank you for the review.

> ...by declaring `Signals = signal.Signals` you just add a node in the ast whose name is `Signals` but `astroid` cannot deduce anything about it. It is `uninferable`.

Ah, I understand. I was trying to be clever as I did not want to add the maintenance burden of keeping the Enums up to date.

Having another look at it now, its possible this issue might be more difficult than I first imagined. The `Signals`/`Handlers`/`Sigmasks` Enums appear to depend on the operating system, and if we are not somehow generating them dynamically then I'm not sure how we want to define them?

For example, on my personal Macbook (Mac OS 11.5.2), the Signals enum contains:
```
>>> import signal
>>> for s in signal.Signals:
>>>   print(repr(s))
 
<Signals.SIGHUP: 1>
<Signals.SIGINT: 2>
<Signals.SIGQUIT: 3>
<Signals.SIGILL: 4>
<Signals.SIGTRAP: 5>
<Signals.SIGABRT: 6>
<Signals.SIGEMT: 7>
<Signals.SIGFPE: 8>
<Signals.SIGKILL: 9>
<Signals.SIGBUS: 10>
<Signals.SIGSEGV: 11>
<Signals.SIGSYS: 12>
<Signals.SIGPIPE: 13>
<Signals.SIGALRM: 14>
<Signals.SIGTERM: 15>
<Signals.SIGURG: 16>
<Signals.SIGSTOP: 17>
<Signals.SIGTSTP: 18>
<Signals.SIGCONT: 19>
<Signals.SIGCHLD: 20>
<Signals.SIGTTIN: 21>
<Signals.SIGTTOU: 22>
<Signals.SIGIO: 23>
<Signals.SIGXCPU: 24>
<Signals.SIGXFSZ: 25>
<Signals.SIGVTALRM: 26>
<Signals.SIGPROF: 27>
<Signals.SIGWINCH: 28>
<Signals.SIGINFO: 29>
<Signals.SIGUSR1: 30>
<Signals.SIGUSR2: 31>
```

and on my Linux laptop (Ubuntu 20.04), the Signals enum contains:
```
>>> import signal
>>> for s in signal.Signals:
>>>   print(repr(s))

<Signals.SIGHUP: 1>
<Signals.SIGINT: 2>
<Signals.SIGQUIT: 3>
<Signals.SIGILL: 4>
<Signals.SIGTRAP: 5>
<Signals.SIGABRT: 6>
<Signals.SIGBUS: 7>
<Signals.SIGFPE: 8>
<Signals.SIGKILL: 9>
<Signals.SIGUSR1: 10>
<Signals.SIGSEGV: 11>
<Signals.SIGUSR2: 12>
<Signals.SIGPIPE: 13>
<Signals.SIGALRM: 14>
<Signals.SIGTERM: 15>
<Signals.SIGCHLD: 17>
<Signals.SIGCONT: 18>
<Signals.SIGSTOP: 19>
<Signals.SIGTSTP: 20>
<Signals.SIGTTIN: 21>
<Signals.SIGTTOU: 22>
<Signals.SIGURG: 23>
<Signals.SIGXCPU: 24>
<Signals.SIGXFSZ: 25>
<Signals.SIGVTALRM: 26>
<Signals.SIGPROF: 27>
<Signals.SIGWINCH: 28>
<Signals.SIGIO: 29>
<Signals.SIGPWR: 30>
<Signals.SIGSYS: 31>
<Signals.SIGRTMIN: 34>
<Signals.SIGRTMAX: 64>
```

Not sure what the best approach is, the options I can think of:
* Dynamically generate the AST/nodes for the Enums at runtime (somehow?)
* Statically define the Enums as suggested, and choose which members to include (include every signal? what values?)

SupImDos:
@hippo91 

What about something like this to generate the Enums dynamically?

```python3
import enum
import signal

def _signals_enum():
    """Generates the signal.Signals IntEnum."""
    # Generate enum source
    src = "class Signals(enum.IntEnum):\n"

    # Loop through signals and add to source
    for sig in signal.Signals:
        src += f"    {sig.name} = {sig.value}\n"

    # Return the source code for the enum
    return src
```

Pierre-Sassoulas:
(This would need an import guard so astroid do not crash if signal is not importable)

hippo91:
@SupImDos i do not agree with the idea of loading dynamically the signal module. IMHO `pylint` should stay, as much as it can, a __static__ linter. The only exception to this rule is for `C` modules when user explicitly allows `astroid` to load them through the `extension-pkg-whitelist` option.

To be honest, some years ago, i had the same idea to fight againt numerous `pylint` false positives messages dealing with `numpy` and @PCManticore explain me why it was not a good idea (#452).

SupImDos:
@hippo91 I understand and agree.

I guess the question is then, is when we define the `Signals(enum.IntEnum)` statically in the `signal` brain, which signals do we include, and what integer values do we assign to them?

Do we try to be clever and detect the architecture of the system? (I'd assume no, as I don't think there is any precedent for this in the current `astroid` codebase.)

This seems to be a good reference for the signals:
https://man7.org/linux/man-pages/man7/signal.7.html

Pierre-Sassoulas:
Maybe it would be acceptable to not get the real int value on all os when inferring ? My guess is that the most useful part of the brain is to fix the no-member false positive, and I don't think pylint will really use the actual int value. Maybe I'm missing something and this value is what we want ?

SupImDos:
> Maybe it would be acceptable to not get the real int value on all os when inferring ? My guess is that the most useful part of the brain is to fix the no-member false positive, and I don't think pylint will really use the actual int value. Maybe I'm missing something and this value is what we want ?

I tend to agree. If the values of the Signals enums are dynamic such that they can't be determined unless we dynamically load the module, and dynamically loading the module is not a viable option then maybe we have to just assign an automatic/meaningless value.

For example, some of the signals such as:
* `SIGRTMIN`
* `SIGRTMAX`

are very difficult to assign a value to, and seem to depend on the `glibc` version installed (at least from what I've skimmed over on the man `signal(7)` page).

If this is the case for those (and some other) signals, maybe the best thing to do is something like the following:

```python3
def _signals_enum():
    """Generates the source code for the Signals int enum."""
    return """
    class Signals(enum.IntEnum):
        SIGABRT = enum.auto()
        SIGEMT  = enum.auto()
        SIGFPE  = enum.auto()
        SIGILL  = enum.auto()
        SIGINFO = enum.auto()
        SIGINT  = enum.auto()
        SIGSEGV = enum.auto()
        SIGTERM = enum.auto()
        if sys.platform != "win32":
            SIGALRM   = enum.auto()
            SIGBUS    = enum.auto()
            SIGCHLD   = enum.auto()
            SIGCONT   = enum.auto()
            SIGHUP    = enum.auto()
            SIGIO     = enum.auto()
            SIGIOT    = enum.auto()
            SIGKILL   = enum.auto()
            SIGPIPE   = enum.auto()
            SIGPROF   = enum.auto()
            SIGQUIT   = enum.auto()
            SIGSTOP   = enum.auto()
            SIGSYS    = enum.auto()
            SIGTRAP   = enum.auto()
            SIGTSTP   = enum.auto()
            SIGTTIN   = enum.auto()
            SIGTTOU   = enum.auto()
            SIGURG    = enum.auto()
            SIGUSR1   = enum.auto()
            SIGUSR2   = enum.auto()
            SIGVTALRM = enum.auto()
            SIGWINCH  = enum.auto()
            SIGXCPU   = enum.auto()
            SIGXFSZ   = enum.auto()
        if sys.platform == "win32":
            SIGBREAK  = enum.auto()
        if sys.platform != "darwin" and sys.platform != "win32":
            SIGCLD    = enum.auto()
            SIGPOLL   = enum.auto()
            SIGPWR    = enum.auto()
            SIGRTMAX  = enum.auto()
            SIGRTMIN  = enum.auto()
    """
```

hippo91:
@Pierre-Sassoulas @SupImDos  i totally agree with your last comments. Let's do this! :thumb:

SupImDos:
A few outstanding questions for review:
1. Do we need to do an `import enum` statement somewhere in the code which is being parsed?
2. Do the Signals/Handlers/Sigmasks enums correctly line up with the Python `typeshed` definitions? (i.e. have I made any mistakes)
3. Do `sys.platform` checks work in the code which is being parsed, or does that need to be checked outside when we are generating the code?
4. Should we note somewhere (docstrings/comments) that the values of the enums are meaningless, or is this conveyed by the use of `enum.auto()`?
5. Changelog entry?
6. Unit tests?

SupImDos:
Did some testing and answered a few of my own questions in the commit above.

It seems I needed the `import enum` statement inside the generated code, and the `sys.platform == ...` checks outside of the generated code (i.e. assisting in building the generated code. This makes the code a bit less clean, but oh well.

Also made a note in the module docstring about the enum values.

Pierre-Sassoulas (commented):
This look really good already, thank you. Regarding unit test, we could add functional tests in pylint to check that there are no ``no-member`` warning anymore, but you could also check the ``unittest_brain_``'s type file in ``tests/``.'"

hippo91:
@SupImDos it seems the brain you propose is ok! Congrats!
As @Pierre-Sassoulas said, it deserve a unit test. IMHO it would be a better option to add it directly in `astroid` in order to be autonomous and not let an error appear by forgetting to run `pylint` tests.
@Pierre-Sassoulas what do you think about it?

SupImDos:
@hippo91
@Pierre-Sassoulas 

What do we think of the test I just added? I tried to follow the other `unittest_brain_xxx.py` files.

Also, the CI failure appears to be a minor formatting issue. Which approach would be better to fix it up?
```python3
if sys.platform not in ("darwin", "win32"):
```

or

```python3
if sys.platform != "darwin" and sys.platform != "win32":  # pylint: disable=consider-using-in
```

Pierre-Sassoulas:
What is the problem with `if sys.platform not in ("darwin", "win32"):` ? (It would be my preferred choice)

SupImDos:
> What is the problem with `if sys.platform not in ("darwin", "win32"):` ? (It would be my preferred choice)

I agree, it was originally copied from the Python `typeshed` stdlib signal.pyi module. Will make that change now.

Pierre-Sassoulas (approved):

PR #1173, opened 2021-09-15
Prefer last same-named function in a class rather than first in `igetattr()`

### LINKED ISSUE #1015: false not-an-iterable for class method with @overload (2.5.7 regression)

nelfin:
## Steps

## Description

Ref #1015. When there are multiple statements defining some attribute
`ClassDef.igetattr` filters out later definitions if they are not in the
same scope as the first (to support cases like "self.a = 1; self.a = 2"
or "self.items = []; self.items += 1"). However, it checks the scope of
the first attribute against the *parent scope* of the later attributes.
For mundane statements this makes no difference, but for
scope-introducing statements such as `FunctionDef` these are not the same.

## Type of Changes

## Related Issue

Closes #1015


tests/unittest_inference.py
```diff
@@ -4005,6 +4005,38 @@ class Test(Outer.Inner):
...
+                def foo(self):
+                    yield
+
+            A().foo()  #@
+        """
+        )
+        inferred = list(node.infer())
+        assert len(inferred) == 2
+        assert isinstance(inferred[0], nodes.Const)
+        assert isinstance(inferred[1], Generator)
```
cdce8p, line 4038: I'm wondering if
```py
class A:
    def foo(self): ...
    def foo(self):
        yield
```
should be inferred as `Const` and `Generator` as Python will always use the last definition.

--
Some test cases to be aware of
```py
if some_test:
    def func():
        return 42
else:
    def func():
        return bool

# should still return two results for `func()`
```

Functools.singledispatch should still work:
https://docs.python.org/3/library/functools.html#functools.singledispatch

nelfin: > Wouldn't it be better to only infer it as `Generator`?

My thinking leans the same way as it's the more "useful" definition.

> Python will always use the last definition

Pretty much, but I wonder what to do about monkeypatching, or this (contrived) example:

```python
class A:
    def foo(self):
        return 'first'
    bar = foo(object())
    def foo(self):
        return 'second'

A().foo()  # 'second'
A().bar  # 'first'
```

cdce8p: Just tested your example: This will already be inferred incorrectly. `A().foo() ->'first'`. Possibly another bug?

As for the issue at hand. Would it work to filter previous definitions only if no other attributes are in between?

Pierre-Sassoulas (commented):
The case with two ``foo`` function one redefined are a little contrived and probably not seen a lot in real "sane" code. The ``@overload`` use case though seems interesting to test. Do you think this is really equivalent ?

DanielNoord:
@nelfin Just going through some stale PRs. Have you given the questions raised here some thoughts? Perhaps somewhat related is my comment [here](https://github.com/PyCQA/pylint/issues/5264#issuecomment-995591983) which also relates to "picking the correct definition".

jacobtylerwalls:
We're closing tons of duplicates for this in pylint. I think special-casing functions (sans properties) is good enough.

Pierre-Sassoulas (approved):

Pierre-Sassoulas (approved):

PR #1174, opened 2021-09-15
Add `Arguments` typing to `Lambda` & `FunctionDef`

DanielNoord:
## Steps

## Description
This should fix some of the problems we were having with `args.???` in #1170.
See this comment https://github.com/PyCQA/astroid/pull/1170#discussion_r709067042
This should make it possible to remove the disable's added in that PR.

/CC @cdce8p 

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -69,7 +69,7 @@ from astroid.exceptions import (
 from astroid.interpreter.dunder_lookup import lookup
 from astroid.interpreter.objectmodel import ClassModel, FunctionModel, ModuleModel
 from astroid.manager import AstroidManager
-from astroid.nodes import Const, node_classes
+from astroid.nodes import Arguments, Const, node_classes
 
 ITER_METHODS = ("__iter__", "__getitem__")
 EXCEPTION_BASE_CLASSES = frozenset({"Exception", "BaseException"})
@@ -1190,7 +1190,6 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
         :returns: 'method' if this is a method, 'function' otherwise.
         :rtype: str
         """
-        # pylint: disable=no-member
         if self.args.arguments and self.args.arguments[0].name == "self":
             if isinstance(self.parent.scope(), ClassDef):
                 return "method"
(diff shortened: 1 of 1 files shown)


astroid/nodes/scoped_nodes.py
```diff
@@ -1214,11 +1213,8 @@ def __init__(self, lineno=None, col_offset=None, parent=None):
         :type: dict(str, NodeNG)
         """
 
-        self.args = []
-        """The arguments that the function takes.
-
-        :type: Arguments or list
-        """
+        self.args: Arguments
```
cdce8p, line 1216: At some point we should probably enforce that `postinit` is actually called. Otherwise `self.args` might not be set when accessed. Task for another time. (It's somewhere on my long list.)

cdce8p (approved):
Looks good overall. I'm going to merge that first. You should be able to remove the `pylint: disable` calls from #1170 then.

astroid/nodes/scoped_nodes.py
DanielNoord: This is already note the case.. Of the two places creating a `FunctioDef` only one does this. Oh well... 😄

cdce8p: Which one do you mean. `raw_building`, `objectmodel`, and `rebuilder` are fine. Am I missing something? I didn't notice any errors while running the tests either.

DanielNoord: https://github.com/PyCQA/astroid/blob/9a7878a78a284571d3788afc42203089185cfcbf/astroid/raw_building.py#L127-L150

The `postinit` is called on `Arguments`, but the `FunctionDef.postinit` method is never called.

cdce8p: At least `func.args` is still assigned. So it won't break (just yet).

DanielNoord: Something to fix in the future I guess. Seems like the `Lambda` and `FunctionDef` class could use a refactor anyway.

cdce8p: Someday... Too many other things open at the moment 😪

Pierre-Sassoulas: > Someday... Too many other things open at the moment 

Yeah let's finish up 2.11 first and the typing MR first. The code is getting better little by little and that's great, but let's merge what we did already. There are a lot of changes in those, and conflicts could become a problem.

DanielNoord: I expect to have #4980 ready for review in 5-10 minutes!

PR #1175, opened 2021-09-15
Fix documentation after changes to astroid.nodes

cdce8p:
## Description

With the recent changes to the `astroid.nodes` module, most `nodes` get printed twice. Furthermore some references have become outdated.

Current: https://pylint.pycqa.org/projects/astroid/en/v2.8.0/api/astroid.nodes.html
The site should start with a list of all nodes as seen here: https://pylint.pycqa.org/projects/astroid/en/v2.5.8/api/astroid.nodes.html

## Type of Changes


Pierre-Sassoulas (approved):
Nice catch !

PR #1180, opened 2021-09-17
Add f-strings with `flynt`

DanielNoord:
## Steps

## Description

Sorry for the size of this PR but it should be fairly easy to review. It's all changes of strings > f-strings.
This is all done with [flynt](https://github.com/ikamensh/flynt) which I also used to do the same for `pylint`. I checked every change and did not see any problems.
There are some possible f-strings in the codebase which didn't get picked up by `flynt` which I will tackle after this has been merged.

## Type of Changes

### DIFF
--- a/tests/unittest_regrtest.py
+++ b/tests/unittest_regrtest.py
@@ -219,16 +219,15 @@ def test():
 
         # In a regular file, "coding: utf-8" would have been used.
         node = extract_node(
-            """
+            f"""
         from __future__ import unicode_literals
 
         class MyClass(object):
             def method(self):
-                "With unicode : %s "
+                "With unicode : {'’'} "
 
         instance = MyClass()
         """
-            % "\u2019"
         )
 
         next(node.value.infer()).as_string()
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -575,12 +575,10 @@ class Generator(BaseInstance):
         return True
 
     def __repr__(self):
-        return "<Generator({}) l.{} at 0x{}>".format(
-            self._proxied.name, self.lineno, id(self)
-        )
+        return f"<Generator({self._proxied.name}) l.{self.lineno} at 0x{id(self)}>"
 
     def __str__(self):
-        return "Generator(%s)" % self._proxied.name
+        return f"Generator({self._proxied.name})"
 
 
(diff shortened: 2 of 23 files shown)


Pierre-Sassoulas:
Well we might want to advise user to use flynt instead of pyupgrade then, because pyupgrade is cautious and do not upgrade everything. (https://github.com/PyCQA/pylint/blob/main/doc/whatsnew/2.11.rst#summary----release-highlights)

DanielNoord:
I'm actually not sure what the best option is. This was the first I found on StackOverflow and which also focuses on this specific task. We can suggest both?

tests/unittest_regrtest.py
```diff
@@ -219,16 +219,15 @@ def test_unicode_in_docstring(self) -> None:
...
         from __future__ import unicode_literals
 
         class MyClass(object):
             def method(self):
-                "With unicode : %s "
+                "With unicode : {'’'} "
 
         instance = MyClass()
         """
-            % "\u2019"
```
Pierre-Sassoulas, line 231: 👌

Pierre-Sassoulas (changes requested):
Great, I did not know about flynt, I'm glad you made me discover that one 👍 We can remove the disable in pylintrc and this is good to go :)

DanielNoord:
@Pierre-Sassoulas Can you test this with `pylint` before merging? My benchmarks are indicating that `pylint` becomes slower with this branch of `astroid`. Could be my computer doing too much at once, but I ran it twice and it was slower on both occasions.

Pierre-Sassoulas:
On my side it's ~=0,3% faster with f-strings, It's close to what I believed would happen.

DanielNoord:
Probably my computer needing weekend then.. 😄

DanielNoord:
Btw, do you normally `squash and merge` or `rebase and merge` when committing PRs?

Pierre-Sassoulas:
Depends on the commits in the merge requests, if there's a pre-commit auto-fix commit or a merge update from master I'll squash, if the commit are clean and well organized (or too big to be squashed) I'll rebase .  Generally someone who does well organized commit will rebase themself so I don't often have to ask for a rebase.

PR #1181, opened 2021-09-17
Fix regression on ``ClassDef`` inference

DanielNoord:
## Steps

## Type of Changes

## Related Issue

Closes PyCQA/pylint#5030
Closes PyCQA/pylint#5036

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -178,7 +178,8 @@ class AsStringVisitor:
         args = [n.accept(self) for n in node.bases]
         if node._metaclass and not node.has_metaclass_hack():
             args.append("metaclass=" + node._metaclass.accept(self))
-        args += [n.accept(self) for n in node.keywords]
+        if node.keywords:
+            args += [n.accept(self) for n in node.keywords]
         args = f"({', '.join(args)})" if args else ""
         docs = self._docs_dedent(node.doc) if node.doc else ""
         return "\n\n{}class {}{}:{}\n{}\n".format(
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -178,8 +178,7 @@ class AsStringVisitor:
         args = [n.accept(self) for n in node.bases]
         if node._metaclass and not node.has_metaclass_hack():
             args.append("metaclass=" + node._metaclass.accept(self))
-        if node.keywords:
-            args += [n.accept(self) for n in node.keywords]
+        args += [n.accept(self) for n in node.keywords]
         args = f"({', '.join(args)})" if args else ""
         docs = self._docs_dedent(node.doc) if node.doc else ""
         return "\n\n{}class {}{}:{}\n{}\n".format(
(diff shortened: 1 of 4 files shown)


DanielNoord:
I'm not sure about the location of the unit test. Might need to be placed in a different place, but I did not know where.

astroid/nodes/as_string.py
```diff
@@ -178,7 +178,8 @@ def visit_classdef(self, node):
         args = [n.accept(self) for n in node.bases]
         if node._metaclass and not node.has_metaclass_hack():
             args.append("metaclass=" + node._metaclass.accept(self))
-        args += [n.accept(self) for n in node.keywords]
+        if node.keywords:
+            args += [n.accept(self) for n in node.keywords]
```
cdce8p, line 182: `node.keywords` should never be `None` in the first place. It should always be a list, even if empty.
This should be fixed in  `ClassDef.postinit` with
```py
if keywords is not None:
    self.keywords = keywords
```
https://github.com/PyCQA/astroid/blob/28411ee8575562b7d7d1d0639d4e61f91ed1586b/astroid/nodes/scoped_nodes.py#L2099

--

```c
ClassDef(identifier name,
             expr* bases,
             keyword* keywords,
             stmt* body,
             expr* decorator_list)
```
https://docs.python.org/3.10/library/ast.html

cdce8p (changes requested):

astroid/nodes/as_string.py
DanielNoord: Before merging this, do we want to identify why this regressed? Or is fixing the `postinit` good enough in your opinion?

cdce8p: The fix is good enough for me. Did some digging nevertheless.

That was an existing issue that only with the last update started causing problems. #979 added inference for `Compare` nodes. The inference method uses a clever workaround to get to the actual value of a node.
https://github.com/PyCQA/astroid/blob/28411ee8575562b7d7d1d0639d4e61f91ed1586b/astroid/inference.py#L811-L814

Calling `as_string` on any `NodeNG` should be safe. Multiple `ClassDef.postinit` calls in `astroid/brain/brain_typing.py` don't pass a `keywords` argument and thus use the default `None`.
https://github.com/PyCQA/astroid/blob/28411ee8575562b7d7d1d0639d4e61f91ed1586b/astroid/brain/brain_typing.py#L350-L354

I wouldn't add `keywords=[]` to these calls as a default value is expected to work fine. Thus this was an error in the `postinit` method which is now fixed.

cdce8p (approved):
LGTM!

astroid/nodes/as_string.py
DanielNoord: Thanks for the deep-dive! Feels good to know that we are not using a work-around to fix a problem but rather a fixing an actual problem, which was what I was afraid of!

cdce8p: I agree

PR #1182, opened 2021-09-17
Small improvements to Compare infer

cdce8p:
## Description

Followup to https://github.com/PyCQA/astroid/pull/979#pullrequestreview-756114803
Small improvements, mainly typing, for `Compare` inference.

/CC: @nelfin

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -802,7 +802,10 @@ COMPARE_OPS: Dict[str, Callable[[Any, Any], bool]] = {
     "in": lambda a, b: a in b,
     "not in": lambda a, b: a not in b,
 }
-UNINFERABLE_OPS = frozenset(("is", "is not"))
+UNINFERABLE_OPS = {
+    "is",
+    "is not",
+}
 
 
 def _to_literal(node: nodes.NodeNG) -> Any:
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -31,7 +31,7 @@ import ast
 import functools
 import itertools
 import operator
-from typing import Any, Iterable
+from typing import Any, Callable, Dict, Iterable, Optional
 
 import wrapt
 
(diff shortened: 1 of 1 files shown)


astroid/inference.py
```diff
@@ -792,7 +792,7 @@ def infer_binop(self, context=None):
 nodes.BinOp._infer_binop = _infer_binop
 nodes.BinOp._infer = infer_binop
 
-COMPARE_OPS = {
+COMPARE_OPS: Dict[str, Callable[[Any, Any], bool]] = {
```
nelfin, line 795: I like this

astroid/inference.py
```diff
@@ -857,7 +854,9 @@ def _do_compare(
     return retval  # it was all the same value
 
 
-def _infer_compare(self: nodes.Compare, context: InferenceContext) -> Any:
+def _infer_compare(
+    self: nodes.Compare, context: Optional[InferenceContext] = None
+) -> Any:
```
nelfin, line 859: Ditto, I like this

astroid/inference.py
```diff
@@ -802,10 +802,7 @@ def infer_binop(self, context=None):
     "in": lambda a, b: a in b,
     "not in": lambda a, b: a not in b,
 }
-UNINFERABLE_OPS = {
-    "is",
-    "is not",
-}
+UNINFERABLE_OPS = frozenset(("is", "is not"))
```
nelfin, line 805: I'm not sold on this. There's no measurable performance improvement, readability is arguably worse off. In my opinion, use of frozenset should be limited to the cases where the intended use of the value is a dictionary key.

nelfin (changes requested):

astroid/inference.py
cdce8p: Using `frozenset` instead of `set` was suggested to me a while ago for a similar case in `pylint`.
After your comment, I did look up the differences and you do have a point here. I'll revert that change.

nelfin (approved):

PR #1183, opened 2021-09-18
Bug pylint 4326

hippo91:
## Steps

## Description
This PR adds type hints support inside `numpy`'s brains. This way subscripting of classes that inherit from `generic` or `ndarray` is possible.

## Type of Changes

## Related Issue
Closes PyCQA/pylint#4326

### DIFF
--- a/astroid/brain/brain_numpy_utils.py
+++ b/astroid/brain/brain_numpy_utils.py
@@ -9,12 +9,12 @@
 
 """Different utilities for the numpy brains"""
 from typing import Tuple
+
 from astroid.builder import extract_node
 from astroid.nodes.node_classes import Attribute, Import, Name, NodeNG
 
-
 # Class subscript is available in numpy starting with version 1.20.0
-NUMPY_VERSION_TYPE_HINTS_SUPPORT = ('1', '20', '0')
+NUMPY_VERSION_TYPE_HINTS_SUPPORT = ("1", "20", "0")
 
 
 def numpy_supports_type_hints() -> bool:
@@ -32,7 +32,8 @@ def _get_numpy_version() -> Tuple[str, str, str]:
     """
     try:
         import numpy  # pylint: disable=unused-import
-        return tuple(numpy.version.version.split('.'))
+
+        return tuple(numpy.version.version.split("."))
     except ImportError:
         return None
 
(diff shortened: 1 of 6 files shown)


tests/unittest_brain_numpy_core_numerictypes.py
```diff
@@ -341,6 +345,59 @@ def test_datetime_astype_return(self):
             ),
         )
 
+    @unittest.skipUnless(
```
Pierre-Sassoulas, line 348: I think we can add a small test to check that we're not crashing if numpy is'nt installed.

astroid/brain/brain_numpy_utils.py
```diff
@@ -8,9 +8,35 @@
...
+    """
+    Return the numpy version number if numpy can be imported. Otherwise returns
+    None
+    """
+    try:
+        import numpy  # pylint: disable=import-outside-toplevel
+
+        return tuple(numpy.version.version.split("."))
+    except ImportError:
+        return None
```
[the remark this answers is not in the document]
hippo91: Absolutely right!

tests/unittest_brain_numpy_core_numerictypes.py
hippo91: @Pierre-Sassoulas i'm sorry but i don't see the point here. Can you explain a bit what you want to prevent?

Pierre-Sassoulas: I think it would have detected the problem of the other remark (TypeError: '>' not supported between instances of 'NoneType' and 'tuple'), by having automated tests for it.

hippo91: @Pierre-Sassoulas i was thinking to add something alone the following lines:
```
@unittest.skipIf(HAS_NUMPY)
class NumpyBrainUtilsTest(unittest.TestCase):
    def test_get_numpy_version_do_not_crash(self):
        """
        Test that the function _get_numpy_version doesn't crash even if numpy is not installed
        """
        self.assertEqual(_get_numpy_version(), ('0', '0', '0'))
```
Is it what you were thinking of? Or do you had in mind something more generic?

Pierre-Sassoulas: Looks good, but we could even check what we get when parsing the code without numpy (that would execute the ``_get_numpy_version`` too, but also much more code).

PR #1186, opened 2021-09-21
Fix ``relative-beyond-top-level`` false positive

DanielNoord:
## Steps

## Description

I believe this solves the false positive as discussed in the `pylint` issue. However, I'm not 100% comfortable with namespace packages so please tell me if I did anything wrong here.

I have not added a test yet, but I have openend a `pylint` PR which includes a test that does pass.
See: https://github.com/PyCQA/pylint/pull/5059

## Type of Changes

## Related Issue

https://github.com/PyCQA/pylint/issues/2967

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,6 +32,10 @@ Release date: TBA
 
 * Improve brain for ``typing.Callable`` and ``typing.Type``.
 
+* Fix bug with importing namespace packages with relative imports
+
+  Closes PyCQA/pylint#5059
+
 
 What's New in astroid 2.8.0?
 ============================
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -43,6 +43,7 @@ Lambda, GeneratorExp, DictComp and SetComp to some extent).
 import builtins
 import io
 import itertools
+import os
 import typing
 from typing import List, Optional, TypeVar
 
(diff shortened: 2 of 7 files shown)


Pierre-Sassoulas (changes requested):
I think we need to add a test in astroid too. Maybe we can use the same one than in pylint ?

DanielNoord:
@Pierre-Sassoulas Would you mind taking a look at this? I have added the necessary files to create the same test as in `pylint` but I don't really understand how the unit tests with data work in `astroid`.

Pierre-Sassoulas:
> I don't really understand how the unit tests with data work in astroid.

There's no handy automated functional tests like in pylint :)  I think you have to parse them manually, there's probably example of reasing data in the testdata directory Maybe we could imagine adding some helper like in pylint but we'd have to create a text representation of the expected AST (?) This would be helpful but probably quite a lot of work too.

DanielNoord:
I used a unittest in `pylint` as well, but it was much easier to understand how I could load extra data.
Searching for some of the directories I found in `testdata` I did not find any hits within the tests. So I couldn't find any examples I could copy.

Pierre-Sassoulas:
@hippo91 how do you handle test using real file in ``testdata`` generally ?

hippo91:
@Pierre-Sassoulas @DanielNoord i never had to use `testdata`. However it seems that one way is to use the `resources` module. 
I have to look a bit deeper to have an idea of how to use it for this specific PR.

DanielNoord:
Sorry for the rebase. I have added a test based on some of the others I saw using the `resources` module.

Pierre-Sassoulas (approved):

cdce8p:
This PR sees to cause a regression with the `too-many-function-args` warning for Home Assistant.
I'll need some more time to isolate the issue before posting code to reproduce it.

cdce8p:
Opened #1200 with an example to reproduce the error.

PR #839, opened 2020-09-07
Extend attrs brain to support provisional APIs

tmr232:
## Description

With attrs 20, new "provisional" APIs were added (see https://www.attrs.org/en/stable/api.html?highlight=field#provisional-apis).

This PR adds support to those new functions in the existing attrs brain.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |

PR #1187, opened 2021-09-21
continuing #839: Extend attrs brain to support provisional APIs

jstriebel:
## Description

This PR continues the original PR #839, which seems to be stale. Only thing missing are formatting and obeying the linter.
From the original PR:
> With attrs 20, new "provisional" APIs were added (see https://www.attrs.org/en/stable/api.html?highlight=field#provisional-apis).
>
> This PR adds support to those new functions in the existing attrs brain.

## Type of Changes

## Related Issue
* related to #656


Pierre-Sassoulas (approved):
The coverage is expected to fail don't worry about it coverall is down right now.

jstriebel:
@Pierre-Sassoulas Thanks for the fast review! I adapted the formatting, CI is green now, but needs an additional OK from a maintainer.

ChangeLog
```diff
@@ -6,7 +6,7 @@ What's New in astroid 2.9.0?
 ============================
 Release date: TBA
 
-
+* Extended attrs brain to support the provisional APIs
```
jstriebel, line 9: I'm not sure if this is supposed to go in 2.9, I think it could easily be marked as a patch for semver. I'll move it to 2.8.1 for now.

jstriebel:
@Pierre-Sassoulas Is there anything blocking here? Or could you merge this PR? Thank you!

Pierre-Sassoulas:
Hey @hippo91 could you review this, please ? I'm not confident enough to merge astroid brains change :)

astroid/brain/brain_attrs.py
```diff
@@ -10,8 +10,18 @@
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib"))
-ATTRS_NAMES = frozenset(("attr.s", "attrs", "attr.attrs", "attr.attributes"))
+ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib", "attr.field", "field"))
```
hippo91, line 13: I do not understand why there is "field" alone? 
By the way i wonder about the existence of `attrib` alone also.

hippo91 (approved):
@jstriebel thanks for this PR. Everything looks ok. As i understand the new attrs API, `define`, `mutable` and `frozen` are in fact `attr.s` with specialization of the default arguments.
`attrs.field` is an alias of `attrs.ib`. So the modifications proposed here are ok for me even if i wonder about the utility of `field` and `attrib` names without `attrs` prefix.

astroid/brain/brain_attrs.py
jstriebel: I guess this would help in cases where they are imported as `from attr import attrib` or `from attr import field`. Since this is only used to detect the attributes after an attrs-class is identified, I think the risk of a name collision should be negligible.

jstriebel:
> As i understand the new attrs API, define, mutable and frozen are in fact attr.s with specialization of the default arguments.

Exactly, this only adds some new decorator- and function-names that were added with the next-generation API (indeed with different defaults than before), the main mechanisms stay exactly the same.

astroid/brain/brain_attrs.py
hippo91: Ok, thanks, that's what i thought too. I also think that there is no risk of name collision.

jstriebel:
@Pierre-Sassoulas @hippo91 Is there anything missing to merge this? Or do you track this via the milestones and merge it later? Sorry to be bothering :pray:

Pierre-Sassoulas:
Thank you for the remainder, let's merge that in 2.8.1 as it's finished earlier than I thought :)

PR #1188, opened 2021-09-21
Fix regression from cache changes

DanielNoord:
## Steps

## Description

I'm still investigating why exactly this works but in my local environment this changes fixes the hang reported in https://github.com/PyCQA/pylint/issues/4972 for the https://github.com/open-telemetry/opentelemetry-python-contrib/ repository.

This commit that introduced this regression is https://github.com/PyCQA/astroid/commit/cf6528cbc158097c4903f0cab68242ff14bb591b#

## Type of Changes

## Related Issue

https://github.com/PyCQA/pylint/issues/4972

### DIFF
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -25,7 +25,7 @@ _INFERENCE_CACHE = {}
 
 
 def _invalidate_cache():
-    _INFERENCE_CACHE = {}
+    pass
 
 
 class InferenceContext:
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -25,7 +25,7 @@ _INFERENCE_CACHE = {}
 
 
 def _invalidate_cache():
-    _INFERENCE_CACHE.clear()
+    _INFERENCE_CACHE = {}
 
 
 class InferenceContext:


DanielNoord:
@nelfin Since you wrote the commit I'm relying on your expertise here. Do you think it might be possible that there is perhaps some sort of race condition where `.clear()` is not quick enough? Or might it be that the global value `_INFERENCE_CACHE` is "lost" at some point?
When I reassigned `_INFERENCE_CACHE` as `{}` the tests passed locally and the bug mentioned in the PR description was fixed. However, apparently `pre-commit` doesn't like that. I'm not sure what the code is exactly trying to achieve, but it seems some bug was introduced with the aforementioned commit.

Do you think there is a better way to make `_INFERENCE_CACHE` empty again? Let me know if you need any more details!

nelfin:
Hey @DanielNoord, could there be a race condition? I don't know enough about pylint internals to answer that at first glance, but my assumption is between the GIL and the fact that it uses `multiprocessing` I would say no.

Not invalidating the inference cache would actually speed up astroid, but it produces incorrect results as the test runs show (note: I should add a specific regression test for this). The gist of it is that transforms (e.g. `astroid.brain`) won't be applied correctly.

Does `pylint instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask` from https://github.com/PyCQA/pylint/issues/4972#issuecomment-924277980 *reliably* hang? And does it hang with `pylintrc:MASTER.jobs=1`? If so, then it sounds more like some kind of non-linear performance issue: some inference chain causing quadratic or worse performance due to things like super wide trees or deep call chains like we saw with pandas and numpy. If it happens intermittently, maybe there's an ordering dependency where inferring one file causes issues with inferring some other file.

DanielNoord:
> Not invalidating the inference cache would actually speed up astroid, but it produces incorrect results as the test runs show (note: I should add a specific regression test for this). The gist of it is that transforms (e.g. `astroid.brain`) won't be applied correctly.

Doesn't assigning a new `dict` to the cache also invalidate it? (see first commit of this pr) `pre-commit` doesn't like that, but that is was did pass the tests for me. Doesn't `a = dict().clear() == a = {}`?

> Does `pylint instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask` from [PyCQA/pylint#4972 (comment)](https://github.com/PyCQA/pylint/issues/4972#issuecomment-924277980) _reliably_ hang? And does it hang with `pylintrc:MASTER.jobs=1`? If so, then it sounds more like some kind of non-linear performance issue: some inference chain causing quadratic or worse performance due to things like super wide trees or deep call chains like we saw with pandas and numpy. If it happens intermittently, maybe there's an ordering dependency where inferring one file causes issues with inferring some other file.

It does hang reliably. After running their set-up script as discussed in that issue `pylint` never finishes correctly. I noticed it had something to do with this commit because a new `InferenceContext` kept being created with `nodes_inferred` being `None`. There seemed to be some sort of infinite loop. I don't know their codebase by heart, but I don't immediately see why the `flask` sub-directory creates this issue and many of the others don't. In set-up of directory and depth they seem familiar.

nelfin:
> Doesn't assigning a new `dict` to the cache also invalidate it? (see first commit of this pr) `pre-commit` doesn't like that, but that did pass the tests for me. Doesn't `a = dict().clear() == a = {}`?

Not quite, since they will be different objects: `.clear()` mutates the original dictionary. Also, the tests may pass for you locally, but they don't pass in CI, see: https://github.com/PyCQA/astroid/pull/1188/checks?sha=114a15e1de965b5e60eb9a7d71d21b4cfa10bddf and https://github.com/PyCQA/astroid/runs/3667423622 for the runs on the first commit in the PR. The behaviour is unchanged when you switch to just `pass` since without adding `global _INFERENCE_CACHE` to the function body, the result of your change is to create a new name `_INFERENCE_CACHE` in the local scope, hence the warning from pylint about the shadowing and the unused variable. See https://docs.python.org/3/reference/executionmodel.html#resolution-of-names for reference, specifically "if a name binding operation [e.g. an assignment] occurs anywhere within a code block, all uses of the name within the block are treated as references to the current block".

> It does hang reliably. After running their set-up script as discussed in that issue `pylint` never finishes correctly.

That is interesting, maybe it has something to do with their contrib/core code trees and editable installs but I don't know yet. After some "fun" I just now got a new venv set up and their tests running so I'll see if I can reproduce that and dive into what's causing the hang.

nelfin:
Couldn't get a profiler running properly, but a little SIGINT/poor man's sampling profiler later it looks like 99% of the time is spent in `ancestors` or specifically https://github.com/PyCQA/pylint/blob/1e675abcc2aa931421d7ce300908e734a93fd790/pylint/checkers/design_analysis.py#L270. If I disable the design-analysis checkers via `--disable=too-many-ancestors,too-many-instance-attributes,too-few-public-methods,too-many-public-methods` then a pylint run on the `opentelemetry-instrumentation-flask` folder runs in a few seconds.

DanielNoord:
@nelfin Sorry for dragging you into this. It seems I erroneously did so. Thanks very much for still doing this thorough analysis!

I looked at the function you pointed to and what is happening within the `opentelemetry` repo. I think the problem is that whenever a `parent` is already analysed previously its `ancestors` still get added to the `to_explore` list. I think in certain inheritance structures with sub- or mixin-classes this might create an infinite loop?

```python
parents: Set[nodes.ClassDef] = set()
    to_explore = cast(List[nodes.ClassDef], list(node.ancestors(recurs=False)))
    explored_parents: Set[nodes.ClassDef] = set()
    while to_explore:
        parent = to_explore.pop()
        if parent.qname() in ignored_parents:
            continue
        parents.add(parent)
        if parent not in explored_parents:
            explored_parents.add(parent)
            to_explore.extend(parent.ancestors(recurs=False))
    return parents
```
Keeping track of already explored (and added ancestors) like this made `pylint` work again on that repository. Does this fit your analysis?

nelfin:
> Sorry for dragging you into this. It seems I erroneously did so.

Don't be sorry, feel free to ping me any time.

> Keeping track of already explored (and added ancestors) like this made `pylint` work again on that repository. Does this fit your analysis?

Your fix is almost identical to the one that I've just been working on, just a slightly different approach:

```python
        if parent not in parents:
            yield parent
            parents.add(parent)
            to_explore.extend(parent.ancestors(recurs=False))
```

DanielNoord:
I am wondering if we truly got to the bottom of the issue though.
```python
    while to_explore:
        parent = to_explore.pop()
        print(parent)
        print(list(parent.ancestors()))
```
gives the following output repeated indefinitely:
```python
ClassDef._InstrumentedFlask(name='_InstrumentedFlask',
                            doc=None,
                            decorators=None,
                            bases=[<Attribute.Flask l.194 at 0x10a5f6460>],
                            keywords=[],
                            body=[ <Assign l.196 at 0x10a5f62b0>,
                              <Assign l.197 at 0x10a5f6700>,
                              <Assign l.198 at 0x10a5f6790>,
                              <Assign l.199 at 0x10a5f6820>,
                              <FunctionDef.__init__ l.201 at 0x10a5f68b0>])
[<ClassDef.Flask l.102 at 0x107819700>, <ClassDef.Scaffold l.61 at 0x10795e6d0>, <ClassDef.object l.0 at 0x1067ae610>, <ClassDef._InstrumentedFlask l.194 at 0x107830730>]
ClassDef._InstrumentedFlask(name='_InstrumentedFlask',
                            doc=None,
                            decorators=None,
                            bases=[<Attribute.Flask l.194 at 0x10784f850>],
                            keywords=[],
                            body=[ <Assign l.196 at 0x10784fbb0>,
                              <Assign l.197 at 0x10784fca0>,
                              <Assign l.198 at 0x10784fd30>,
                              <Assign l.199 at 0x10784fdc0>,
                              <FunctionDef.__init__ l.201 at 0x10784fe50>])
[<ClassDef.Flask l.102 at 0x107819700>, <ClassDef.Scaffold l.61 at 0x10795e6d0>, <ClassDef.object l.0 at 0x1067ae610>, <ClassDef._InstrumentedFlask l.194 at 0x10a5ca6d0>]
ClassDef._InstrumentedFlask(name='_InstrumentedFlask',
                            doc=None,
                            decorators=None,
                            bases=[<Attribute.Flask l.194 at 0x10a5f6460>],
                            keywords=[],
                            body=[ <Assign l.196 at 0x10a5f62b0>,
                              <Assign l.197 at 0x10a5f6700>,
                              <Assign l.198 at 0x10a5f6790>,
                              <Assign l.199 at 0x10a5f6820>,
                              <FunctionDef.__init__ l.201 at 0x10a5f68b0>])
[<ClassDef.Flask l.102 at 0x107819700>, <ClassDef.Scaffold l.61 at 0x10795e6d0>, <ClassDef.object l.0 at 0x1067ae610>, <ClassDef._InstrumentedFlask l.194 at 0x107830730>]
```

Going into the `flask` library I can follow the ancestors being `Flask`, `Scaffold` and `object` but I don't understand where `_InstrumentedFlask` comes from? Is `ancestors()` supposed to also return the own class?

nelfin:
`_InstrumentedFlask` comes from opentelemetry: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/14b0a3fdafa5f38b1558a4019280ce3ed90c1d24/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L194.

> Is ancestors() supposed to also return the own class?

I don't think so.

DanielNoord:
Update:

https://github.com/PyCQA/astroid/blob/4ffdf1108a6084b85e282b5835427c1d747bb22c/astroid/nodes/scoped_nodes.py#L2344-L2359

At some point, `if self.name == "_InstrumentedFlask" and self.bases == [<Attribute.Flask l.194 at 0x104fe6910>]` 
line 2354 `stmt.infer(context)` starts returning:
`[<ClassDef.Flask l.102 at 0x104fbadc0>, <ClassDef._InstrumentedFlask l.194 at 0x104fd05b0>, Uninferable, <ClassDef._InstrumentedFlask l.194 at 0x107c174c0>]` 
This creates a loop as then later on `.ancestors()` will be called on `_InstrumentedFlask` again.

So, it might be something to do with the `context` being passed to `infer()` or `infer()` incorrectly returning the own class.

nelfin:
I think it's this: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/14b0a3fdafa5f38b1558a4019280ce3ed90c1d24/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L257

nelfin:
Anyway, good work. I think it's probably safe to close this PR. Are you planning on submitting a fix to pylint for the cycle check or would you like me to push it through?

DanielNoord:
You can make the PR (if you want to) as you did most of the base investigative work. Although I think it will be good to include a good write-up of this bug for the maintainers. I'm not sure if stopping the cycle there is the right way forward as I'm wondering if it shouldn't go somewhere in `ancestors()` or in `NodeNG.infer()`.

nelfin:
> I'm not sure if stopping the cycle there is the right way forward as I'm wondering if it shouldn't go somewhere in ancestors() or in NodeNG.infer().

There's a long-standing FIXME about cycles in `ancestors()`, we should probably raise an issue to track it. What I see is that it's not the *same* ClassDef that is inferred so it's pretty tricky to sort it out (note object IDs):

```python
>>> baseobj
<ClassDef._InstrumentedFlask l.194 at 0x7fdb57f04d30>
>>> yielded
{<ClassDef._InstrumentedFlask l.194 at 0x7fdb64b7edf0>}
>>> baseobj in yielded
False
```

DanielNoord:
Ah, looking at the age of the `fixme` I think closing the cycle in `_get_parents()` is probably okay 😅

Pierre-Sassoulas:
> it's not the same ClassDef  that is inferred

Could it be fixed by defining ``__eq__`` and ``__hash__`` based on the attribute of ``ClassDef`` ? It makes sense for me that a class definition is equal to another instance if it come from the same file at the same line.

DanielNoord:
Sounds like a good idea, but we would probably also want this for `FuncDef`?

Pierre-Sassoulas:
Maybe even every node instance if we add the column to the mix.

PR #1189, opened 2021-09-24
Support "is None" constraints from if statements during inference

### LINKED ISSUE #791: Notice `is not None` and similar checks and prevent `BadUnaryOperationMessage`
### Steps to reproduce

Run pylint on this file:

```
"""Demonstrate invalid-unary-operand-type deficiency."""

class Klass:
    """Demonstration."""

    def __init__(self, param=None):
        self._param = param

    def method(self):
        """Returns -param or None."""
        if self._param is not None:
            return -self._param
        return None
```

### Current behavior
`example.py:12:19: E1130: bad operand type for unary -: NoneType (invalid-unary-operand-type)`

### Expected behavior

The `invalid-unary-operand-type` message should've been supressed because the code block it is coming from is guarded by an `is not None` check on the attribute in question.  astroid has enough information to infer that within this file.

_(confirming that... change the default value from the constructor parameter from None to any number and the error disappears)_

I suppose you can argue that anything _could_ have been changed to another type at any time by threads or some intervening call that somehow twiddles with this instances state, but the natural intent of code like this is that it hasn't which seems like the ideal thing for a best effort inference engine like astroid to infer.

### Versions

pylint 2.5.2
astroid 2.4.1

david-yz-liu:
## Steps

## Description

Alright, this is a big one so please don't hold back. This PR adds infrastructure to support *constraints* on values during inference. Here's a simple example:

```python

node = extract_node("""
def f(b):
    if b:
        x = 1
    else:
        x = None
    
    if x is not None:
        x #@
""")
print(node.inferred())
```

This prints `[<Const.int l.4 at 0x14232a6e6d0>, <Const.NoneType l.6 at 0x142329b38b0>]`, using both possible values from the assignment statements. Of course, because of the `x is not None` if condition, only the int is possible, but astroid currently doesn't take this condition into account.

This PR changes how inference works so that only `[<Const.int l.4 at 0x14232a6e6d0>]` is printed.

### Scope

I reviewed the existing pylint issues related to this (they are generally labelled as "topic-control-flow"). There were four types of constraints that were brought up:

1. `x is None`/`x is not None`
2. `x`/`not x` (truthy/falsey value)
3. `isinstance(x, ...)`
4. `x == ...`
5. Combinations of the above with `and`/`or`

There were also three primary locations where these constraints could be inferred (orthogonal to the constraints themselves):

1. if statement (as in above example)
2. `and`/`or` clauses. For example, `x is not None and x[0] == 1`; the second `x` should not be inferred as `None`
3. ternary if expressions. For example, `x[0] if x is not None else 1`; the first `x` should not be inferred as `None`
4. reassignment within an if statement, e.g.

    ```python
    def f():
        a = None
        if a is None:
            a = 1
        return -a
    ```
5. early return/raises within an if statement.
6. assert statements

This pull request addresses just the item 1's in each list. I used an approach that can handle 1-5 of the constraint types, and 1-3 of the locations. If this approach is approved, I think it would be pretty straightforward to make additional PRs for these items. (More on the other constraint locations below.) 

### Implementation overview

I added a new `Constraint` class to represent a constraint on a variable, like `x is not None`.
- This is an abstract class, and can be subclassed for different constraint types.

When a variable is inferred (see `inference.py`, `infer_name`), we search the ancestor chain from the node and look for containing if statements, using these to accumulate constraints on the variable. These constraints are added to the inference context (added a new attribute to `InferenceContext`).

Then, in `_infer_stmts` (`bases.py`), we filter the inferred values based on whether or not they satisfy all constraints for that variable. 
- `Uninferable` is considered to satisfy all constraints.
- ⚠️ If there is at least one inferred value, but *all* inferred values are filtered out, a new `Uninferable` value is yielded, rather than yielding nothing at all. I think this is the conservative approach, to avoid errors from the `raise_if_nothing_inferred` decorator.

The above also applies to attributes (see `infer_attribute`).

### Some comments/questions

1. Because constraints are gathered by searching up the ancestor chain, this PR doesn't handle real control flow issues, e.g. with the reassignment example from above.
2. Because the filtering occurs in `_infer_stmts`, it only applies to variables and attributes. So for example, subscript nodes (e.g. `my_array[0]`) won't have their inferred values filtered.
3. Gathering constraints occurs every time `infer_name`/`infer_attribute` is called, and since this requires going up the ancestor tree, this adds some overhead to inference. In principle constraints can be gathered through an AST visitor, but I thought that would be more complex so didn't choose it. (Thoughts welcome of course!)
4. I think adding the constraints to `InferenceContext` is correct, but to be honest I'm fuzzy on the nuances of this class and how it's used.
5. Do you like the name "constraint" here? I go back on forth on it, but haven't come up with anything better.

## Type of Changes

## Related Issue

Closes #791
Closes PyCQA/pylint#157
Closes PyCQA/pylint#1472
Closes PyCQA/pylint#2016
Closes PyCQA/pylint#2631
Closes PyCQA/pylint#2880

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,14 @@ What's New in astroid 2.9.0?
 ============================
 Release date: TBA
 
+* Support "is None" constraints from if statements during inference.
+
+  Closes #791
+  Closes PyCQA/pylint#157
+  Closes PyCQA/pylint#1472
+  Closes PyCQA/pylint#2016
+  Closes PyCQA/pylint#2631
+  Closes PyCQA/pylint#2880
 
 
 What's New in astroid 2.8.1?
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -137,12 +137,15 @@ class Proxy:
 def _infer_stmts(stmts, context, frame=None):
     """Return an iterator on statements inferred by each statement in *stmts*."""
     inferred = False
+    constraint_failed = False
     if context is not None:
         name = context.lookupname
         context = context.clone()
+        constraints = context.constraints.get(name, {})
     else:
         name = None
         context = InferenceContext()
+        constraints = {}
 
     for stmt in stmts:
         if stmt is Uninferable:
@@ -151,15 +154,26 @@ def _infer_stmts(stmts, context, frame=None):
             continue
         context.lookupname = stmt._infer_name(frame, name)
         try:
+            stmt_constraints = {
+                constraint
+                for constraint_stmt, constraint in constraints.items()
+                if not constraint_stmt.parent_of(stmt)
+            }
             for inf in stmt.infer(context=context):
-                yield inf
-                inferred = True
+                if all(constraint.satisfied_by(inf) for constraint in stmt_constraints):
+                    yield inf
+                    inferred = True
+                else:
+                    constraint_failed = True
         except NameInferenceError:
             continue
         except InferenceError:
             yield Uninferable
             inferred = True
-    if not inferred:
+
+    if not inferred and constraint_failed:
+        yield Uninferable
+    elif not inferred:
         raise InferenceError(
             "Inference failed for all members of {stmts!r}.",
             stmts=stmts,
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -1,0 +1,137 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+"""Classes representing different types of constraints on inference values."""
+
+from typing import Any, Dict, Optional
+
+from astroid import nodes, util
+
+
+class Constraint:
+    """Represents a single constraint on a variable."""
+
+    node: nodes.NodeNG
+    negate: bool
+
+    def __init__(self, node: nodes.NodeNG, negate: bool) -> None:
+        self.node = node
+        """The node that this constraint applies to."""
+        self.negate = negate
+        """True if this constraint is negated. E.g., "is not" instead of "is"."""
+
+    def invert(self) -> None:
+        """Invert this constraint."""
+        self.negate = not self.negate
+
+    @classmethod
+    def match(
+        cls, node: nodes.NodeNG, expr: nodes.NodeNG, negate: bool = False
+    ) -> Optional["Constraint"]:
+        """Return a new constraint for node matched from expr, if expr matches
+        the constraint pattern.
+
+        If negate is True, negate the constraint.
+        """
+
+    def satisfied_by(self, inferred: Any) -> bool:
+        """Return True if this constraint is satisfied by the given inferred value."""
+        return True
+
+
+class NoneConstraint(Constraint):
+    """Represents an "is None" or "is not None" constraint."""
+
+    @classmethod
+    def match(
+        cls, node: nodes.NodeNG, expr: nodes.NodeNG, negate: bool = False
+    ) -> Optional[Constraint]:
+        """Return a new constraint for node matched from expr, if expr matches
+        the constraint pattern.
+
+        Negate the constraint based on the value of negate.
+        """
+        if isinstance(expr, nodes.Compare) and len(expr.ops) == 1:
+            left = expr.left
+            op, right = expr.ops[0]
+            const_none = nodes.Const(None)
+            if op in {"is", "is not"} and (
+                matches(left, node)
+                and matches(right, const_none)
+                or matches(left, const_none)
+                and matches(right, node)
+            ):
+                negate = (op == "is" and negate) or (op == "is not" and not negate)
+                return cls(node=node, negate=negate)
+
+        return None
+
+    def satisfied_by(self, inferred: Any) -> bool:
+        """Return True if this constraint is satisfied by the given inferred value."""
+        if inferred is util.Uninferable:
+            return True
+
+        if self.negate and matches(inferred, nodes.Const(None)):
+            return False
+        if not self.negate and not matches(inferred, nodes.Const(None)):
+            return False
+
+        return True
+
+
+def matches(node1: nodes.NodeNG, node2: nodes.NodeNG) -> bool:
+    """Returns True if the two nodes match."""
+    if isinstance(node1, nodes.Name) and isinstance(node2, nodes.Name):
+        return node1.name == node2.name
+    if isinstance(node1, nodes.Attribute) and isinstance(node2, nodes.Attribute):
+        return node1.attrname == node2.attrname and matches(node1.expr, node2.expr)
+    if isinstance(node1, nodes.Const) and isinstance(node2, nodes.Const):
+        return node1.value == node2.value
+
+    return False
+
+
+def get_constraints(
+    expr: nodes.NodeNG, frame: nodes.NodeNG
+) -> Dict[nodes.NodeNG, Constraint]:
+    """Returns the constraints for the given expression.
+
+    The returned dictionary maps the node where the constraint was generated to the
+    corresponding constraint.
+
+    Constraints are computed statically by analysing the code surrounding expr.
+    Currently this only supports constraints generated from if conditions.
+    """
+    current_node = expr
+    constraints = {}
+    while current_node is not None and current_node is not frame:
+        parent = current_node.parent
+        if isinstance(parent, nodes.If):
+            branch, _ = parent.locate_child(current_node)
+            if branch == "body":
+                constraint = match_constraint(expr, parent.test)
+            elif branch == "orelse":
+                constraint = match_constraint(expr, parent.test, invert=True)
+            else:
+                constraint = None
+
+            if constraint:
+                constraints[parent] = constraint
+        current_node = parent
+
+    return constraints
+
+
+ALL_CONSTRAINTS = (NoneConstraint,)
+"""All supported constraint types."""
+
+
+def match_constraint(
+    node: nodes.NodeNG, expr: nodes.NodeNG, invert: bool = False
+) -> Optional[Constraint]:
+    """Returns a constraint pattern for node, if one matches."""
+    for constraint_cls in ALL_CONSTRAINTS:
+        constraint = constraint_cls.match(node, expr, invert)
+        if constraint:
+            return constraint
+
+    return None
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -41,6 +41,7 @@ class InferenceContext:
         "callcontext",
         "boundnode",
         "extra_context",
+        "constraints",
         "_nodes_inferred",
     )
 
@@ -89,6 +90,13 @@ class InferenceContext:
 
         Context that needs to be passed down through call stacks
         for call arguments
+        """
+
+        self.constraints = {}
+        """
+        :type: dict
+
+        The constraints on nodes
         """
 
     @property
@@ -145,6 +153,7 @@ class InferenceContext:
         clone.callcontext = self.callcontext
         clone.boundnode = self.boundnode
         clone.extra_context = self.extra_context
+        clone.constraints = self.constraints.copy()
         return clone
 
     @contextlib.contextmanager
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -35,7 +35,9 @@ from typing import Any, Callable, Dict, Iterable, Optional
 
 import wrapt
 
-from astroid import bases, decorators, helpers, nodes, protocols, util
+from astroid import bases
+from astroid import constraint as constraintmod
+from astroid import decorators, helpers, nodes, protocols, util
 from astroid.context import (
     CallContext,
     InferenceContext,
@@ -215,6 +217,8 @@ def infer_name(self, context=None):
             )
     context = copy_context(context)
     context.lookupname = self.name
+    context.constraints[self.name] = constraintmod.get_constraints(self, frame)
+
     return bases._infer_stmts(stmts, context, frame)
 
 
@@ -317,6 +321,11 @@ def infer_attribute(self, context=None):
         old_boundnode = context.boundnode
         try:
             context.boundnode = owner
+            if isinstance(owner, (nodes.ClassDef, bases.Instance)):
+                frame = owner if isinstance(owner, nodes.ClassDef) else owner._proxied
+                context.constraints[self.attrname] = constraintmod.get_constraints(
+                    self, frame=frame
+                )
             yield from owner.igetattr(self.attrname, context)
         except (
             AttributeInferenceError,
--- a/tests/unittest_inference_constraints.py
+++ b/tests/unittest_inference_constraints.py
@@ -1,0 +1,569 @@
+"""Tests for inference involving constraints"""
+
+from typing import Any
+
+import pytest
+
+from astroid import builder, nodes
+from astroid.util import Uninferable
+
+
+def common_params(node):
+    return pytest.mark.parametrize(
+        ("condition", "satisfy_val", "fail_val"),
+        (
+            (f"{node} is None", None, 3),
+            (f"{node} is not None", 3, None),
+        ),
+    )
+
+
+@common_params(node="x")
+def test_if_single_statement(condition: str, satisfy_val: Any, fail_val: Any) -> None:
+    """Test constraint for a variable that is used in the first statement of an if body."""
+    node1, node2 = builder.extract_node(
+        f"""
+    def f1(x = {fail_val}):
+        if {condition}:  # Filters out default value
+            return (
+                x  #@
+            )
+
+    def f2(x = {satisfy_val}):
+        if {condition}:  # Does not filter out default value
+            return (
+                x  #@
+            )
+    """
+    )
+
+    inferred = node1.inferred()
+    assert len(inferred) == 1
+    assert inferred[0] is Uninferable
+
+    inferred = node2.inferred()
+    assert len(inferred) == 2
+    assert isinstance(inferred[0], nodes.Const)
+    assert inferred[0].value == satisfy_val
+
+    assert inferred[1] is Uninferable
+
+
+@common_params(node="x")
+def test_if_multiple_statements(
+    condition: str, satisfy_val: Any, fail_val: Any
+) -> None:
+    """Test constraint for a variable that is used in an if body with multiple statements."""
+    node1, node2 = builder.extract_node(
+        f"""
+    def f1(x = {fail_val}):
+        if {condition}:  # Filters out default value
+            print(x)
+            return (
+                x  #@
+            )
+
+    def f2(x = {satisfy_val}):
+        if {condition}:  # Does not filter out default value
+            print(x)
+            return (
+                x  #@
+            )
+    """
+    )
+
+    inferred = node1.inferred()
+    assert len(inferred) == 1
+    assert inferred[0] is Uninferable
+
+    inferred = node2.inferred()
+    assert len(inferred) == 2
+    assert isinstance(inferred[0], nodes.Const)
+    assert inferred[0].value == satisfy_val
+
+    assert inferred[1] is Uninferable
+
+
+@common_params(node="x")
+def test_if_irrelevant_condition(
+    condition: str, satisfy_val: Any, fail_val: Any
+) -> None:
+    """Test that constraint for a different variable doesn't apply."""
+    nodes_ = builder.extract_node(
+        f"""
+    def f1(x, y = {fail_val}):
+        if {condition}:  # Does not filter out fail_val
+            return (
+                y  #@
+            )
+
+    def f2(x, y = {satisfy_val}):
+        if {condition}:
+            return (
+                y  #@
+            )
+    """
+    )
+    for node, val in zip(nodes_, (fail_val, satisfy_val)):
+        inferred = node.inferred()
+        assert len(inferred) == 2
+        assert isinstance(inferred[0], nodes.Const)
+        assert inferred[0].value == val
+
+        assert inferred[1] is Uninferable
+
+
+@common_params(node="x")
+def test_outside_if(condition: str, satisfy_val: Any, fail_val: Any) -> None:
+    """Test that constraint in an if condition doesn't apply outside of the if."""
+    nodes_ = builder.extract_node(
+        f"""
+    def f1(x = {fail_val}):
+        if {condition}:
+            pass
+        return (
+            x  #@
+        )
+
+    def f2(x = {satisfy_val}):
+        if {condition}:
+            pass
+
+        return (
+            x  #@
+        )
+    """
+    )
+    for node, val in zip(nodes_, (fail_val, satisfy_val)):
+        inferred = node.inferred()
+        assert len(inferred) == 2
+        assert isinstance(inferred[0], nodes.Const)
+        assert inferred[0].value == val
+
+        assert inferred[1] is Uninferable
+
+
+@common_params(node="x")
+def test_nested_if(condition: str, satisfy_val: Any, fail_val: Any) -> None:
+    """Test that constraint in an if condition applies within inner if statements."""
+    node1, node2 = builder.extract_node(
+        f"""
+    def f1(y, x = {fail_val}):
+        if {condition}:
+            if y is not None:
+                return (
+                    x  #@
+                )
+
+    def f2(y, x = {satisfy_val}):
+        if {condition}:
+            if y is not None:
+                return (
+                    x  #@
+                )
+    """
+    )
+    inferred = node1.inferred()
+    assert len(inferred) == 1
+    assert inferred[0] is Uninferable
+
+    inferred = node2.inferred()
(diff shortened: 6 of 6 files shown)


DanielNoord:
Hi @david-yz-liu, thanks for all of the work here. I think the size (and importance) of this PR has put off people from reviewing this but I think we should try and start the process of getting this merged.

The first question I would have before doing a full review is how we could expand this to include conditional reassignments. I think it would be every important to support the following code pattern:
```python
def func():
    if True:
        to_infer = 1
    else:
        to_infer = 2

    print(to_infer)
```

This does not need to be added in this PR, but we should at least have a general idea of how we could implement this within the same system. Do/did you have any ideas about this yourself?

david-yz-liu:
@DanielNoord wow, thank you! So sorry for the delay, I think I saw your message and meant to reply months ago but then... forgot. My bad!

Your question is a great one. It's been a while since I've thought about this, but I'll try to give a decent answer now (and perhaps a more thorough one later). I believe my current approach piggybacks off of the way `InferenceContext`s are passed around when doing inference, so the general rule for what this PR can or can't do is "we can accumulate Contraints whenever we can accumulate other things in `InferenceContext`s". The examples I gave in the PR description are places where a context is passed from a parent down to a child node in the AST, so they're pretty straightforward.

*However* as far I as can remember, existing `InferenceContexts` don't get passed across statements in a block. I think this is fundamentally why pylint's [`VariablesChecker`](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/variables.py) is as complex as it is?

But anyway to answer your question, I think that this PR _can_ be extended to handle the example you provided, but it wouldn't be by adding to my code very much, but rather by changing how these `InferenceContext`s (or some other "context" state) is passed around, using techniques similar to what's happening in the `VariablesChecker`. Or, using control flow graphs (😅) to precisely model the possible execution paths, and then accumulating Contraints across paths in the graph.

Please let me know if that answers your question, or if there are any follow-ups! Happy to be of service :)

DanielNoord:
> @DanielNoord wow, thank you! So sorry for the delay, I think I saw your message and meant to reply months ago but then... forgot. My bad!
> 
> Your question is a great one. It's been a while since I've thought about this, but I'll try to give a decent answer now (and perhaps a more thorough one later). I believe my current approach piggybacks off of the way `InferenceContext`s are passed around when doing inference, so the general rule for what this PR can or can't do is "we can accumulate Contraints whenever we can accumulate other things in `InferenceContext`s". The examples I gave in the PR description are places where a context is passed from a parent down to a child node in the AST, so they're pretty straightforward.
> 
> _However_ as far I as can remember, existing `InferenceContexts` don't get passed across statements in a block. I think this is fundamentally why pylint's [`VariablesChecker`](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/variables.py) is as complex as it is?
> 
> But anyway to answer your question, I think that this PR _can_ be extended to handle the example you provided, but it wouldn't be by adding to my code very much, but rather by changing how these `InferenceContext`s (or some other "context" state) is passed around, using techniques similar to what's happening in the `VariablesChecker`. Or, using control flow graphs (😅) to precisely model the possible execution paths, and then accumulating Contraints across paths in the graph.
> 
> Please let me know if that answers your question, or if there are any follow-ups! Happy to be of service :)

I think that answers it pretty clearly. As you can see I spent some time familiarising myself with the code 😄 I think this could indeed be a very good first step towards some better control flow. I have been thinking about doing something with `frames` and line numbering to do some more difficult constraints, but at the moment I don't see how we couldn't change the current approach to do that. It all seems flexible enough.

Let me know what you think of my changes. I made the tests pass on 3.7 and made some of the typing a little more strict. I appreciate that you had the constraints allow for other nodes than `If`, but for ease of later expansion I think it makes sense to keep it tighter now and expand the typing if needed later on.

david-yz-liu:
> Let me know what you think of my changes. I made the tests pass on 3.7 and made some of the typing a little more strict. I appreciate that you had the constraints allow for other nodes than `If`, but for ease of later expansion I think it makes sense to keep it tighter now and expand the typing if needed later on.

Will do. I'll try to look at this more closely tonight or tomorrow. 👍

ChangeLog
```diff
@@ -6,12 +6,20 @@ What's New in astroid 2.12.0?
...
+  Closes PyCQA/pylint#1472
+  Closes PyCQA/pylint#2016
+  Closes PyCQA/pylint#2631
+  Closes PyCQA/pylint#2880
 
 What's New in astroid 2.11.1?
 =============================
 Release date: TBA
 
+
```
Pierre-Sassoulas, line 22:
```suggestion
```

david-yz-liu:
@DanielNoord Okay I reviewed all of your changes, they look good to me! Yeah I was going for more general types but happy to stay strict for now and expand later if needed. Also thanks for making that Python 3.7 fix, that would have taken me a while to track down 😅.

Anyway as far as I can tell, this is good from my end, just waiting for more detailed review/feedback.

astroid/constraint.py
```diff
@@ -0,0 +1,142 @@
...
+        # Assume true if uninferable
+        if inferred is util.Uninferable:
+            return True
+
+        if self.negate and matches(inferred, nodes.Const(None)):
+            return False
+        if not self.negate and not matches(inferred, nodes.Const(None)):
+            return False
+
+        return True
```
Pierre-Sassoulas, line 83:
```suggestion
        return bool(self.negate) ^ matches(inferred, nodes.Const(None))
```

```suggestion
        return bool(self.negate) ^ bool(matches(inferred, nodes.Const(None)))
```

I guess a xor would be slightly more efficient with a comparison less ? Not sure if we need the cast to bool on ``matches``.

Pierre-Sassoulas (approved):
LGTM. The tests in particular are diversified and very meticulous. This will close a ton of issues and open a lot of possibilities in pylint too.

astroid/constraint.py
david-yz-liu: Thanks @Pierre-Sassoulas, I made a commit to use XOR. `self.negate` is a bool, and `matches` returns a `bool`, so we didn't need the explicit casting.

astroid/constraint.py
```diff
@@ -0,0 +1,137 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
```
DanielNoord, line 1: Just thinking in terms of future expandability:
Should we make this a module with a `__init__.py` and put `get_constraints` in a `__all__`.

I can see these classes and functions change over time as we might expand the control flow inference and by putting things explicitly in `__all__` we don't need to worry about backwards incompatibility.

astroid/constraint.py
```diff
@@ -0,0 +1,137 @@
...
+) -> Dict[nodes.If, Constraint]:
+    """Returns the constraints for the given expression.
+
+    The returned dictionary maps the node where the constraint was generated to the
+    corresponding constraint.
+
+    Constraints are computed statically by analysing the code surrounding expr.
+    Currently this only supports constraints generated from if conditions.
+    """
+    current_node: nodes.NodeNG = expr
```
DanielNoord, line 104:
```suggestion
    current_node = expr
```

This is probably left over from my previous changes. Forgot to remove it.

astroid/constraint.py
```diff
@@ -0,0 +1,137 @@
...
+    constraints = {}
```
DanielNoord, line 105:
```suggestion
    constraints: Dict[nodes.If, Constraint] = {}
```

DanielNoord:
@cdce8p Do you still want to review this before we merge this? I think with my last comments addressed this is almost good to go!

cdce8p:
> @cdce8p Do you still want to review this before we merge this? I think with my last comments addressed this is almost good to go!

If you're ok with waiting a bit longer, I could take a look in about 2-3 weeks. Will have much more time then to catch up on everything.

DanielNoord:
> > @cdce8p Do you still want to review this before we merge this? I think with my last comments addressed this is almost good to go!
> 
> If you're ok with waiting a bit longer, I could take a look in about 2-3 weeks. Will have much more time then to catch up on everything.

Fine with me! Should be around the time that I finish with `argparse` so I'll have time to assist with anything that comes up then 😄

cdce8p:
@DanielNoord Sounds good to me!

jacobtylerwalls (approved):
Looks great! I can help with merging main or resolving the last set of comments.

Pierre-Sassoulas (approved):
👌

david-yz-liu:
Sorry for the commit spam, I'm currently only running 3.10 locally. If this continues I'll fix it properly...

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Negate the constraint based on the value of negate.
+        """
+        if isinstance(expr, nodes.Compare) and len(expr.ops) == 1:
+            left = expr.left
+            op, right = expr.ops[0]
+            if op in {"is", "is not"} and (
+                matches(left, node)
+                and matches(right, cls.CONST_NONE)
+                or matches(left, cls.CONST_NONE)
```
cdce8p, line 65: `cls.CONST_NONE` is a type error since `cls` is bound to `Constraint`. Here it can only ever be a `NoneConstraint` or a subclass of it. To reflect that, you need to create a new TypeVar unfortunately. At least until mypy supports the new `Self` typing. I.e.
```py
_NoneConstraintT = TypeVar("_NoneConstraintT", bound="NoneConstraint")
```

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+                and matches(right, node)
```
cdce8p, line 66: Do we need to check both directions? How many people are going to write `None is x.y`? The constant test is usually last.

_This could even be a new pylint checker, to detect the reversed order. Tbh. though I have only ever seen the "correct" one `x.y is None`._

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+            ):
+                negate = (op == "is" and negate) or (op == "is not" and not negate)
+                return cls(node=node, negate=negate)
+
+        return None
+
+    def satisfied_by(self, inferred: nodes.NodeNG | type[util.Uninferable]) -> bool:
+        """Return True if this constraint is satisfied by the given inferred value."""
+        # Assume true if uninferable
```
cdce8p, line 75:
```suggestion
        # Assume true if Uninferable
```

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+        if inferred is util.Uninferable:
+            return True
+
+        return self.negate ^ matches(inferred, nodes.Const(None))
```
cdce8p, line 79: Can you add a small comment here?
It's correct but at least I don't see binary XORs all that often in Python. A comment would help to understand it quicker.

cdce8p, line 79:
```suggestion
        return self.negate ^ matches(inferred, self.CONST_NONE)
```
Or `NoneConstraint.CONST_NONE`

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+    while current_node is not None and current_node is not frame:
+        parent = current_node.parent
+        if isinstance(parent, nodes.If):
+            branch, _ = parent.locate_child(current_node)
+            if branch == "body":
+                constraint = match_constraint(expr, parent.test)
+            elif branch == "orelse":
+                constraint = match_constraint(expr, parent.test, invert=True)
+            else:
+                constraint = None
```
cdce8p, line 116:
```suggestion
            constraint: Constraint | None = None
            if branch == "body":
                constraint = match_constraint(expr, parent.test)
            elif branch == "orelse":
                constraint = match_constraint(expr, parent.test, invert=True)
```

astroid/constraint/constraint.py
```diff
@@ -0,0 +1,138 @@
...
+
+            if constraint:
+                constraints[parent] = constraint
+        current_node = parent
+
+    return constraints
+
+
+ALL_CONSTRAINTS = (NoneConstraint,)
+"""All supported constraint types."""
```
cdce8p, line 126: This can be a set.

You could also explore using `__init_subclass__` to auto-register all Constraint subclasses with the base class. Although I'm not sure how it interacts with `abc.ABC`.

astroid/bases.py
```diff
@@ -129,12 +129,15 @@ def _infer_stmts(
...
     inferred = False
+    constraint_failed = False
     if context is not None:
         name = context.lookupname
         context = context.clone()
+        constraints = context.constraints.get(name, {})
     else:
         name = None
         context = InferenceContext()
+        constraints = {}
```
cdce8p, line 140: I like to initialize container types unconditionally. That opens the possibility of adding a type hint which will help type checkers infer other variables later. Like `stmt_constraints` with is currently `set[Unknown]`. That should get better with
```py
constraints: dict[str, dict[nodes.If, constraints.Constraint]] = {}
```

cdce8p (changes requested):
Thanks for your patients @david-yz-liu! It's immediately visible that you've put a lot of work into it. So far I only had time for a surface level view. Never the less, I wanted to give you a first feedback as this is pretty easy to implement and will make the subsequent review a bit easier.

astroid/constraint/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
+
+"""Representations of logical constraints on values used during inference.
+"""
+
+from .constraint import get_constraints
+
+__all__ = ("get_constraints",)
```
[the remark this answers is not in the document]
cdce8p: Just saw the comment from @DanielNoord https://github.com/PyCQA/astroid/pull/1189#discussion_r834192353

IMO it's fine to leave everything in one module for now. If it get's too big, we can always go back and split it but for 140 lines, it just adds unnecessary complexity.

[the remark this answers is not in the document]
DanielNoord: I recommended this because of the many "moving" commits in astroid that interfere with git blame. It's a little premature, but it would avoid such churn later on.

But if both of you think it should stay in one file, then please go ahead. And sorry for the churn I caused then.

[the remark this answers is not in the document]
cdce8p: The git blame issue is an inconvenience, true. However we shouldn't overcomplicate things now just in anticipation of what could be. We're far way from any necessary split up, can add at least another 2000 lines 😄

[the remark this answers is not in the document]
DanielNoord: Then @david-yz-liu please revert the change. As I said, sorry for the inconvenience. Your work on this PR is much appreciated as it is one of the longer request features of `astroid` and `pylint`.

david-yz-liu:
Hello! I apologize for the delay again in making these updates (just have had a lot of other things going on). Notably, I've moved the constraint code back into a single file (`constraint.py`) and removed the `__all__` and instead used the underscore prefix where appropriate.

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, node: nodes.NodeNG, negate: bool) -> None:
+        self.node = node
+        """The node that this constraint applies to."""
+        self.negate = negate
+        """True if this constraint is negated. E.g., "is not" instead of "is"."""
+
+    @classmethod
+    @abstractmethod
+    def match(
+        cls: _ConstraintT, node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
```
cdce8p, line 29: Since this is a `classmethod`, the type of the first argument should be a class type not an instance.
```suggestion
    @classmethod
    @abstractmethod
    def match(
        cls: type[_ConstraintT], node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
```

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class NoneConstraint(Constraint):
+    """Represents an "is None" or "is not None" constraint."""
+
+    CONST_NONE: nodes.Const = nodes.Const(None)
+
+    @classmethod
+    def match(
+        cls: _ConstraintT, node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
```
cdce8p, line 49:
```suggestion
    @classmethod
    def match(
        cls: type[_ConstraintT], node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
```
Same here

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> _ConstraintT | None:
+        """Return a new constraint for node matched from expr, if expr matches
+        the constraint pattern.
+
+        If negate is True, negate the constraint.
+        """
+
+    @abstractmethod
+    def satisfied_by(self, inferred: nodes.NodeNG | type[util.Uninferable]) -> bool:
```
cdce8p, line 38:
```suggestion
    def satisfied_by(self, inferred: InferenceResult) -> bool:
```
See https://github.com/PyCQA/astroid/pull/1189#discussion_r903134872

`InferenceResult` can be imported from `astroid/typing.py`.

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+                and _matches(right, cls.CONST_NONE)
+                or _matches(left, cls.CONST_NONE)
+                and _matches(right, node)
+            ):
+                negate = (op == "is" and negate) or (op == "is not" and not negate)
+                return cls(node=node, negate=negate)
+
+        return None
+
+    def satisfied_by(self, inferred: nodes.NodeNG | type[util.Uninferable]) -> bool:
```
cdce8p, line 70:
```suggestion
    def satisfied_by(self, inferred: InferenceResult) -> bool:
```

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+
+        Negate the constraint based on the value of negate.
+        """
+        if isinstance(expr, nodes.Compare) and len(expr.ops) == 1:
+            left = expr.left
+            op, right = expr.ops[0]
+            if op in {"is", "is not"} and (
+                _matches(left, node)
+                and _matches(right, cls.CONST_NONE)
+                or _matches(left, cls.CONST_NONE)
```
cdce8p, line 62: `cls.CONST_NONE` is a type error. See: https://github.com/PyCQA/astroid/pull/1189#discussion_r903136395

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+                and _matches(right, node)
```
cdce8p, line 63: I would suggest we remove the `None is ...` check. That should not be used in code.
See: https://github.com/PyCQA/astroid/pull/1189#discussion_r903138606

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+
+            if constraint:
+                constraints[parent] = constraint
+        current_node = parent
+
+    return constraints
+
+
+ALL_CONSTRAINT_CLASSES = (NoneConstraint,)
+"""All supported constraint types."""
```
cdce8p, line 111:
```suggestion
ALL_CONSTRAINT_CLASSES = frozenset((NoneConstraint,))
"""All supported constraint types."""
```
See: https://github.com/PyCQA/astroid/pull/1189#discussion_r903145312

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _matches(node1: nodes.NodeNG, node2: nodes.NodeNG) -> bool:
```
cdce8p, line 114:
```suggestion
def _matches(node1: nodes.NodeNG | bases.Proxy, node2: nodes.NodeNG) -> bool:
```
With `InferenceResult`, `bases.Proxy` will also be passed as `node1`.

astroid/constraint.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _match_constraint(
+    node: _NameNodes, expr: nodes.NodeNG, invert: bool = False
+) -> Constraint | None:
+    """Returns a constraint pattern for node, if one matches."""
+    for constraint_cls in ALL_CONSTRAINT_CLASSES:
+        constraint = constraint_cls.match(node, expr, invert)
+        if constraint:
+            return constraint
```
cdce8p, line 133: Is there a case in which multiple constraint classes could match (in the future)? Does it make sense to change it to an Iterator then, i.e. `yield constraint` instead of `return`?

astroid/bases.py
```diff
@@ -146,9 +149,12 @@ def _infer_stmts(
 ) -> collections.abc.Generator[InferenceResult, None, None]:
     """Return an iterator on statements inferred by each statement in *stmts*."""
     inferred = False
+    constraint_failed = False
+    constraints: dict[str, dict[nodes.If, Constraint]] = {}
     if context is not None:
         name = context.lookupname
         context = context.clone()
+        constraints = context.constraints.get(name, {})
```
cdce8p, line 157: It's better to avoid assigning a different type to an existing variable. That will mess with type checking and make it more difficult to understand. Better to choose a unique name for it.

cdce8p (commented):
> Hello! I apologize for the delay again in making these updates (just have had a lot of other things going on).

Not necessary 🙂 We all do.

> Notably, I've moved the constraint code back into a single file (constraint.py) and removed the __all__ and instead used the underscore prefix where appropriate.

👍🏻

--
Left some more style comments.

astroid/constraint.py
DanielNoord: Changed this to be `Self` which I think makes more sense here, let me know if I need to revert

astroid/constraint.py
DanielNoord: I think this might be fixed with using `Self`?

astroid/constraint.py
DanielNoord: Removed!

DanielNoord:
I handled the review by Marc and incorporate his comments to help move this PR along. I think we're all eager to get this merged 😄

david-yz-liu:
Thanks @DanielNoord and @cdce8p! I wasn't sure exactly when I'd have time to look at this again so very much appreciate the updates. 👍 

BTW on the question of whether multiple constraint *classes* may match a given node (in the future), I think... maybe but not necessarily. For example, if we imagine a hypotheical "`AndConstraint`" from

```python
if x is not None and isinstance(x, int):
    ... x ...
```

and so the if node's `test` is an `and` of two expressions, I could see either yielding two distinct constraints ("not none constraint" and "is int instance" constraint), or a single `AndConstraint` whose children are the two constraints. (Personally I'd prefer going with the latter approach.)

DanielNoord:
> BTW on the question of whether multiple constraint _classes_ may match a given node (in the future), I think... maybe but not necessarily. For example, if we imagine a hypotheical "`AndConstraint`" from
> 
> ```python
> if x is not None and isinstance(x, int):
>     ... x ...
> ```
> 
> and so the if node's `test` is an `and` of two expressions, I could see either yielding two distinct constraints ("not none constraint" and "is int instance" constraint), or a single `AndConstraint` whose children are the two constraints. (Personally I'd prefer going with the latter approach.)

I think the concept of `AndConstraint` makes sense. This would remove the need for the `yield` behaviour.

I implemented the `yielding` behaviour for now but that could easily be reverted. @cdce8p what do you think?

astroid/constraint.py
cdce8p: `Self` would be awesome. Unfortunately it isn't supported by mypy yet.
The PR has already been merged, but we need to wait for the next release (1.0.0).

DanielNoord: Yeah, because it was already merged I thought it would be okay to use. `pyright` understands it and I don't think we will turn on `mypy` in `astroid` before `1.0.0` gets released so this doesn't give us any additional work (instead it prevents some :smile: )

astroid/constraint/constraint.py
Pierre-Sassoulas: It's done in latest versions: ``# Return the XOR of self.negate and matches(inferred, self.CONST_NONE)``

astroid/constraint/constraint.py
Pierre-Sassoulas: It's not a frozenset : ``ALL_CONSTRAINT_CLASSES = frozenset((NoneConstraint,))``

astroid/constraint.py
Pierre-Sassoulas: This was done at some point.

astroid/constraint.py
Pierre-Sassoulas: This has been done.

astroid/bases.py
Pierre-Sassoulas: I think it's fixed in the latest version ``context.constraints`` return a dict.

david-yz-liu:
💯 @Pierre-Sassoulas thank you! And thank you all for the efforts in reviewing this and keeping the branch up to date. I'm super excited to see how this evolves!

Issue #1145, opened 2021-08-23
Feature: Persistent caching of modules


joshua-cannon-techlabs:
### Steps to reproduce
For fairly large repositories, using inference can be incredibly slow. Running `pylint` with a rule which leverages inference can take multiple seconds, while running a rule without inference is almost immediate. 

From profiling, most of the time taken is traversing the file's dependencies and parsing them as well. This information is cached (see `astroid.manager.AtroidManager.brain`), however the cache is in-memory and therefore doesn't last between runs. This is likely due to the fact that `astroid` classes can't be easily serialized (pickling them is unfortunately not a drop-in solution).

Ideally, there is an option to enable persistent caching which is file-timestamp aware. If a particular module hasn't been changed, `astroid` could pull from the cache (exactly like Python's caching of bytecode).

Of course, a lot of thought has to be put into caching, but the benefits would be on the order of dev-years if not dev-decades IMO.

### Current behavior

- Run `pylint` on a single file in a large repo with `assert-on-tuple` as the only rule: Maybe ~0.5s on a bad day
- Run `pylint` on the same file with `abstract-class-instantiated` as the only rule: Roughly 7.4s
- Run `pylint` __twice__ in the same python process with `abstract-class-instantiated` as the only rule: Also roughly 7.4s

### Expected behavior

Running `pylint` multiple times on the same file ideally doesn't take the same amount of time each time.

### `python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"` output
2.4.2

joshua-cannon-techlabs:
FWIW I'm fully aware that this is no easy undertaking. However it should at least be documented as a feature request and talked about :smile:

joshua-cannon-techlabs:
I'll add too, AFAICT `mypy` runs speedily on the same codebase. I assume since it has a `cache-dir` flag it's caching post-parse information.

Pierre-Sassoulas:
Thank you for opening the issue, I agree with everything you said. I would add that making astroid performance better would save gigawatt of electricity. What I'm saying it that AWS should sponsor this issue :smile:

joshua-cannon-techlabs:
In the meantime I think I'll play around with running `pylint` as part of a daemon in order to get the performance improvement on dev boxes (since it'll keep the manager brain in-memory). Then in the daemon process, eject modules from the brain when files change.

thejcannon:
I've been playing around with this in my free time and am getting close to a really good Proof-of-Concept/MvP.

[...]

PR #1194, opened 2021-10-01
DO NOT MERGE - WIP: Add ability to serialize/deserialze astroid.Module (and friends)

thejcannon:
## Steps

## Description

This PR adds the ability to serialize and deserialize `astroid.Module` instances (and by association, every other kind of instance it may reference) to-and-from JSON. 

### Why persistence?

Persisting processed modules is useful primarily when parsing a large number of files, or files with a lot of interdependencies. So long as desrializing form the cache is quicker than parsing the files, it's a net win.

(I'll do some profiling with `pylint` on our codebase of about 1k files when this is more polished) 

### Why JSON, why not YAML, XML, or `pickle`?

JSON was chosen as it is compact (as opposed to yml or xml or similar) and we can choose what and how to store information. Although pickling might be simpler in terms of complexity, it is inherently insecure as a bad actor only needs to infect the `astroid` pickled data on disk to infect the next parse. Using JSON, since we control the data, the actor would have to infect the cache, and the Python environment the next parse runs in.

And since we aren't inventing anything new here, it should also be stated `mypy`'s cache is also JSON :wink:

### Changes

The bulk of the relevant changes have been localized to the `_persistence` module, which has been specially implemented in order to reduce the amount of changes to all of the relevant `astroid` classes. Additional changes have been made in order to either serialize/deserialize in a special way or to enable the default way.

A future change will have the changes to the module cache to leverage this new feature.

### Testing

TODO

## Type of Changes

## Related Issue

Related #1145


tests/conftest.py
```diff
@@ -0,0 +1,30 @@
+from functools import wraps
+
+import astroid.builder
+import astroid.nodes
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--test-roundtrip-persistence", type=bool, default=True,
```
thejcannon: Eventually, I think we'll add a tox section to re-run all tests with this enabled. For now it's just enabled so I can test easily.

thejcannon (commented):
Not all tests pass right now. Latest errors are from inference not working (and since inference is quite the beasty, I thought I'd publish this PR as-is for initial feedback)

thejcannon:
I'll use this PR to update progress and dump thoughts as I work through the remaining problems.

The current test is failing AFAICT due to `_explicit_inference` not persisting after serialization. I assume this should come from the 
`inference_tip` cache? :thinking:

astroid/nodes/scoped_nodes.py
```diff
@@ -2096,8 +2155,7 @@ def postinit(
         :param keywords: The keywords given to the class definition.
         :type keywords: list(Keyword) or None
         """
-        if keywords is not None:
-            self.keywords = keywords
+        self.keywords = keywords
```
Pierre-Sassoulas, line 2158: I wonder why it was that way at first. Thanks you for the cleanup.

astroid/nodes/scoped_nodes.py
```diff
@@ -3045,3 +3103,12 @@ def _get_assign_nodes(self):
             child_node._get_assign_nodes() for child_node in self.body
         )
         return list(itertools.chain.from_iterable(children_assign_nodes))
+
+    def __dump__(self, dumper):
+        data = super().__dump__(dumper)
```
Pierre-Sassoulas, line 3108: Is this a cultural reference to dumb and dumber ;) ?

Pierre-Sassoulas (commented):
Thank you for opening this. Sorry I don't have time for an in depth review right now. I agree with the need to improve the caching though. Could you be more specific about the advantages of JSON compared to pickle and about what make the persistance necessary ? Is it for very big code base or low ram machine where the cache would be bigger than ram ? Also could you add typing on newly added functions, please ?

astroid/nodes/scoped_nodes.py
thejcannon: Haha no, but we can pretend it is.
I originally was passing the remap around but thought that was kind of dirty, so wanted to pass a function that did the dumping and remembered the remap so "dumper" was the first word that came to mind 😂

thejcannon:
I agree the AST parsing caching itself wouldn't be sufficient (hence why there isn't any code that actually persists the AST to disk yet). I assumed this would be the first step that would eventually lead to inference caching (since the inference results would need to be persistable).

Closing, though as I won't have much free time anymore to walk this along. It can be used as a starting point for that future effort though :tada:

PR #1204, opened 2021-10-06
Fix regression introduced by #1186 and add tests

### LINKED ISSUE #1200: Regression for #1186
#1186 added logic for parsing package names. This seems to have caused a regression visible as `too-many-function-args` error in the example below.

File structure
```
b
├── __init__.py
├── a.py
└── level1
    ├── __init__.py
    └── b.py
```

```py
# b/a.py

# pylint: disable=missing-docstring

from .level1.b import func

def do_something(var, some_other_var):  # error
    func(var, some_other_var)

```

```py
# b/level1/__init__.py

def func(var):
    pass
```

```py
# b/level1/b.py

def func(var, some_other_var):
    pass
```

Runing pylint: `pylint -sn b/a.py`
```
************* Module b.a
b/a.py:8:4: E1121: Too many positional arguments for function call (too-many-function-args)
```

The error seems to happen as `func` is wrongly inferred as coming from `b/level1/__init__.py`, thus having one argument less than passed. The inference happens here: [typecheck.py -> visit_call](https://github.com/PyCQA/pylint/blob/7bc962ba81966b61b5b626cbfce36dd9279226c7/pylint/checkers/typecheck.py#L1274-L1279).

--
@DanielNoord As you authored #1186, would you mind taking a look at this?

DanielNoord:
## Steps

## Description

This is based on @cdce8p's suggestion. This also includes the test files he came up with.

I think we should also reconsider the tests that are currently here. Also the one for #1186. I don't think we actually reached consensus in that issue about whether we are fully testing everything we want. It seems we're missing some knowledge about how `astroid` `resources` are tested.

I did test this with https://github.com/PyCQA/pylint/pull/5059 and this suggestion works.

## Type of Changes

## Related Issue

This closes #1200

### DIFF
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -752,7 +752,7 @@ class Module(LocalsDictNodeNG):
         if package_name:
             if not modname:
                 return package_name
-            return f"{package_name}.{modname.split('.')[0]}"
+            return f"{package_name}.{modname}"
         return modname
 
     def wildcard_import_names(self):
--- a/tests/testdata/python3/data/beyond_top_level_two/a.py
+++ b/tests/testdata/python3/data/beyond_top_level_two/a.py
@@ -1,0 +1,7 @@
+# pylint: disable=missing-docstring
+
+from .level1.beyond_top_level_two import func
+
+
+def do_something(var, some_other_var):  # error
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (approved):
👍

PR #1207, opened 2021-10-07
Prefer using the module loader to get source in builder.

gpshead:
## Steps

## Description

Prefer the module loader get_source() method in AstroidBuilder's module_build() when possible to avoid assumptions about source code being available on a filesystem.  Otherwise the source cannot be found and application behavior changes when running within an embedded hermetic interpreter environment (pyoxidizer, etc.).

Without this change, pylint packaged up as such would fail to infer that functions like `bz2.compress` returned
something other than None, triggering `assignment-from-no-return` when it clearly should not.

At the pylint level, this works as an application integration regression test - _when you run the test build within such a no-filesystem environment._

```python
    def test_safe_infer_on_bz2_compress_return(self):
        src_node = astroid.extract_node('import bz2\nunused = bz2.compress(b"")')
        assignment = next(src_node.nodes_of_class(astroid.Assign))
        inferred = pylint.checkers.utils.safe_infer(assignment.value.func)
        self.assertIsInstance(inferred, astroid.FunctionDef, msg=str(inferred))
        try:
            next(inferred.nodes_of_class(astroid.Return))
        except StopIteration:
            self.fail(f'No return found in body of:\n{str(inferred)}.')
```

I did not try to come up with a smaller targeted astroid unittest faking the "The path in .\_\_file\_\_ does not exist" environment.  That would probably be useful.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@ What's New in astroid 2.9.0?
 ============================
 Release date: TBA
 
+* Prefer the module loader get_source() method in AstroidBuilder's
+  module_build() when possible to avoid assumptions about source
+  code being available on a filesystem.  Otherwise the source cannot
+  be found and application behavior changes when running within an
+  embedded hermetic interpreter environment (pyoxidizer, etc.).
 
 
 What's New in astroid 2.8.3?
--- a/ChangeLog
+++ b/ChangeLog
@@ -92,6 +92,7 @@ Release date: 2021-10-25
   nodes are now correctly added to the locals of the ``FunctionDef``,
   ``ClassDef``, or ``Comprehension``.
 
+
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
Merging without tests as pylint's test suite is passing and the actual testing will be done in https://github.com/PyCQA/pylint/issues/5341

Congratulation on becoming an astroid contributor :)

gpshead:
Thanks!  I had stalled on responding while pondering how to construct a meaningful regression test case without being overcomplicated! :)

Pierre-Sassoulas:
You're welcome to contribute to PyCQA/pylint#5341 if you have new ideas :)

PR #1209, opened 2021-10-09
Add recognition of previous partial args/kwargs when chaining

### LINKED ISSUE #1208: Nested partial functions lose track of previously-filled args/kwargs
### Steps to reproduce

Cause of issue [#2462](https://github.com/PyCQA/pylint/issues/2462) for `pylint`

### Current behavior

```python
node = astroid.extract_node(
    """
from functools import partial
def test(a, b, c):
    return a + b + c

test1 = partial(test, 1)
test2 = partial(test1, 2)
test2(3)
"""

call_site = astroid.arguments.CallSite.from_call(node)
called_func = next(node.func.infer())
called_args = called_func.filled_args + call_site.positional_arguments
```

The `called_args` above will contain args for `2` and `3`, but not `1`, because the second call to `partial` overwrites the first, dropping `1` from the list of `filled_args`.

### Expected behavior

All args/kwargs locked in by calls to `partial` should be present in the final function. I've written a test similar to the code example above and have implemented a fix, so I'll open a PR shortly.

`2.6.6`

cfranklin11:
## Steps

## Description

Repeat use of functools.partial on the same function would only
keep the args & kwargs locked in by the last call to partial.
By checking if the wrapped function is itself a partial, and
including its filled args/kwargs if it is, we can keep track
of all locked args/kwargs across any number of nested partial
calls.

## Type of Changes

## Related Issue

Closes #1208

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,7 +17,7 @@ Release date: TBA
 * Fixes handling of nested partial functions
 
   Closes pyCQA/pylint#2462
-  Closes pyCQA/astroid#1208
+  Closes #1208
 
 
 What's New in astroid 2.8.2?


Pierre-Sassoulas:
Thanks a lot for the fix :)

PR #1211, opened 2021-10-12
Fix regression with import resolver

DanielNoord:
## Steps

## Description

This fixes the regression reported in https://github.com/PyCQA/pylint/issues/5131

Basically the previous iteration of the import resolver wrongly assumed that the current path was also the directory of the module which was checked for existence of `__init__.py`. We now check whether the `nodes.Module` has a path and then check the directory of that path for a `__init__.py`.
The check for existence of `self.path` is because `test_relative_to_absolute_name` in `tests/unittest_scoped_nodes.py` failed. I wonder if it is actually relevant for any real examples as I think a `nodes.Module` should always have a `path`, but I think that is something for another time.

While I was working on this I also added basic typing to the function in the new format.

## Type of Changes

## Related Issue

Ref. https://github.com/PyCQA/astroid/pull/1186, https://github.com/PyCQA/astroid/issues/1200, https://github.com/PyCQA/pylint/pull/5059

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,12 +16,17 @@ Release date: TBA
 
 * Fixes handling of nested partial functions
 
-  Closes pyCQA/pylint#2462
+  Closes PyCQA/pylint#2462
   Closes #1208
 
 * Fix regression with the import resolver
 
   Closes PyCQA/pylint#5131
+
+* Fix crash with invalid dataclass field call
+
+  Closes PyCQA/pylint#5153
+
 
 What's New in astroid 2.8.2?
 ============================
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -308,6 +308,7 @@ def _looks_like_dataclass_field_call(node: Call, check_scope: bool = True) -> bo
         scope = stmt.scope()
         if not (
             isinstance(stmt, AnnAssign)
+            and stmt.value is not None
             and isinstance(scope, ClassDef)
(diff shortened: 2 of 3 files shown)


cdce8p:
@DanielNoord I think you accidentally committed the first commit to `main` this morning. I reverted it, as black was failing.
https://github.com/PyCQA/astroid/commits/main

The changes here seem to work, I'll take a closer look next.

cdce8p:
> Maybe it's the time to put "real world" test in place where we run pylint on open-source python code available online. numpy, pandas, requests, colorama, botocore, setuptools, ... there's a lot of choices and probably some variety and some namespace packages too. We could run those only for "risky" changes. Real world examples are hard to do well. black is doing something similar I don't remember how they call it. It would reduce the risk we take during release greatly. Plus there are already some slow tests that we're not running in the CI for pylint (acceptance).

https://github.com/PyCQA/pylint/issues/4413 😄
Unfortunately, it's easier said than done. Mypy has [mypy_primer](https://github.com/hauntsaninja/mypy_primer), black has [black_primer](https://github.com/psf/black/tree/main/src/black_primer). The difference is that both are focused on maintaining stability whereas we constantly add new things as well. I see it with `Home Assistant` where I do something similar just manually. It's a **lot** of effort just keeping the repos in sync, comparing new error messages, checking if they are intended or false-positives, updating the existing code and pushing the changes back to `Home Assistant`.

What I'm trying to say: I would love it if we could automate it, just not sure it's actually possible / useful for our development cycle.

PR #1213, opened 2021-10-19
Add assignment expressions to correct scope for automatic evaluation

DanielNoord:
## Steps

## Description

See also https://github.com/PyCQA/pylint/pull/5188 and first report of https://github.com/PyCQA/pylint/issues/3688.

## Type of Changes

### DIFF
--- a/tests/unittest_nodes.py
+++ b/tests/unittest_nodes.py
@@ -1557,9 +1557,7 @@ def test_assignment_expression_in_functiondef() -> None:
     class_assign = module.body[8]
     assert "assignment_nine" in class_assign.locals
     assert "assignment_nine" not in module.locals
-    assert isinstance(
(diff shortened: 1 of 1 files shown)


DanielNoord:
It just came to me that this is not doing what it is supposed to do.

I think we only need to elevate `NamedExpr` in `Arguments` nodes to the upper scope as without doing so they will remain in the scope of the `FunctionDef`. The rest should stay the same, and thus we can revert to the first commit here.
The other commits I have written can be used for the `used-before-assignment` checker. Instead of checking whether a `NamedExpr` should be added to `locals` this actually checks whether they are already defined. With how `pylint` is set up, we also add still undefined variables to `locals`. I mixed this up while working on this.

Sorry for all of the spam here. Tomorrow I will revert to the version with only `NamedExpr` in `Arguments` and we save the work done for `used-before-assignment` in a draft PR on `pylint`.

PR #1215, opened 2021-10-20
Refactor: Stop adding arbitrary attributes to module obj when building

thejcannon:
## Steps

## Description

Moving code from `_post_build` into `_data_build` so that we don't have to attach arbitrary objects to the module (namely `_import_from_nodes`, and `_delayed_assattr`).

Alternatively we could:
- Keep the code in `_post_build` and pass the info along
- Same, but pass the builder

## Type of Changes

|   | :scroll: Docs          |

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -147,17 +147,6 @@ class AstroidBuilder(raw_building.InspectBuilder):
     def _post_build(self, module, encoding):
         """Handles encoding and delayed nodes after a module has been built"""
         module.file_encoding = encoding
-        self._manager.cache_module(module)
-        # post tree building steps after we stored the module in the cache:
-        for from_node in module._import_from_nodes:
-            if from_node.modname == "__future__":
-                for symbol, _ in from_node.names:
-                    module.future_imports.add(symbol)
-            self.add_from_names_to_locals(from_node)
-        # handle delayed assattr nodes
-        for delayed in module._delayed_assattr:
-            self.delayed_assattr(delayed)
-
         # Visit the transforms
         if self._apply_transforms:
             module = self._manager.visit_transforms(module)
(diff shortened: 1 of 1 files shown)


astroid/builder.py
```diff
@@ -190,8 +179,18 @@ def _data_build(self, data, modname, path):
             )
         builder = rebuilder.TreeRebuilder(self._manager, parser_module)
         module = builder.visit_module(node, modname, node_file, package)
-        module._import_from_nodes = builder._import_from_nodes
-        module._delayed_assattr = builder._delayed_assattr
+
+        self._manager.cache_module(module)
```
DanielNoord, line 183: Shouldn't the `cache_module` stay in `_post_build`? If we ever do something different in `_data_build` after this line or we add a method between `_data_build` and `_post_build` it won't be cached?

DanielNoord (commented):
Based on the tests and the actual code I this change seems fine? It doesn't actually do much different, other than making the code more readable since `_post_build` is called after `_data_build` anyway.

Edit: And by the way @thejcannon sorry for letting this PR drop down into "stale-PR-hell" 😅

astroid/builder.py
cdce8p: I think it wouldn't matter. The `module` object reference is cached, ie. if the object changes, those changes are also reflected in the cache.

cdce8p (dismissed):
I would prefer if we kept the logic inside `_post_build`. IMO passing the `builder` along is probably the better alternative.

cdce8p:
I've updated the code to pass `builder` along instead. Didn't want to drag this on any longer.

cdce8p (approved):

Pierre-Sassoulas (approved):
LGTM thank you @thejcannon and @cdce8p

PR #1217, opened 2021-10-22
Add typing to ``NodeNG.statement``

DanielNoord:
## Steps

## Description

`statement()` is only defined twice. Once on `NodeNG` and once on `Module`. The one on `NodeNG` refers to `self.is_statement` which is only set to `True` on `Statement`. As a result `statement()` can either return `Module` or `Statement`.

Added because of comment by @cdce8p here:
https://github.com/PyCQA/pylint/pull/5163#pullrequestreview-786711662

I think I will not wait for this version to be added to `pylint` to update the typing there.

## Type of Changes

### DIFF
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -272,6 +272,19 @@ class ParentMissingError(AstroidError):
         super().__init__(message=f"Parent not found on {target!r}.")
 
 
+class StatementMissing(ParentMissingError):
+    """Raised when a call to node.statement() does not return a node. This is because
+    a node in the chain does not have a parent attribute and therefore does not
+    return a node for statement().
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    def __init__(self, target: "nodes.NodeNG") -> None:
+        self.target = target
+        super().__init__(message=f"Statement not found for {target!r}.")
+
 # Backwards-compatibility aliases
 OperationError = util.BadOperationMessage
 UnaryOperationError = util.BadUnaryOperationMessage
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -267,9 +267,9 @@ class ParentMissingError(AstroidError):
         target: The node for which the parent lookup failed.
     """
 
-    def __init__(self, target: "nodes.NodeNG") -> None:
+    def __init__(self, target: "nodes.NodeNG", missing_element: str = "Parent") -> None:
         self.target = target
-        super().__init__(message=f"Parent not found on {target!r}.")
+        super().__init__(message=f"{missing_element} not found on {target!r}.")
 
 
 class StatementMissing(ParentMissingError):
@@ -282,8 +282,8 @@ class StatementMissing(ParentMissingError):
     """
 
     def __init__(self, target: "nodes.NodeNG") -> None:
-        self.target = target
-        super().__init__(message=f"Statement not found for {target!r}.")
+        super().__init__(target, "Statement")
+
 
 # Backwards-compatibility aliases
 OperationError = util.BadOperationMessage
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -265,7 +265,6 @@ class ParentMissingError(AstroidError):
 
     Standard attributes:
         target: The node for which the parent lookup failed.
-        missing_element: The type of attribute/method that is missing because of the missing parent
     """
 
     def __init__(self, target: "nodes.NodeNG") -> None:
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -265,6 +265,7 @@ class ParentMissingError(AstroidError):
 
     Standard attributes:
         target: The node for which the parent lookup failed.
+        missing_element: The type of attribute/method that is missing because of the missing parent
     """
 
     def __init__(self, target: "nodes.NodeNG", missing_element: str = "Parent") -> None:
@@ -283,7 +284,6 @@ class StatementMissing(ParentMissingError):
 
     def __init__(self, target: "nodes.NodeNG") -> None:
         super().__init__(target, "Statement")
-
 
 
 # Backwards-compatibility aliases
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -5,8 +5,8 @@
 # Copyright (c) 2016 Derek Gustafson <degustaf@gmail.com>
 # Copyright (c) 2018 Bryce Guinta <bryce.paul.guinta@gmail.com>
 # Copyright (c) 2020-2021 hippo91 <guillaume.peillex@gmail.com>
-# Copyright (c) 2021 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
 # Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
+# Copyright (c) 2021 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
 # Copyright (c) 2021 Andrew Haigh <hello@nelf.in>
 
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -285,6 +285,7 @@ class StatementMissing(ParentMissingError):
         super().__init__(target, "Statement")
 
 
+
 # Backwards-compatibility aliases
 OperationError = util.BadOperationMessage
 UnaryOperationError = util.BadUnaryOperationMessage
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -282,9 +282,9 @@ class StatementMissing(ParentMissingError):
     """
 
     def __init__(self, target: "nodes.NodeNG") -> None:
-        # See: https://github.com/PyCQA/pylint/issues/2903
-        # and: https://github.com/PyCQA/astroid/pull/1217#discussion_r744149027
         # pylint: disable-next=bad-super-call
+        # https://github.com/PyCQA/pylint/issues/2903
+        # https://github.com/PyCQA/astroid/pull/1217#discussion_r744149027
         super(ParentMissingError, self).__init__(
             message=f"Statement not found on {target!r}"
         )
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -282,6 +282,9 @@ class StatementMissing(ParentMissingError):
     """
 
     def __init__(self, target: "nodes.NodeNG") -> None:
+        # See: https://github.com/PyCQA/pylint/issues/2903
+        # and: https://github.com/PyCQA/astroid/pull/1217#discussion_r744149027
+        # pylint: disable-next=bad-super-call
         super(ParentMissingError, self).__init__(
             message=f"Statement not found on {target!r}"
         )
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -268,9 +268,9 @@ class ParentMissingError(AstroidError):
         missing_element: The type of attribute/method that is missing because of the missing parent
     """
 
-    def __init__(self, target: "nodes.NodeNG", missing_element: str = "Parent") -> None:
+    def __init__(self, target: "nodes.NodeNG") -> None:
         self.target = target
-        super().__init__(message=f"{missing_element} not found on {target!r}.")
+        super().__init__(message=f"Parent not found on {target!r}.")
 
 
 class StatementMissing(ParentMissingError):
@@ -283,7 +283,9 @@ class StatementMissing(ParentMissingError):
     """
 
     def __init__(self, target: "nodes.NodeNG") -> None:
-        super().__init__(target, "Statement")
+        super(ParentMissingError, self).__init__(
+            message=f"Statement not found on {target!r}"
+        )
 
 
 # Backwards-compatibility aliases
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -30,10 +30,6 @@ from astroid.nodes.const import OP_PRECEDENCE
 if TYPE_CHECKING:
     from astroid import nodes
 
-if sys.version_info >= (3, 6, 2):
-    from typing import NoReturn
-else:
-    from typing_extensions import NoReturn
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -264,23 +260,24 @@ class NodeNG:
 
     @overload
     def statement(
-        self, future: Literal[False] = ...
+        self, future: Literal[None] = ...
     ) -> Union["nodes.Statement", "nodes.Module"]:
         ...
 
     @overload
-    def statement(self, future: Literal[True]) -> Union[NoReturn, "nodes.Statement"]:
+    def statement(self, future: Literal[True]) -> "nodes.Statement":
         ...
 
     def statement(
-        self, future: bool = False
-    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
+        self, future: Optional[Literal[True]] = None
+    ) -> Union["nodes.Statement", "nodes.Module"]:
         """The first parent node, including self, marked as statement node.
 
         TODO: Deprecate the future parameter and only raise StatementMissing and return
         nodes.Statement
 
-        :raises StatementMissing: If no self has no parent attribute
+        :raises AttributeError: If self has no parent attribute
+        :raises StatementMissing: If self has no parent attribute and future is True
         """
         if self.is_statement:
             return cast("nodes.Statement", self)
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -10,6 +10,7 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    cast,
     overload,
 )
 
@@ -25,7 +26,7 @@ from astroid.nodes.as_string import AsStringVisitor
 from astroid.nodes.const import OP_PRECEDENCE
 
 if TYPE_CHECKING:
-    from astroid.nodes import LocalsDictNodeNG
+    from astroid import nodes
 
 # Types for 'NodeNG.nodes_of_class()'
 T_Nodes = TypeVar("T_Nodes", bound="NodeNG")
@@ -248,14 +249,17 @@ class NodeNG:
                 return True
         return False
 
-    def statement(self):
+    def statement(self) -> Union["nodes.Statement", "nodes.Module"]:
         """The first parent node, including self, marked as statement node.
 
         :returns: The first parent statement.
         :rtype: NodeNG
         """
         if self.is_statement:
+            self = cast("nodes.Statement", self)
             return self
+        if not self.parent:
+            raise ParentMissingError(target=self)
         return self.parent.statement()
 
     def frame(self):
@@ -269,7 +273,7 @@ class NodeNG:
         """
         return self.parent.frame()
 
-    def scope(self) -> "LocalsDictNodeNG":
+    def scope(self) -> "nodes.LocalsDictNodeNG":
         """The first parent node defining a new scope.
         These can be Module, FunctionDef, ClassDef, Lambda, or GeneratorExp nodes.
 
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -1,6 +1,7 @@
 import pprint
 import sys
 import typing
+import warnings
 from functools import singledispatch as _singledispatch
 from typing import (
     TYPE_CHECKING,
@@ -288,6 +289,12 @@ class NodeNG:
         if not self.parent:
             if future:
                 raise StatementMissing(target=self)
+            warnings.warn(
+                "In astroid 3.0.0 NodeNG.statement() will return either a nodes.Statement "
+                "or raise a StatementMissing exception. This behaviour can already be triggered "
+                "by passing 'future=True' to a statement() call.",
+                DeprecationWarning,
+            )
             raise AttributeError(f"{self} object has no attribute 'parent'")
         return self.parent.statement(future)
 
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -19,6 +19,7 @@ from astroid.exceptions import (
     AstroidError,
     InferenceError,
     ParentMissingError,
+    StatementMissing,
     UseInferenceDefault,
 )
 from astroid.manager import AstroidManager
@@ -249,17 +250,19 @@ class NodeNG:
                 return True
         return False
 
-    def statement(self) -> Union["nodes.Statement", "nodes.Module"]:
+    def statement(self) -> "nodes.Statement":
         """The first parent node, including self, marked as statement node.
 
         :returns: The first parent statement.
         :rtype: NodeNG
+
+        :raises StatementMissing: If no self has no parent attribute
         """
         if self.is_statement:
             self = cast("nodes.Statement", self)
             return self
         if not self.parent:
-            raise ParentMissingError(target=self)
+            raise StatementMissing(target=self)
         return self.parent.statement()
 
     def frame(self):
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -30,6 +30,11 @@ from astroid.nodes.const import OP_PRECEDENCE
 if TYPE_CHECKING:
     from astroid import nodes
 
+if sys.version_info >= (3, 6, 2):
+    from typing import NoReturn
+else:
+    from typing_extensions import NoReturn
+
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
@@ -259,23 +264,17 @@ class NodeNG:
 
     @overload
     def statement(
-        self, future: Literal[None] = ...
+        self, future: Literal[False] = ...
     ) -> Union["nodes.Statement", "nodes.Module"]:
         ...
 
     @overload
-    def statement(
-        self, future: Literal[False]
-    ) -> Union["nodes.Statement", "nodes.Module"]:
-        ...
-
-    @overload
-    def statement(self, future: Literal[True]) -> "nodes.Statement":
+    def statement(self, future: Literal[True]) -> Union[NoReturn, "nodes.Statement"]:
         ...
 
     def statement(
         self, future: bool = False
-    ) -> Union["nodes.Statement", "nodes.Module"]:
+    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
         """The first parent node, including self, marked as statement node.
 
         TODO: Deprecate the future parameter and only raise StatementMissing and return
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -274,7 +274,7 @@ class NodeNG:
         ...
 
     def statement(
-        self, future: Literal[None, True] = None
+        self, *, future: Literal[None, True] = None
     ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
         """The first parent node, including self, marked as statement node.
 
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -1,4 +1,5 @@
 import pprint
+import sys
 import typing
 from functools import singledispatch as _singledispatch
 from typing import (
@@ -28,6 +29,12 @@ from astroid.nodes.const import OP_PRECEDENCE
 
 if TYPE_CHECKING:
     from astroid import nodes
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
 
 # Types for 'NodeNG.nodes_of_class()'
 T_Nodes = TypeVar("T_Nodes", bound="NodeNG")
@@ -250,11 +257,29 @@ class NodeNG:
                 return True
         return False
 
-    def statement(self) -> "nodes.Statement":
+    @overload
+    def statement(
+        self, future: Literal[None] = ...
+    ) -> Union["nodes.Statement", "nodes.Module"]:
+        ...
+
+    @overload
+    def statement(
+        self, future: Literal[False]
+    ) -> Union["nodes.Statement", "nodes.Module"]:
+        ...
+
+    @overload
+    def statement(self, future: Literal[True]) -> "nodes.Statement":
+        ...
+
+    def statement(
+        self, future: bool = False
+    ) -> Union["nodes.Statement", "nodes.Module"]:
         """The first parent node, including self, marked as statement node.
 
-        :returns: The first parent statement.
-        :rtype: NodeNG
+        TODO: Deprecate the future parameter and only raise StatementMissing and return
+        nodes.Statement
 
         :raises StatementMissing: If no self has no parent attribute
         """
@@ -262,8 +287,10 @@ class NodeNG:
             self = cast("nodes.Statement", self)
             return self
         if not self.parent:
-            raise StatementMissing(target=self)
-        return self.parent.statement()
+            if future:
+                raise StatementMissing(target=self)
+            raise AttributeError(f"{self} object has no attribute 'parent'")
+        return self.parent.statement(future)
 
     def frame(
         self,
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -291,12 +291,13 @@ class NodeNG:
                 raise StatementMissing(target=self)
             warnings.warn(
                 "In astroid 3.0.0 NodeNG.statement() will return either a nodes.Statement "
-                "or raise a StatementMissing exception. This behaviour can already be triggered "
+                "or raise a StatementMissing exception. AttributeError will no longer be raised. "
+                "This behaviour can already be triggered "
                 "by passing 'future=True' to a statement() call.",
                 DeprecationWarning,
             )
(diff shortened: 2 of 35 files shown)


astroid/nodes/node_ng.py
```diff
@@ -248,14 +249,17 @@ def parent_of(self, node):
...
         """The first parent node, including self, marked as statement node.
 
         :returns: The first parent statement.
         :rtype: NodeNG
         """
         if self.is_statement:
+            self = cast("nodes.Statement", self)
             return self
+        if not self.parent:
+            raise ParentMissingError(target=self)
```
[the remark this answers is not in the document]
DanielNoord: @cdce8p Pinging you again. Please consider unsent if you just haven't had time to look at this!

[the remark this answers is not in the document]
cdce8p: > (Just passing a `future` argument to functions doesn't really work because we would then deprecate the future parameter afterwards.)

Isn't that how `SQLAlechemy` will do it, too? I would recommend to add it as argument to `statement()`. Passing it during initialization has some drawbacks which I'd like to avoid:
* Difficult to pass around to all nodes, could introduce issues with circular dependencies. Although we might be able to make it work in the end.
* What if we want to add a `future` argument to another function? We would break code that has already added the flag.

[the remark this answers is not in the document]
DanielNoord: https://docs.sqlalchemy.org/en/14/core/future.html
https://github.com/sqlalchemy/sqlalchemy/blob/8656050241766688fee8363b5bb950d3945756c1/lib/sqlalchemy/engine/create.py#L627-L632

Sorry, I should have linked the documentation.
Seems to me that they pass the argument to two of their base classes as a "flag" and then methods/classes get changed depending on that flag. However, I have never used their code before, so it might not be true. 

I'm against passing it as an argument to `statement()` because then we would need to deprecate the `future` parameter again at some point. That seems like double-work?

astroid/nodes/scoped_nodes.py
```diff
@@ -65,13 +66,19 @@
...
     MroError,
+    StatementMissing,
     TooManyLevelsError,
 )
 from astroid.interpreter.dunder_lookup import lookup
 from astroid.interpreter.objectmodel import ClassModel, FunctionModel, ModuleModel
 from astroid.manager import AstroidManager
 from astroid.nodes import Arguments, Const, node_classes
 
+if sys.version_info >= (3, 6, 2):
```
Pierre-Sassoulas, line 77: I just realized that we also need to do https://github.com/PyCQA/pylint/issues/5065 for astroid.

Pierre-Sassoulas (approved):
👍

astroid/nodes/scoped_nodes.py
DanielNoord: I believe this is the only one. Shall I open an issue so we don't forget about this?

Pierre-Sassoulas: Might as well create an issue from the [unofficial todo list](https://github.com/PyCQA/pylint/issues/5065#issuecomment-935428793) (hidden in dozen of comment 😄) and add a "release astroid 2.8.5 for > 3.6.2 only" ?:

astroid/nodes/node_ng.py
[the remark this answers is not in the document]
cdce8p: > Sorry, I should have linked the documentation. Seems to me that they pass the argument to two of their base classes as a "flag" and then methods/classes get changed depending on that flag. However, I have never used their code before, so it might not be true.

Yes, they pass it to the base functions. However, as I explained earlier, that doesn't really work for us. Should we add `future2`, `future3`, ... for any subsequent change to other functions? `SQLAlchemy` is different as they have just one big change that they what to do, not many small once.

> I'm against passing it as an argument to `statement()` because then we would need to deprecate the `future` parameter again at some point. That seems like double-work?

We need to deprecate it either way. If you don't like that, the only other alternatives I can think of would be to not add a breaking change or create a new method with a different name that replaces `statement()`. Not sure these are any better, though.

--
It would be a bit easier if we were changing function arguments. Then we could use something similar to what I've done for the removal of `None` as default values and use `inspect.signature` among other things. Without any `arguments` that doesn't work unfortunately.
https://github.com/PyCQA/astroid/blob/cdf8a2a9813128db5732a792434b89967c6fb701/astroid/decorators.py#L155-L209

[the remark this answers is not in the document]
DanielNoord: > We need to deprecate it either way. If you don't like that, the only other alternatives I can think of would be to not add a breaking change or create a new method with a different name that replaces `statement()`. Not sure these are any better, though.

Would the `future` parameter be deprecated at the same time as we change the normal behaviour? I think I (wrongly) assumed that the flow would be:
1) Allow `future` parameter
2) Bump to 3.0 > change normal behaviour
3) Warn of deprecation for `future` parameter
4) Bump to 4.0 > disallow `future` parameter
 
Thinking about this you probably meant to do:
1) Allow `future` parameter
2) Bump to 3.0 > change normal behaviour and disallow `future` parameter

The second option seems fine to me!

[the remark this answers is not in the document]
cdce8p: I believe `SQLAlchemy` uses the first flow, although they might not wait till the next major release to disallow it.
For us, the second option should work. It's just a matter of removing the `future=True` at that point.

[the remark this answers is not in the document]
DanielNoord: I'll create a commit that implements this!

astroid/exceptions.py
```diff
@@ -267,9 +267,23 @@ class ParentMissingError(AstroidError):
         target: The node for which the parent lookup failed.
```
DanielNoord, line 267:
```suggestion
        target: The node for which the parent lookup failed.
        missing_element: The type of attribute/method that is missing because of the missing parent
```

DanielNoord:
2 questions:
- Is the `pylint` message about `self = cast("nodes.Statement", self)` correct?
- How should I overload `nodes.Module.statement()`? I thought this was correct but both `pylint` and `mypy` complain..

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
                 return True
         return False
 
-    def statement(self):
+    @overload
+    def statement(
+        self, future: Literal[False] = ...
+    ) -> Union["nodes.Statement", "nodes.Module"]:
+        ...
```
cdce8p, line 269: I believe we should add an additional overload before that.
```py
    @overload
     def statement(self) -> Union["nodes.Statement", "nodes.Module"]:
         ...
```
This should match those cases were `future` isn't passed.

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
...
+
+    @overload
+    def statement(self, future: Literal[True]) -> Union[NoReturn, "nodes.Statement"]:
+        ...
```
cdce8p, line 273: `NoReturn` should not be a return type here. As I understand it, it should only be used to annotate functions that **never** return. I.e. unconditional `raise` or `sys.exit`.

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
...
+
+    def statement(
+        self, future: bool = False
+    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
```
cdce8p, line 277:
```suggestion
    ) -> Union["nodes.Statement", "nodes.Module"]:
```

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
...
-        :returns: The first parent statement.
-        :rtype: NodeNG
+        TODO: Deprecate the future parameter and only raise StatementMissing and return
+        nodes.Statement
+
+        :raises StatementMissing: If no self has no parent attribute
         """
         if self.is_statement:
+            self = cast("nodes.Statement", self)
             return self
```
cdce8p, line 287:
```suggestion
            return cast("nodes.Statement", self)
```
The `pylint` warning is correct. In that case it should work by retuning the result of `cast` directly.
Ideally we wouldn't even need it. It might not be worth it as we risk circular dependencies, but would
`isinstance(self, nodes.Statement)` work? Might also be a breaking change if someone depends on manually assigning `is_statement`.

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,25 @@ def fully_defined(self):
         """
         return self.file is not None and self.file.endswith(".py")
 
-    def statement(self):
+    @overload
+    def statement(self, future: Literal[False] = ...) -> "Module":
+        ...
```
cdce8p, line 656: Same here. Add an overload without `future`.

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,25 @@ def fully_defined(self):
...
+
+    @overload
+    def statement(self, future: Literal[True] = ...) -> NoReturn:
+        ...
```
cdce8p, line 660: Here it's correct.

cdce8p (commented):
Just some initial thoughts. Need to come back to it later when I've more time.
What do you think about making `future` a keyword only argument?

astroid/nodes/node_ng.py
DanielNoord: I think the `...` after `Literal[False]` does this? I played around with it and this does seem to work. VS Code picks up on it and `mypy` does as well I think.

The screenshot is from `nodes.Module`, but I think it should work similarly.
Schermafbeelding 2021-11-03 om 16 35 52

astroid/nodes/node_ng.py
DanielNoord: I added it because I thought it would create problems with `nodes.Module.statement()` if its return type did not match either of the possibilities of `NodeNG`.
`Mypy` is turned off for `astroid`, but won't it complain when a class that inherits from `NodeNG` overwrites a method and returns something that is not included in the possibilities for `NodeNG`?

astroid/nodes/node_ng.py
DanielNoord: It already creates a circular dependency, that's why I opted to use `cast`.

I thought about the `pylint` warning and couldn't figure out why it would be correct. `typing.cast` returns the second argument without any changes, so that should be fine, right?
I opened https://github.com/PyCQA/pylint/pull/5254 (prematurely). If the warning is indeed correct I will change that PR.

astroid/nodes/node_ng.py
cdce8p: Tested a few things. What do you think about this solution?
```py
    @overload
    def statement(
        self, future: Literal[None] = ...
    ) -> Union["nodes.Statement", "nodes.Module"]:
        ...

    @overload
    def statement(self, future: Literal[True]) -> "nodes.Statement":
        ...

    def statement(
        self, future: Optional[Literal[True]] = None
    ) -> Union["nodes.Statement", "nodes.Module"]:
```

And this for `Module`:
```py
    @overload
    def statement(self, future: Literal[None] = ...) -> "Module":
        ...

    @overload
    def statement(self, future: Literal[True]) -> NoReturn:
        ...

    def statement(self, future: Literal[True] | None = None) -> Union[NoReturn, "Module"]:
```

astroid/nodes/node_ng.py
cdce8p: I see what you mean. Tested it with both `mypy` and `pylance` and both seem to except it. Might be that `NoReturn` is special cased somehow.

astroid/nodes/node_ng.py
cdce8p: Note that I also changed the default to `None`.

DanielNoord: Looks good to me! I'll create a commit.

astroid/nodes/node_ng.py
DanielNoord: We learn something new every day :)

astroid/nodes/node_ng.py
DanielNoord: Did you test these with `pylint`?
I'm getting the following error with pre-commit:

```shell
************* Module astroid.nodes.scoped_nodes
astroid/nodes/scoped_nodes.py:659:4: W0222: Signature differs from overridden 'statement' method (signature-differs)
```

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,27 @@ def fully_defined(self):
...
+    @overload
+    def statement(self, future: Literal[None] = ...) -> "Module":
+        ...
+
+    @overload
+    def statement(self, future: Literal[True]) -> NoReturn:
+        ...
+
+    def statement(
+        self, future: Union[Literal[True]] = None
```
cdce8p, line 663:
```suggestion
        self, future: Optional[Literal[True]] = None
```

DanielNoord: Oops, I replaced the `|` from your suggestion. Should have payed more attention.

astroid/nodes/node_ng.py
cdce8p: Turns out, you were right all along. Just tested it with strict mode and now `pylance` complains. Should be fixed by adding `NoReturn` just to the function signature in `NodeNG` (not as part of overloads).

astroid/nodes/node_ng.py
cdce8p: I did not. It's not immediately obvious to me what's causing the issue. Need to investigate that a bit more.

astroid/nodes/scoped_nodes.py
cdce8p: I should have used the correct syntax for `astroid`.

astroid/nodes/node_ng.py
DanielNoord: Added with
https://github.com/PyCQA/astroid/pull/1217/commits/6026560846c00ade3c6f2d2f89f71331300e10e4

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
...
+        self, future: Literal[None] = ...
+    ) -> Union["nodes.Statement", "nodes.Module"]:
+        ...
+
+    @overload
+    def statement(self, future: Literal[True]) -> "nodes.Statement":
+        ...
+
+    def statement(
+        self, future: Optional[Literal[True]] = None
```
cdce8p, line 276:
```suggestion
        self, future: Literal[None, True] = None
```
Might be a bit easier to understand, especially with the first overload.

astroid/nodes/node_ng.py
```diff
@@ -248,15 +262,34 @@ def parent_of(self, node):
...
+        :raises StatementMissing: If self has no parent attribute and future is True
         """
         if self.is_statement:
-            return self
-        return self.parent.statement()
+            return cast("nodes.Statement", self)
+        if not self.parent:
+            if future:
+                raise StatementMissing(target=self)
+            raise AttributeError(f"{self} object has no attribute 'parent'")
```
cdce8p, line 291: Can we emit a `DeprecationWarning` if `future==None` and `AttributeError` is emitted?

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,27 @@ def fully_defined(self):
         """
         return self.file is not None and self.file.endswith(".py")
 
-    def statement(self):
+    @overload
+    def statement(self, future: Literal[None] = ...) -> "Module":
+        ...
+
+    @overload
+    def statement(self, future: Literal[True]) -> NoReturn:
```
cdce8p, line 659:
```suggestion
    # pylint: disable-next=signature-differs
    # https://github.com/PyCQA/pylint/issues/5264
    @overload
    def statement(self, future: Literal[True]) -> NoReturn:
```

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,27 @@ def fully_defined(self):
...
+        ...
+
+    def statement(
+        self, future: Optional[Literal[True]] = None
```
cdce8p, line 663:
```suggestion
        self, future: Literal[None, True] = None
```

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +651,27 @@ def fully_defined(self):
...
-        :rtype: NodeNG
+        When called on a :class:`Module` with the future parameter this raises an error.
+
+        TODO: Deprecate the future parameter and only raise StatementMissing
+
+        :raises StatementMissing: If no self has no parent attribute and future is True
         """
+        if future:
+            raise StatementMissing(target=self)
         return self
```
cdce8p, line 675: Same here: `DeprecationWarning`

tests/unittest_builder.py
```diff
@@ -614,7 +615,9 @@ def test_module_base_props(self) -> None:
         self.assertEqual(module.package, 0)
         self.assertFalse(module.is_statement)
         self.assertEqual(module.statement(), module)
-        self.assertEqual(module.statement(), module)
+        self.assertEqual(module.statement(False), module)
```
cdce8p, line 618: After the recent change `statement(False)` should never be called. Either `None` or `True`.

tests/unittest_nodes.py
```diff
@@ -626,6 +627,12 @@ def _test(self, value: Any) -> None:
         self.assertIs(node.value, value)
         self.assertTrue(node._proxied.parent)
         self.assertEqual(node._proxied.root().name, value.__class__.__module__)
+        with self.assertRaises(AttributeError):
+            node.statement()
+        with self.assertRaises(AttributeError):
+            node.statement(False)
```
cdce8p, line 633:
```suggestion
```

tests/unittest_scoped_nodes.py
```diff
@@ -323,6 +323,8 @@ def test_default_value(self) -> None:
     def test_navigation(self) -> None:
         function = self.module["global_access"]
         self.assertEqual(function.statement(), function)
+        self.assertEqual(function.statement(False), function)
```
cdce8p, line 326:
```suggestion
```

tests/unittest_scoped_nodes.py
```diff
@@ -821,6 +823,8 @@ def test_instance_special_attributes(self) -> None:
     def test_navigation(self) -> None:
         klass = self.module["YO"]
         self.assertEqual(klass.statement(), klass)
+        self.assertEqual(klass.statement(False), klass)
```
cdce8p, line 826:
```suggestion
```

cdce8p (commented):
I think it would be useful if we emit `DeprecationWarnings` in case `future=True` isn't passed to `statement()`. For test cases, `unittest_decorators.py` could be helpful.

astroid/exceptions.py
```diff
@@ -265,11 +265,25 @@ class ParentMissingError(AstroidError):
...
     Standard attributes:
         target: The node for which the parent lookup failed.
+        missing_element: The type of attribute/method that is missing because of the missing parent
     """
 
-    def __init__(self, target: "nodes.NodeNG") -> None:
+    def __init__(self, target: "nodes.NodeNG", missing_element: str = "Parent") -> None:
         self.target = target
-        super().__init__(message=f"Parent not found on {target!r}.")
+        super().__init__(message=f"{missing_element} not found on {target!r}.")
```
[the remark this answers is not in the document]
DanielNoord: Good idea!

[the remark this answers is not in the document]
DanielNoord: This creates a new pylint error 😅 

Is this the correct way to do this?

tests/unittest_builder.py
```diff
@@ -614,7 +615,11 @@ def test_module_base_props(self) -> None:
         self.assertEqual(module.package, 0)
         self.assertFalse(module.is_statement)
         self.assertEqual(module.statement(), module)
-        self.assertEqual(module.statement(), module)
+        with pytest.warns(DeprecationWarning) as warnings:
+            module.statement()
+            assert len(warnings) == 1
```
cdce8p, line 620:
```suggestion
        with pytest.warns(DeprecationWarning) as records:
            module.statement()
            assert len(records) == 1
```
`warnings` is also the name of the stdlib module, so it might be better to use something different instead.

tests/unittest_nodes.py
```diff
@@ -626,6 +627,12 @@ def _test(self, value: Any) -> None:
         self.assertIs(node.value, value)
         self.assertTrue(node._proxied.parent)
         self.assertEqual(node._proxied.root().name, value.__class__.__module__)
+        with self.assertRaises(AttributeError):
+            with pytest.warns(DeprecationWarning) as warnings:
+                node.statement()
+                assert len(warnings) == 1
```
cdce8p, line 633:
```suggestion
        with (
            self.assertRaises(AttributeError),
            pytest.warns(DeprecationWarning) as records
        ):
            node.statement()
            assert len(records) == 1
```

astroid/nodes/node_ng.py
```diff
@@ -248,15 +263,40 @@ def parent_of(self, node):
...
+    ) -> Union["nodes.Statement", "nodes.Module"]:
+        ...
+
+    @overload
+    def statement(self, future: Literal[True]) -> "nodes.Statement":
+        ...
+
+    def statement(
+        self, future: Literal[None, True] = None
+    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
```
cdce8p, line 278:
```suggestion
    def statement(
        self, *, future: Literal[None, True] = None
    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
```

I should have brought that up earlier but I somehow missed it. What do you think about making `future` a keyword only arg? That will make it explicit, prevent issue if we decide to add additional arguments, and make it easier to deprecate / remove.

cdce8p (commented):
I've seen that `pytest` will emit `DeprecationWarnings` for `astroid` itself. Could you add `future=True` to existing statement in `astroid`, too?
https://github.com/PyCQA/astroid/runs/4126645345?check_suite_focus=true#step:6:81

astroid/exceptions.py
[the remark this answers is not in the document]
cdce8p: We seem to be hitting quite a streak with false-positives 😂
This is valid, although at the limit of what one should reasonable do.
```py
class Parent:
    def __init__(self):
        print("Parent")

class Child(Parent):
    def __init__(self):
        print("Child")
        super().__init__()

class GrandChild(Child):
    def __init__(self):
        print("Grandchild")
        super(Child, self).__init__()  # will skip 'Child.__init__` and call 'Parent.__init__' instead

GrandChild()
```

**Links and related issues**
* https://realpython.com/python-super/#a-super-deep-dive
* https://github.com/PyCQA/pylint/issues/2903 (We might what to consider reopening this one)
* https://github.com/PyCQA/pylint/issues/4922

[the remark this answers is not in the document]
cdce8p: If you don't want to add `pylint: disable`, I think the alternative would be to add `message: Optional[str] = None` to `ParentMissingError`. And pass conditionally pass either `message` or the new constructed string to `super().__init__(...)`.

astroid/nodes/node_ng.py
```diff
@@ -248,15 +263,40 @@ def parent_of(self, node):
...
-            return self
-        return self.parent.statement()
+            return cast("nodes.Statement", self)
+        if not self.parent:
+            if future:
+                raise StatementMissing(target=self)
+            warnings.warn(
+                "In astroid 3.0.0 NodeNG.statement() will return either a nodes.Statement "
+                "or raise a StatementMissing exception. This behaviour can already be triggered "
+                "by passing 'future=True' to a statement() call.",
```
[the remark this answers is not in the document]
DanielNoord: I added something about `AttributeError` but think it might be helpful to fully explain what `statement()` will now do. Saves the maintainer/contributor changing the code some time!

DanielNoord:
> I've seen that `pytest` will emit `DeprecationWarnings` for `astroid` itself. Could you add `future=True` to existing statement in `astroid`, too?
> https://github.com/PyCQA/astroid/runs/4126645345?check_suite_focus=true#step:6:81

This started to fail some of the tests. I wanted to do this in a separate PR because I wondered if some of the functions/methods calling `statement()` are considered `public` and should thus keep raising `AttributeError` or returning `Module`.

astroid/exceptions.py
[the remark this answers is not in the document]
DanielNoord: I would be in favour of opening PyCQA/pylint#2903. I also googled online (and found the example you use) and some others. Seems that this is considered valid pythonic code.

Shall I reopen and add a `disable` + reference to the issue?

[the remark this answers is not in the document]
DanielNoord: I would be in favour of opening PyCQA/pylint#2903. I also googled online (and found the example you use) and some others. Seems that this is considered valid pythonic code.

Shall I reopen and add a `disable` + reference to the issue?

tests/unittest_nodes.py
DanielNoord: `black` failed on these combined `with` statements. It seems like it can't parse those statements together. I reverted in the latest commit but kept using `records`.

astroid/exceptions.py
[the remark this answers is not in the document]
cdce8p: Sounds good to me!

astroid/nodes/node_ng.py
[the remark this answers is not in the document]
cdce8p: 👌🏻

astroid/nodes/node_ng.py
```diff
@@ -248,15 +263,41 @@ def parent_of(self, node):
                 return True
         return False
 
-    def statement(self):
+    @overload
+    def statement(
+        self, future: Literal[None] = ...
```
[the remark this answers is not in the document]
DanielNoord: Ah forgot about that!

tests/unittest_nodes.py
cdce8p: I realized that it also requires at least Python 3.9. So even if black would work, we couldn't use it.
For reference: https://github.com/psf/black/issues/1948

This would work, but looks rather ugly with black formatting.
```py
        with self.assertRaises(AttributeError), pytest.warns(
            DeprecationWarning
        ) as records:
            node.statement()
            assert len(records) == 1
```
Let's leave it like it is then.

astroid/nodes/node_ng.py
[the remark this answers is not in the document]
DanielNoord: Done now!

astroid/exceptions.py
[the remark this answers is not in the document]
DanielNoord: Done!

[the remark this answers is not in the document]
cdce8p: Thanks!

astroid/exceptions.py
```diff
@@ -272,6 +272,24 @@ def __init__(self, target: "nodes.NodeNG") -> None:
...
+    return a node for statement().
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    def __init__(self, target: "nodes.NodeNG") -> None:
+        # See: https://github.com/PyCQA/pylint/issues/2903
+        # and: https://github.com/PyCQA/astroid/pull/1217#discussion_r744149027
+        # pylint: disable-next=bad-super-call
```
cdce8p, line 287:
```suggestion
        # pylint: disable-next=bad-super-call
        # https://github.com/PyCQA/pylint/issues/2903
        # https://github.com/PyCQA/astroid/pull/1217#discussion_r744149027
```
That's usually how I do it.

cdce8p (approved):
Overall, LGTM! Thanks @DanielNoord 🚀
A bit unfortunate, that we came across multiple false-positives in the process.

@Pierre-Sassoulas Do you want to take a final look at it? Maybe with focus on the Exception and deprecation message?

astroid/exceptions.py
DanielNoord: Oh wow, does this work?
That's good to know for https://github.com/PyCQA/pylint/issues/5258

DanielNoord: Oh wow, does this work?
That's good to know for https://github.com/PyCQA/pylint/issues/5258

DanielNoord:
Thanks for all of your help here @cdce8p!

Indeed a bit unfortunate, but I learned a lot from this PR and it should help with some of the future PR's and issues we might come across.

astroid/exceptions.py
```diff
@@ -272,6 +272,24 @@ def __init__(self, target: "nodes.NodeNG") -> None:
...
+    a node in the chain does not have a parent attribute and therefore does not
+    return a node for statement().
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    def __init__(self, target: "nodes.NodeNG") -> None:
+        # pylint: disable-next=bad-super-call
+        # https://github.com/PyCQA/pylint/issues/2903
```
Pierre-Sassoulas, line 286: I think the spirit of the check is to signal the problem to people that don't really know how to use super(). Here it's a false positive because calling the father's function is what you really want to do, but maybe in most case and especially for beginner  the check is right. If we remove this "false positive" the check will only have really obvious cases where the class called is not even an ancestor ?

Pierre-Sassoulas (approved):
I only have a small remark regarding the "bad super call" but I don't see anything wrong with the MR otherwise. To be fair I did not find the time to read the whole conversation and this seem really polished :) Look like you took care of a lot of problems, thanks a lot for handling this !

astroid/nodes/scoped_nodes.py
```diff
@@ -637,12 +652,36 @@ def fully_defined(self):
         """
         return self.file is not None and self.file.endswith(".py")
 
-    def statement(self):
+    @overload
+    def statement(self, *, future: Literal[None] = ...) -> "Module":
+        ...
+
+    # pylint: disable-next=signature-differs
+    # https://github.com/PyCQA/pylint/issues/5264
```
cdce8p, line 660:
```suggestion
```
It seems like your luck didn't outran us yet. This actual now triggers a `useless-suppression` message. Guess we found a `false-negative` by adding `*` 😅 https://github.com/PyCQA/pylint/issues/5270

astroid/exceptions.py
cdce8p: 🤔 Not sure about that. What it comes done to is if we want to emit errors for valid Python code. Even if it's probably not what was intended, there are still situations where it's useful and has it's place, especially with multi-inheritance.

The check also emits errors in cases that could cause recursion loops:
```py
class SuperWithType(object):
    """type(self) may lead to recursion loop in derived classes"""
    def __init__(self):
        super(type(self), self).__init__()  # [bad-super-call]

class SuperWithSelfClass(object):
    """self.__class__ may lead to recursion loop in derived classes"""
    def __init__(self):
        super(self.__class__, self).__init__()  # [bad-super-call]
```
Any modifications should also make sure to still emit errors if the class is not part of the MRO.
```py

class Getattr(object):
    """ crash """
    name = NewAaaa

class CrashSuper(object):
    """ test a crash with this checker """
    def __init__(self):
        super(Getattr.name, self).__init__()  # [bad-super-call]
```

Test cases taken from the `pylint` testsuit: [super_checks.py](https://github.com/PyCQA/pylint/blob/0a1ebd488fcdf7bf306615aba29e401ce49e3e10/tests/functional/s/super/super_checks.py)

Pierre-Sassoulas: OK, I was trying to search for the initial issue where this was implemented for the rational but I can't find it. It makes sense to not warn for valid code, and the check has a lot of valid test cases, so let's fix this false positive.

astroid/nodes/scoped_nodes.py
DanielNoord: Is `useless-suppression` not enabled for `astroid`? CI is passing.

We might want to enable that in another PR.

cdce8p: It's enabled, but not a failure condition! I don't think we should change that.

DanielNoord: Sure? Seems like something that will be easily missed in review

Pierre-Sassoulas: I also think we should make it a failure condition so it's handled in the proper MR each time.

cdce8p (approved):
Let's (squash) merge it!

astroid/nodes/scoped_nodes.py
cdce8p: > We might want to enable that in another PR.

It would be as easy as adding `--fail-on=useless-suppression` to the `pylint` command.

The issue is that some errors might be Python version specific. It could create a situation where only one, `pre-commit` or `CI`, will succeed when tested with different versions. The `py-version` option should help, but I don't know if it's enough.

Additionally, this would prevent us from having the latest `pylint` version installed in a local env as this is likely to have fixed some false-positives. -> Which would then trigger `useless-suppression`.

DanielNoord: That last point will indeed be difficult to avoid. I'll try and think of something to fix that, but in the meantime let's let this rest!

cdce8p: Do you maybe want to track this as issue in `pylint`?

> in the meantime let's let this rest!

I completely agree!

DanielNoord: Shouldn't this be tracked in `astroid`? `pylint` has its own issues with `useless-suppression` (https://github.com/PyCQA/pylint/issues/5040) 😄

cdce8p: 😂 Though if we enable it, we'll likely want to do it for `pylint` too. Tbh I don't really care about it due to the issues mentioned earlier.

PR #1233, opened 2021-11-06
Improve filtering of ``NamedExpr``, particularly within ``If`` nodes

DanielNoord:
## Steps

## Description

The crash discussed in the issue was (I think) the result of the original code only doing `_stmts = [node]` without changing `_stmt_parents`.

I did this, but while working on this I also found a number of other issues related to if-statements and assignment expressions. To fix this I have added the `If.is_orelse` attribute and added some basic control flow checks. This should improve the inference of these nodes overall.

## Type of Changes

## Related Issue

PyCQA/pylint#5178

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -486,8 +486,28 @@ class LookupMixIn:
                 continue
 
             if isinstance(assign_type, NamedExpr):
-                _stmts = [node]
-                continue
+                # If the NamedExpr is in an if statement we do some basic control flow inference
+                if_parent = _get_if_statement_ancestor(assign_type)
+                if if_parent:
+                    # If the if statement is within another if statement we append the node
+                    # to possible statements
+                    if _get_if_statement_ancestor(if_parent):
+                        optional_assign = False
+                        _stmts.append(node)
+                        _stmt_parents.append(stmt.parent)
+                    # If the if statement is first-level and not within an orelse block
+                    # we know that it will be evaluated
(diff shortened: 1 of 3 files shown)


tests/unittest_regrtest.py
```diff
@@ -156,6 +159,38 @@ class B(compiler.__class__):
         base = next(result._proxied.bases[0].infer())
         self.assertEqual(base.name, "int")
 
+    @pytest.mark.skipif(not PY38_PLUS, reason="needs assignment expressions")
```
Pierre-Sassoulas, line 162: Congrats ! First try this time !

astroid/nodes/node_classes.py
```diff
@@ -4795,3 +4825,11 @@ def is_from_decorator(node):
         if isinstance(parent, Decorators):
             return True
     return False
+
+
+def _get_if_statement_ancestor(node: NodeNG) -> Optional[If]:
+    """Return True if the given node is the child of a If node"""
```
Pierre-Sassoulas, line 4831:
```suggestion
    """Return the first parent node that is a if Node (or None)"""
```

Pierre-Sassoulas (approved):
LGTM !

tests/unittest_regrtest.py
DanielNoord: I guess you didn't see the third commit 😄

cdce8p:
Thanks @DanielNoord 👍🏻

PR #1234, opened 2021-11-06
Fix crash on inference of ``__len__``

DanielNoord:
## Steps

## Description

This is a weird one.

I found this while working on the primer tests for `pylint` where running on the codebase of `numpy` created a crash. I issued https://github.com/PyCQA/pylint/issues/5244 for this. Initially I thought it had something to do with the brain for `numpy` as the crash only occurred when the directory name starts with `numpy`.
However, removing this single line seems to fix the issue. Furthermore, (locally) all tests for `astroid` and `pylint` pass with this change. I have looked at the commit that added this line (https://github.com/PyCQA/astroid/commit/25384d4bebf0187b6704c818c7df64945793362c) but did not immediately understand why this was added. In any case, no tests seems to have been added or this line is redundant and the removing it is the way forward.

## Type of Changes

## Related Issue

Would close https://github.com/PyCQA/pylint/issues/5244

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,10 @@ What's New in astroid 2.8.5?
 ============================
 Release date: TBA
 
+* Fix crash on inference of ``__len__``
+
+  Closes PyCQA/pylint#5244
+
 * Added missing ``kind`` (for ``Const``) and ``conversion`` (for ``FormattedValue``) fields to repr.
 
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,7 @@ What's New in astroid 2.8.5?
 ============================
 Release date: TBA
 
-* Fix crash on inference of ``__len__``
+* Fix crash on inference of ``__len__``.
 
   Closes PyCQA/pylint#5244
 
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -3113,7 +3113,7 @@ def test_no_recursionerror_on_self_referential_length_check() -> None:
     """
     Regression test for https://github.com/PyCQA/astroid/issues/777
 
-    This inference should only raise an InferenceError and no RecursionError
+    This test should only raise an InferenceError and no RecursionError.
     """
     with pytest.raises(InferenceError):
         node = astroid.extract_node(
@@ -3133,7 +3133,7 @@ def test_inference_on_outer_referential_length_check() -> None:
     Regression test for https://github.com/PyCQA/pylint/issues/5244
     See also https://github.com/PyCQA/astroid/pull/1234
 
-    This inference should succeed without any error
+    This test should succeed without any error.
     """
     node = astroid.extract_node(
         """
@@ -3160,7 +3160,7 @@ def test_no_attributeerror_on_self_referential_length_check() -> None:
     Regression test for https://github.com/PyCQA/pylint/issues/5244
     See also https://github.com/PyCQA/astroid/pull/1234
 
-    This inference should only raise an InferenceError and no AttributeError
+    This test should only raise an InferenceError and no AttributeError.
     """
     with pytest.raises(InferenceError):
         node = astroid.extract_node(
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -3112,6 +3112,8 @@ def test_str_and_bytes(code, expected_class, expected_value):
 def test_no_recursionerror_on_self_referential_length_check() -> None:
     """
     Regression test for https://github.com/PyCQA/astroid/issues/777
+
+    This inference should only raise an InferenceError and no RecursionError
     """
     with pytest.raises(InferenceError):
         node = astroid.extract_node(
@@ -3129,6 +3131,9 @@ def test_no_recursionerror_on_self_referential_length_check() -> None:
 def test_inference_on_outer_referential_length_check() -> None:
     """
     Regression test for https://github.com/PyCQA/pylint/issues/5244
+    See also https://github.com/PyCQA/astroid/pull/1234
+
+    This inference should succeed without any error
     """
     node = astroid.extract_node(
         """
@@ -3153,6 +3158,9 @@ def test_inference_on_outer_referential_length_check() -> None:
 def test_no_attributeerror_on_self_referential_length_check() -> None:
     """
     Regression test for https://github.com/PyCQA/pylint/issues/5244
+    See also https://github.com/PyCQA/astroid/pull/1234
+
+    This inference should only raise an InferenceError and no AttributeError
     """
     with pytest.raises(InferenceError):
         node = astroid.extract_node(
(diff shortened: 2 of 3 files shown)


DanielNoord:
I'm marking this as a draft because I'm not 100% sure this is the right way forward, but I would like some input for any of the experienced `astroid` maintainers. I can't really continue this on my own.

Pierre-Sassoulas:
@brycepg @hippo91 do you have an opinion on this change ?

hippo91:
@DanielNoord @Pierre-Sassoulas i had a look to this PR. Honestly i do not understand why this line was added.
Maybe @brycepg will be more helpful than me. 
IMHO if this line is not tested and all `astroid` and `pylint` tests are ok we can remove it.
Maybe we can ask @cdce8p to test it again home-assistant?

Pierre-Sassoulas:
> Maybe we can ask @cdce8p to test it again home-assistant?

We're also putting primer tests in place in https://github.com/PyCQA/pylint/pull/5173 so maybe we'll get to be more confident in small changes like this in the near future :)

cdce8p (changes requested):
I had some time to review this PR. A few notes.

The `_proxied` comparison is necessary. Otherwise this would also match.
```py
class A:
    def __len__(self) -> int:
        return 42

class Crash:
    def __len__(self) -> int:
        a = A()
        return len(a)
```

Furthermore, I was able to reproduce the issue without `numpy`.
```py
class MyClass:
    def some_func(self):
        return lambda: 42

    def __len__(self):
        return len(self.some_func())
```

Those two tests could probably be added to `unittest_brain.py`, similar to the existing one.
https://github.com/PyCQA/astroid/blob/b62f243b16b7f435c8be869577959e95a7927a91/tests/unittest_brain.py#L3112-L3126

--
As for `numpy`, the whole issue was only triggered because `self.array` was inferred as numpy `FunctionDef`. Not sure this is correct in this instance or if the `numpy` brain needs to be updated. Maybe @hippo91 know more about that? Anyway, if it needs to be updated, this should happen in another PR.
https://github.com/PyCQA/astroid/blob/b62f243b16b7f435c8be869577959e95a7927a91/astroid/brain/brain_numpy_core_multiarray.py#L41-L42
https://github.com/PyCQA/astroid/blob/b62f243b16b7f435c8be869577959e95a7927a91/astroid/brain/brain_numpy_utils.py#L90

DanielNoord:
Fixed the code with @cdce8p's suggestions and added test cases. The last test will fail without the changes in this PR, the first test is to make sure the inference on these kind of functions actually works.

As for the `numpy` error, I have not looked into it but we might want to open a separate issue to channel discussion. @cdce8p did you happen to find any hints as to why this inference was (probably) wrong? And if so, would you mind opening an issue?

tests/unittest_brain.py
```diff
@@ -3126,5 +3126,50 @@ def __len__(self) -> int:
         node.inferred()
 
 
+def test_inference_on_outer_referential_length_check() -> None:
+    """
+    Regression test for https://github.com/PyCQA/pylint/issues/5244
+    """
```
cdce8p, line 3132: Can you add a small note here that the inference should succeed? Might make it a bit more obvious.

tests/unittest_brain.py
```diff
@@ -3126,5 +3126,50 @@ def __len__(self) -> int:
...
+    inferred = node.inferred()
+    assert len(inferred) == 1
+    assert isinstance(inferred[0], nodes.Const)
+    assert inferred[0].value == 42
+
+
+def test_no_attributeerror_on_self_referential_length_check() -> None:
+    """
+    Regression test for https://github.com/PyCQA/pylint/issues/5244
+    """
```
cdce8p, line 3156: And here that the code should only ever raise an `InferenceError` and not `AttributeError`.

cdce8p (commented):
Just some last comments. I know that the test names are already descriptive, but adding one more line in the docstring might be useful.

DanielNoord:
Added docstrings, also for the existing test

ChangeLog
```diff
@@ -12,6 +12,10 @@ What's New in astroid 2.8.5?
 ============================
 Release date: TBA
 
+* Fix crash on inference of ``__len__``
```
cdce8p, line 15:
```suggestion
* Fix crash on inference of ``__len__``.
```

tests/unittest_brain.py
```diff
@@ -3112,6 +3112,8 @@ def test_str_and_bytes(code, expected_class, expected_value):
 def test_no_recursionerror_on_self_referential_length_check() -> None:
     """
     Regression test for https://github.com/PyCQA/astroid/issues/777
+
+    This inference should only raise an InferenceError and no RecursionError
```
cdce8p, line 3116:
```suggestion
    This test should only raise an InferenceError and no RecursionError.
```

tests/unittest_brain.py
```diff
@@ -3126,5 +3128,56 @@ def __len__(self) -> int:
         node.inferred()
 
 
+def test_inference_on_outer_referential_length_check() -> None:
+    """
+    Regression test for https://github.com/PyCQA/pylint/issues/5244
+    See also https://github.com/PyCQA/astroid/pull/1234
+
+    This inference should succeed without any error
```
cdce8p, line 3136:
```suggestion
    This test should succeed without any error.
```

tests/unittest_brain.py
```diff
@@ -3126,5 +3128,56 @@ def __len__(self) -> int:
...
+    assert isinstance(inferred[0], nodes.Const)
+    assert inferred[0].value == 42
+
+
+def test_no_attributeerror_on_self_referential_length_check() -> None:
+    """
+    Regression test for https://github.com/PyCQA/pylint/issues/5244
+    See also https://github.com/PyCQA/astroid/pull/1234
+
+    This inference should only raise an InferenceError and no AttributeError
```
cdce8p, line 3163:
```suggestion
    This test should only raise an InferenceError and no AttributeError.
```

cdce8p (approved):

cdce8p:
As for the numpy issue, I don't think it's worth opening an issue for it as long as we don't encounter it in practice.
Maybe @hippo91 knows the answer to it, but not too important if not.

--
File placed in a folder which starts with `numpy`.
```py
class MyClass:
    def __init__(self, array):
        self.array = array
        self.array  #@
```

`self.array` will (wrongly ?) be inferred as `FunctionDef`.
It seems to be related to this https://github.com/PyCQA/astroid/blob/b62f243b16b7f435c8be869577959e95a7927a91/astroid/brain/brain_numpy_core_multiarray.py#L41-L42
https://github.com/PyCQA/astroid/blob/e1ecb6d3d5a6555a807ab3db0f2f938d8d73e95c/astroid/brain/brain_numpy_core_multiarray.py#L90-L95

Issue #1072, opened 2021-06-23
Remove deprecated ``astroid.node_classes`` and ``astroid.scoped_nodes``


Pierre-Sassoulas:
Those two files have been deprecated in 2.7.0 and need to be removed in 3.0.0.

belm0:
astroid seems to be triggering its own deprecation warnings, which is quite annoying

https://github.com/PyCQA/astroid/blob/4ffdf1108a6084b85e282b5835427c1d747bb22c/astroid/__init__.py#L54-L55

```
  File "/.../site-packages/astroid/__init__.py", line 54, in <module>
    from astroid import node_classes  # Deprecated, to remove later
  File "/.../site-packages/astroid/node_classes.py", line 90, in <module>
    warnings.warn(
DeprecationWarning: The 'astroid.node_classes' module is deprecated and will be replaced by 'astroid.nodes' in astroid 3.0.0
```

Pierre-Sassoulas:
@belm0 thank you for the coment, I opened #1190 to fix that, should be available in 2.8.1.

PR #1235, opened 2021-11-07
Add ``future`` argument to all ``NodeNG.statement()`` calls

### LINKED ISSUE #1239: TypeError: Plain typing.NoReturn is not valid as type argument in Astroid 2.8.5 + Python 3.7.0/1
### Steps to reproduce

1. Use Python 3.7.0 or 3.7.1
2. pip install astroid==2.8.5
3. python -c "import astroid"

### Current behavior

An error is raised
```
TypeError: Plain typing.NoReturn is not valid as type argument
```

In this commit https://github.com/PyCQA/astroid/commit/b62f243b16b7f435c8be869577959e95a7927a91#diff-039e09546480e9ce70fc9a23e975f1ac2364f6d0192e330311d204653947a216R34 there is a conditional import that checks if Python >= 3.6.2, but that should also include Python 3.7.0 and 3.7.1 (which have the same issue as Python 3.6.0/1).

### Expected behavior

There should be no error

### python -c "from astroid import __pkginfo__; print(__pkginfo__.version)" output

This raises the same error as above

DanielNoord:
## Steps

## Description

Follow up to #1217

This makes all calls to `NodeNG.statement` use the `future` keyword argument added in #1217.

There is one case where this is not as straightforward because the code depends on `nodes.Module.statement()` returning itself. This is why I have marked this as a draft. I will comment on the relevant line.

Closes #1239 

## Type of Changes

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -54,7 +54,6 @@ from astroid.exceptions import (
     InferenceError,
     NoDefault,
     ParentMissingError,
-    StatementMissing,
 )
 from astroid.manager import AstroidManager
 from astroid.nodes.const import OP_PRECEDENCE
@@ -445,16 +444,15 @@ class LookupMixIn:
             ):
                 myframe = myframe.parent.frame()
 
-        # nodes.Module don't have a parent attribute
-        try:
+        mystmt = None
+        if self.parent:
             mystmt = self.statement(future=True)
-        except StatementMissing:
-            mystmt = self
+
         # line filtering if we are in the same frame
         #
         # take care node may be missing lineno information (this is the case for
         # nodes inserted for living objects)
-        if myframe is frame and mystmt.fromlineno is not None:
+        if myframe is frame and mystmt and mystmt.fromlineno is not None:
             assert mystmt.fromlineno is not None, mystmt
             mylineno = mystmt.fromlineno + offset
         else:
@@ -555,7 +553,7 @@ class LookupMixIn:
                         _stmt_parents = []
                     else:
                         continue
-                elif not optional_assign and stmt.parent is mystmt.parent:
+                elif not optional_assign and mystmt and stmt.parent is mystmt.parent:
                     _stmts = []
                     _stmt_parents = []
             elif isinstance(node, DelName):
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -54,6 +54,7 @@ from astroid.exceptions import (
     InferenceError,
     NoDefault,
     ParentMissingError,
+    StatementMissing,
 )
 from astroid.manager import AstroidManager
 from astroid.nodes.const import OP_PRECEDENCE
@@ -387,7 +388,7 @@ class LookupMixIn:
         return _infer_stmts(stmts, context, frame)
 
     def _get_filtered_node_statements(self, nodes):
-        statements = [(node, node.statement()) for node in nodes]
+        statements = [(node, node.statement(future=True)) for node in nodes]
         # Next we check if we have ExceptHandlers that are parent
         # of the underlying variable, in which case the last one survives
         if len(statements) > 1 and all(
@@ -437,10 +438,18 @@ class LookupMixIn:
             #
             # def test(b=1):
             #     ...
-
-            if self.statement() is myframe and myframe.parent:
+            if (
+                self.parent
+                and self.statement(future=True) is myframe
+                and myframe.parent
+            ):
                 myframe = myframe.parent.frame()
-        mystmt = self.statement()
+
+        # nodes.Module don't have a parent attribute
+        try:
+            mystmt = self.statement(future=True)
+        except StatementMissing:
+            mystmt = self
         # line filtering if we are in the same frame
         #
         # take care node may be missing lineno information (this is the case for
@@ -1816,7 +1825,7 @@ class Comprehension(NodeNG):
             if isinstance(lookup_node, (Const, Name)):
                 return [lookup_node], True
 
-        elif self.statement() is mystmt:
+        elif self.statement(future=True) is mystmt:
             # original node's statement is the assignment, only keeps
             # current node (gen exp, list comp)
 
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -389,9 +389,7 @@ class LookupMixIn:
     def _get_filtered_node_statements(
         self, nodes: typing.List[NodeNG]
     ) -> typing.List[typing.Tuple[NodeNG, Statement]]:
-        statements: typing.List[typing.Tuple[NodeNG, Statement]] = [
-            (node, node.statement(future=True)) for node in nodes
-        ]
+        statements = [(node, node.statement(future=True)) for node in nodes]
         # Next we check if we have ExceptHandlers that are parent
         # of the underlying variable, in which case the last one survives
         if len(statements) > 1 and all(
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1846,7 +1846,7 @@ class Const(mixins.NoChildrenMixin, NodeNG, Instance):
     <Const.bytes l.1 at 0x7f23b2e35a20>]
     """
 
-    _other_fields = ("value",)
+    _other_fields = ("value", "kind")
 
     def __init__(
         self,
@@ -4076,6 +4076,7 @@ class FormattedValue(NodeNG):
     """
 
     _astroid_fields = ("value", "format_spec")
+    _other_fields = ("conversion",)
 
     def __init__(
         self,
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -24,8 +24,8 @@
 # Copyright (c) 2020 Raphael Gaschignard <raphael@rtpg.co>
 # Copyright (c) 2020 Bryce Guinta <bryce.guinta@protonmail.com>
 # Copyright (c) 2021 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
-# Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
+# Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
 # Copyright (c) 2021 David Liu <david@cs.toronto.edu>
 # Copyright (c) 2021 Alphadelta14 <alpha@alphaservcomputing.solutions>
 # Copyright (c) 2021 Andrew Haigh <hello@nelf.in>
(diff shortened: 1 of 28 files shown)


astroid/nodes/node_classes.py
```diff
@@ -437,10 +438,18 @@ def _filter_stmts(self, stmts, frame, offset):
...
+                and myframe.parent
+            ):
                 myframe = myframe.parent.frame()
-        mystmt = self.statement()
+
+        # nodes.Module don't have a parent attribute
+        try:
+            mystmt = self.statement(future=True)
+        except StatementMissing:
+            mystmt = self
```
DanielNoord, line 452: @cdce8p These are the lines I referred to. Both calls to `self.statement()` will raise a `StatementMissing` exception when called on a `nodes.Module`. By adding a checking for `self.parent`, which `nodes.Module` don't have and a `try ... except` we can catch this.
However, similar to the actual change in #1217 I wonder if `mystmt` should actually become a `nodes.Module` in the first place (as it would still be with this change). You probably have a better idea if this is the way to go about fixing this.

Note that `nodes.Module` can't be imported due to circular imports so we can't do `isinstance()` here.

tests/unittest_builder.py
```diff
@@ -614,7 +614,9 @@ def test_module_base_props(self) -> None:
         self.assertEqual(module.pure_python, 1)
         self.assertEqual(module.package, 0)
         self.assertFalse(module.is_statement)
-        self.assertEqual(module.statement(), module)
+        with pytest.warns(DeprecationWarning) as records:
+            self.assertEqual(module.statement(), module)
+            assert len(records) == 1
         with pytest.warns(DeprecationWarning) as records:
             module.statement()
             assert len(records) == 1
```
DanielNoord, line 622:
```suggestion
```

astroid/nodes/node_classes.py
cdce8p: I had some time to think about it. Did I make made a mistake by suggesting to raise an error if `statement()` is called on `Module`?

Some points to consider
* `statement()` is often used together with `frame()` and thus doesn't quite represent an actual `ast.Statement`. It's rather used to compare if the `statement()` of a node is also equal to the `frame()`. Thus returning `Module` made sense in that context.
* During a normal `pylint` run you didn't encounter an `AttributeError` as a parent had always been defined or it was a `Module` which returned itself. To update `pylint`, we would need to add exception handling in at least some / most (?) cases.
* Just from looking at the changes here, I'm not sure it's an actual improvement.

@DanielNoord What do you think?

cdce8p: @DanielNoord Sorry to ping you again so soon. If we follow though and revert the change, I think if would be good to do so before pylint `2.12.0` is released. Reverting it would also address #1239.

/CC: @Pierre-Sassoulas

Pierre-Sassoulas: If we need to revert that we might as well do it quick as the deprecation warning exists in 2.8.5 that has been released, it's better if it's reverted fast. I did not follow this close enough to have an opinion about it, so I trust your decision on this. Re-releasing astroid with a revert is cheap.

cdce8p: I would recommend doing that. Would like to know what @DanielNoord thinks first as he also spend considerable time with the original change. One or two days more doesn't really matter too much, just the issue with the `2.12` release.

DanielNoord: I haven't gotten the time to look at this sufficiently. Should be able to do so tomorrow.

My initial feeling is that we should probably try and find a way to allow the precious behaviour but not in `statement`. `statement` returning Modules is completely unexpected based on the methods name and that confusion is what caused this in the first place. I think it is valuable to avoid such unlogical behaviour. 

I haven't looked into why we need to return modules exactly in the first place, @cdce8p already mentioned it but haven't fully read the initial comment. But if this is indeed a valid use case, perhaps we need a `module_and_statement()`?

cdce8p: I do agree. The name `statement` isn't ideal.
However as for changing it, I don't think it's worth it. We don't gain anything and just break code in the process. If there were a clear advantage, maybe. But even then there is a point to be made that the function should stay the same and we should instead add a new one that only returns actual Statement nodes. That would be backwards compatible.

DanielNoord: Just as this PR my response is WIP. But;

I have looked at `_filter_stmts`, which is the only function causing problems in `astroid` with `future=True`.
https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_classes.py#L401

`mystmt` is used on some occasions, which I will discuss to show what happens when `mystmt` is in fact a `nodes.Module`.
https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_classes.py#L448

`nodes.Module.fromlineno` is always `0`. Therefore, this check will never be `True`.

https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_classes.py#L465

`stmt` will never be a `nodes.Module` since `_filter_stmts` is only called with `self.locals[name]`. This will never include a `nodes.Module` so this check will never be `True`.

https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_classes.py#L476

`_get_filtered_stmts` is not defined on `nodes.Module`. All three definitions of `_get_filtered_stmts` rely on comparing the `self` of `_get_filtered_stmts` to `mystmt` or calling `self.statement()` and comparing to `mystmt`. Since `self` is never `nodes.Module` this will never return `True` and `done` will never be `True`.

https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_classes.py#L569

Since `nodes.Module` doesn't have a `parent` this will never be `True`.


So, to me it seems that the return of `self` for `nodes.Module` is done to avoid the `AttributeError` but is not meaningful in any way. Initialising `mystmt` as `None` and only reassigning `if self.parent` as I do in the commit I just added seems fine for `astroid`.

I tested these changes locally with `pylint` and all tests pass.

----

Then there is the issue of whether the behaviour of `future` is compatible with `pylint`. I am going to investigate this now and see whether we really need `statement()` to return `nodes.Module`. I agree that adding `try ... except` everywhere would not be good, but seeing as the change to `astroid` is relatively simple I think we might get away with it.

DanielNoord: I think https://github.com/PyCQA/pylint/pull/5310 shows that the effect of no longer returning `nodes.Module` for `pylint` is minimal as well.
That PR adds one `isinstance(nodes.Module)` (related to `nodes.Module.fromlineno`), one `if XXX.parent` and changes one `except AttributeError` to `except StatementMissing`. I tested that PR against the latest commit of `astroid` in this branch and all tests passed.

---

Based on this I think we do not need to revert the earlier change.

We never need `statement()` to return `nodes.Module`, both in `pylint` and in `astroid`. There is no if statement in any of the projects that evaluate to `True` whenever it returns a `nodes.Module`. I think the `return self` was added to avoid creating crashes when `AttributeError` was raised or might have served a purpose it no longer does.

Pierre-Sassoulas (approved):
Very nice ! I just realized that we should have done that before releasing 2.8.5 as we already had this issue [signaled to us previousely](https://github.com/PyCQA/astroid/issues/1072#issuecomment-925471310)

cdce8p (commented):
I don't have the bandwidth at the moment to think about this further. It looks like you have done the research and it'll work out. So let's finish it then.

Could you also add quotes around `NoReturn` in the two places linked below? That would fix #1239  https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/node_ng.py#L278
https://github.com/PyCQA/astroid/blob/30711bc012592844a902d1d5912b3b2b6f2dce0f/astroid/nodes/scoped_nodes.py#L665

astroid/nodes/node_ng.py
```diff
@@ -32,7 +32,7 @@
     from astroid import nodes
 
 if sys.version_info >= (3, 6, 2):
-    from typing import NoReturn
+    from typing import NoReturn  # pylint: disable=unused-import
```
cdce8p, line 35: Should we move the import in the `if TYPE_CHECKING` block instead?

DanielNoord: Doesn't work sadly. The `TYPE_CHECKING` check for `unused-import` only checks one level. I'll investigate and open an issue/create a fix.

cdce8p: You could remove the `sys.version_info` check then. We likely won't run `mypy` with Python < `3.6.2` anyway. At runtime `TYPE_CHECKING` is always `False`.

cdce8p: But yeah, maybe we'll need to update the detection logic in `pylint`.

Pierre-Sassoulas: > We likely won't run mypy with Python < 3.6.2

Should we change the required python version like in pylint ?

cdce8p: > > We likely won't run mypy with Python < 3.6.2
> 
> Should we change the required python version like in pylint ?

Isn't that already planned (on some todo list)?

Pierre-Sassoulas: Yes, I also realized I released 2.8.5 without taking care of that.

Pierre-Sassoulas: Tracked this here now : https://github.com/PyCQA/pylint/issues/5250

astroid/nodes/node_classes.py
```diff
@@ -386,8 +386,12 @@ def ilookup(self, name):
         context = InferenceContext()
         return _infer_stmts(stmts, context, frame)
 
-    def _get_filtered_node_statements(self, nodes):
-        statements = [(node, node.statement()) for node in nodes]
+    def _get_filtered_node_statements(
+        self, nodes: typing.List[NodeNG]
+    ) -> typing.List[typing.Tuple[NodeNG, Statement]]:
+        statements: typing.List[typing.Tuple[NodeNG, Statement]] = [
```
cdce8p, line 392:
```suggestion
        statements = [
```
The type can be inferred from the list comprehension. It's only really necessary for empty lists, if the type can't be inferred, or if you would like to accept a parent class.

```py
l: list[Base] = [childA]
l.append(childB)
```
Usually `l` would have been inferred as `list[ChildA]` and thus the `append` call would have failed.

astroid/nodes/node_ng.py
```diff
@@ -275,7 +276,7 @@ def statement(self, *, future: Literal[True]) -> "nodes.Statement":
 
     def statement(
         self, *, future: Literal[None, True] = None
-    ) -> Union["nodes.Statement", "nodes.Module", NoReturn]:
+    ) -> Union["nodes.Statement", "nodes.Module", "NoReturn"]:
```
cdce8p, line 279: If this PR should wait for the primer tests, maybe it's worth addressing the `NoReturn` issue separately before then.

cdce8p (commented):
Overall this looks good. Could you add a one line regarding the `NoReturn` fix? Either here or, as mentioned in the comment, in a new PR.

astroid/nodes/node_ng.py
DanielNoord: See #1252

DanielNoord:
@Pierre-Sassoulas Should we do this before the primer? Or do you want to wait?

astroid/nodes/node_ng.py
cdce8p: Left a small comment there. I think the separation does make sense weather we continue or wait with this one. Thanks for doing it!

DanielNoord:
@Pierre-Sassoulas Now that `primer` is in place I think this can be merged! I'll leave that to you :)

PR #1242, opened 2021-11-16
Add common caching class and global cache flushing function

### LINKED ISSUE #792: Continuously increasing memory usage when pylint is run via its API

keichi:
## Description

This PR adds an LRU cache class `astroid.cache.LRUCache` and modifies the inference cache, inference tip cache, generator cache and `@lru_cache` to use this new cache. `LRUCache` is bounded by default and can also be manually flushed by calling `astroid.cache.clear_caches()` to address the memory leak issue discussed in #792.

## Type of Changes

## Related Issue

Closes #792

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -234,16 +234,6 @@ Release date: 2021-11-21
 
   Closes #1239
 
-* Resolve symlinks in the import path
-  Fixes inference error when the import path includes symlinks (e.g. Python
-  installed on macOS via Homebrew).
-
-  Closes #823
-  Closes PyCQA/pylint#3499
-  Closes PyCQA/pylint#4302
-  Closes PyCQA/pylint#4798
-  Closes PyCQA/pylint#5081
-
 
 What's New in astroid 2.8.5?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,8 @@ What's New in astroid 2.11.3?
 =============================
 Release date: TBA
 
+* Add a common caching mechanism (`astroid.cache.LRUCache`) and a global
+  cache flushing function (``astroid.cache.clear_caches()``).
 
 
 What's New in astroid 2.11.2?
@@ -231,6 +233,16 @@ Release date: 2021-11-21
 * Fix bug with Python 3.7.0 / 3.7.1 and ``typing.NoReturn``.
 
   Closes #1239
+
+* Resolve symlinks in the import path
+  Fixes inference error when the import path includes symlinks (e.g. Python
+  installed on macOS via Homebrew).
+
+  Closes #823
+  Closes PyCQA/pylint#3499
+  Closes PyCQA/pylint#4302
+  Closes PyCQA/pylint#4798
+  Closes PyCQA/pylint#5081
 
 
 What's New in astroid 2.8.5?
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -47,7 +47,7 @@ class LRUCache:
             cache.clear()
 
 
-def lru_cache(arg=None):
+def lru_cache_astroid(arg=None):
     """A decorator to cache the results of a function. Similar to
     functools.lru_cache but uses astroid.cache.LRUCache as its internal cache.
     """
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -1,0 +1,50 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
+import functools
+
+import wrapt
+
+LRU_CACHES = set()
+GENERATOR_CACHES = {}
+INFERENCE_CACHE = {}
+
+
+@wrapt.decorator
+def cached_generator(func, instance, args, kwargs):
+    node = args[0]
+    try:
+        result = GENERATOR_CACHES[func, node]
+    except KeyError:
+        result = GENERATOR_CACHES[func, node] = list(func(*args, **kwargs))
+    return iter(result)
+
+
+def lru_cache(maxsize=128, typed=False):
+    if maxsize is None:
+        maxsize = 128
+
+    def decorator(f):
+        cached_func = functools.lru_cache(maxsize, typed)(f)
+
+        LRU_CACHES.add(cached_func)
+
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            return cached_func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def clear_inference_cache():
+    INFERENCE_CACHE.clear()
+
+
+def clear_caches():
+    for c in LRU_CACHES:
+        c.cache_clear()
+
+    GENERATOR_CACHES.clear()
+    INFERENCE_CACHE.clear()
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -19,7 +19,7 @@ V = typing.TypeVar("V")
 class LRUCache(typing.Generic[K, V]):
     """An LRU cache that keeps track of its instances."""
 
-    instances: WeakSet["LRUCache[typing.Any, typing.Any]"] = WeakSet()
+    instances: "WeakSet[LRUCache[typing.Any, typing.Any]]" = WeakSet()
 
     def __init__(self) -> None:
         self.cache: typing.OrderedDict[K, V] = OrderedDict()
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -83,37 +83,6 @@ def lru_cache_astroid(arg: typing.Optional[F] = None) -> F:
     return typing.cast(F, decorator)
 
 
-_GENERATOR_CACHE: LRUCache[typing.Any, typing.Any] = LRUCache()
-
-
-def cached_generator(arg: typing.Optional[F] = None) -> F:
-    """A decorator to cache the elements returned by a generator. The input
-    generator is consumed and cached as a list.
-    """
-
-    @wrapt.decorator  # type: ignore[misc] # wrapt.decorator is untyped
-    def decorator(
-        func: F,
-        instance: typing.Any,
-        args: typing.Tuple[typing.Any, ...],
-        kwargs: typing.Dict[str, typing.Any],
-    ) -> typing.Any:
-        key = func, args[0]
-
-        if key in _GENERATOR_CACHE:
-            result = _GENERATOR_CACHE[key]
-        else:
-            result = _GENERATOR_CACHE[key] = list(func(*args, **kwargs))
-
-        return iter(result)
-
-    if callable(arg):
-        # pylint: disable=no-value-for-parameter
-        return typing.cast(F, decorator(arg))
-
-    return typing.cast(F, decorator)
-
-
 def clear_caches() -> None:
     """Clears all caches."""
     LRUCache.clear_all()
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -1,13 +1,65 @@
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
-import functools
+
+from collections import OrderedDict
+from weakref import WeakSet
 
 import wrapt
 
-LRU_CACHES = set()
+LRU_CACHE_CAPACITY = 128
 GENERATOR_CACHES = {}
-INFERENCE_CACHE = {}
+
+
+class LRUCache:
+    instances = WeakSet()
+
+    def __init__(self):
+        self.cache = OrderedDict()
+        type(self).instances.add(self)
+
+    def __setitem__(self, key, value):
+        self.cache[key] = value
+
+        if len(self.cache) > LRU_CACHE_CAPACITY:
+            self.cache.popitem(last=False)
+
+    def __getitem__(self, key):
+        if key in self.cache:
+            self.cache.move_to_end(key)
+
+        return self.cache[key]
+
+    def __contains__(self, key):
+        return key in self.cache
+
+    def clear(self):
+        self.cache.clear()
+
+    @classmethod
+    def clear_all(cls):
+        for cache in cls.instances:
+            cache.clear()
+
+
+def lru_cache():
+    cache = LRUCache()
+
+    @wrapt.decorator
+    def decorator(func, instance, args, kwargs):
+        key = (instance,) + args
+
+        for kv in kwargs:
+            key += kv
+
+        if key in cache:
+            return cache[key]
+
+        result = cache[key] = func(*args, **kwargs)
+
+        return result
+
+    return decorator
 
 
 @wrapt.decorator
@@ -20,31 +72,6 @@ def cached_generator(func, instance, args, kwargs):
     return iter(result)
 
 
-def lru_cache(maxsize=128, typed=False):
-    if maxsize is None:
-        maxsize = 128
-
-    def decorator(f):
-        cached_func = functools.lru_cache(maxsize, typed)(f)
-
-        LRU_CACHES.add(cached_func)
-
-        @functools.wraps(f)
-        def wrapper(*args, **kwargs):
-            return cached_func(*args, **kwargs)
-
-        return wrapper
-
-    return decorator
-
-
-def clear_inference_cache():
-    INFERENCE_CACHE.clear()
-
-
 def clear_caches():
-    for c in LRU_CACHES:
-        c.cache_clear()
-
+    LRUCache.clear_all()
     GENERATOR_CACHES.clear()
-    INFERENCE_CACHE.clear()
--- a/astroid/cache.py
+++ b/astroid/cache.py
@@ -2,6 +2,9 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
+""" Caching utilities used in various places."""
+
+
 from collections import OrderedDict
 from weakref import WeakSet
 
@@ -11,11 +14,13 @@ LRU_CACHE_CAPACITY = 128
 
 
 class LRUCache:
+    """An LRU cache that keeps track of its instances."""
+
     instances = WeakSet()
 
     def __init__(self):
         self.cache = OrderedDict()
-        type(self).instances.add(self)
+        LRUCache.instances.add(self)
 
     def __setitem__(self, key, value):
         self.cache[key] = value
@@ -37,11 +42,15 @@ class LRUCache:
 
     @classmethod
     def clear_all(cls):
+        """Clears all LRUCache instances."""
         for cache in cls.instances:
             cache.clear()
 
 
 def lru_cache(arg=None):
+    """A decorator to cache the results of a function. Similar to
+    functools.lru_cache but uses astroid.cache.LRUCache as its internal cache.
+    """
     cache = LRUCache()
 
     @wrapt.decorator
@@ -52,9 +61,9 @@ def lru_cache(arg=None):
             key += kv
 
         if key in cache:
-            return cache[key]
-
-        result = cache[key] = func(*args, **kwargs)
+            result = cache[key]
+        else:
+            result = cache[key] = func(*args, **kwargs)
 
         return result
 
@@ -68,6 +77,10 @@ _GENERATOR_CACHE = LRUCache()
 
 
 def cached_generator(arg=None):
+    """A decorator to cache the elements returned by a generator. The input
+    generator is consumed and cached as a list.
+    """
+
     @wrapt.decorator
     def decorator(func, instance, args, kwargs):
         key = func, args[0]
@@ -86,4 +99,5 @@ def cached_generator(arg=None):
 
 
 def clear_caches():
+    """Clears all caches."""
     LRUCache.clear_all()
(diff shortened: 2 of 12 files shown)


ChangeLog
```diff
@@ -12,6 +12,8 @@ What's New in astroid 2.8.6?
 ============================
 Release date: TBA
 
+* Add a common caching mechanism (`astroid.cache.LRUCache`) and a global
```
DanielNoord, line 15: Due to the review taking a little longer this will need to be moved to `2.10`.

astroid/cache.py
```diff
@@ -0,0 +1,100 @@
+""" Caching utilities used in various places."""
+
+from collections import OrderedDict
+from weakref import WeakSet
+
+import wrapt
+
+LRU_CACHE_CAPACITY = 128
```
DanielNoord, line 8: Do we want to this to be set-able somehow? I can imagine instances where somebody would want to cache to be larger?

astroid/cache.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class LRUCache:
```
DanielNoord, line 11: Could you add typing to all methods of this class? We're trying to turn on `mypy` for `astroid` and adding typing as we're adding new functionality would be very beneficial in the future.
The best would be to run `mypy --strict` over this file to fix any potential typing related issues in the future.

astroid/cache.py
```diff
@@ -0,0 +1,100 @@
...
+        self.cache.clear()
+
+    @classmethod
+    def clear_all(cls):
+        """Clears all LRUCache instances."""
+        for cache in cls.instances:
+            cache.clear()
+
+
+def lru_cache(arg=None):
```
DanielNoord, line 45: I wonder if we should rename this to `lru_cache_astroid` just so we don't create any confusion with the `functools` one.

DanielNoord (commented):
I have not looked at everything yet and I am by no means a caching expert, but I'd like to make an effort to getting some of the stale `astroid` PRs merged. 

These are just some things I was thinking about before this could be merged.

I also wonder if we should add any form of tests. We have `LruCacheModelTest` right now. Does that work in itself for this change?

keichi:
@DanielNoord Thank you for your review. I am addressing your comments one by one.

> I also wonder if we should add any form of tests. We have LruCacheModelTest right now. Does that work in itself for this change?

Sure, I will add tests for `LRUCache`. `LruCacheModelTest` tests if `@functools.lru_cache` is inferred correctly so it does not (directly) test `LRUCache`.

astroid/cache.py
keichi: Renamed it to `@lru_cache_astroid` following your suggestion.

astroid/cache.py
keichi: I've added typing to all methods. I've left `@lru_cache_astroid` and `@cached_generator` untyped for now because `@wrapt.decorator` is untyped.

ChangeLog
keichi: I've moved the change log to 2.10.0.

astroid/cache.py
keichi: @DanielNoord I'm not sure what you mean by "set-able"; it is settable. Would you like to add a setter method? Or make it overridable by instances?

Pierre-Sassoulas: We probably need to document the variable and explains explicitely how to modify it (assign a new value to it, I suppose). We can add the option in pylint config later but astroid users will need the info.

astroid/cache.py
```diff
@@ -0,0 +1,104 @@
+""" Caching utilities used in various places."""
+
+import typing
+from collections import OrderedDict
+from weakref import WeakSet
+
+import wrapt
+
+LRU_CACHE_CAPACITY = 128
```
Pierre-Sassoulas, line 9:
```suggestion
LRU_CACHE_CAPACITY = 2**30
```

Reading the test it seems it's a number of entry in the cache and not the sizeof taken in Mb of RAM. I suppose calculating the size in ram would take performance away ? Or is it impossible ? This is counter intuitive to me as I expected the cache capacity to be the space taken  in RAM. What do you think?

Pierre-Sassoulas (commented):
This looks pretty good already, thank you @keichi !

astroid/cache.py
keichi: I agree it's a bit counter-intuitive, but as far as I know, there is no efficient way to calculate the memory footprint of objects.
 
[`sys.getsizeof(obj)`](https://docs.python.org/3/library/sys.html#sys.getsizeof) is an option, but it only returns the size of the object itself and does not include the size of the objects that are referenced. So if we want to calculate the total size of an AST, we would need to traverse the entire tree and accumulate the size of the tree nodes. I believe this would be pretty expensive.

Pierre-Sassoulas: I think taking only the size of the current node into account is fine. We're not really storing the whole ast in the cache, only this node, right ? The node permit to navigate to other part of the tree if we want, but there's no additional cost for it. 

Some random ideas : we could check the way `functool.lru_cache` handle this ? Because there is a maxsize when using it. Also maybe it's possible to use it directly, I'm no expert in caching but I suppose there is a way to reuse the existing and override it.

astroid/cache.py
DanielNoord: Preferably we could change this setting via an option, but now that I think of it I don't think we really have a good way to do so.

I'll create a "create option-setting system" project after this gets merged and add this global to the "to be configurable" list.

astroid/transforms.py
```diff
@@ -25,12 +24,10 @@ class TransformVisitor:
     transforms for each encountered node.
     """
 
-    TRANSFORM_MAX_CACHE_SIZE = 10000
-
     def __init__(self):
         self.transforms = collections.defaultdict(list)
 
-    @lru_cache(maxsize=TRANSFORM_MAX_CACHE_SIZE)
+    @lru_cache_astroid
```
DanielNoord, line 30: How does the `10000` from `lru_cache` translate to the `128` of `LRUCache`? Is it similar in size?

astroid/cache.py
```diff
@@ -0,0 +1,104 @@
...
+    functools.lru_cache but uses astroid.cache.LRUCache as its internal cache.
+    """
+    cache = LRUCache()
+
+    @wrapt.decorator
+    def decorator(func, instance, args, kwargs):
+        key = (instance,) + args
+
+        for kv in kwargs:
+            key += kv
```
DanielNoord, line 60: Based on this wouldn't the `key` for a `LRUCache` always be a `Tuple[Any]`?

astroid/cache.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def lru_cache_astroid(arg=None):
+    """A decorator to cache the results of a function. Similar to
+    functools.lru_cache but uses astroid.cache.LRUCache as its internal cache.
+    """
+    cache = LRUCache()
+
+    @wrapt.decorator
+    def decorator(func, instance, args, kwargs):
```
DanielNoord, line 56: I'm not totally sure, but I think we can a bit better typing-wise based on the documentation [here](https://wrapt.readthedocs.io/en/latest/decorators.html).

`func` is probably a `Callable`? `Instance` `Any`? And `args` and `kwargs` something like `List` and `Dict`. If we can't narrow them down we should at least add `Any` (same goes for the return type) as `mypy` will eventually complain about coming across an untyped function definition.

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
+import pytest
+
+import astroid.cache
+from astroid.cache import LRUCache
+
+
+class TestLRUCache:
+    """
+    Test astroid.cache.LRUCache
```
DanielNoord, line 9:
```suggestion
    """Test astroid.cache.LRUCache
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    def test_set_and_get(self):
+        """
+        Test __setitem__, __getitem__, and __contains__
```
DanielNoord, line 14:
```suggestion
        """Test __setitem__, __getitem__, and __contains__
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        cache = LRUCache()
+        cache["a"] = 123
+
+        assert cache["a"] == 123
+        assert "a" in cache
+        assert "b" not in cache
+
+    def test_clear(self):
+        """
+        Test LRUCache.clear()
```
DanielNoord, line 25:
```suggestion
        """Test LRUCache.clear()
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        assert "c" in cache
+
+        cache.clear()
+
+        assert "c" not in cache
+
+    @pytest.fixture
+    def reduce_cache_capacity(self):
+        """
+        A fixture to temporarily decrease the cache capactiy
```
DanielNoord, line 39:
```suggestion
        """A fixture to temporarily decrease the cache capactiy
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        astroid.cache.LRU_CACHE_CAPACITY = 3
+        yield
+        astroid.cache.LRU_CACHE_CAPACITY = 128
+
+    def test_eviction(self, reduce_cache_capacity):
+        """
+        Test cache eviction behavior
```
DanielNoord, line 47:
```suggestion
        """Test cache eviction behaviour
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        cache["b"] = 2
+        cache["c"] = 3
+
+        assert "a" in cache
+        assert "b" in cache
+        assert "c" in cache
+
+        cache["d"] = 3
+
+        # "a" is evicted since it's the least recelty used item
```
DanielNoord, line 60:
```suggestion
        # "a" is evicted since it's the least recenlty used item
```

tests/unittest_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        cache["e"] = 4
+        assert "b" in cache
+        # "c" is evicted since "b" was recently referenced
+        assert "c" not in cache
+        assert "d" in cache
+        assert "e" in cache
+
+    def test_clear_caches(self):
+        """
+        Test clear_caches() (global cache flush)
```
DanielNoord, line 76:
```suggestion
        """Test clear_caches() (global cache flush)
```

DanielNoord (changes requested):
Thanks @keichi! This is becoming a very nice addition.

The tests are also well written.

astroid/cache.py
keichi: > We're not really storing the whole ast in the cache, only this node, right ? The node permit to navigate to other part of the tree if we want, but there's no additional cost for it.

I'm not 100% sure, but it seems AST nodes could reference other AST nodes via various fields. Anyway, AST nodes are just an example. Any object that is referenced via a field is not accounted by `sys.getsizeof()`.

> Some random ideas : we could check the way functool.lru_cache handle this ? Because there is a maxsize when using it. 

`maxsize` limits the number of entries in the cache, not the memory footprint. Please see:
https://github.com/python/cpython/blob/8d7644fa64213207b8dc6f555cb8a02bfabeced2/Lib/functools.py#L479

The main reason I couldn't reuse `functool.lru_cache` is because it hold the cache as a closure variable, and thus there is no way to clear it from outside.

astroid/transforms.py
keichi: Nope, the current `LRU_CACHE_CAPACITY` is arbitrary. Should we change it to 10000?

astroid/cache.py
keichi: You are right, I've typed it as `typing.Tuple[typing.Any, ...]`.

astroid/cache.py
keichi: Sure, I've added the typings!

astroid/cache.py
Pierre-Sassoulas: Ho ok, it's in fact exactly the same than functools.lru_cache, thanks for correcting me.

astroid/cache.py
DanielNoord: You can add a `type: ignore[misc] # wrapt.decorator is untyped` on both lines, we can't do much about it sadly.

astroid/transforms.py
DanielNoord: I'm not sure.. 10000 seems a bit excessive, but the difference with 128 is so large that I wonder if this has unforeseen effects.

astroid/cache.py
```diff
@@ -0,0 +1,117 @@
+""" Caching utilities used in various places."""
```
DanielNoord, line 1:
```suggestion
"""Caching utilities used in various places."""
```

astroid/cache.py
```diff
@@ -0,0 +1,117 @@
...
+        self.cache.clear()
+
+    @classmethod
+    def clear_all(cls) -> None:
+        """Clears all LRUCache instances."""
+        for cache in cls.instances:
+            cache.clear()
+
+
+F = typing.TypeVar("F", bound=typing.Callable[..., typing.Any])
```
DanielNoord, line 49: Could you move this to the top of the module? We like to keep all typing variables together.

astroid/cache.py
```diff
@@ -0,0 +1,117 @@
...
+    functools.lru_cache but uses astroid.cache.LRUCache as its internal cache.
+    """
+    cache: LRUCache[typing.Tuple[typing.Any, ...], typing.Any] = LRUCache()
+
+    @wrapt.decorator
+    def decorator(
+        func: F,
+        instance: typing.Any,
+        args: typing.Tuple[typing.Any, ...],
+        kwargs: typing.Dict[typing.Any, typing.Any],
```
DanielNoord, line 63:
```suggestion
        kwargs: typing.Dict[str, typing.Any],
```

right?

DanielNoord (commented):
Thanks @keichi 

I think I have done all I can with respect to this PR. Sadly, I'm not proficient enough with caching and some of the intricacies of typing generics to confidently say that this is now 100% perfect though. Hopefully some more experienced developer finds time to look at this soon.. 😄

astroid/transforms.py
keichi: In principle, a smaller cache size means more recomputation so it could lower the performance. However when I actually measure how long the unit tests take, I don't see any slowdown:

main (d2a5b3c7):
pytest  19.51s user 0.96s system 108% cpu 18.893 total

this branch:
pytest  17.93s user 0.82s system 109% cpu 17.171 total

So at least for the unit tests, I don't see any negative side effects.

Pierre-Sassoulas: The real problem will appear on a very large code base where the RAM won't be enough to cache everything. The tests are supposed to run fast so they run on small examples with a few nodes and they won't cause caching problem. As this can create a cache for every nodes, and because the number of nodes is close to infinite for large code base, I think we need to get a sense on what is the median weight of a node in the cache. Then probably limit the default value to a number of cache entry value corresponding to a low value of RAM used like 1Gb to be safe. (RAM is what is expensive on the cloud, so I suppose that by default jobs runs on low RAM machines).

keichi: I can parse a large project and find out how the distribution of the node size looks like. Are there any projects you usually use for benchmarking?

Pierre-Sassoulas: We have some project in pylint's test suite that we call primer tests. django is pretty huge for example. See https://pylint.pycqa.org/en/latest/development_guide/testing.html#primer-tests

Pierre-Sassoulas:
Hello @keichi, sorry for taking a long time to check this merge request. We currently have flake8-bugbear who (rightfully I think) warn us of potential cache issue on the main branch because we cache methods that will never be garbage collected. So merging this would solve two problems in one. I rebased the code on the latest main branch, but there's a recursion error in one of the test. Do you think you would have some time to check the issue in the near future ?

ChangeLog
```diff
@@ -232,6 +234,16 @@ Release date: 2021-11-21
...
+* Resolve symlinks in the import path
+  Fixes inference error when the import path includes symlinks (e.g. Python
+  installed on macOS via Homebrew).
+
+  Closes #823
+  Closes PyCQA/pylint#3499
+  Closes PyCQA/pylint#4302
+  Closes PyCQA/pylint#4798
+  Closes PyCQA/pylint#5081
+
```
Pierre-Sassoulas, line 246:
```suggestion
```
Duplicate from line 202 to 210 probably following a rebase.

Pierre-Sassoulas (approved):
LGTM, thank you @keichi

DanielNoord:
I haven't completed my review yet, but I see that two added lines are uncovered. Is there a way to cover these? Or did we already decide not to do so?

keichi:
@DanielNoord How can I locate those two lines? The Coverall page does not show the source code for some reason.
Screen Shot 2022-04-11 at 18 59 02

DanielNoord:
> @DanielNoord How can I locate those two lines? The Coverall page does not show the source code for some reason.
> Screen Shot 2022-04-11 at 18 59 02

They are L70 and L83 in `astroid/cache.py`. Coveralls doesn't show source code when you aren't logged in to reduce API calls on their "own" account so they require you to login/create an account to show annotated source code.

jacobtylerwalls (commented):
Thanks for working on `pylint`/`astroid`'s performance. It's really important.  :)

I'm wondering what the motivation was for reimplementing the basic LRU cache functionality. If it was chiefly to have a central call to flush all caches, we could just have the `lru_cache_astroid` decorator do that one thing: collect a reference to every function using `functools.lru_cache` so they can all be flushed.

Speaking of, we have a documented method `clear_cache()` already, so I'm worried we're expanding the API for no benefit and causing confusion. Better for _that_ to call `for c in cached_methods:  c.clear_cache()`, maybe? Just one flusher?

The reimplementation looks clean, so it shouldn't be hard to maintain. Still, I'm thinking we should be able to avoid merging it in. Especially if someday there comes a C-[version](https://github.com/python/cpython/blob/326ae71f1d93c12100150baa1173ea7ce7a96ea0/Lib/functools.py#L502) in the stdlib that we would get for free by staying on the `functools` one.

If I'm missing something, let me know!
*** 
👀 _cache by cache_ 

- [ ] `LookupMixin.lookup()`:  I checked again on `django`: even a maxsize of 128 still generates 79K hits (instead of 85K), so I'm good with 128. This was the only `lru_cache` that we left unbounded after #1503, so bounding it seems important.
- [ ] `TransformVisitor._transform()`: we realized in #1503 that this was generating 0 hits in typical scenarios and removed caching. So let's be sure not to add it back :-)
- [ ] `_cache_normalize_path()`: honestly this doesn't look very important, 128 is fine
- [ ] `ObjectModel.attributes`: same
- [ ] `infer_functiondef()` and `_inference_tip_cached()`. These two look like the ballgame. If we already had these factored out to use `functools.lru_cache`, then we could already be using `cache_info()` to reason about them. I think that's my final argument in favor of not reimplementing this -- we're kind of _losing_ some tooling by losing `cache_info()`.

Happy to hear any thoughts on these directions! And thanks again for pushing this forward, as I said, it's really important.

Pierre-Sassoulas:
Regarding the use of functool we discussed it here: https://github.com/PyCQA/astroid/pull/1242#discussion_r776690809

> The main reason I couldn't reuse functool.lru_cache is because it hold the cache as a closure variable, and thus there is no way to clear it from outside.

Maybe it's not an issue if we choose a low cache size and it's enough for big project.

jacobtylerwalls:
Oh, I was thinking some light refactors might be involved to make them all reachable. I'll try to put together a little proof of concept to see if I'm suggesting something impossible.

jacobtylerwalls:
Wow, `infer_functiondef()` is only cached as part of a workaround for an issue very similar to #1490. With a workaround can remove caching completely and still pass that unitt test in 17a5ee681bcf4aacffcc4ec5afbc3436cfdc4537. We may eventually want a keyword argument `set_parent_local` or something if we have three places we have to avoid doing that.

jacobtylerwalls:
Thanks for pushing this discussion forward. We decided to go with #1521 in order to hew more closely to the current documented way to clear caches. Always eager to hear more proposals for improving performance. Thanks again.

PR #1243, opened 2021-11-16
Use ``functools.cached_property`` instead of ``cachedproperty``

DanielNoord:
## Steps

## Description

@cdce8p Pinging you as you wrote https://github.com/python/mypy/pull/10408 which prompted me to do this.

We could remove the deprecation but I don't see why we would include this decorator in our codebase if we're not using it ourselves. The added benefit of `mypy` supporting the `functools` one is necessary for us to enable `mypy` at some point.

## Type of Changes

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -70,6 +70,12 @@ class cachedproperty:
     __slots__ = ("wrapped",)
 
     def __init__(self, wrapped):
+        warnings.warn(
+            "astroid.decorators.cachedproperty will be deprecated in astroid 3.0.0. "
+            "functools.cached_property offers similar behaviour and is supported by mypy and "
+            "other projects.",
+            DeprecationWarning,
+        )
         try:
             wrapped.__name__
         except AttributeError as exc:
--- a/astroid/mixins.py
+++ b/astroid/mixins.py
@@ -16,6 +16,7 @@
 """This module contains some mixins for the different nodes.
 """
 import itertools
+from functools import cached_property
 
 from astroid import decorators
 from astroid.exceptions import AttributeInferenceError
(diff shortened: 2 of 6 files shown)


DanielNoord:
Should have looked up when `cached_property` was added to `functools`.

@cdce8p What do you think about extracting our `cachedproperty` decorator from `decorators.py` into `typed_decorators.py` so we can add a `typed_decorators.pyi` file?
See the solution proposed here: https://github.com/pydanny/cached-property/issues/172

I could also add `type: ignore` to all occurrences of `cachedproperty`, but I think this solution could actually be quite elegant.

Pierre-Sassoulas:
Was it closed by mistake ? It looked like a nice evolution to me.

DanielNoord:
Only added in 3.8 😉

cdce8p:
> @cdce8p What do you think about extracting our `cachedproperty` decorator from `decorators.py` into `typed_decorators.py` so we can add a `typed_decorators.pyi` file? See the solution proposed here: [pydanny/cached-property#172](https://github.com/pydanny/cached-property/issues/172)

I don't think we should mix type annotations and stub files.
Have you tried this? (after the definition in `decorators.py`)
```py
if TYPE_CHECKING:
    cachedproperty = property
```

DanielNoord:
Returns:
```shel
astroid/decorators.py:94: error: Cannot assign to a type
astroid/decorators.py:94: error: Incompatible types in assignment (expression has type "Type[property]", variable has type "Type[cachedproperty]")
```

cdce8p:
I thought about this PR recently again. Typing `cachedproperty` might not work, but it can be replaced with `functools.cached_property`. We only need to `sys` guard it for `< 3.8`. This should work
```py
    from functools import cached_property
else:
    from astroid.decorators import cachedproperty as cached_property
```

Additionally, the deprecation warning can be limited to `>= 3.8`.

cdce8p:
I've created #1410 to track this. Github milestones don't really work for closed PRs.

PR #1244, opened 2021-11-16
Fix ``mypy`` warnings for ``astroid/rebuilder``

DanielNoord:
## Steps

## Description

Only 130 warnings left before we can turn on `mypy`.

## Type of Changes

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4489,7 +4489,7 @@ class MatchSingleton(Pattern):
     def __init__(
         self,
         *,
-        value: Literal[True, False, None],  # type: ignore[assignment] # See https://github.com/python/mypy/pull/10389
+        value: Literal[True, False, None],
         lineno: Optional[int] = None,
         col_offset: Optional[int] = None,
         parent: Optional[NodeNG] = None,
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4489,12 +4489,12 @@ class MatchSingleton(Pattern):
     def __init__(
         self,
         *,
-        value: Literal[True, False, None],
+        value: Optional[bool],
         lineno: Optional[int] = None,
         col_offset: Optional[int] = None,
         parent: Optional[NodeNG] = None,
     ) -> None:
-        self.value: Literal[True, False, None] = value
+        self.value = value
         super().__init__(lineno=lineno, col_offset=col_offset, parent=parent)
 
 
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4489,7 +4489,7 @@ class MatchSingleton(Pattern):
     def __init__(
         self,
         *,
-        value: Optional[bool],
+        value: Literal[True, False, None],  # type: ignore[assignment] # See https://github.com/python/mypy/pull/10389
         lineno: Optional[int] = None,
         col_offset: Optional[int] = None,
         parent: Optional[NodeNG] = None,
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -48,6 +48,7 @@ from typing import (
 from astroid import nodes
 from astroid._ast import ParserModule, get_parser_module, parse_function_type_comment
 from astroid.const import PY37_PLUS, PY38_PLUS, PY39_PLUS, Context
+from astroid.exceptions import ParentMissingError
 from astroid.manager import AstroidManager
 from astroid.nodes import NodeNG
 
@@ -106,7 +107,7 @@ class TreeRebuilder:
     def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
         try:
             if PY37_PLUS and hasattr(node, "docstring"):
-                doc = node.docstring
+                doc = node.docstring  # type: ignore # mypy doesn't recognize hasattr
                 return node, doc
             if node.body and isinstance(node.body[0], self._module.Expr):
 
@@ -803,8 +804,10 @@ class TreeRebuilder:
         """save assignement situation since node.parent is not available yet"""
         if self._global_names and node.name in self._global_names[-1]:
             node.root().set_local(node.name, node)
-        else:
+        elif node.parent:
             node.parent.set_local(node.name, node)
+        else:
+            raise ParentMissingError(target=node)
 
     def visit_arg(self, node: "ast.arg", parent: NodeNG) -> nodes.AssignName:
         """visit an arg node by returning a fresh AssName instance"""
@@ -873,8 +876,12 @@ class TreeRebuilder:
         )
         # save argument names in locals:
         if vararg:
+            if not newnode.parent:
+                raise ParentMissingError(target=newnode)
             newnode.parent.set_local(vararg, newnode)
         if kwarg:
+            if not newnode.parent:
+                raise ParentMissingError(target=newnode)
             newnode.parent.set_local(kwarg, newnode)
         return newnode
 
@@ -933,6 +940,9 @@ class TreeRebuilder:
             type_comment_ast = parse_function_type_comment(type_comment)
         except SyntaxError:
             # Invalid type comment, just skip it.
+            return None
+
+        if not type_comment_ast:
             return None
 
         returns: Optional[NodeNG] = None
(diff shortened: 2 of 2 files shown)


astroid/nodes/node_classes.py
```diff
@@ -4489,12 +4489,12 @@ class MatchSingleton(Pattern):
     def __init__(
         self,
         *,
-        value: Literal[True, False, None],
+        value: Optional[bool],
```
DanielNoord, line 4492: The `Literal` doesn't work when the `init` is called via `super()` when that inheritor does use `bool`. I don't think `Literal` provides any added benefit here.

astroid/rebuilder.py
```diff
@@ -873,8 +876,12 @@ def visit_arguments(self, node: "ast.arguments", parent: NodeNG) -> nodes.Argume
         )
         # save argument names in locals:
         if vararg:
+            if not newnode.parent:
```
DanielNoord, line 879: `Mypy` doesn't recognise:

```python
if vararg or kwarg and not new node.parent:
    raise ParentMissingError(target=newnode)
```

astroid/nodes/node_classes.py
cdce8p: This should really stay as `Literal`. A PR recently should have added support for it. Just don't know if it's already included in the next release or the one after that. Please add `type: ignore` if necessary.
https://github.com/python/mypy/pull/10389

astroid/rebuilder.py
cdce8p: Shouldn't `ast.argument` nodes always have a parent anyway?
Why not do:
```py
if not newnode.parent:
    raise ParentMissingError(target=newnode)
```

cdce8p (changes requested):

astroid/rebuilder.py
```diff
@@ -1354,7 +1364,8 @@ def visit_attribute(
             newnode = nodes.AssignAttr(node.attr, node.lineno, node.col_offset, parent)
             # Prohibit a local save if we are in an ExceptHandler.
             if not isinstance(parent, nodes.ExceptHandler):
-                self._delayed_assattr.append(newnode)
+                # mypy doesn't recognize that newnode has to be AssignAttr
+                self._delayed_assattr.append(newnode)  # type: ignore
```
cdce8p, line 1368: Only for the moment. `AssignAttr.__init__` has a decorator that uses `ParamSpec` which is unsupported currently.
I believe this will also be fixed with the next release.

astroid/rebuilder.py
```diff
@@ -106,7 +107,7 @@ def __init__(
     def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
         try:
             if PY37_PLUS and hasattr(node, "docstring"):
-                doc = node.docstring
+                doc = node.docstring  # type: ignore # mypy doesn't recognize hasattr
```
cdce8p, line 110:
```suggestion
                doc = node.docstring  # type: ignore[attr-defined]  # mypy doesn't recognize hasattr
```
https://github.com/PyCQA/astroid/pull/1247#discussion_r750803599

astroid/nodes/node_classes.py
DanielNoord: Didn't know there is a difference between the two.

I'll add a ignore.

astroid/rebuilder.py
DanielNoord: I thought so but didn't want to make any false assumptions. This was the safer option. I'll add the error.

astroid/rebuilder.py
```diff
@@ -875,13 +875,11 @@ def visit_arguments(self, node: "ast.arguments", parent: NodeNG) -> nodes.Argume
             type_comment_posonlyargs=type_comment_posonlyargs,
         )
         # save argument names in locals:
+        if not newnode.parent:
+            raise ParentMissingError(target=newnode)
```
Pierre-Sassoulas, line 879:
```suggestion
        if not newnode.parent and (vararg or kwarg):
            raise ParentMissingError(target=newnode)
```
Otherwise I think this change the behavior ?

astroid/rebuilder.py
```diff
@@ -106,7 +107,7 @@ def __init__(
     def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
         try:
             if PY37_PLUS and hasattr(node, "docstring"):
-                doc = node.docstring
+                doc = node.docstring  # type: ignore[union-attr, attr-defined] # mypy doesn't recognize hasattr
```
DanielNoord, line 110:
```suggestion
                doc = node.docstring  # type: ignore[union-attr,attr-defined] # mypy doesn't recognize hasattr
```
```suggestion
                doc = node.docstring  # type: ignore[union-attr, attr-defined] # mypy doesn't recognize hasattr
```

astroid/rebuilder.py
```diff
@@ -873,9 +876,9 @@ def visit_arguments(self, node: "ast.arguments", parent: NodeNG) -> nodes.Argume
         )
         # save argument names in locals:
         if vararg:
-            newnode.parent.set_local(vararg, newnode)
+            newnode.parent.set_local(vararg, newnode)  # type: ignore[union-attr] # newnode is initialized with a parent
         if kwarg:
-            newnode.parent.set_local(kwarg, newnode)
+            newnode.parent.set_local(kwarg, newnode)  # type: ignore[union-attr] # newnode is initialized with a parent
```
cdce8p, line 881: An idea: Instead of `type: ignore`, try `assert newnode.parent` above `if vararg`. As it's always initialized with a parent, this should be a no-op.

astroid/rebuilder.py
```diff
@@ -803,8 +804,10 @@ def _save_assignment(self, node: Union[nodes.AssignName, nodes.DelName]) -> None
         """save assignement situation since node.parent is not available yet"""
         if self._global_names and node.name in self._global_names[-1]:
             node.root().set_local(node.name, node)
-        else:
+        elif node.parent:
             node.parent.set_local(node.name, node)
+        else:
+            raise ParentMissingError(target=node)
```
cdce8p, line 810: Same here as `node.parent` will always be set.

astroid/rebuilder.py
```diff
@@ -1354,7 +1360,8 @@ def visit_attribute(
             newnode = nodes.AssignAttr(node.attr, node.lineno, node.col_offset, parent)
             # Prohibit a local save if we are in an ExceptHandler.
             if not isinstance(parent, nodes.ExceptHandler):
-                self._delayed_assattr.append(newnode)
+                # mypy doesn't recognize that newnode has to be AssignAttr
+                self._delayed_assattr.append(newnode)  # type: ignore [arg-type]
```
cdce8p, line 1364:
```suggestion
                self._delayed_assattr.append(newnode)  # type: ignore[arg-type]
```

cdce8p, line 1364: I would suggest to add that it's because mypy doesn't understand `ParamSpec` yet.

astroid/nodes/node_classes.py
```diff
@@ -4489,12 +4489,12 @@ class MatchSingleton(Pattern):
     def __init__(
         self,
         *,
-        value: Literal[True, False, None],
+        value: Literal[True, False, None],  # type: ignore[assignment] # See https://github.com/python/mypy/pull/10389
```
cdce8p, line 4492: Why is the `type: ignore` comment necessary here at all? I didn't notice any errors without it.

cdce8p (changes requested):

astroid/nodes/node_classes.py
DanielNoord: That was indeed misplaced. Should have been in `rebuilder`. See latest commit, `node.value` is `Optional[bool]`.

cdce8p (approved):
LGTM! I also tested it with Home-Assistant just to make sure since we added the asserts. As expected they everything still works.

Pierre-Sassoulas (commented):
LGTM but the assert makes me want to wait for primers to be available in pylint. The "pylint tested" label will means a lot more then.

DanielNoord:
@Pierre-Sassoulas primer is in place :)

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Maybe we can put it in 2.9.1 too ?

DanielNoord:
I would thinks so! Marc already approved so I'll merge!

PR #1246, opened 2021-11-16
Fix `mypy` warnings for `astroid/builder` and `astroid/nodes/as_string`

DanielNoord:
## Steps

## Type of Changes

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -263,7 +263,7 @@ class AstroidBuilder(raw_building.InspectBuilder):
             pass
 
 
-def build_namespace_package_module(name, path: List[str]) -> nodes.Module:
+def build_namespace_package_module(name: str, path: List[str]) -> nodes.Module:
     return nodes.Module(name, doc="", path=path, package=True)
 
 
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -263,8 +263,8 @@ class AstroidBuilder(raw_building.InspectBuilder):
             pass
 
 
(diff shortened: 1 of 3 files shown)


astroid/builder.py
```diff
@@ -264,7 +264,7 @@ def delayed_assattr(self, node):
 
 
 def build_namespace_package_module(name, path: str) -> nodes.Module:
-    return nodes.Module(name, doc="", path=path, package=True)
+    return nodes.Module(name, doc="", path=[path], package=True)
```
cdce8p, line 267: This is probably a true bug fix. If `path` was set as `str`, `path[0]` would access just the first char and not the complete string. Can you create a separate PR for it, ideally with changelog entry. We should also test how this effects existing code.

cdce8p (changes requested):

astroid/builder.py
Pierre-Sassoulas: If this is a bug fix maybe we can put it in 2.8.6.

DanielNoord: I looked into it. I don't think `path: str` is correct here. Looking at how this is called the `path` always comes from `ModuleSpec.submodule_search_locations` which is always a `List[str]`. 

You can run the test `test_nested_namespace_import` to debug this.

Thus, this change is actually incorrect and the typing should be updated. I will push a new commit.

cdce8p: That does make sense! I guess otherwise namespace packages would be broken completely.

cdce8p (approved):

PR #1247, opened 2021-11-16
Add ``type: ignore`` to method reassignments

DanielNoord:
## Steps

## Type of Changes

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -72,11 +72,13 @@ def infer_end(self, context=None):
     yield self
 
 
-nodes.Module._infer = infer_end  # type: ignore # See https://github.com/python/mypy/issues/2427
-nodes.ClassDef._infer = infer_end  # type: ignore
-nodes.Lambda._infer = infer_end  # type: ignore
-nodes.Const._infer = infer_end  # type: ignore
-nodes.Slice._infer = infer_end  # type: ignore
+# We add ignores to all these assignments in this file
+# See https://github.com/python/mypy/issues/2427
+nodes.Module._infer = infer_end  # type: ignore[assignment]
+nodes.ClassDef._infer = infer_end  # type: ignore[assignment]
+nodes.Lambda._infer = infer_end  # type: ignore[assignment]
+nodes.Const._infer = infer_end  # type: ignore[assignment]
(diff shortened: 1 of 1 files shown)


DanielNoord:
This is the last one for now.

I think together with the other PRs this should bring the amount of warnings for `mypy astroid` from 135 to 100.

astroid/inference.py
```diff
@@ -72,11 +72,11 @@ def infer_end(self, context=None):
     yield self
 
 
-nodes.Module._infer = infer_end
-nodes.ClassDef._infer = infer_end
-nodes.Lambda._infer = infer_end
-nodes.Const._infer = infer_end
-nodes.Slice._infer = infer_end
+nodes.Module._infer = infer_end  # type: ignore # See https://github.com/python/mypy/issues/2427
```
cdce8p, line 75:
```suggestion
nodes.Module._infer = infer_end  # type: ignore[assignment]  # See https://github.com/python/mypy/issues/2427
```
It's better if we only ignore the specific mypy error.

I haven't gotten around to it, but we should consider adding the `--show-error-codes` flag to `mypy` in both astroid and pylint. Maybe even update our existing `type: ignore` statements.

cdce8p (commented):
Hmm 🤔 Tbh. this looks rather ugly, but it seems like it's the only way at the moment (aside from `setattr`).

astroid/inference.py
DanielNoord: > Hmm 🤔 Tbh. this looks rather ugly, but it seems like it's the only way at the moment (aside from `setattr`).

Would you prefer using `setattr`? I opted for `ignore` as this seems a "bug" in `mypy` and I thought this should thus be fixed by telling `mypy` to ignore this rather than to change our own code.

cdce8p: > Would you prefer using `setattr`?

Not really. `type: ignore[assignment]` still allows for other type checks, `setattr` not (afaik).
Maybe it's enough to link the issue ones for the complete file though. (above `nodes.Module._infer = ...)

cdce8p (approved):
👍🏻

Pierre-Sassoulas (approved):
👍

PR #1248, opened 2021-11-17
Add additional flags to ``mypy`` config and update ``type: ignore``'s

DanielNoord:
## Steps

## Description

@cdce8p I think these are the flags we can turn on right now without much trouble.

I also added an error code to every single `type: ignore` currently in the code.
I couldn't find anything about it online, but do you know if there is a way to require error codes in these ignore messages?

## Type of Changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,6 +65,10 @@ skip_glob = tests/testdata
 
 [mypy]
 scripts_are_modules = True
+no_implicit_optional = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+show_error_codes = True
 
 [mypy-setuptools]
 ignore_missing_imports = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,6 @@ skip_glob = tests/testdata
 scripts_are_modules = True
 no_implicit_optional = True
 warn_redundant_casts = True
-warn_unused_ignores = True
 show_error_codes = True
 
 [mypy-setuptools]
--- a/tests/unittest_protocols.py
+++ b/tests/unittest_protocols.py
@@ -297,7 +297,7 @@ class TestPatternMatching:
                 pass
         """
         )
-        match_mapping: nodes.MatchMapping = assign_stmts.pattern  # type: ignore
+        match_mapping: nodes.MatchMapping = assign_stmts.pattern
         assert match_mapping.rest
         assigned = next(match_mapping.rest.assigned_stmts())
         assert assigned == Uninferable
@@ -316,7 +316,7 @@ class TestPatternMatching:
                 pass
         """
         )
-        match_sequence: nodes.MatchSequence = assign_stmts.pattern  # type: ignore
+        match_sequence: nodes.MatchSequence = assign_stmts.pattern
         match_star = match_sequence.patterns[2]
         assert isinstance(match_star, nodes.MatchStar) and match_star.name
         assigned = next(match_star.name.assigned_stmts())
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -65,6 +65,10 @@ skip_glob = tests/testdata
 
 [mypy]
 scripts_are_modules = True
+no_implicit_optional = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+show_error_codes = True
```
Pierre-Sassoulas, line 71: 🎉

Pierre-Sassoulas (approved):
Smell like progress !

Pierre-Sassoulas:
> I couldn't find anything about it online, but do you know if there is a way to require error codes in these ignore messages?

[I think it's ``--show-error-codes``](https://mypy.readthedocs.io/en/stable/error_codes.html#displaying-error-codes), but we need mypy 9.10 at least.

DanielNoord:
I don't think that requires you to specify the error you're ignoring in a `type: ignore` it only shows the error code in the output in the terminal.

Pierre-Sassoulas:
Ho, sorry I misunderstood your question.  I think this is not yet in mypy : https://github.com/python/mypy/issues/11509

tests/unittest_protocols.py
```diff
@@ -297,7 +297,7 @@ def test_assigned_stmts_match_mapping():
                 pass
         """
         )
-        match_mapping: nodes.MatchMapping = assign_stmts.pattern  # type: ignore
+        match_mapping: nodes.MatchMapping = assign_stmts.pattern
```
cdce8p, line 300: The `type: ignores` here are only unused since the method isn't annotated and thus not type checked.

setup.cfg
```diff
@@ -65,6 +65,10 @@ skip_glob = tests/testdata
 
 [mypy]
 scripts_are_modules = True
+no_implicit_optional = True
+warn_redundant_casts = True
+warn_unused_ignores = True
```
cdce8p, line 70: I would remove `warn_unused_ignores = True` just for now. It isn't an issue at the moment. Instead it's more likely to cause false positives (see comment above). Other false-positives are the two `type: ignores` in `rebuilder.py`. They are only necessary in Python 3.9+, thus when checked with 3.8 they are marked as unnecessary.

cdce8p (changes requested):

tests/unittest_protocols.py
```diff
@@ -337,10 +337,10 @@ def test_assigned_stmts_match_as():
...
         """
         )
-        subject: nodes.Const = assign_stmts[0].subject  # type: ignore
-        match_or: nodes.MatchOr = assign_stmts[1].pattern  # type: ignore
-        match_as_with_pattern: nodes.MatchAs = assign_stmts[2].pattern  # type: ignore
-        match_as: nodes.MatchAs = assign_stmts[3].pattern  # type: ignore
+        subject: nodes.Const = assign_stmts[0].subject  # type: ignore[index, union-attr]
+        match_or: nodes.MatchOr = assign_stmts[1].pattern  # type: ignore[index, union-attr]
+        match_as_with_pattern: nodes.MatchAs = assign_stmts[2].pattern  # type: ignore[index, union-attr]
+        match_as: nodes.MatchAs = assign_stmts[3].pattern  # type: ignore[index, union-attr]
```
cdce8p, line 343:
```suggestion
        subject: nodes.Const = assign_stmts[0].subject  # type: ignore[index,union-attr]
        match_or: nodes.MatchOr = assign_stmts[1].pattern  # type: ignore[index,union-attr]
        match_as_with_pattern: nodes.MatchAs = assign_stmts[2].pattern  # type: ignore[index,union-attr]
        match_as: nodes.MatchAs = assign_stmts[3].pattern  # type: ignore[index, union-attr]
```
I would recommend no spaces between error codes. Otherwise this looks like a python list.

DanielNoord:
```suggestion
        subject: nodes.Const = assign_stmts[0].subject  # type: ignore[index,union-attr]
        match_or: nodes.MatchOr = assign_stmts[1].pattern  # type: ignore[index,union-attr]
        match_as_with_pattern: nodes.MatchAs = assign_stmts[2].pattern  # type: ignore[index,union-attr]
        match_as: nodes.MatchAs = assign_stmts[3].pattern  # type: ignore[index,union-attr]
```

cdce8p (approved):

setup.cfg
Pierre-Sassoulas: Look like mypy has the same problem than us for useless-suppression. We could mimic how they handle it ?

cdce8p: > Look like mypy has the same problem than us for useless-suppression. We could mimic how they handle it ?

We already do ;) They don't handle it.

Pierre-Sassoulas: I mean they have an option to activate it or not with ``warn_unused_ignores = True``, like we do when we enable useless-suppression, but do you know how they handle discrepancies that appear between python interpreter when unused ignores are activated ?

cdce8p: If the code is evaluated, they emit an error. I.e. not if it's guarded behind a `if sys.version_info` check what is quite common when using typing.

PR #1249, opened 2021-11-17
Add parameter typing to ``assigned_stmts``

DanielNoord:
## Steps

## Description

This should probably be refactored. ``protocols.py`` gets imported by in ``inference.py`` and by that import a number of methods get assigned to various nodes. I don't immediately see why these methods can't be defined on the ``nodes`` themselves.

However, since I think in the short term there is more benefit of being able to turn on `mypy` instead of trying to refactor a complete module I think this could be a good temporary solution.

## Type of Changes


Pierre-Sassoulas:
>  I don't immediately see why these methods can't be defined on the nodes themselves.

There's a lot of things done like this, I think to avoid cyclic import that were removed before (?) Now it's just creating it's own problem imo. Example [here](https://github.com/PyCQA/astroid/pull/972/files), this one was fixable and we did not get any complaint about it since it was fixed. Imo method defined in another file than the class makes the code really complicated and hard to follow, if we can, we should fix them. But there is a lot of work...

cdce8p:
> This should probably be refactored. `protocols.py` gets imported by in `inference.py` and by that import a number of methods get assigned to various nodes. I don't immediately see why these methods can't be defined on the `nodes` themselves.

This was done so we don't have to deal with circular dependencies. We already tried to do 'some' (limited) work there but it usually didn't end too well. My personal opinion: We should be lucky that part works and keep our hand of it (regarding a major refactoring). _Maybe_ if astroid is fully typed, but as long as there isn't a clear need for it, I wouldn't recommend to touch it.

> However, since I think in the short term there is more benefit of being able to turn on `mypy` instead of trying to refactor a complete module I think this could be a good temporary solution.

I agree. This has been another part I've been thinking about but never actually got around too 😅

astroid/nodes/node_classes.py
```diff
@@ -766,6 +766,7 @@ class Arguments(mixins.AssignTypeMixin, NodeNG):
     )
 
     _other_fields = ("vararg", "kwarg")
+    assigned_stmts: Callable
```
cdce8p, line 769: Ideally, we should be more specific.

All the `assigned_stmts` method have basically the same signature. Maybe we can figure out the actual types for them.
```py
def xxx_assigned_stmts(self, node=None, context=None, assign_path=None):
    ...
```

cdce8p (changes requested):
This PR mixes `assigned_stmts` and `infer_xxx` methods. There are some still missing
```py
nodes.AssignName.assigned_stmts = assend_assigned_stmts
nodes.AssignAttr.assigned_stmts = assend_assigned_stmts
...  # (?)
```

I would recommend to focus on `assigned_stmts` for now and remove the `infer_xxx`. With a more specific `Callable` type, it should be trivial to type the method signatures as well. That should be the focus here. The `infer` functions can follow later.

DanielNoord:
I have broken this down into little steps which should help with the review process and allow @cdce8p some time to work on his own projects 😄 

In order for the `Callable` to be added to each class we will need to do the following steps:
1) Add parameter typing
2) Add return typing
3) Fix ``mypy`` issues that arise from `mypy` now checking these functions as well
4) Add `Callable` to the class

Because `mypy` isn't turned on we can do this in separate PRs without breaking the CI. This PR now does the first step of this process.

astroid/protocols.py
```diff
@@ -274,7 +274,14 @@ def _resolve_looppart(parts, assign_path, context):
 
 
 @decorators.raise_if_nothing_inferred
-def for_assigned_stmts(self, node=None, context=None, assign_path=None):
+def for_assigned_stmts(
+    self: Union[nodes.For, nodes.Comprehension],
+    node: Union[
+        nodes.List, nodes.Tuple, nodes.AssignName, nodes.AssignAttr, None
+    ] = None,
```
cdce8p, line 281: How did you narrow it down to these types? I would need to spend more time looking at it.

astroid/protocols.py
```diff
@@ -274,7 +274,14 @@ def _resolve_looppart(parts, assign_path, context):
...
+    context: Optional[InferenceContext] = None,
+    assign_path: Optional[List[int]] = None,
+):
```
cdce8p, line 284: It does make sense not to include return types for now. What about adding `Any` until then?

cdce8p (commented):
Thanks for splitting this up! It does especially make sense not to do the return types here.
I'm not so sure about the `Callable` types though. It could make sense to include them. That might enable basic type checking for it and reduce the number of errors we need to address afterwards. You can use `Any` as return type.

astroid/protocols.py
DanielNoord: I looked at all calls to `assigned_stmts` and looked at what was provided as `node`. I'll show them here:

1)
https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/inference.py#L955
No `node`.

2)
https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/protocols.py#L294-L314
`node` is either `Tuple` or `List`.

3)
https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/protocols.py#L317-L322
`node` is either `AssignName` or `AssignAttr`

4)
https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/protocols.py#L420-L425
Just passes whatever `node` currently is -> no new types.

5)
https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/nodes/scoped_nodes.py#L803
No `node`.


I checked and `assigned_stmts` is not used in `pylint`.

cdce8p: Beware that `assigned_stmts` is called on `self.parent` and not `self`. Thus I would think technically it can be `NodeNG`.

DanielNoord: > Beware that `assigned_stmts` is called on `self.parent` and not `self`. Thus I would think technically it can be `NodeNG`.

I have taken another look, but I'm not seeing how it can be anything other than the four types in there now. However, you have often proven me wrong so I won't claim I'm 100% sure. My thought process is as follows:
Case 1 and 5 can be disregarded immediately as they don't pass a `node`.
Case 4 does pass a `node`, but only passes the one already defined or gives the standard `None`. Because `assign_assigned_stmts` does nothing with `node` this doesn't allow any new types other than the ones possibly being passed to `assign_annassigned_stmts` to become `node`.
In case 2 and 3 only pass the `self` as `node`. Therefore we need only need to determine what their `self` can be to determine what `node` can be in any of the calls.
`sequence_assigned_stmts` (case 2) is never called directly and only a method on `Tuple` and `List`. Referring to your comment, we can only enter this function if `self.parent` is `Tuple` or `List`. I don't think it is possible to enter `sequence_assigned_stmts` with `self` not being `Tuple` or `List` and therefore the `node` created here is always `Tuple` or `List`.
A similar case can be made for `assend_assigned_stmts`. It is only accessed as method on `AssignName` and `AssignAttr`. `self` should always be either of them and therefore the `node` "created" in that function also needs to be.

cdce8p: You are right! Thanks for the explanation. I missed that `self` is passed to node. That it's called on the parent doesn't change anything.

cdce8p:
@DanielNoord What do you think about adding the `Callable` types in this one (with `Any` as return). I think that could make sense contextually and still be easy to do.

DanielNoord:
Fine with me! Do you approve of the typing as is? Then I can copy it to the class definitions.

cdce8p:
> Fine with me! Do you approve of the typing as is? Then I can copy it to the class definitions.

Looks good so far. Might still be that something will come up later once we start adding the return types, but that can be addressed then.

astroid/nodes/node_classes.py
```diff
@@ -612,6 +612,19 @@ class AssignName(
...
+            Union["List", "Tuple", "AssignName", "AssignAttr", None],
+            Optional[InferenceContext],
+            Optional[typing.List[int]],
+        ],
+        Any,
+    ]
+    """Returns the assigned statement (non inferred) according to the assignment type.
+    See astroid/protocols.py for actual implementation.
+    """
+
```
cdce8p, line 627: What do you think about moving method types below `__init__` and `postinit`? I feel like where it's now, it could easily be mistaken for an attribute.

astroid/nodes/node_classes.py
```diff
@@ -612,6 +612,19 @@ class AssignName(
 
     _other_fields = ("name",)
 
+    assigned_stmts: Callable[
+        [
+            Union["AssignName", "AssignAttr"],
```
cdce8p, line 617:
```suggestion
            "AssignName",
```
The first argument here will always be `AssignName` or a subclass, so this should work. The function type is more permissive.

astroid/nodes/node_classes.py
```diff
@@ -1230,6 +1269,19 @@ class Assign(mixins.AssignTypeMixin, Statement):
     _astroid_fields = ("targets", "value")
     _other_other_fields = ("type_annotation",)
 
+    assigned_stmts: Callable[
+        [
+            Union["AugAssign", "Assign", "AnnAssign"],
```
cdce8p, line 1274:
```suggestion
            "Assign",
```

astroid/nodes/node_classes.py
```diff
@@ -1372,6 +1437,19 @@ class AugAssign(mixins.AssignTypeMixin, Statement):
     _astroid_fields = ("target", "value")
     _other_fields = ("op",)
 
+    assigned_stmts: Callable[
+        [
+            Union["AugAssign", "Assign", "AnnAssign"],
```
cdce8p, line 1442:
```suggestion
            "AugAssign",
```

astroid/nodes/node_classes.py
```diff
@@ -1779,6 +1857,19 @@ class Comprehension(NodeNG):
     optional_assign = True
     """Whether this node optionally assigns a variable."""
 
+    assigned_stmts: Callable[
+        [
+            Union["For", "Comprehension"],
```
cdce8p, line 1862:
```suggestion
            "Comprehension",
```

astroid/nodes/node_classes.py
```diff
@@ -2498,6 +2602,19 @@ class For(
     This is always ``True`` for :class:`For` nodes.
     """
 
+    assigned_stmts: Callable[
+        [
+            Union["For", "Comprehension"],
```
cdce8p, line 2607:
```suggestion
            "For",
```

astroid/nodes/node_classes.py
```diff
@@ -3735,6 +3878,19 @@ class Tuple(BaseContainer):
 
     _other_fields = ("ctx",)
 
+    assigned_stmts: Callable[
+        [
+            Union["Tuple", "List"],
```
cdce8p, line 3883:
```suggestion
            "Tuple",
```

astroid/nodes/node_classes.py
```diff
@@ -612,6 +612,19 @@ class AssignName(
 
     _other_fields = ("name",)
 
+    assigned_stmts: Callable[
+        [
+            Union["AssignName", "AssignAttr"],
+            Union["List", "Tuple", "AssignName", "AssignAttr", None],
```
cdce8p, line 618: Does it make sense to create an alias type for it?

astroid/nodes/node_classes.py
DanielNoord: Ah, didn't think of this!

astroid/protocols.py
```diff
@@ -51,6 +51,10 @@
 else:
     from typing_extensions import Literal
 
+AssignedStmtsPossibleNode = Union[
```
DanielNoord, line 54: @cdce8p Added here as well!

DanielNoord:
CI isn't running. Github status is not reporting anything (as of yet)..

cdce8p:
> CI isn't running. Github status is not reporting anything (as of yet)..

Saw that too. @Pierre-Sassoulas Could you take a look?

Pierre-Sassoulas:
> Could you take a look?

Seems like it's green but no github actions checks are launched ? I'll check that.

Pierre-Sassoulas:
When in doubt, reboot :D !

DanielNoord:
CodeQL is having trouble for 48 hours I think? Has something changed in the settings? Or did we reach a limit?

Pierre-Sassoulas:
> CodeQL is having trouble for 48 hours

Yes, also it never gave an useful warning so we can ignore it for now and I'll remove it definitely it if it fails for too long.

astroid/protocols.py
```diff
@@ -595,7 +639,12 @@ def __enter__(self):
 
 
 @decorators.raise_if_nothing_inferred
-def named_expr_assigned_stmts(self, node, context=None, assign_path=None):
+def named_expr_assigned_stmts(
+    self: nodes.NamedExpr,
+    node: mixins.AssignTypeMixin,
```
cdce8p, line 644: It has been awhile since I last looked at that. Is this right?

astroid/nodes/node_classes.py
```diff
@@ -635,6 +638,19 @@ def __init__(
...
 
+    assigned_stmts: Callable[
+        [
+            "AssignName",
+            AssignedStmtsPossibleNode,
+            Optional[InferenceContext],
+            Optional[typing.List[int]],
+        ],
+        Any,
+    ]
```
cdce8p, line 649: The type is almost the same for every node. We could try to use a generic type alias.
https://mypy.readthedocs.io/en/latest/generics.html#generic-type-aliases

If PEP 673 gets accepted (and eventually implemented) we could even use `Self` for the node type. https://www.python.org/dev/peps/pep-0673/

astroid/protocols.py
DanielNoord: Incorrect indeed, fixed!

astroid/nodes/node_classes.py
```diff
@@ -72,6 +72,19 @@ def _is_const(value):
     return isinstance(value, tuple(CONST_CLS))
 
 
+AssignedStmtsPossibleNode = Union["List", "Tuple", "AssignName", "AssignAttr", None]
+T = TypeVar("T")
+AssignedStmtsCall = Callable[
```
cdce8p, line 77:
```suggestion
T = TypeVar("T")

AssignedStmtsPossibleNode = Union["List", "Tuple", "AssignName", "AssignAttr", None]
AssignedStmtsCall = Callable[
```

We shouldn't directly mix TypeVars and Type Aliases. That just makes it difficult to read.

--
Another point, would it make sense to bound `T` to `nodes.NodeNG`? I.e.
```py
T_Nodes = TypeVar("T_Nodes", bound=NodeNG)
```
instead of `T`. That way we would limit the allowed types.

astroid/nodes/node_classes.py
```diff
@@ -4567,15 +4650,10 @@ def postinit(
...
 
-    assigned_stmts: Callable[
-        [
-            "MatchMapping",
-            AssignName,
-            Optional[InferenceContext],
-            Literal[None],
-        ],
-        Generator[NodeNG, None, None],
-    ]
```
cdce8p, line 4578: Hmm, technically this and the once for `MatchClass` and `MatchAs` are more concrete than `AssignedStmtsCall`. I would suggest to leave them as is for now until we come back and refine (especially the return type) for `AssignedStmtsCall`.

astroid/nodes/node_classes.py
DanielNoord: I'll update!

astroid/protocols.py
```diff
@@ -51,6 +51,10 @@
 else:
     from typing_extensions import Literal
 
+AssignedStmtsPossibleNode = Union[
+    nodes.List, nodes.Tuple, nodes.AssignName, nodes.AssignAttr, None
+]
```
cdce8p, line 56: Do we need it here or can it be imported from `astroid.nodes`?

DanielNoord: We're already importing `nodes` from `astroid` so I thought it would be better to use it in a similar way as we do in `pylint`, ie. `List` is `typing.List` and `nodes.List` is `astroid.nodes.List`.

cdce8p: What I meant is that we define `AssignedStmtsPossibleNode` in `astroid.nodes` already. Can't we reuse it in `astroid/protocols`?

DanielNoord: Ah sorry! Does it make sense to creating `astroid/typing.py` as we did for `pylint`?

cdce8p: I would not recommend it at this stage. Such a file only really makes sense for complex type definitions like NamedTuples or TypedDicts. This is just a simple type alias used in two files.

DanielNoord: See https://github.com/PyCQA/astroid/pull/1249/commits/d095f3202bac96c603460953d1824178c46188df

astroid/nodes/__init__.py
```diff
@@ -31,6 +31,7 @@
     Assert,
     Assign,
     AssignAttr,
+    AssignedStmtsPossibleNode,
```
cdce8p, line 34: Sorry for the confusion so far.

It doesn't really make sense to expose the type here as it's only used internally. We can use it from `astroid.nodes.node_classes` directly.

astroid/protocols.py
```diff
@@ -274,7 +275,12 @@ def _resolve_looppart(parts, assign_path, context):
 
 
 @decorators.raise_if_nothing_inferred
-def for_assigned_stmts(self, node=None, context=None, assign_path=None):
+def for_assigned_stmts(
+    self: Union[nodes.For, nodes.Comprehension],
+    node: nodes.AssignedStmtsPossibleNode = None,
```
cdce8p, line 280:
```suggestion
    node: node_classes.AssignedStmtsPossibleNode = None,
```
`node_classes` is already imported, so this should work.

astroid/protocols.py
```diff
@@ -51,6 +51,7 @@
 else:
     from typing_extensions import Literal
 
+
```
cdce8p, line 54:
```suggestion
```
Empty line change

DanielNoord:
No worries!

cdce8p (approved):
Looks good now!
Thanks @DanielNoord 🚀

Pierre-Sassoulas (approved):
LGTM !

DanielNoord:
@Pierre-Sassoulas Why is this blocked?

Pierre-Sassoulas:
The milestone indicated was 2.10 so we need to release 2.9.1 first.

cdce8p:
> The milestone indicated was 2.10 so we need to release 2.9.1 first.

This is just a small typing improvement so in theory I think we could merge this now (even for 2.9.1).
@Pierre-Sassoulas What do you think?

PR #1251, opened 2021-11-17
Add ``DeprecationWarning`` contextmanager to tests of deprecated methods

DanielNoord:
## Steps

## Description

Removes some warnings from the `pytest` output that were bugging me.

## Type of Changes

### DIFF
--- a/tests/unittest_nodes.py
+++ b/tests/unittest_nodes.py
@@ -369,6 +369,7 @@ class IfNodeTest(_NodeTest):
         self.assertEqual(self.astroid.body[1].orelse[0].block_range(8), (8, 8))
 
     @staticmethod
+    @pytest.mark.filterwarnings("ignore:.*is_sys_guard:DeprecationWarning")
     def test_if_sys_guard() -> None:
         code = builder.extract_node(
             """
@@ -394,6 +395,7 @@ class IfNodeTest(_NodeTest):
         assert code[2].is_sys_guard() is False
 
     @staticmethod
+    @pytest.mark.filterwarnings("ignore:.*is_typing_guard:DeprecationWarning")
     def test_if_typing_guard() -> None:
         code = builder.extract_node(
             """


Pierre-Sassoulas (commented):
Shouldn't we remove the use of the deprecated function inside ``is_sys_guard`` instead ?

DanielNoord:
> Shouldn't we remove the use of the deprecated function inside `is_sys_guard` instead ?

As long as the functions are still in our codebase the tests are still relevant. We don't use the functions ourselves, we only test whether they do what they are supposed to be doing even if they are deprecated.

Pierre-Sassoulas:
> As long as the functions are still in our codebase the tests are still relevant

Ho my bad I thought it was pylint.

Pierre-Sassoulas (approved):

tests/unittest_nodes.py
```diff
@@ -386,12 +386,19 @@ def test_if_sys_guard() -> None:
         assert isinstance(code, list) and len(code) == 3
 
         assert isinstance(code[0], nodes.If)
-        assert code[0].is_sys_guard() is True
+        with pytest.warns(DeprecationWarning) as records:
+            assert code[0].is_sys_guard() is True
+            assert len(records) == 1
```
cdce8p, line 391: These to methods will be removed soon, so no real need to assert each Deprecation warning. Why not use a filter decorator on the test functions?

```py
    @staticmethod
    @pytest.mark.filterwarnings("ignore:.*is_sys_guard:DeprecationWarning")
    def test_if_sys_guard() -> None:
        ...

    @staticmethod
    @pytest.mark.filterwarnings("ignore:.*is_typing_guard:DeprecationWarning")
    def test_if_typing_guard() -> None:
        ...
```

cdce8p (changes requested):

cdce8p (approved):

PR #1252, opened 2021-11-17
Change ``NoReturn`` to be encapsulated in strings

### LINKED ISSUE #1239: TypeError: Plain typing.NoReturn is not valid as type argument in Astroid 2.8.5 + Python 3.7.0/1

DanielNoord:
## Steps

## Description

Follow up to #1235

Closes #1239

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,9 @@ What's New in astroid 2.8.6?
 ============================
 Release date: TBA
 
-* Fix bug on Python 3.7.0 and 3.7.1 by removing unallowed usage of ``typing.NoReturn``
+* Fix bug with Python 3.7.0 / 3.7.1 and ``typing.NoReturn``.
+
+   Closes #1239
 
 What's New in astroid 2.8.5?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,7 @@ What's New in astroid 2.8.6?
 ============================
 Release date: TBA
 
-
+* Fix bug on Python 3.7.0 and 3.7.1 by removing unallowed usage of ``typing.NoReturn``
 
 What's New in astroid 2.8.5?
 ============================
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -12,7 +12,7 @@ What's New in astroid 2.8.6?
 ============================
 Release date: TBA
 
-
+* Fix bug on Python 3.7.0 and 3.7.1 by removing unallowed usage of ``typing.NoReturn``
```
cdce8p, line 15:
```suggestion
* Fix bug with Python 3.7.0 / 3.7.1 and ``typing.NoReturn``.

   Closes #1239
```
Maybe something like this?

ChangeLog
```diff
@@ -12,7 +12,9 @@ What's New in astroid 2.8.6?
 ============================
 Release date: TBA
 
+* Fix bug with Python 3.7.0 / 3.7.1 and ``typing.NoReturn``.
 
+   Closes #1239
```
cdce8p, line 17:
```suggestion
  Closes #1239
  
```

cdce8p (approved):
Thanks @DanielNoord 🐬

Pierre-Sassoulas (approved):
🎉

PR #1255, opened 2021-11-18
Remove forgotten debug ``print`` statement from tests

DanielNoord:
## Steps

## Description

This print is shown when running `pytest -s`.

## Type of Changes

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -6027,7 +6027,6 @@ def test_property_inference() -> None:
     assert inferred.type == "property"
 
     inferred = next(prop_result.infer())
-    print(prop_result.as_string())
     assert isinstance(inferred, nodes.Const)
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Maybe we could introduce a check for "no output" on a reasonably diverse code to analyses ?

DanielNoord:
> Maybe we could introduce a test for "no output" on a reasonably diverse code to analyses ?

Not quite sure what you mean. We would only find this forgotten `print` if we were to run our complete test suite with a `no-output` rule. I don't think that is necessary. This also doesn't impact `astroid` itself, it just gives a cleaner development environment without any left over debug statements.

Pierre-Sassoulas:
I was thinking of a forgotten print in the code of astroid itself.

PR #1258, opened 2021-11-19
Add  `end_lineno` and `end_col_offset` to nodes

cdce8p:
## Description

In Python 3.8, `end_lineno` and `end_col_offset` where added to all AST nodes which already had `lineno`  and `col_offset` information. This PR adds them to the astoid nodes. Part of https://github.com/PyCQA/pylint/issues/5336

## Type of Changes

## Todo list
- [x] Add docstrings for `end_lineno`, `end_col_offset`
- [x] Write changelog entry
- [x] Write tests for _all_ nodes
- [ ] Decide if `sys.version_info >= (3, 8)` checks are ok, or if `getattr(node, 'end_lineno', None)` should be used.
Note: ` getattr` can't be checked with `mypy`.
- [x] Resolve open `TODOs` in code
- [ ] ...

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -47,7 +47,7 @@ from typing import (
 
 from astroid import nodes
 from astroid._ast import ParserModule, get_parser_module, parse_function_type_comment
-from astroid.const import PY37_PLUS, PY38_PLUS, Context
+from astroid.const import PY37_PLUS, PY38, PY38_PLUS, Context
 from astroid.manager import AstroidManager
 from astroid.nodes import NodeNG
 
@@ -830,6 +830,15 @@ class TreeRebuilder:
         if node.kwarg:
             kwarg = node.kwarg.arg
             kwargannotation = self.visit(node.kwarg.annotation, newnode)
+
+        if PY38:
+            # In Python 3.8 'end_lineno' and 'end_col_offset'
+            # for 'kwonlyargs' don't include the annotation.
+            for arg in node.kwonlyargs:
+                if arg.annotation is not None:
+                    arg.end_lineno = arg.annotation.end_lineno
+                    arg.end_col_offset = arg.annotation.end_col_offset
+
         kwonlyargs = [self.visit(child, newnode) for child in node.kwonlyargs]
         kw_defaults = [self.visit(child, newnode) for child in node.kw_defaults]
         annotations = [self.visit(arg.annotation, newnode) for arg in node.args]
@@ -1401,7 +1410,7 @@ class TreeRebuilder:
         self, node: "ast.ExtSlice", parent: nodes.Subscript
     ) -> nodes.Tuple:
         """visit an ExtSlice node by returning a fresh instance of Tuple"""
-        # TODO: add line and column info?
+        # ExtSlice doesn't have lineno or col_offset information
         newnode = nodes.Tuple(ctx=Context.Load, parent=parent)
         newnode.postinit([self.visit(dim, newnode) for dim in node.dims])  # type: ignore[attr-defined]
         return newnode
@@ -2046,8 +2055,16 @@ class TreeRebuilder:
 
     def visit_slice(self, node: "ast.Slice", parent: nodes.Subscript) -> nodes.Slice:
         """visit a Slice node by returning a fresh instance of it"""
-        # TODO add lineno and col offset info?
-        newnode = nodes.Slice(parent=parent)
+        if sys.version_info >= (3, 9):
+            newnode = nodes.Slice(
+                lineno=node.lineno,
+                col_offset=node.col_offset,
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
+        else:
+            newnode = nodes.Slice(parent=parent)
         newnode.postinit(
             lower=self.visit(node.lower, newnode),
             upper=self.visit(node.upper, newnode),
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -1607,7 +1607,7 @@ class TreeRebuilder:
             if not isinstance(parent, nodes.ExceptHandler):
                 self._delayed_assattr.append(newnode)
         else:
-            if sys.version_info >= (3, 8):
+            if sys.version_info >= (3, 8):  # pylint: disable=else-if-used
                 newnode = nodes.Attribute(
                     attrname=node.attr,
                     lineno=node.lineno,
@@ -1859,7 +1859,7 @@ class TreeRebuilder:
                     node.id, node.lineno, node.col_offset, parent
                 )
         else:
-            if sys.version_info >= (3, 8):
+            if sys.version_info >= (3, 8):  # pylint: disable=else-if-used
                 newnode = nodes.Name(
                     name=node.id,
                     lineno=node.lineno,
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -2120,7 +2120,6 @@ class TreeRebuilder:
 
     def visit_tryexcept(self, node: "ast.Try", parent: NodeNG) -> nodes.TryExcept:
         """visit a TryExcept node by returning a fresh instance of it"""
-        # TODO check lineno and col_offset
         if sys.version_info >= (3, 8):
             newnode = nodes.TryExcept(
                 lineno=node.lineno,
@@ -2143,7 +2142,6 @@ class TreeRebuilder:
     ) -> Union[nodes.TryExcept, nodes.TryFinally, None]:
         # python 3.3 introduce a new Try node replacing
         # TryFinally/TryExcept nodes
-        # TODO check lineno and col_offset
         if node.finalbody:
             if sys.version_info >= (3, 8):
                 newnode = nodes.TryFinally(
@@ -2168,7 +2166,6 @@ class TreeRebuilder:
 
     def visit_tryfinally(self, node: "ast.Try", parent: NodeNG) -> nodes.TryFinally:
         """visit a TryFinally node by returning a fresh instance of it"""
-        # TODO check lineno and col_offset
         if sys.version_info >= (3, 8):
             newnode = nodes.TryFinally(
                 lineno=node.lineno,
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -1607,7 +1607,9 @@ class TreeRebuilder:
             if not isinstance(parent, nodes.ExceptHandler):
                 self._delayed_assattr.append(newnode)
         else:
-            if sys.version_info >= (3, 8):  # pylint: disable=else-if-used
+            # pylint: disable-next=else-if-used
+            # Preserve symmetry with other cases
+            if sys.version_info >= (3, 8):
                 newnode = nodes.Attribute(
                     attrname=node.attr,
                     lineno=node.lineno,
@@ -1859,7 +1861,9 @@ class TreeRebuilder:
                     node.id, node.lineno, node.col_offset, parent
                 )
         else:
-            if sys.version_info >= (3, 8):  # pylint: disable=else-if-used
+            # pylint: disable-next=else-if-used
+            # Preserve symmetry with other cases
+            if sys.version_info >= (3, 8):
                 newnode = nodes.Name(
                     name=node.id,
                     lineno=node.lineno,
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -47,7 +47,7 @@ from typing import (
 
 from astroid import nodes
 from astroid._ast import ParserModule, get_parser_module, parse_function_type_comment
-from astroid.const import PY37_PLUS, PY38_PLUS, PY39_PLUS, Context
+from astroid.const import PY37_PLUS, PY38_PLUS, Context
 from astroid.manager import AstroidManager
 from astroid.nodes import NodeNG
 
@@ -880,7 +880,16 @@ class TreeRebuilder:
 
     def visit_assert(self, node: "ast.Assert", parent: NodeNG) -> nodes.Assert:
         """visit a Assert node by returning a fresh instance of it"""
-        newnode = nodes.Assert(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
+            newnode = nodes.Assert(
+                lineno=node.lineno,
+                col_offset=node.col_offset,
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
+        else:
+            newnode = nodes.Assert(node.lineno, node.col_offset, parent)
         msg: Optional[NodeNG] = None
         if node.msg:
             msg = self.visit(node.msg, newnode)
@@ -953,7 +962,16 @@ class TreeRebuilder:
         return self._visit_for(nodes.AsyncFor, node, parent)
 
     def visit_await(self, node: "ast.Await", parent: NodeNG) -> nodes.Await:
-        newnode = nodes.Await(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
+            newnode = nodes.Await(
+                lineno=node.lineno,
+                col_offset=node.col_offset,
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
+        else:
+            newnode = nodes.Await(node.lineno, node.col_offset, parent)
         newnode.postinit(value=self.visit(node.value, newnode))
         return newnode
 
@@ -962,7 +980,16 @@ class TreeRebuilder:
 
     def visit_assign(self, node: "ast.Assign", parent: NodeNG) -> nodes.Assign:
         """visit a Assign node by returning a fresh instance of it"""
-        newnode = nodes.Assign(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
+            newnode = nodes.Assign(
+                lineno=node.lineno,
+                col_offset=node.col_offset,
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
+        else:
+            newnode = nodes.Assign(node.lineno, node.col_offset, parent)
         type_annotation = self.check_type_comment(node, parent=newnode)
         newnode.postinit(
             targets=[self.visit(child, newnode) for child in node.targets],
@@ -973,7 +1000,16 @@ class TreeRebuilder:
 
     def visit_annassign(self, node: "ast.AnnAssign", parent: NodeNG) -> nodes.AnnAssign:
         """visit an AnnAssign node by returning a fresh instance of it"""
-        newnode = nodes.AnnAssign(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
+            newnode = nodes.AnnAssign(
+                lineno=node.lineno,
+                col_offset=node.col_offset,
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
+        else:
+            newnode = nodes.AnnAssign(node.lineno, node.col_offset, parent)
         newnode.postinit(
             target=self.visit(node.target, newnode),
             annotation=self.visit(node.annotation, newnode),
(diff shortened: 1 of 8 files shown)


astroid/rebuilder.py
```diff
@@ -1607,7 +1607,7 @@ def visit_attribute(
             if not isinstance(parent, nodes.ExceptHandler):
                 self._delayed_assattr.append(newnode)
         else:
-            if sys.version_info >= (3, 8):
+            if sys.version_info >= (3, 8):  # pylint: disable=else-if-used
```
Pierre-Sassoulas, line 1610: As discussed in the other issue, I don't think it's a false positive.

astroid/rebuilder.py
```diff
@@ -880,7 +880,16 @@ def visit_arguments(self, node: "ast.arguments", parent: NodeNG) -> nodes.Argume
 
     def visit_assert(self, node: "ast.Assert", parent: NodeNG) -> nodes.Assert:
         """visit a Assert node by returning a fresh instance of it"""
-        newnode = nodes.Assert(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 883:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -953,7 +962,16 @@ def visit_asyncfor(self, node: "ast.AsyncFor", parent: NodeNG) -> nodes.AsyncFor
         return self._visit_for(nodes.AsyncFor, node, parent)
 
     def visit_await(self, node: "ast.Await", parent: NodeNG) -> nodes.Await:
-        newnode = nodes.Await(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 965:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -962,7 +980,16 @@ def visit_asyncwith(self, node: "ast.AsyncWith", parent: NodeNG) -> nodes.AsyncW
 
     def visit_assign(self, node: "ast.Assign", parent: NodeNG) -> nodes.Assign:
         """visit a Assign node by returning a fresh instance of it"""
-        newnode = nodes.Assign(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 983:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -973,7 +1000,16 @@ def visit_assign(self, node: "ast.Assign", parent: NodeNG) -> nodes.Assign:
 
     def visit_annassign(self, node: "ast.AnnAssign", parent: NodeNG) -> nodes.AnnAssign:
         """visit an AnnAssign node by returning a fresh instance of it"""
-        newnode = nodes.AnnAssign(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1003:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
         """
         if node_name is None:
             return None
-        newnode = nodes.AssignName(
-            node_name,
-            node.lineno,
-            node.col_offset,
-            parent,
-        )
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1042:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
...
 
     def visit_augassign(self, node: "ast.AugAssign", parent: NodeNG) -> nodes.AugAssign:
         """visit a AugAssign node by returning a fresh instance of it"""
-        newnode = nodes.AugAssign(
-            self._parser_module.bin_op_classes[type(node.op)] + "=",
-            node.lineno,
-            node.col_offset,
-            parent,
-        )
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1063:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
...
 
     def visit_binop(self, node: "ast.BinOp", parent: NodeNG) -> nodes.BinOp:
         """visit a BinOp node by returning a fresh instance of it"""
-        newnode = nodes.BinOp(
-            self._parser_module.bin_op_classes[type(node.op)],
-            node.lineno,
-            node.col_offset,
-            parent,
-        )
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1086:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
...
 
     def visit_boolop(self, node: "ast.BoolOp", parent: NodeNG) -> nodes.BoolOp:
         """visit a BoolOp node by returning a fresh instance of it"""
-        newnode = nodes.BoolOp(
-            self._parser_module.bool_op_classes[type(node.op)],
-            node.lineno,
-            node.col_offset,
-            parent,
-        )
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1109:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
...
+                node.lineno,
+                node.col_offset,
+                parent,
+            )
         newnode.postinit([self.visit(child, newnode) for child in node.values])
         return newnode
 
     def visit_break(self, node: "ast.Break", parent: NodeNG) -> nodes.Break:
         """visit a Break node by returning a fresh instance of it"""
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1130:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1003,59 +1039,116 @@ def visit_assignname(
...
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
         return nodes.Break(node.lineno, node.col_offset, parent)
 
     def visit_call(self, node: "ast.Call", parent: NodeNG) -> nodes.Call:
         """visit a CallFunc node by returning a fresh instance of it"""
-        newnode = nodes.Call(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1142:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1068,7 +1161,20 @@ def visit_classdef(
     ) -> nodes.ClassDef:
         """visit a ClassDef node to become astroid"""
         node, doc = self._get_doc(node)
-        newnode = nodes.ClassDef(node.name, doc, node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1164:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1091,11 +1197,28 @@ def visit_classdef(
 
     def visit_continue(self, node: "ast.Continue", parent: NodeNG) -> nodes.Continue:
         """visit a Continue node by returning a fresh instance of it"""
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1200:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1091,11 +1197,28 @@ def visit_classdef(
...
+                end_lineno=node.end_lineno,
+                end_col_offset=node.end_col_offset,
+                parent=parent,
+            )
         return nodes.Continue(node.lineno, node.col_offset, parent)
 
     def visit_compare(self, node: "ast.Compare", parent: NodeNG) -> nodes.Compare:
         """visit a Compare node by returning a fresh instance of it"""
-        newnode = nodes.Compare(node.lineno, node.col_offset, parent)
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1212:
```suggestion
        if PY38_PLUS:
```

astroid/rebuilder.py
```diff
@@ -1134,18 +1257,37 @@ def visit_decorators(
             return None
         # /!\ node is actually an _ast.FunctionDef node while
         # parent is an astroid.nodes.FunctionDef node
-        if PY38_PLUS:
+        if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 1260: Huhu. I'm going to stop suggesting for every test then. But why though ? I'd say it's a good thing to use PY38+ if  only for the performance increase of not having to recalculate the bool everytime ?

Pierre-Sassoulas (commented):
Looks great, I only have two minor remarks.

DanielNoord:
Based on the documentation of `ast` I think only `nodes.Statement` and subclasses can have an `end_lineno`. `nodes.Expr` is a subclass so that is covered as well. Does it make sense to make 1) `nodes.NodeNG` have `end_lineno = None`, 2) add an `__init__` to `nodes.Statement` and 3) handle adding `end_lineno` there?
This would give better IntelliSense as `nodes.Module` will have `end_lineno: None` instead of `end_lineno: Optional[int]`. Because of the way we `visit_` nodes in `pylint` this information would be accessible to use. I'm not sure if we will need but as it seems to reflect reality it might be good to adhere to `ast` here.

(A similar point can be made for `lineno` and `col_offset` but let's not worry about that now)

astroid/rebuilder.py
cdce8p: True, I don't really like the performance impact either. On the positive side `sys.version_info` is special cased by type checkers so they know how to infer it. That doesn't work with `PY38_PLUS`.

If performance is a real concern, the only other option, I think, would be to conditional define the `visit_xxx` methods itself. That would mean a lot of code duplications though. Not sure we could maintain that, at least not easily.

In the end, it's only the node tree. So the performance impact might be acceptable.

--
```py
        def visit_constant(self, node: "ast.Constant", parent: NodeNG) -> nodes.Const:
            """visit a Constant node by returning a fresh instance of Const"""
            return nodes.Const(
                value=node.value,
                kind=node.kind,
                lineno=node.lineno,
                col_offset=node.col_offset,
                end_lineno=node.end_lineno,
                parent=parent,
            )
    else:
        def visit_constant(self, node: "ast.Constant", parent: NodeNG) -> nodes.Const:
            """visit a Constant node by returning a fresh instance of Const"""
            return nodes.Const(
                node.value,
                node.lineno,
                node.col_offset,
                parent,
                node.kind,
            )
```

astroid/rebuilder.py
cdce8p: Even if not a false-positive, I would recommend to keep the `pylint: disable` as this preserves the symmetry with all other cases and will make removing the check once `3.8` is the default much easier.

Pierre-Sassoulas (approved):

cdce8p:
@DanielNoord

> Based on the documentation of `ast` I think only `nodes.Statement` and subclasses can have an `end_lineno`. `nodes.Expr` is a subclass so that is covered as well.

`nodes.Expr` is, but what about all `ast.expr` nodes? `BoolOp`, `Set`, `Dict`, `List`, just to name a few.

> Does it make sense to make 1) `nodes.NodeNG` have `end_lineno = None`, 2) add an `__init__` to `nodes.Statement` and 3) handle adding `end_lineno` there? This would give better IntelliSense as `nodes.Module` will have `end_lineno: None` instead of `end_lineno: Optional[int]`. Because of the way we `visit_` nodes in `pylint` this information would be accessible to use. I'm not sure if we will need but as it seems to reflect reality it might be good to adhere to `ast` here.
> 
> (A similar point can be made for `lineno` and `col_offset` but let's not worry about that now)

If I understand you correctly, what you would like is for IntelliSense to know that some nodes will always have `end_lineno = None` (same for `lineno`, `col_offset`, and `end_col_offset`). That can still be archived by overriding the type. E.g. for `nodes.Module`:
```py
class Module(LocalsDictNodeNG):
    lineno: None
    col_offset: None
    end_lineno: None
    end_col_offset: None
    parent: None
```

DanielNoord:
Ah I didn't know not all `ast.Expr` nodes actually subclass `nodes.Expr`. That makes this more difficult indeed. Your latest commit is a nice solution!

cdce8p:
> Ah I didn't know not all `ast.Expr` nodes actually subclass `nodes.Expr`.

Actually, no node subclasses `ast.Expr`, some subclass `ast.expr` but we don't represent that in `astroid` currently.

cdce8p:
Took a moment to get this ready, especially writing the tests. I believe it's now ready for review.

Some additional comments regarding things I noticed while working on it:
1. `end_lineno` and `end_col_offset` are usually set regardless the annotation `int | None`. With the exception of the few nodes that truly don't have these information.
2. The `ast` module is usually considered unstable, that's even more apparent in the handling of `end_lineno` and `end_col_offset`.
3. Sometimes there are just bugs that might or might not have been fixed from one version to the next. This could make testing a bit more interesting in pylint if we **require** those and they vary between versions.
4. At least until `3.10` the `lineno` and `col_offset` information for `JoinedStr` and `FormattedValue` nodes don't make much sense.
5. Many `AssignName` nodes don't have own `lineno` and `col_offset` info. Instead they fall back to that of the `parent`. The reason for it is the `AST` sometimes only stores strings (`identifier`) instead of `ast.Name` nodes. The `TreeRebuilder` still interprets them as `AssignName` but doesn't have any individual node information. E.g. `MatchMapping.rest`, `MatchAs.name`, `arg.arg` (argument names).
https://github.com/PyCQA/astroid/blob/40d556070fd566b7bb43cb3ed0e498045b88f2d0/astroid/rebuilder.py#L997-L1013
6. The handling of `ast.Try` nodes introduced in Python `3.3` should be adopted in `TreeRebuilder`. At the moment we split it into `TryExcept` and `TryFinally` the ast nodes for which have long been removed. I was thinking about adding a `version` argument to the `TreeRebuilder` initialization. That would allow us to make this a non-breaking change. Only downside being that it also adds additional maintenance and test requirements.
https://github.com/PyCQA/astroid/blob/40d556070fd566b7bb43cb3ed0e498045b88f2d0/astroid/rebuilder.py#L1622-L1638

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+@pytest.mark.skipif(
+    not PY38_PLUS, reason="end_lineno and end_col_offset were added in PY38"
+)
+class TestLinenoColOffset:
+    """Test 'lineno', 'col_offset', 'end_lineno', and 'end_col_offset' for all nodes."""
+
+    @staticmethod
+    def test_end_lineno_container() -> None:
+        """Container nodes: List, Tuple, Set."""
+        code = textwrap.dedent(
```
Pierre-Sassoulas, line 18: 👍 dedent make the tests looks nicer.

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+        assert (c1.decorators.lineno, c1.decorators.col_offset) == (1, 0)
+        assert (c1.decorators.end_lineno, c1.decorators.end_col_offset) == (2, 11)
+        assert (c1.bases[0].lineno, c1.bases[0].col_offset) == (3, 8)
+        assert (c1.bases[0].end_lineno, c1.bases[0].end_col_offset) == (3, 14)
+        if PY39_PLUS:
+            # 'lineno' and 'col_offset' information only added in Python 3.9
+            assert (c1.keywords[0].lineno, c1.keywords[0].col_offset) == (3, 16)
+            assert (c1.keywords[0].end_lineno, c1.keywords[0].end_col_offset) == (3, 22)
+        assert (c1.body[0].lineno, c1.body[0].col_offset) == (4, 4)
+        assert (c1.body[0].end_lineno, c1.body[0].end_col_offset) == (4, 8)
```
Pierre-Sassoulas, line 1185: I just reviewed the content of the test, not the expected column. But it comes from the ast, if it's wrong there isn't a lot we can do, right ?

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+        assert isinstance(f1.args.posonlyargs[0], nodes.AssignName)
+        assert (f1.args.posonlyargs[0].lineno, f1.args.posonlyargs[0].col_offset) == (2, 4)
+        assert (f1.args.posonlyargs[0].end_lineno, f1.args.posonlyargs[0].end_col_offset) == (2, 10)
+        assert isinstance(f1.args.posonlyargs_annotations[0], nodes.Name)
+        assert (
+            f1.args.posonlyargs_annotations[0].lineno, f1.args.posonlyargs_annotations[0].col_offset
+        ) == (2, 7)
+        assert (
+            f1.args.posonlyargs_annotations[0].end_lineno, f1.args.posonlyargs_annotations[0].end_col_offset
+        ) == (2, 10)
```
Pierre-Sassoulas, line 603:
```suggestion
        annotation = f1.args.posonlyargs_annotations[0]
        assert (annotation.lineno, annotation.col_offset) == (2, 7)
        assert (annotation.end_lineno, annotation.end_col_offset) == (2, 10)
```

There's very long line with long name that are duplicated, maybe doing that would help if we need to modify those tests later on ?

Pierre-Sassoulas (commented):
Amazing work ! Do you want to wait for 2.9.0 ? Should we use astroid 2.9.0 in pylint 2.12 ? We could bump the other 2.9.0 features in 2.10 so we can release the end column/end line.

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
+import textwrap
+
+import pytest
+
+from astroid import builder, nodes
+from astroid.const import PY38_PLUS, PY39_PLUS, PY310_PLUS
+
+
+@pytest.mark.skipif(
```
DanielNoord, line 9: Do we want a test for < `3.8` to see if `end_lineno` and `end_col_offset` are correctly initiliazed as `None`?

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+
+        assert (c1.lineno, c1.col_offset) == (1, 0)
+        assert (c1.end_lineno, c1.end_col_offset) == (1, 22)
+        assert (c1.func.lineno, c1.func.col_offset) == (1, 0)
+        assert (c1.func.end_lineno, c1.func.end_col_offset) == (1, 4)
+
+        assert (c1.args[0].lineno, c1.args[0].col_offset) == (1, 5)
+        assert (c1.args[0].end_lineno, c1.args[0].end_col_offset) == (1, 9)
+
+        if PY39_PLUS:
```
DanielNoord, line 163: Should we test that these are `None` for < `3.9`?

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+        assert (s2.slice.upper.end_lineno, s2.slice.upper.end_col_offset) == (2, 7)
+        assert (s2.slice.step.lineno, s2.slice.step.col_offset) == (2, 8)
+        assert (s2.slice.step.end_lineno, s2.slice.step.end_col_offset) == (2, 9)
+
+        s3 = ast_nodes[2]
+        assert isinstance(s3, nodes.Subscript)
+        assert isinstance(s3.slice, nodes.Tuple)
+        assert (s3.lineno, s3.col_offset) == (3, 0)
+        assert (s3.end_lineno, s3.end_col_offset) == (3, 11)
+        if PY39_PLUS:
```
DanielNoord, line 804: Similar as above.

tests/unittest_nodes_lineno.py
```diff
@@ -0,0 +1,1185 @@
...
+        assert isinstance(c1.keywords[0], nodes.Keyword)
+        assert isinstance(c1.body[0], nodes.Pass)
+
+        assert (c1.lineno, c1.col_offset) == (3, 0)
+        assert (c1.end_lineno, c1.end_col_offset) == (4, 8)
+        assert (c1.decorators.lineno, c1.decorators.col_offset) == (1, 0)
+        assert (c1.decorators.end_lineno, c1.decorators.end_col_offset) == (2, 11)
+        assert (c1.bases[0].lineno, c1.bases[0].col_offset) == (3, 8)
+        assert (c1.bases[0].end_lineno, c1.bases[0].end_col_offset) == (3, 14)
+        if PY39_PLUS:
```
DanielNoord, line 1180: Similar as above.

DanielNoord (commented):
Great work @cdce8p! Writing those tests must not have been an easy task.

tests/unittest_nodes_lineno.py
cdce8p: True, at least not without significant effort. One area I was able to fix fairly easily was `Arguments`. See the snipped below. We might also be able to address the `AssignName` issue I mentioned by parsing the source code ourselves but that is definitely an issue for another time.

Usually the worse that can happen is that the lineno and col_offset also include the parent node. This is still correct just not as accurate as we would like it to be for error highlighting.

The true value of these tests is to know when a new version changed something. Like knowing that `Keywords` and `Slice` only include `end_lineno` in 3.9, or the issue with `Arguments`.

--
https://github.com/PyCQA/astroid/blob/7dd55fe8d96846617b2827e5272582ffc865190f/astroid/rebuilder.py#L834-L840

tests/unittest_nodes_lineno.py
cdce8p: Don't know about that. I think I prefer the current option as it doesn't hide the path. Otherwise you just end up introducing a lot of helper variables  which itself add complexity.

I explicitly chose to disable black in two functions to not have line breaks. Those made it even more difficult to read.

tests/unittest_nodes_lineno.py
cdce8p: I'll add one function to just test one or two nodes. That should be enough. I wouldn't expect issues with it.

tests/unittest_nodes_lineno.py
cdce8p: I'll add `else` blocks. If we manually fix those issue for Python 3.8 in ` TreeRebuilder` these should let us know.

cdce8p:
> Amazing work ! Do you want to wait for 2.9.0 ? Should we use astroid 2.9.0 in pylint 2.12 ? We could bump the other 2.9.0 features in 2.10 so we can release the end column/end line.

I would consider this a new feature that probably deserves its own minor version. I.e. release `2.9.0` next and bump the other features to `2.10.0` if that's ok with you.

Pierre-Sassoulas:
> I would consider this a new feature that probably deserves its own minor version. I.e. release 2.9.0 next and bump the other features to 2.10.0 

Sounds good, do we still release 2.8.6 first ?

cdce8p:
> > I would consider this a new feature that probably deserves its own minor version. I.e. release 2.9.0 next and bump the other features to 2.10.0
> 
> Sounds good, do we still release 2.8.6 first ?

Could make sense if it isn't too much effort. What is your opinion?
If we decide to, I just think we should do the release from the current main and do `2.9.0` directly afterwards with just this one. Everything currently targeted for  `2.8.6` could be moved to `2.9.1`.

Pierre-Sassoulas:
Hmm, I forgot to deprecate python 3.6.0 and 3.6.1. I'm going release 2.8.7 with the deprecation and suggest astroid 2.6.6 (as it's the one working with pylint 2.9.3).

Issue #1250, opened 2021-11-17
Should we change how attributes are defined on certain nodes classes?


DanielNoord:
I ran into this while working on the typing errors that we need to solve before turning on `mypy`.

Some of our ``nodes`` classes have attributes that get defined in a `postinit`. Take for example `func` on `nodes.Call`:
https://github.com/PyCQA/astroid/blob/907818246c89ef938b23448e0cecc52b7c8693e0/astroid/nodes/node_classes.py#L1635-L1642

https://github.com/PyCQA/astroid/blob/907818246c89ef938b23448e0cecc52b7c8693e0/astroid/nodes/node_classes.py#L1660-L1664

We know that this `func` attribute can't technically be `None`. If it were `None` the `Call` wouldn't have a function to call. In the `postinit` you can clearly see the difference between attributes that can and can not be `None`. For example, `args` and `keywords` can be `None`. (They should probably be `[]` in that case to allow iteration, but that's another issue).

I wonder what would happen if we were to change `self.func: Optional[NodeNG] = None` to `self.func: NodeNG`. This would simplify a lot of our typing as we wouldn't need to add checks for the existence of `Call.func` everywhere. Perhaps there are solutions to this which we should consider, but I think there must be a way to show that `nodes.Call` **must** have a `func` attribute, but that we don't know it at initialisation time.
One issue I already thought of is what would happen if something accessed `func` before the `postinit` is called. I don't think this is an issue and any exception raised is correct: the instance shouldn't be accessed before the `postinit` is complete.

I have created https://github.com/DanielNoord/astroid/pull/3 to show that this does indeed pass our test suite.

I'm pinging @cdce8p as you often have valuable insights for typing related issues.

cdce8p:
I've already spend some time working on this exact issue 😅
The next step is to deprecate additional default values for arguments in `postinit`. I.e. don't allow a default value of `None` for `func`. If you're ok with it taking a bit longer, I would appreciate if you let me do it. (Or just open fewer PRs, maybe then I have some time left to work on my own ones 😉)

DanielNoord:
I'll dim down the typing issues for now 😄 

Let me know if you need any help!

Pierre-Sassoulas:
Don't panic, there's enough issues for everyone 😄

DanielNoord:
> Don't panic, there's enough issues for everyone 😄

[...]

PR #1263, opened 2021-11-20
Fix frame() error on inferred node

tushar-deepsource:
## Description

Ran into this issue today. The `parent` of an inferred node was `None`, so `node.frame()` was crashing. This should fix that.

## Type of Changes


DanielNoord:
On which commit of `astroid` did you experience this issue? We have recently worked on `NodeNG.frame()` in #1225 and #1221, but it seems we have not fixed all issues.

In any case, whenever `NodeNG.parent` is missing we should raise the new `ParentMissingError`, or perhaps a new `FrameMissing` inheriting from it as we did for `NodeNG.statement()` in https://github.com/PyCQA/astroid/pull/1217.

This was quite difficult to do correctly and took some tries to get right, but you're more than welcome to start working towards it! I'm happy to provide support a long the way and answer any questions you might have!

tushar-deepsource:
I encountered this on 2.7.3, with pylint 2.10.2 I believe.

tushar-deepsource:
Actually, in my case the latest master fixes this issue. Thanks though!

tushar-deepsource:
It came back. I needed this fix to make inference work locally again today.

tushar-deepsource:
Does this seem good?
Should I do the same for `.statement()` as well?

astroid/nodes/node_ng.py
```diff
@@ -311,6 +310,13 @@ def frame(
 
         :returns: The first parent frame node.
         """
+        if self.parent is None:
+            if isinstance(
```
DanielNoord, line 314: I don't think this check should be here. If `self` is either of these `nodes` we shouldn't be in this function anyway as they have their `frame()` defined in their class definitions.

I think just raising the exception is fine.

I looked at `NamedExpr.frame()` which seems to raise `ParentMissingError`. I think if we go ahead with raising `FrameMissing` we might use this exception there as well. It shouldn't be a breaking change since `FrameMissing` inherits from `ParentMissingError`.

DanielNoord (commented):
Thanks for reopening the issue!

Would you mind adding a test for this as well? You can probably base it on the code that is currently throwing an error for you or look at how we tested `StatementMissing` in #1217!

astroid/exceptions.py
```diff
@@ -290,6 +290,22 @@ def __init__(self, target: "nodes.NodeNG") -> None:
...
+
+    Standard attributes:
+        target: The node for which the parent lookup failed.
+    """
+
+    def __init__(self, target: "nodes.NodeNG") -> None:
+        # pylint: disable-next=bad-super-call
+        super(ParentMissingError, self).__init__(
+            message=f"Frame not found on {target!r}"
+        )
```
cdce8p, line 306: I'm wondering if it's actually necessary to create a new exception subclass. All existing `frame()` and `scope()` methods raise just a `ParentMissingError`.

@DanielNoord I don't actually remember the exact details why we decided to add `StatementMissingError` maybe you do? Was there an actual need to have a more specific error or was it just an idea that we adopted?

astroid/nodes/node_ng.py
cdce8p: The `isinstance` check is indeed unnecessary.  Just this would be enough:
```py
        if not self.parent:
            raise ParentMissingError(target=self)

        return self.parent.frame()
```

--
Technically this is a small breaking change as users could be fine with catching `AttributeError`.
We added `future=True` to `statement()` so just for completeness we should probably do that here, too. You can take a look at #1217 as example.

cdce8p (changes requested):
> Should I do the same for .statement() as well?

@tushar-deepsource What do you mean with that? @DanielNoord mentioned it already but we reworked `statement()` in #1217. With `future=True` this should already work.

This change will also need a changelog entry and tests.

Would you feel confident doing these changes? It should be relatively straightforward if you follow the lead of #1217. Please let us know if you need help or if we should take this one over. For a first PR this is quite complex.

astroid/exceptions.py
DanielNoord: I think this followed from the logic that we were catching the `AttributeError` ourselves so it sort of naturally followed that we had to create a new error to catch. 
It might be a bit overkill, but I think it makes sense to make a new subclass: the more fine grained the error is the better users will know how to handle it.

DanielNoord:
> To double check: I'd have to essentially replicate the type stubs and tests for `statement`, for `frame` in this PR?
> 
> If the changes so far make sense, I can go ahead and do that.

I'll wait for @cdce8p to comment on whether he thinks that subclassing `ParentMissingError` makes sense here. Either way we should raise something for `if not node.parent`.

Other than that I think the following todo list should bring this PR a long way towards getting merged:
1) Add keyword only argument `future` to `frame()`
2) Add check for `future` argument and raise `Frame/ParentMissingError` for `if not node.parent`
3) Add `DeprecationWarning` and `TODO` for calls to `frame()` which raise `AttributeError` (also raise `AttributeError`)
4) Add overloads to `frame()`
5) Add tests for all variants of `frame()` introduced by this PR. See #1217 for examples.

Follow-up PR:
6) Add `future` keyword to all internal calls to `frame()` (see #1235)

For all of these changes #1217 and #1235 and the discussions within those PR could help.

cdce8p:
Thanks for the summery @DanielNoord! That looks about right to me.

> I'll wait for @cdce8p to comment on whether he thinks that subclassing `ParentMissingError` makes sense here.

Usually just subclassing an exception doesn't really hurt much. However in this case it also doesn't add any more information. I personally don't really see a need to be even more specific than `ParentMissingError` already is. Adding `FrameMissing` just introduces unnecessary complexity both for users and us. So I would recommend to stick with the `ParentMissingError`.

DanielNoord:
> Usually just subclassing an exception doesn't really hurt much. However in this case it also doesn't add any more information. I personally don't really see a need to be even more specific than `ParentMissingError` already is. Adding `FrameMissing` just introduces unnecessary complexity both for users and us. So I would recommend to stick with the `ParentMissingError`.

Fine with me! 

@tushar-deepsource Let me know if you need any help with implementing this. Happy to help as always 😄

DanielNoord:
@tushar-deepsource I think you might need to revert those latest commits and force push/rebase your branch on `main`. So revert back to https://github.com/PyCQA/astroid/pull/1263/commits/b380cfbb335fee90e1f0159fb0fa88f39b96ae14 and then rebase. That should also help fix the merge conflict and clean up the commit history.

astroid/nodes/as_string.py
```diff
@@ -99,6 +99,9 @@ def _should_wrap(self, node, child, is_left):
 
     # visit_<node> methods ###########################################
 
+    def visit_unknown(self, node):
+        return "astroid.Unknown()"
+
```
tushar-deepsource, line 104:
```suggestion\
```

tushar-deepsource:
@DanielNoord as far as I can see, `frame()` doesn't need overloads.

DanielNoord:
> @DanielNoord as far as I can see, `frame()` doesn't need overloads.

I have not done a full review, but I'll look into this!

> I also realised that the `NoReturn` in the union for `statement()` types was unnecessary, so I got rid of that as well. `NoReturn` is intended to only be used when the function never returns anything.

This is not actually true, see: https://github.com/PyCQA/astroid/pull/1217#discussion_r742896662

It seems unnecessary as we're not running `mypy` in strict mode, but it should actually be there.

tushar-deepsource:
> It seems unnecessary as we're not running mypy in strict mode, but it should actually be there.

@DanielNoord mypy will never complain about this. I believe what you mentioned there is a pylance bug.

```console
$ cat code.py  
import sys
from typing import NoReturn

class C:
    def f(self) -> int:
        return 5

class D:
    def f(self) -> NoReturn:
        sys.exit()

$ mypy --strict code.py
Success: no issues found in 1 source file
```

DanielNoord:
> > It seems unnecessary as we're not running mypy in strict mode, but it should actually be there.
> 
> @DanielNoord mypy will never complain about this. I believe what you mentioned there is a pylance bug.
> 
> ```
> $ cat code.py  
> import sys
> from typing import NoReturn
> 
> class C:
>     def f(self) -> int:
>         return 5
> 
> class D(C):
>     def f(self) -> NoReturn:
>         sys.exit()
> 
> $ mypy --strict code.py
> Success: no issues found in 1 source file
> ```

Ah, you're right! Sorry about that!

Would you mind opening a separate PR for this? We're likely going to squash merge the change in this PR and I think this change deserves its own commit.

astroid/nodes/node_classes.py
```diff
@@ -5006,7 +5007,9 @@ def postinit(self, target: NodeNG, value: NodeNG) -> None:
     See astroid/protocols.py for actual implementation.
     """
 
-    def frame(self):
+    def frame(
+        self, *, future: Literal[None, True] = None
+    ) -> Union["nodes.FunctionDef", "nodes.Module", "nodes.ClassDef", "nodes.Lambda"]:
```
DanielNoord, line 5012: Seems like we want ahead and raised `ParentMissingError` here without any deprecation.. Oh well, damage is done already 😄

DanielNoord (approved):
Thanks @tushar-deepsource 👍 

Let's wait for @cdce8p to review this, but this looks good to me.

Do you want to do the follow-up PR of adding `future` to all internal `frame()` calls as well?

astroid/nodes/node_classes.py
tushar-deepsource: Yeah.

tushar-deepsource:
> Do you want to do the follow-up PR of adding future to all internal frame() calls as well?

@DanielNoord Sure

tushar-deepsource:
@DanielNoord If someone could begin with sorting out the mypy issues with astroid as of now, I could also help with wrapping all of that up. I'd really like astroid to be fully type checked.

DanielNoord:
> @DanielNoord If someone could begin with sorting out the mypy issues with astroid as of now, I could also help with wrapping all of that up.

We need to fix (or start with) https://github.com/PyCQA/astroid/issues/1250 first for this. Some of the remaining issues can be resolved by adding assertions and `isinstance` checks, but those often shouldn't be necessary. I'm hesitant/against adding these now temporarily as we'll likely lose track of them and never remove them afterwards.

If you want you can have a direct effort on the typing efforts by looking at https://github.com/PyCQA/astroid/pull/1292#discussion_r769118968. You seem to have more knowledge of `mypy` and typing than I do, so perhaps you can come up with a neat solution.

> I'd really like astroid to be fully type checked.

I think we all really want this 😄

astroid/nodes/node_ng.py
```diff
@@ -323,7 +322,19 @@ def frame(
 
         :returns: The first parent frame node.
         """
-        return self.parent.frame()
+        if self.parent is None:
+            if future:
+                raise ParentMissingError(target=self)
+            warnings.warn(
+                "In astroid 3.0.0 NodeNG.frame() will return either a Frame, "
```
cdce8p, line 329:
```suggestion
                "In astroid 3.0.0 NodeNG.frame() will return either a Frame node, "
```
Technically those could be `FunctionDef`, `Module`, `ClassDef`, and `Lambda`. `Frame node` is probably enough though.

astroid/nodes/node_ng.py
```diff
@@ -323,7 +322,19 @@ def frame(
...
+                "or raise ParentMissingError. AttributeError will no longer be raised. "
+                "This behaviour can already be triggered "
+                "by passing 'future=True' to a statement() call.",
```
cdce8p, line 332:
```suggestion
                "by passing 'future=True' to a frame() call.",
```

cdce8p (changes requested):
Just some minor comments. Looks good otherwise!

cdce8p:
As for the typing discussion, let me at my opinion.

> I'd really like astroid to be fully type checked.

Although I would also like for astroid (and eventually pylint) to be fully typed, it's not the silver bullet that you might image it is.

> We need to fix (or start with) #1250 first for this. Some of the remaining issues can be resolved by adding assertions and isinstance checks, but those often shouldn't be necessary. I'm hesitant/against adding these now temporarily as we'll likely lose track of them and never remove them afterwards.

Please don't! Assertions don't guarantee anything and can easily be disabled with a Python flag. Those should not be used for typing!

astroid/nodes/node_ng.py
```diff
@@ -314,7 +313,7 @@ def statement(
         return self.parent.statement(future=future)
 
     def frame(
-        self,
+        self, *, future: Literal[None, True] = None
```
Pierre-Sassoulas, line 316: Why not
```suggestion
        self, *, future: bool = False
```
?

Pierre-Sassoulas (commented):
Looks good, I have a small question :)

DanielNoord:
> Please don't! 

I'll promise I won't 😄

astroid/nodes/node_ng.py
cdce8p: The idea behind it was to only allow two specific call options:
* `node.frame()`
* `node.frame(future=True)`

It doesn't really make sense to add `future=None` or even `future=False` if you don't support the future case.

--
After your comment I noticed that we probably need to add overloads here to make that happen.

```py
@overload
def frame(self) -> Union[<all types>]:
    ...
@overload
def frame(self, *, future: Literal[True]) -> Union[<all types>]:
    ...
```
This needs to be added to **all** frame definitions as overloads aren't inherited.

_The overloads for `statement()` also need to be fixed, but I can do that in a separate PR._

cdce8p: This probably also need a `pylint: disable` comment, as there is a false-positive with the `arguments-differ` message and overloads: https://github.com/PyCQA/pylint/issues/5264

```py
    # pylint: disable-next=arguments-differ
    # https://github.com/PyCQA/pylint/issues/5264
```

cdce8p: For reference, the PR to fix the `node.statement()` overload typing: #1308

cdce8p:
@tushar-deepsource Looking at the file changes, it seems the rebase didn't quite work. Would you mind taking another look at it?

astroid/nodes/node_ng.py
```diff
@@ -282,7 +286,7 @@ def statement(self, *, future: Literal[True]) -> "nodes.Statement":
 
     def statement(
         self, *, future: Literal[None, True] = None
-    ) -> Union["nodes.Statement", "nodes.Module"]:
+    ) -> Union["nodes.Statement", "nodes.Module", "NoReturn"]:
```
tushar-deepsource, line 289:
```suggestion
    ) -> Union["nodes.Statement", "nodes.Module"]:
```

astroid/nodes/node_ng.py
```diff
@@ -32,12 +32,16 @@
...
 
+    if sys.version_info >= (3, 6, 2):
+        from typing import NoReturn
+    else:
+        from typing_extensions import NoReturn
+
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
```
tushar-deepsource, line 44:
```suggestion


```

tushar-deepsource:
Should be good now

astroid/nodes/node_ng.py
```diff
@@ -32,6 +32,11 @@
 if TYPE_CHECKING:
     from astroid import nodes
 
+    if sys.version_info >= (3, 6, 2):
+        pass
+    else:
+        pass
+
```
tushar-deepsource, line 39:
```suggestion
```

astroid/nodes/node_ng.py
cdce8p: Turns out the overload idea doesn't actually work as I though. Sorry for the confusion. The default with `None` is needed for the parent calls.
```py
...
return self.parent.frame(future=future)
```
I've already opened another PR to fix the `node.statement` calls #1317.

cdce8p (approved):
Looks good overall!

@Pierre-Sassoulas @DanielNoord What do you guys think, is a changelog entry necessary? We didn't add one for #1217. In theory everyone that depends on the `AttributeError` should get a `DeprecationWarning`. We can (or maybe should ?) include a note when we eventually release `3.0.0`.

Pierre-Sassoulas (changes requested):
>  is a changelog entry necessary? 

I think so, could you add a notice about the deprecation in the changelog @tushar-deepsource , please ? :)

Pierre-Sassoulas (approved):
Thank you ! This is blocked by the release of 2.9.1, we'll merge as soon as 2.9.1 is out :)

DanielNoord:
This should go in `2.9.1` as it is technically a bug fix 😉

Pierre-Sassoulas (changes requested):
Thanks for the comment @DanielNoord  :) Could you move the changelog to 2.9.1 @tushar-deepsource ? I can't suggest/apply it in interface.

Pierre-Sassoulas (approved):
👌

cdce8p:
I think we could merge this one as is. The test errors are unrelated to the changes made here.

PR #1266, opened 2021-11-21
Always treat __class_getitem__ as a classmethod

mitya57:
## Steps

## Description

Consider this code:
```python
class Foo:
    def __class_getitem__(cls, *args, **kwargs):
        return cls
```
    
Without this change, astroid would treat `cls` as an instance of Foo, and thus it will treat Foo[bar] also as an instance, while really it is the same as Foo.

In my project, it fixes pylint warnings like this one:

    conferences/serializers.py:18:0: E0239: Inheriting 'serializers.ModelSerializer[models.Conference]', which is not a class. (inherit-non-class)

where ModelSerializer is a django-rest-framework class having the same definition of `__class_getitem__` as in my example.

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -1570,6 +1570,8 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
                 return "classmethod"
             if self.name == "__init_subclass__":
                 return "classmethod"
+            if self.name == "__class_getitem__":
+                return "classmethod"
 
             type_name = "method"
 
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -4612,7 +4612,7 @@ class TestBool(unittest.TestCase):
             def __class_getitem__(cls, *args, **kwargs):
                 return cls
 
-        Foo[True]
+        Foo[int]
         """
         )
         inferred = next(node.infer())
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -4605,6 +4605,20 @@ class TestBool(unittest.TestCase):
             inferred = next(node.infer())
             self.assertEqual(inferred, util.Uninferable)
 
+    def test_class_subscript(self) -> None:
+        node = extract_node(
+            """
+        class Foo:
+            def __class_getitem__(cls, *args, **kwargs):
+                return cls
+
+        Foo[True]
+        """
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (changes requested):
Thank you for opening this MR, could you add a test in astroid and provide a snippet with the pylint false positive using django-rest framework that failed before and succeed after your fix, please ?

mitya57:
Added a test. Without my change, it fails with:

    AssertionError: <Instance of .Foo at 0x139650032721584> is not an instance of <class 'astroid.nodes.scoped_nodes.ClassDef'>

Snippet:

```python
from rest_framework import serializers
from . import models

class Conference(serializers.ModelSerializer[models.Conference]):
    class Meta:
        model = models.Conference
        fields = ["id", "name"]
```

tests/unittest_inference.py
```diff
@@ -4605,6 +4605,20 @@ class LenInvalid(object):
...
             self.assertEqual(inferred, util.Uninferable)
 
+    def test_class_subscript(self) -> None:
+        node = extract_node(
+            """
+        class Foo:
+            def __class_getitem__(cls, *args, **kwargs):
+                return cls
+
+        Foo[True]
```
cdce8p, line 4615:
```suggestion
        Foo[int]
```
This shouldn't change much, but usually the subscript is a type.

cdce8p (approved):
Thanks @mitya57! This looks great 🚀
I'll just add two small commits, as we retarget this one to the `2.9.0` release.

astroid/nodes/scoped_nodes.py
```diff
@@ -1570,6 +1570,8 @@ def type(
                 return "classmethod"
             if self.name == "__init_subclass__":
                 return "classmethod"
+            if self.name == "__class_getitem__":
+                return "classmethod"
```
cdce8p, line 1574: @Pierre-Sassoulas Just noticed that `__class_getitem__` was added in Python `3.7`. Adding a `sys.version_info` check here doesn't make much sense. It would block the usage of `py-version` for it in `pylint`. However we might want to consider updating the `getitem` logic (in pylint). The test case I wrote in https://github.com/PyCQA/pylint/pull/5354 shouldn't pass if `py-version = 3.6`.

Pierre-Sassoulas (approved):
Thank you, for the MR, congratulation on your first contribution :) !

PR #1267, opened 2021-11-21
Add an exception at install for python < 3.6.2

Pierre-Sassoulas:
## Type of Changes

## Related Issue

See https://github.com/PyCQA/pylint/issues/5065 and https://github.com/PyCQA/pylint/issues/5250

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@ What's New in astroid 2.9.0?
 ============================
 Release date: TBA
 
+* The last version compatible with python '3.6.0' and '3.6.1' is astroid '2.6.6'. We did not
+  realize that when adding incompatible typing at the time, and versions might be broken
+  for this interpreter. 2.9.0 meta-information will permit to download pylint on those
+  interpreters but the installation will fail and tell you to install '2.6.6' instead.
+  astroid 2.9.1 will require python >= 3.6.2.
 
 
 What's New in astroid 2.8.6?
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,19 @@
+import sys
+
 from setuptools import setup
 
+
+class AstroidIncompatiblePythonError(Exception):
+    def __init__(self) -> None:
+        super().__init__(
+            "The last version compatible with python '3.6.0' and '3.6.1' is astroid '2.6.6'"
+            f"You're using {'.'.join([str(v) for v in sys.version_info[:3]])}. "
+            "Please install astroid 2.6.6 explicitly or upgrade your python interpreter "
+            "to at least 3.6.2. Remember that Python 3.6 end life is December 2021. "
+            "See https://github.com/PyCQA/pylint/issues/5065 for more detail."
+        )
+
+
+if sys.version_info < (3, 6, 2):
+    raise AstroidIncompatiblePythonError()
 setup()


cdce8p (approved):
Looks good! It does even make sense to do it in `2.9.0`. Pylint `2.11.1` has a limit for astroid `<2.9.0` and the warning there is only included in `2.12.0`.

DanielNoord:
Should probably update this as well:
https://github.com/PyCQA/astroid/blob/935ff12e0d65899d3d48d1c7be9cbeeb1d9e6089/setup.cfg#L44

Pierre-Sassoulas:
> Should probably update this as well:

That's the tricky part, if we do that python users 3.6.0 will download 2.8.6 which is broken. So we're making our own error asking them to download astroid 2.6.6 before changing this value. (Same reasoning than with pylint 2.12.0 / 2.12.1)

cdce8p:
> Should probably update this as well:
> 
> https://github.com/PyCQA/astroid/blob/935ff12e0d65899d3d48d1c7be9cbeeb1d9e6089/setup.cfg#L44

No. We want this version to still be installable on `3.6` otherwise users would need see the error. That's also how we do it in pylint.

Something else I just realized, do we need this at all? AFAIK we don't have anything in `astroid` that breaks with Python 3.6.0.

Pierre-Sassoulas:
> do we need this at all? AFAIK we don't have anything in astroid that breaks with Python 3.6.0.

Well no-one complained yet 😄 To be fair I don't know exactly that's why I chose 2.6.6 (the astroid for thae last working pylint). I think @DanielNoord was upgrading the typing everywhere like a machine so the probability of astroid being broken was high in my mind.

cdce8p:
> > do we need this at all? AFAIK we don't have anything in astroid that breaks with Python 3.6.0.
> 
> Well no-one complained yet 😄 To be fair I don't know exactly that's why I chose 2.6.6 (the astroid for thae last working pylint). I think @DanielNoord was upgrading the typing everywhere like a machine so the probability of astroid being broken was high in my mind.

Typing itself isn't really a problem. The main issue in `pylint` is `typing.NamedTuple` that has some incompatibilities.

We should probably still update the min Python version to `3.6.2` but aside from that I don't think we need to do anything else.

cdce8p:
> We should probably still update the min Python version to `3.6.2` but aside from that I don't think we need to do anything else.

We could do this for astroid `2.9.0` as this will require pylint `2.12.0`.

cdce8p:
> > We should probably still update the min Python version to `3.6.2` but aside from that I don't think we need to do anything else.
> 
> ~We could do this for astroid `2.9.0` as this will require pylint `2.12.0`.~

**Edit**: If we require `3.6.2` now, the `pip` install will fail for pylint `2.12.0` and users will never see the exception message.

Pierre-Sassoulas:
Plus, if astroid is still compatible with 3.6.0 / 3.6.1 let's close this issue altogether ?

cdce8p:
> Plus, if astroid is still compatible with 3.6.0 / 3.6.1 let's close this issue altogether ?

Yes! I've opened #1269 as a followup, when it's time to require Python 3.6.2.

PR #1269, opened 2021-11-21
Require Python 3.6.2

cdce8p:
## Description

Require Python 3.6.2 going forward to better support typing. Similar to planned change in pylint with `2.12.1` https://github.com/PyCQA/pylint/pull/5068. There should be at least one version supported by pylint `2.12.0` without this change! I.e. if pylint `2.12.0` requires `2.9.0`, this should only go into `>= 2.9.1`.

Related #1267


Pierre-Sassoulas (approved):
👍

PR #1270, opened 2021-11-21
Resolve dependency conflicts with pylint

cdce8p:
## Description

`astroid` and `pylint` create circular dependencies which sometimes causes CI to fail completely. This happens if we update the minor version of astroid while still using a pylint version which is incompatible.

This PR doesn't completely fix the issue. It just separates the installs of `pylint` and `astroid` so that the `pip` dependency resolver doesn't complain.

## Type of Changes

## Related Issue
https://github.com/PyCQA/astroid/runs/4279233543?check_suite_focus=true#step:6:70

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,6 +52,7 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt -r requirements_test_brain.txt
+          pip install -e .
       - name: Generate pre-commit restore key
         id: generate-pre-commit-key
         run: >-
@@ -155,6 +156,7 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt -r requirements_test_brain.txt
+          pip install -e .
 
   pytest-linux:
     name: Run tests Python ${{ matrix.python-version }} (Linux)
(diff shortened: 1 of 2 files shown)


cdce8p:
This change still results in an error, but it isn't blocking anymore.
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed.
    This behaviour is the source of the following dependency conflicts.
Successfully installed astroid-2.9.1.dev0
pylint 2.11.1 requires astroid<2.9,>=2.8.0, but you have astroid 2.9.1.dev0 which is incompatible.
```

For now that should work. Will be interesting when we decide to do the major version jump 😄

Pierre-Sassoulas (commented):
Should we use ``pip install --no-deps pylint==2.11.1`` when installing pylint instead ? (We know what we're doing when we use the latest astroid with a past pylint version, so we can silence pip)

Edit : https://github.com/pypa/pip/issues/7339 seems to suggest that we need to isolate pylint to signle it out for no-deps

cdce8p:
> Should we use `pip install --no-deps pylint==2.11.1` when installing pylint instead ? (We know what we're doing when we use the latest astroid with a past pylint version, so we can silence pip)

We would then need to make sure to install the `pylint` dependencies ourselves too. Don't know if it's worth it.
```cfg
install_requires =
    platformdirs>=2.2.0
    astroid>=2.9.0,<2.10  # obvious
    isort>=4.2.5,<6
    mccabe>=0.6,<0.7
    toml>=0.9.2
    colorama;sys_platform=="win32"
    typing-extensions>=3.10.0;python_version<"3.10"  # required for astroid too

Pierre-Sassoulas (approved):

PR #1272, opened 2021-11-22
fix deque.insert() signature

### LINKED ISSUE #1260: deque.insert() has reversed args
### Steps to reproduce

https://docs.python.org/3/library/collections.html#collections.deque.insert

```python
class MyDeque(collections.deque):
    def insert(i, x): pass
```

### Current behavior
pylint says:
```
W0237: Parameter 'x' has been renamed to 'i' in overridden 'MyDeque.insert' method (arguments-renamed)
W0237: Parameter 'i' has been renamed to 'x' in overridden 'MyDeque.insert' method (arguments-renamed)
```
### Expected behavior
no error

### Version
2.8.4

belm0:
## Description

The collections brain had the arguments of `deque.insert()` reversed, which would case a false `arguments-renamed` error if an overriding class uses the correct order.

## Type of Changes

Closes #1260

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,10 @@ Release date: TBA
 
 * Require Python 3.6.2 to use astroid.
 
+* Fix ``deque.insert()`` signature in ``collections`` brain.
+
+  Closes #1260
+
 
 What's New in astroid 2.9.0?
 ============================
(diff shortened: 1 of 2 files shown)


cdce8p (approved):
Looks good! Thanks you @belm0 🐬

PR #1276, opened 2021-11-26
Add doc_node attribute

kasium:
## Steps

## Description

This change adds a new attribute `doc_node` to `ClassDef`, `FunctionDef`, `AsyncFunctionDef` and `Module`. This is needed, since some checks in pylint not only need the docstring as a string, but the node containing it. For details see https://github.com/PyCQA/pylint/issues/3077

## Type of Changes

## Related Issue

https://github.com/PyCQA/pylint/issues/3077


tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,7 @@ def func():
     """
     )
     assert node.doc == "Docstring"
+    assert node.doc_node
```
Pierre-Sassoulas, line 1586: It would be interesting to test more than that here. In particular information about the lineno, line_end, column, column end is important.

Pierre-Sassoulas (changes requested):
Thank you for this, I expected more work to be fair, this is rather elegant. As it's a "big" features I'll wait for opinion on it and it's going to be in astroid 2.10.

DanielNoord:
I haven't reviewed the code yet, but can't this be added to all nodes?

```python
VARIABLE = 1
"""A variable I will use somewhere"""
```

IDEs know that this docstring belongs to the variable. If we're creating this new attribute we might as well try and do it for all nodes? This would allows allow us to call `NodeNG.doc_node` instead of having to do an `isinstance` first.

DanielNoord:
I haven't reviewed the code yet, but can't this be added to all nodes?

```python
VARIABLE = 1
"""A variable I will use somewhere"""
```

IDEs know that this docstring belongs to the variable. If we're creating this new attribute we might as well try and do it for all nodes? This would allows allow us to call `NodeNG.doc_node` instead of having to do an `isinstance` first.

tests/unittest_nodes.py
kasium: Added tests; end line/column are optional and depend on the python version

kasium:
> I haven't reviewed the code yet, but can't this be added to all nodes?
> 
> ```python
> VARIABLE = 1
> """A variable I will use somewhere"""
> ```
> 
> IDEs know that this docstring belongs to the variable. If we're creating this new attribute we might as well try and do it for all nodes? This would allows allow us to call `NodeNG.doc_node` instead of having to do an `isinstance` first.

To doc_node is only used for four nodes, so does it really make sense to add it to all. Especially if ``doc`` is also just present for four nodes?

DanielNoord:
> To doc_node is only used for four nodes, so does it really make sense to add it to all. Especially if `doc` is also just present for four nodes?

As I said, one of the benefits would be that we can do `node.doc_node` without first inferring the type of node without running into issues. This is quite a nice benefit!
A second thing is that we never know when we might need this information. Perhaps somebody would like to make a highly opinionated checker that forces all global constants to have a docstring? We might as well implement this in its fulllest and most productive form 😄 

By the way, I'm happy to assist in changing this so it is added to all `node` types! Just let me know if I can help!

kasium:
@DanielNoord okay, the let me add it to all nodes. Which is the right class for that?

DanielNoord:
> @DanielNoord okay, the let me add it to all nodes. Which is the right class for that?

`NodeNG` is the base class used by all other `nodes`. That would be the first place to start looking.

kasium:
> > @DanielNoord okay, the let me add it to all nodes. Which is the right class for that?
> 
> `NodeNG` is the base class used by all other `nodes`. That would be the first place to start looking.

done

.gitignore
```diff
@@ -15,3 +15,4 @@ astroid.egg-info/
 .pytest_cache/
 .mypy_cache/
 venv
+.vscode/
```
DanielNoord, line 18: @Pierre-Sassoulas I saw an old discussion on `pylint` about this. Do we want to commit this to the project?

astroid/nodes/node_ng.py
```diff
@@ -84,6 +84,8 @@ class NodeNG:
     """Attributes that contain AST-dependent fields."""
     # instance specific inference function infer(node, context)
     _explicit_inference = None
+    doc_node: Optional["nodes.Const"] = None
+    """The doc node."""
```
DanielNoord, line 88:
```suggestion
    """The doc node associated with this node."""
```

astroid/nodes/scoped_nodes.py
```diff
@@ -484,6 +484,7 @@ def __init__(
         package=None,
         parent=None,
         pure_python=True,
+        doc_node=None,
```
DanielNoord, line 487: Please add typing to this parameter here, like it has been done for `path`. The typing in the docstring can be removed as we're moving to using typing within the code.
Same goes for the other occurrences of this pattern.

tests/unittest_nodes.py
DanielNoord: Could we also test (for example) the type of `node.doc_node`? And perhaps the value of the node?
You should also be able to add tests for `end_lineno`, either by creating a special test for `3.8`+ or with something like we do ourselves in `pylint` see:
https://github.com/PyCQA/pylint/blob/6caab13816e8098063b6e077ab0044b94aa87557/tests/testutils/test_output_line.py#L55-L62

tests/unittest_nodes.py
```diff
@@ -1592,6 +1595,7 @@ def func():
     """
     )
     assert node.doc is None
+    assert node.doc_node is None
```
DanielNoord, line 1598: Some additional tests would be appreciated 😄

astroid/rebuilder.py
```diff
@@ -119,10 +120,16 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
                 ):
                     doc = first_value.value if PY38_PLUS else first_value.s
                     node.body = node.body[1:]
-                    return node, doc
+                    doc_node = nodes.Const(
+                        value=doc,
+                        lineno=first_value.lineno,
+                        col_offset=first_value.col_offset,
+                        parent=node,
```
DanielNoord, line 127: This should include the `end_line` and `end_column`. You can look at what is passed when we create other `Const` nodes here:
https://github.com/PyCQA/astroid/blob/6691a9e97373b28e15a76b747f78c041f2064df4/astroid/rebuilder.py#L1939-L1957

DanielNoord (changes requested):
Sorry for taking a little longer to review this!

Although this wouldn't actually add `doc_node` to `Const` nodes, I think it is a good start. Adding it as an attribute to `NodeNG` makes sense to me anyway.

Please let me know if you have any questions! Happy to help move this along 😄

.gitignore
kasium: reset and ignored locally

tests/unittest_nodes.py
kasium: any idea what to test?

tests/unittest_nodes.py
kasium: done

astroid/nodes/scoped_nodes.py
```diff
@@ -504,9 +505,12 @@ def __init__(
         :param parent: The parent node in the syntax tree.
         :type parent: NodeNG or None
 
+        :param doc_node: The doc node.
```
DanielNoord, line 508:
```suggestion
        :param doc_node: The doc node associated with this node.
```

astroid/nodes/scoped_nodes.py
```diff
@@ -1566,6 +1572,8 @@ def __init__(
...
```
DanielNoord, line 1575:
```suggestion
        :param doc_node: The doc node associated with this node.
```

astroid/nodes/scoped_nodes.py
```diff
@@ -2205,6 +2215,8 @@ def __init__(
...
```
DanielNoord, line 2218:
```suggestion
        :param doc_node: The doc node associated with this node.
```

astroid/rebuilder.py
```diff
@@ -104,13 +104,14 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(
+        self, node: T_Doc
+    ) -> Tuple[T_Doc, Optional[str], Optional[nodes.Const]]:
```
DanielNoord, line 109: Just picking up on this now:
Why do we return `node` here? Looking at the three instances in which `_get_doc` is called I don't immediately see why we would return an unchanged value.. Do you know this? If not, would you be so kind as to remove it? We could then add a docstring saying that this:
`Returns the docstring of a node both in str and nodes.Const` representation`.

astroid/rebuilder.py
```diff
@@ -119,10 +120,26 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
...
                     doc = first_value.value if PY38_PLUS else first_value.s
                     node.body = node.body[1:]
-                    return node, doc
+                    if PY38_PLUS:
+                        doc_node = nodes.Const(
+                            value=doc,
+                            lineno=first_value.lineno,
+                            col_offset=first_value.col_offset,
+                            parent=node,
+                            end_lineno=node.end_lineno,
```
DanielNoord, line 129: Shouldn't this be `first_value`? Looking at the test this does not seem correct. The docstring only spans one line, but `lineno` and `end_lineno` are not the same.

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,12 @@ def func():
     """
     )
     assert node.doc == "Docstring"
+    assert isinstance(node.doc_node, nodes.Const)
+    assert node.doc_node.lineno == 3
+    assert node.doc_node.col_offset == 4
+    if PY38_PLUS:
+        assert node.doc_node.end_lineno == 4
+        assert node.doc_node.end_col_offset == 12
```
DanielNoord, line 1591:
```suggestion
        assert node.doc_node.end_col_offset == 12
    else:
        assert node.doc_node.end_lineno == None
        assert node.doc_node.end_col_offset == None
```

tests/unittest_nodes.py
DanielNoord: Sorry I misread this test. This is indeed correct.

Could you add three more tests scenarios? One for `ClassDef` (both with and without docstring), `Module` (with and without) and `NodeNG` that asserts that `doc_node is None`. That way we know we initialise the attribute correctly on all other nodes!

DanielNoord (changes requested):

astroid/rebuilder.py
kasium: done

tests/unittest_nodes.py
kasium: So, should I just add test classes to unittest_nodes.py? Or is there a better way/place?

DanielNoord: Yeah although `ClassDef` and `Module` are likely in `tests/unittest_scoped_nodes.py`. All nodes have a existing test class so you should be able to append to them.

kasium: I just added the tests for module node. Can you please check if they are fine? I didn't check the line, column etc. since this is already done in test_get_doc

Pierre-Sassoulas (approved):

tests/unittest_nodes.py
DanielNoord: Please do add these as well. We never know what intricacies could break the expected behaviour. Especially considering `Module` is such a different compared to `ClassDef` and `FunctionDef`.
Let's make sure we test all expected attributes and values of the `doc_node` attribute. It adds a little bit of extra code to our test suite but makes sure they are much more robust. We could then also add these to `ClassDef`.

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,15 @@ def func():
...
     assert node.doc == "Docstring"
+    assert isinstance(node.doc_node, nodes.Const)
+    assert node.doc_node.lineno == 3
+    assert node.doc_node.col_offset == 4
+    if PY38_PLUS:
+        assert node.doc_node.end_lineno == 3
+        assert node.doc_node.end_col_offset == 15
+    else:
+        assert node.doc_node.end_lineno is None
+        assert node.doc_node.end_col_offset is None
```
DanielNoord, line 1594: Could you add a multiline example? I'm would like to check that `end_lineno` is still correct

DanielNoord (changes requested):
Code itself looks good. We only need make the tests a little more robust 😄

tests/unittest_nodes.py
kasium: added

Pierre-Sassoulas (approved):
The tests looks complete and covering a lot of cases properly, thank you :)

DanielNoord (approved):
Thanks @kasium! Tests look very good indeed, thanks for taking the time the write these. Much appreciated! 😄

DanielNoord:
Adding `blocked` because this is scheduled for `2.10`.

DanielNoord:
Seems the tests are actually catching a real bug. @kasium it seems like we're not building the `doc_node` correctly on < `3.8`

kasium:
No clue why some tests are failing...

kasium:
@DanielNoord will take a look

kasium:
@DanielNoord no clue why, but on 3.7 the docstrings lineno is the last line instead of the first line....

kasium:
@DanielNoord @Pierre-Sassoulas so depending on the python version and implementation, the lineno of the node is either  the starting line or the end line. I managed now to include all beside pypy. When running pypy3 which is not 3.8, then the lineno is the start line. How can I managed to represent that in tests?

DanielNoord:
> @DanielNoord no clue why, but on 3.7 the docstrings lineno is the last line instead of the first line....

I saw @cdce8p requested a self-review and he know much more about the ast parser and line numbering. So I would like to see his opinion on this.

That being said, I don't think we should settle for showing the end line as the first line on 3.6 and 3.7. We know that that is incorrect so it seems counter-intuitive to me to do so. I'm not sure, but perhaps it would be better to then only initialise the `doc_node` on 3.8+? What extra information does `doc_node` give us in 3.7 compared to `doc`? We know the value of the `Const` and the line numbering is incorrect, so its benefit seems minimal?

kasium:
@DanielNoord we need the doc_node especially for 3.7/3.6 for https://github.com/PyCQA/pylint/issues/3077. In 3.8 it's not longer needed 😄. However, what about to say lineno=-1?

Pierre-Sassoulas:
>  I don't think we should settle for showing the end line as the first line on 3.6 and 3.7. We know that that is incorrect so it seems counter-intuitive to me to do so.

Sadly the issue come from the ast module that we do not control. End of life approach slowly for python 3.7 (27 Jun 2023) and faster for python 3.6 (23 Dec 2021). Doing better than the lib we depend can be a huge time investment at time. That's why we chose to ignore the line/col for pylint's functional test when under python < 3.8.

Here I think @kasium wanted to do the [workaround from pyflakes](https://github.com/PyCQA/pyflakes/pull/273) to support python 3.6/3.7 even though [the upstream fix](https://github.com/python/cpython/pull/9731) was not available in python 3.7, so this is a special case (but generally we do not try to have better line/col than the ast).

If the time investment for the workaround is too high I would favor a "won't fix", as it's an issue with an old version of python.

kasium:
So in the end either the end line or the start line helps to receive the original quotes...

DanielNoord:
Are you sure that issue still exists? Running on `pylint` `main` I get:

```bash
❯ cat test.py
from unittest import mock

@mock.patch('mock.'
            'foo')
class Foo:
    """Blah"""
❯ pylint --load-plugins=pylint.extensions.docstyle test.py

Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

DanielNoord:
Okay, so I played around with this for a bit and although I don't have the solution yet I don't think using the actual ending lineno as lineno is the right way forward. There would be no way to discern that when trying to access it in `pylint` and this crucial difference can be easily forgotten. Thus it seems to be a really hacky way to solve this issue, which could potentially cause us headaches down the line.

```python
from unittest import mock

@mock.patch('mock.'
            'foo')
class Foo:
    """Blah"""
```

The problem is that on `3.7` the ast parser doesn't recognise/note that the `ClassDef` node actually "starts" on line 5 instead of at the decorator on line 3. The `bad-docstring-quotes` takes the `lineno` from `ClassDef`, adds one (which makes it 4), retrieves the string on that line and checks the quotes. Therefore in this example it sees `'foo'` and flags as incorrect.
Clearly this approach doesn't work for any code parsed on `3.7`. For example, if the decorator would be one line the checker actually considers `class Foo:` to be the docstring.

We could store the ending lineno on `3.7` in an attribute called `part_of_the_docstring_is_on_this_line`, but the following code would still trip up the checker as it thinks the docstring starts with `'`:
```python
from unittest import mock

@mock.patch('mock.'
            'foo')
class Foo:
    """
    'Blah'"""
``` 

I would like your opinion on this, but perhaps the best way forward is to create the `doc_node` on `3.8`+ as it actually gives us additional information compared to the `doc` attribute we have now and then restrict the checker to `3.8`+. Or we could do something like `part_of_the_docstring_is_on_this_line` but then we would need to rewrite the checker and think about how to check for quotes (ie, do we check the quotes at the end of the line? At the beginning of the line?)

------
```python
class ParentClass:
    pass
class ParentClass1:
    pass
class ParentClass2:
    pass
class ParentClass3:
    pass
class ParentClass4:
    pass
class ParentClass5:
    pass

class Foo(
    ParentClass, ParentClass1, ParentClass2, ParentClass3, ParentClass4, ParentClass5
):
    """Blah"""
```
Code like this also wouldn't be correctly recognised...

kasium:
@DanielNoord thanks for the analysis. Personally, I think your proposal makes sense

* doc_node 3.8+
* disable the checker for <= 3.7

cdce8p:
I didn't completely follow the discussion so far. Just to clarify, what is the actual goal you're trying to archive? Is it already possible by some other means, i.e. checking the first body node? If some features are only available on 3.8+, I would recommend not to do any workarounds for 3.6 and 3.7. In case users what the best linting results, they should stick to a more recent version. Workarounds only introduce additional maintenance requirements and can break quite easily.

DanielNoord:
> I didn't completely follow the discussion so far. Just to clarify, what is the actual goal you're trying to archive? Is it already possible by some other means, i.e. checking the first body node? If some features are only available on 3.8+, I would recommend not to do any workarounds for 3.6 and 3.7. In case users what the best linting results, they should stick to a more recent version. Workarounds only introduce additional maintenance requirements and can break quite easily.

This PR is adding a `Const` node to `Module`, `ClassDef` and `FunctionDef` nodes which represents the docstring of the particular node. 
Currently we add the docstring as a string attribute to the top node via the `doc` attribute. We lose information about line numbers that way. We also strip the node containing the docstring from the top `node.body`.
On `3.8+` we have access to good line number information so we can create a `Const` node which contains relevant information about the docstring. On `3.6` and `3.7` the line information is inconsistent so we have now decided to not initialise the `Const` node on those versions.

This PR was spawned by PyCQA/pylint#3077 and originally aimed to provide `Const` nodes on all versions so the `bad-docstring-quotes` checker could access them. However, after my investigation I think we have now also decided to disable this checker on `3.6` and `3.7`: the information about line numbers is so inconsistent that we cannot reliably determine on which line the docstring starts. This makes it hard/impossible to guarantee no false positives on these versions. Or put differently, on `3.8+` we can actually try and fix false positives, on the other versions we sometimes cannot due to limited information.

@kasium Can I ask you to change this PR into 1) only creating node on `3.8+`, 2) test that these nodes are correct + correct line numbers on `3.8+`, 3) test that they are not created on `3.6` and `3.7` (I think `doc_node` should be `None` on these versions).

kasium:
@DanielNoord sure I'll do that
@Pierre-Sassoulas should we then maybe disable the pylint check on <= 3.7

astroid/nodes/scoped_nodes.py
```diff
@@ -504,9 +505,12 @@ def __init__(
         :param parent: The parent node in the syntax tree.
         :type parent: NodeNG or None
 
+        :param doc_node: The doc node associated with this node.
+
         :param pure_python: Whether the ast was built from source.
         :type pure_python: bool or None
         """
+
```
DanielNoord, line 513:
```suggestion
```

astroid/rebuilder.py
```diff
@@ -104,13 +104,13 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(self, node: T_Doc) -> Tuple[Optional[str], Optional[nodes.Const]]:
+        """Returns the docstring of a node both in str and nodes.Const."""
```
DanielNoord, line 108: Let's also update the docstring here to reflect `nodes.Const` only being made on 3.8+.

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,14 @@ def func():
...
     )
     assert node.doc == "Docstring"
+    if PY38_PLUS:
+        assert isinstance(node.doc_node, nodes.Const)
+        assert node.doc_node.lineno == 3
+        assert node.doc_node.col_offset == 4
+        assert node.doc_node.end_lineno == 3
+        assert node.doc_node.end_col_offset == 15
+    else:
+        assert not node.doc_node
```
DanielNoord, line 1593: Let's do:
```suggestion
        assert node.doc_node is None
```

If `pylint` starts complaining let me know as then I think we should disable that message here (or in pylintrc). We want to know `node.doc_node is None` and not `node.doc_node == []` (which would pass this assertion).

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,46 @@ def test_stream_api(self) -> None:
...
+        """
+        module = builder.parse(data, __name__)
+        if PY38_PLUS:
+            assert isinstance(module.doc_node, nodes.Const)
+            assert module.doc_node.lineno == 2
+            assert module.doc_node.col_offset == 0
+            assert module.doc_node.end_lineno == 2
+            assert module.doc_node.end_col_offset == 17
+        else:
+            assert not module.doc_node
```
DanielNoord, line 307:
```suggestion
            assert module.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,46 @@ def test_stream_api(self) -> None:
...
+        """
+        module = builder.parse(data, __name__)
+        if PY38_PLUS:
+            assert isinstance(module.doc_node, nodes.Const)
+            assert module.doc_node.lineno == 2
+            assert module.doc_node.col_offset == 0
+            assert module.doc_node.end_lineno == 5
+            assert module.doc_node.end_col_offset == 3
+        else:
+            assert not module.doc_node
```
DanielNoord, line 325:
```suggestion
            assert module.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,46 @@ def test_stream_api(self) -> None:
...
+
+    def test_without_docstring(self) -> None:
+        data = """
+            foo = 1
+        """
+        module = builder.parse(data, __name__)
+        assert not module.doc_node
```
DanielNoord, line 332:
```suggestion
        assert module.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -734,6 +774,52 @@ def test(cls):
...
+        """
+        )
+        if PY38_PLUS:
+            assert isinstance(func.doc_node, nodes.Const)
+            assert func.doc_node.lineno == 3
+            assert func.doc_node.col_offset == 4
+            assert func.doc_node.end_lineno == 3
+            assert func.doc_node.end_col_offset == 21
+        else:
+            assert not func.doc_node
```
DanielNoord, line 792:
```suggestion
            assert func.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -734,6 +774,52 @@ def test(cls):
...
+        """
+        )
+        if PY38_PLUS:
+            assert isinstance(func.doc_node, nodes.Const)
+            assert func.doc_node.lineno == 3
+            assert func.doc_node.col_offset == 4
+            assert func.doc_node.end_lineno == 6
+            assert func.doc_node.end_col_offset == 7
+        else:
+            assert not func.doc_node
```
DanielNoord, line 812:
```suggestion
            assert func.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -734,6 +774,52 @@ def test(cls):
...
+
+    def test_without_docstring(self) -> None:
+        func = builder.extract_node(
+            """
+            def foo():
+                bar = 1
+        """
+        )
+        assert not func.doc_node
```
DanielNoord, line 821:
```suggestion
        assert func.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -2023,6 +2109,52 @@ def update(self):
...
+        """
+        )
+        if PY38_PLUS:
+            assert isinstance(node.doc_node, nodes.Const)
+            assert node.doc_node.lineno == 3
+            assert node.doc_node.col_offset == 4
+            assert node.doc_node.end_lineno == 6
+            assert node.doc_node.end_col_offset == 7
+        else:
+            assert not node.doc_node
```
DanielNoord, line 2147:
```suggestion
            assert node.doc_node is None
```

tests/unittest_scoped_nodes.py
```diff
@@ -2023,6 +2109,52 @@ def update(self):
...
+
+    def test_without_docstring(self) -> None:
+        node = builder.extract_node(
+            """
+            class Foo:
+                bar = 1
+        """
+        )
+        assert not node.doc_node
```
DanielNoord, line 2156:
```suggestion
        assert node.doc_node is None
```

DanielNoord (changes requested):
Thanks @kasium!

Some minor remarks, but this is almost good to go imo 👍

astroid/nodes/scoped_nodes.py
```diff
@@ -504,6 +505,8 @@ def __init__(
         :param parent: The parent node in the syntax tree.
         :type parent: NodeNG or None
 
+        :param doc_node: The doc node associated with this node.
```
DanielNoord, line 508: Could you also add the updated docstring to all these docstrings?

astroid/rebuilder.py
```diff
@@ -104,13 +104,16 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(self, node: T_Doc) -> Tuple[Optional[str], Optional[nodes.Const]]:
+        """Returns the docstring of a node both in str and nodes.Const.
+
+        The constant will only be returned in python 3.8+.
```
DanielNoord, line 110:
```suggestion
        The constant will only be returned in python 3.8+ because the improved ast parsers
        gives as additional information about line numbers.
```

DanielNoord (changes requested):
Small set of changes and then this is good to go!

astroid/rebuilder.py
```diff
@@ -119,10 +123,21 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
...
+                        doc_node = nodes.Const(
+                            value=doc,
+                            lineno=first_value.lineno,
+                            col_offset=first_value.col_offset,
+                            parent=node,
+                            end_lineno=first_value.end_lineno,
+                            end_col_offset=first_value.end_col_offset,
+                        )
+                    else:
+                        doc_node = None
```
cdce8p, line 136: Even if the check might not really work for Python 3.7, why can't we create the node for it? It might be useful for some other check.

astroid/rebuilder.py
```diff
@@ -119,10 +123,21 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
...
-                    return node, doc
+                    if PY38_PLUS:
+                        doc_node = nodes.Const(
+                            value=doc,
+                            lineno=first_value.lineno,
+                            col_offset=first_value.col_offset,
+                            parent=node,
+                            end_lineno=first_value.end_lineno,
+                            end_col_offset=first_value.end_col_offset,
+                        )
```
cdce8p, line 134: I believe there is an easier way to create a `Const` node. Try
```py
doc_node = self.visit(node.body[0], node)
```
It obviously needs to happen, before the node is removed from the body, but then it should work. Even for Python 3.7.

Furthermore, you should be able to simplify the `doc` definition by using the newly create node.
```py
doc = doc_node.value
```

astroid/rebuilder.py
```diff
@@ -104,13 +104,17 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(self, node: T_Doc) -> Tuple[Optional[str], Optional[nodes.Const]]:
+        """Returns the docstring of a node both in str and nodes.Const.
+
+        The constant will only be returned in python 3.8+ because the improved ast parsers
+        gives as additional information about line numbers.
```
cdce8p, line 111: This isn't true anymore.

astroid/rebuilder.py
```diff
@@ -104,13 +104,17 @@ def __init__(
...
+        """
         try:
             if PY37_PLUS and hasattr(node, "docstring"):
                 doc = node.docstring  # type: ignore[union-attr,attr-defined] # mypy doesn't recognize hasattr
-                return node, doc
+                return doc, None
```
cdce8p, line 116: When is this called? AFAIK no AST nodes have a `docstring` attribute.
It was added here https://github.com/PyCQA/astroid/commit/2c655de34603e5a0f88b369d3e4b68556fdb0641 together with `typed_ast` but even that doesn't seem to use `docstring`.

Am I missing something?

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,46 @@ def test_stream_api(self) -> None:
...
+        """
+        module = builder.parse(data, __name__)
+        if PY38_PLUS:
+            assert isinstance(module.doc_node, nodes.Const)
+            assert module.doc_node.lineno == 2
+            assert module.doc_node.col_offset == 0
+            assert module.doc_node.end_lineno == 2
+            assert module.doc_node.end_col_offset == 17
+        else:
+            assert module.doc_node is None
```
cdce8p, line 307: Only `doc_node.end_lineno` and `doc_node.end_col_offset` should be `PY38_PLUS` exclusive.

cdce8p (changes requested):
I finally had time to look at your PR. Great work so far @kasium!
I left a few comments, how I think we can make it even better.

One additional thing, should we deprecate the `doc` attribute?
We can't remove it as it would be a breaking change, but deprecation would certainly be possible if the doc string can be accessed through `doc_node`.

For that I would suggest to create a new mixin class with a custom `doc` property.
```py
import warnings
from typing import Optional

class Mixin:
    __doc: Optional[str]

    @property
    def doc(self) -> Optional[str]:
        warnings.warn("<some message>", DeprecationWarning)
        return self.__doc

    @doc.setter
    def doc(self, value) -> None:
        # Should this emit a DeprecationWarning, too?
        self.__doc = value
```
_Just an idea. I haven't tested if it would actually work._

Pierre-Sassoulas (approved):
`LGTM, we can merge once @DanielNoord comment have been addressed.

astroid/nodes/scoped_nodes.py
```diff
@@ -515,6 +518,7 @@ def __init__(
         self.parent = parent
         self.pure_python = pure_python
         self.locals = self.globals = {}
+        self.doc_node = doc_node
```
cdce8p, line 521: `doc_node` should be added to `_astroid_fields` above as it can contain a astroid node.

astroid/rebuilder.py
```diff
@@ -104,13 +104,17 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(self, node: T_Doc) -> Tuple[Optional[str], Optional[nodes.Const]]:
```
cdce8p, line 107: If we deprecate `doc`, we should consider if we want to switch the return values.
I.e. `doc_node, doc` instead of `doc, doc_node`.

astroid/rebuilder.py
cdce8p: I just checked the coverage report. So far these lines are not covered.

astroid/nodes/node_ng.py
```diff
@@ -84,6 +84,12 @@ class NodeNG:
     """Attributes that contain AST-dependent fields."""
     # instance specific inference function infer(node, context)
     _explicit_inference = None
+    doc_node: Optional["nodes.Const"] = None
+    """The doc node associated with this node.
+
+    On Python < 3.8 this attribute will always be None as the ast parser
+    does not allow us to extract the correct line numbering.
+    """
```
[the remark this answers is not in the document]
DanielNoord: This was added based on my suggestion. I suggested this as I thought `doc_node` could be added to a lot of other nodes (`Const` for example). However, looking back at this I think you're right and we should only add this to classes as needed.

Sorry about that @kasium

astroid/rebuilder.py
DanielNoord: I suggested against this because the line numbering of these nodes is (extremely) incorrect on `3.7`. See: https://github.com/PyCQA/astroid/pull/1276/commits/e61409f68ec297434a49ad83c7688facba301d99

`lineno` is actually `end_lineno`. Because of this I don't think it makes sense to make this node on `3.7` as the added benefit of `doc_node` over `doc` is incorrect and I suggested not adding it on `3.6` and `3.7` in https://github.com/PyCQA/astroid/pull/1276#issuecomment-986616113

astroid/rebuilder.py
DanielNoord: We might be good with removing this? Should probably be done in a separate PR though.

tests/unittest_scoped_nodes.py
DanielNoord: This depends on how we resolve https://github.com/PyCQA/astroid/pull/1276#discussion_r767055173

kasium:
Hey all, I'll try to work on your comments tomorrow. As a side note, I would like to keep the current change as small as possible. Therefore I see the deprecation of `doc` in another change, if that is fine for you

astroid/rebuilder.py
cdce8p: > I suggested against this because the line numbering of these nodes is (extremely) incorrect on `3.7`

Ok. Good point. Although it seems to work in some cases, eg.
```py
class X:
    """Some docstring."""
    var = 42
```

Multiline docstrings are an issue though.

--
Even with that, what I would like to prevent is having both `doc` and `doc_node` in parallel just because of it. It should just be one. If the line numbers are wrong, we should reset them manually. (Multiline docstrings are easy to identify as they set `col_offset = -1`.) In pylint you would then know to not rely on the line information. The docstring itself could still be accessed through `doc_node.value`.

I believe long term this is the better solution. Especially considering that Python 3.7 will also be deprecated at some point. We don't really need to talk about 3.6 anymore.

DanielNoord: Although I like that approach as it improves the information we provide to `pylint` and other users of `astroid` I wonder if this wouldn't create problems down the line.

What if we `add_message` on the docstring node and pass the node as `node` argument. That would trickle down to the final output and even the visual underlining in IDE's, as in: how should they handle a message with `col_offset = -1`. 
I wonder what the best place to address this issue is:
1) Make `pylint` check for "informational"/non-sensical `line_no` and `col_offset` (etc.) and handle them
2) Avoid `astroid` creating these "informational"/non-sensical attributes and handle them more gracefully here

I'm currently leaning towards the latter with a similar reasoning as to why I argued against setting `col_offset` to `0` in `pylint` when we don't have the correct value. Namely, I think we shouldn't add data that is not representative of the actual code and rather let users/IDE's handle missing data how they want to, ie. let them handle `col_offset = None`.

astroid/rebuilder.py
cdce8p: Opened #1289

astroid/rebuilder.py
cdce8p: IDEs already need to deal with both `col_offset = -1` and `col_offset = None`. There is actual a comment in the `vscode-python` code that mentioned this 😅 [vscode-python -> baseLinter.ts](https://github.com/microsoft/vscode-python/blob/44ef89560597ec319402b9361a37b8b344c04c41/src/client/linters/baseLinter.ts#L28-L30). So it wouldn't break anything (that isn't already broken).

> I wonder what the best place to address this issue is

My hope and what I know so far is that these issues are mostly fixed in newer versions. It would thus not be the biggest issue if it happen to be wrong, say in `3.6`.

In the end, I believe it's best if we can handle the remaining issues in astroid itself. In some cases a fix in pylint is just a lot quicker / easier.

For the particular issue here, my suggestion would be to set `col_offset = None` on those nodes. `end_lineno` and `end_col_offset` will already be `None` for 3.6 and 3.7.

cdce8p:
> Hey all, I'll try to work on your comments tomorrow. As a side note, I would like to keep the current change as small as possible. Therefore I see the deprecation of `doc` in another change, if that is fine for you

@kasium Works for me! I just merged #1289 to remove the used code patch I mentioned above. You'll probably need to rebase the PR to resolve the conflicts.

astroid/rebuilder.py
DanielNoord: > IDEs already need to deal with both `col_offset = -1` and `col_offset = None`. There is actual a comment in the `vscode-python` code that mentioned this 😅 [vscode-python -> baseLinter.ts](https://github.com/microsoft/vscode-python/blob/44ef89560597ec319402b9361a37b8b344c04c41/src/client/linters/baseLinter.ts#L28-L30). So it wouldn't break anything (that isn't already broken).

😅 

> For the particular issue here, my suggestion would be to set `col_offset = None` on those nodes. `end_lineno` and `end_col_offset` will already be `None` for 3.6 and 3.7.

Agreed!

What about `lineno`? For multi-line docstrings it seems to be the actual `end_lineno`...

cdce8p: > What about `lineno`? For multi-line docstrings it seems to be the actual `end_lineno`...

True, but we can't really determine a better one with complete certainty. At least that's still a valid `lineno` for the docstring. _The check probably be disabled for those versions anyway._

astroid/nodes/scoped_nodes.py
kasium: Updated was what?

astroid/rebuilder.py
kasium: As said, deprecation should go into another PR

astroid/rebuilder.py
kasium: So, the decision is what?

cdce8p: > So, the decision is what?

Create the `doc_node` regardless of Python version. After that, we should take another look at the test output.
For the multiline docstring, we should probably do a conditional overwrite on 3.7 to replace `col_offset = -1` with `None`.

kasium: done

astroid/nodes/scoped_nodes.py
DanielNoord: The updated docstring has become outdated 😄

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,14 @@ def func():
...
     )
     assert node.doc == "Docstring"
+    if PY38_PLUS:
+        assert isinstance(node.doc_node, nodes.Const)
+        assert node.doc_node.lineno == 3
+        assert node.doc_node.col_offset == 4
+        assert node.doc_node.end_lineno == 3
+        assert node.doc_node.end_col_offset == 15
+    else:
+        assert node.doc_node is None
```
DanielNoord, line 1593: The `if` statement should be moved to only cover some of the line numbering. We want to have the `doc_node` attribute regardless of python version.

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,46 @@ def test_stream_api(self) -> None:
             with open(path, "rb") as file_io:
                 self.assertEqual(stream.read(), file_io.read())
 
+    def test_with_singleline_docstring(self) -> None:
+        data = """
+            '''Hello World'''
+            foo = 1
+        """
+        module = builder.parse(data, __name__)
+        if PY38_PLUS:
```
DanielNoord, line 300: Same goes for this `if` statement, and the others in this file.

DanielNoord (changes requested):
We'll need to update the tests to cover the new expected results. Code in `rebuilder` itself looks good upon first inspection. I'll do a better check after the tests pass/have been updated.

astroid/rebuilder.py
kasium: @DanielNoord @cdce8p so, what is now the plan for python3.7 and below?

DanielNoord: Create the `Const` as well and fix line numbers as we see fit (based on the test output). 

@cdce8p Already hinted towards one solution for multi line strings and their column number, but there might be other problems.

kasium: done

astroid/rebuilder.py
```diff
@@ -113,12 +114,14 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
                     and isinstance(first_value, self._module.Constant)
                     and isinstance(first_value.value, str)
                 ):
-                    doc = first_value.value if PY38_PLUS else first_value.s
+                    doc_node = self.visit(node.body[0], node)
                     node.body = node.body[1:]
-                    return node, doc
+                    if not PY38_PLUS and doc_node.value.col_offset == -1:
```
DanielNoord, line 119:
```suggestion
                    # The ast parser of python < 3.8 sets col_offset of multi-line strings to -1
                    # as it is unable to determine the value correctly. We reset this to None.
                    if doc_node.value.col_offset == -1:
```

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,15 @@ def func():
     """
     )
     assert node.doc == "Docstring"
+    assert isinstance(node.doc_node, nodes.Const)
+    assert node.doc_node.lineno == 3
```
DanielNoord, line 1587:
```suggestion
    assert node.doc_node.value == "Docstring"
    assert node.doc_node.lineno == 3
```

tests/unittest_nodes.py
```diff
@@ -1583,6 +1583,15 @@ def func():
     """
     )
     assert node.doc == "Docstring"
+    assert isinstance(node.doc_node, nodes.Const)
```
DanielNoord, line 1586: Can you add a multi-line docstring to test their behaviour for `FunctionDef` nodes?

tests/unittest_scoped_nodes.py
```diff
@@ -291,6 +291,51 @@ def test_stream_api(self) -> None:
...
+        module = builder.parse(data, __name__)
+
+        assert isinstance(module.doc_node, nodes.Const)
+        if PY38_PLUS:
+            assert module.doc_node.lineno == 2
+            assert module.doc_node.col_offset == 0
+            assert module.doc_node.end_lineno == 5
+            assert module.doc_node.end_col_offset == 3
+        else:
+            assert module.doc_node.lineno == 5
```
DanielNoord, line 327: I'm still not sure whether this is the best approach... I foresee issues with pylint contributors not knowing that this is an incorrect value..

DanielNoord (changes requested):
Small changes, but I'd like @cdce8p to take a look at give his opinion on how to handle `line_no` of multi line strings.

I agree that deprecation of `doc` should be done in a separate PR.

tests/unittest_nodes.py
kasium: This is already tested in test_with_multiline_docstring for FunctionDef

tests/unittest_scoped_nodes.py
kasium: Oh great, forgot about pypy. In pypy the lines are correct, even with python 3.7. How should we handle that?

tristanlatr: This is how we've handled that in pydoctor: 

```python
_string_lineno_is_end = sys.version_info < (3,8) \
                    and platform.python_implementation() != 'PyPy'
"""True iff the 'lineno' attribute of an AST string node points to the last
line in the string, rather than the first line.
"""

def _set_docstring(self, ob: _model.ApiObject, node) -> None:
     docstring_lineno = node.lineno
        
     if _string_lineno_is_end:
            # In older CPython versions, the AST only tells us the end line
            # number and we must approximate the start line number.
            # This approximation is correct if the docstring does not contain
            # explicit newlines ('\n') or joined lines ('\' at end of line).
            docstring_lineno -= doc.count('\n')
```

Hope it helps

DanielNoord: That's really interesting actually, I didn't think about this approach. I guess we could do something as well?

I would think we could then reflect that `lineno` is an approximate in the docstring of the attribute and use `end_lineno` in `pylint` for the `bad-docstring-quotes` check. 

@kasium Would you be willing to explore if counting line breaks is a feasible solution for us?

kasium: @DanielNoord I'm sorry, but I'm currently overloaded with other tasks. I might can take a look in the 1/2 weeks, if this is fine.

DanielNoord: That's more than fine! We're all volunteers here, we're grateful for every minute you put into this project! 😄

kasium: Hey, sorry for taking so long. So, the solution works but has also some limitions:
```python

@mock.patch('mock.'
            'foo')
class Foo:
    """
    Blah
    """
```
--> will still return the wrong start line number. If we have no 100% solution I would rather leave it and wait for the deprecation for python 3.7. What do you think?

Pierre-Sassoulas:
Could you fix the conflict with the latest main branch @kasium, please ? It's not obvious to do :)

kasium:
> Could you fix the conflict with the latest main branch @kasium, please ? It's not obvious to do :)

Have not seen that 😄 Thanks! Will do that

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -449,6 +451,8 @@ def __init__(
 
         :param parent: The parent node in the syntax tree.
 
+        :param doc_node: The doc node associated with this node.
```
cdce8p, line 454: This should be after the docstring for `pure_python`.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1507,6 +1514,7 @@ def __init__(
         col_offset=None,
         parent=None,
         *,
+        doc_node: Optional[Const] = None,
```
cdce8p, line 1517: Can you move it after `end_col_offset`? Both `end_lineno` and `end_col_offset` are direct ast attributes so I think it makes sense to keep them first.

Please move the docstring as well.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1557,6 +1567,7 @@ def __init__(
         if parent:
             frame = parent.frame(future=True)
             frame.set_local(name, self)
+        self.doc_node = doc_node
```
cdce8p, line 1570: With the other nodes, the assignments (from arguments) are usually above the `super().__init__(...)` call.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2143,6 +2154,7 @@ def __init__(
         col_offset=None,
         parent=None,
         *,
+        doc_node: Optional[Const] = None,
```
cdce8p, line 2157: Same here, I think it would be good to move it after `end_col_offset`. Move the docstring too.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2219,6 +2233,7 @@ def __init__(
             end_col_offset=end_col_offset,
             parent=parent,
         )
+        self.doc_node = doc_node
```
cdce8p, line 2236: Same here with the assignment. In this particular case below `self.doc = doc` would make the most sense IMO.

astroid/rebuilder.py
```diff
@@ -113,12 +114,16 @@ def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
                     and isinstance(first_value, self._module.Constant)
                     and isinstance(first_value.value, str)
                 ):
-                    doc = first_value.value if PY38_PLUS else first_value.s
+                    doc_node = self.visit(node.body[0], node)
```
cdce8p, line 117: This won't work unfortunately. `node` here is still an `ast` node. However the `self.visit` method needs an astroid node which can be assigned as `parent` later.

I.e. the `doc_node` can't be build here. Instead just return the `node.body[0]` ast node for now. You can keep the `col_offset` handling, by doing it on the ast node directly.

astroid/rebuilder.py
```diff
@@ -140,14 +145,15 @@ def visit_module(
 
         Note: Method not called by 'visit'
         """
-        node, doc = self._get_doc(node)
+        doc, doc_node = self._get_doc(node)
```
cdce8p, line 148: Here `doc_node` will still be an ast node. Since we need to create the parent, i.e. `Module` node first, we can't pass it as argument to `Module.__init__`. Instead move `doc_node` to `postinit`.
The argument should be something like this
```py
newnode.postinit(
    ...,
    doc_node=self.visit(doc_node, newnode),
)

File: astroid/rebuilder.py [hunk 48]
```diff
@@ -1171,7 +1177,7 @@ def visit_classdef(
         self, node: "ast.ClassDef", parent: NodeNG, newstyle: bool = True
     ) -> nodes.ClassDef:
         """visit a ClassDef node to become astroid"""
-        node, doc = self._get_doc(node)
+        doc, doc_node = self._get_doc(node)
```
cdce8p, line 1180: Same issue here

astroid/rebuilder.py
```diff
@@ -1508,7 +1515,7 @@ def _visit_functiondef(
     ) -> T_Function:
         """visit an FunctionDef node to become astroid"""
         self._global_names.append({})
-        node, doc = self._get_doc(node)
+        doc, doc_node = self._get_doc(node)
```
cdce8p, line 1518: And here unfortunately.

cdce8p (changes requested):
@kasium Sorry for the long delays. I too didn't remember where exactly we left off last time.

Looked at the whole code change now. Unfortunately, we missed an issue so far. Due to the `parent` attribute, `doc_node` can't be assigned in `__init__`. It needs to be moved to `postinit` instead. I've left some comments below. Please take a look at those for `rebuilder.py` first, it might be that the others are outdated after you modified the code.

Just to be catch up, would you mind merging upstream in again?

astroid/rebuilder.py
kasium: I get what you mean but not sure how teh code should look like. Can you please assist

cdce8p: Here, inside the if block, it should be something like this
```py
                    doc = first_value.value if PY38_PLUS else first_value.s
                    doc_node = node.body[0]
                    node.body = node.body[1:]
                    # The ast parser of python < 3.8 sets col_offset of multi-line strings to -1
                    # as it is unable to determine the value correctly. We reset this to None.
                    if doc_node.col_offset == -1:
                        doc_node.col_offset = None
                    return doc, doc_node
```
Note that `doc_node` is just `node.body[0]` instead of the return from `self.visit`.
For the other changes, see the next comments.

tests/unittest_scoped_nodes.py
cdce8p: @DanielNoord With the last changes, we now have access to the source code inside `TreeRebuilder`. Haven't tested it yet, but it should be possible to use tokenize to extract the correct positions.

DanielNoord:
@kasium Due to some unforeseen issues with another PR this PR has become a little more important for `pylint` as the latest version of `astroid` has become unusable for it. Landing this PR would be one way to solve that issue and I think from all other options that might actually be the best way forward.

I'd gladly help with any questions you might have to help move this PR along. I think @cdce8p is right that with recent changes the reservations I had previously about this PR can now be solved 😄

astroid/rebuilder.py
```diff
@@ -113,7 +115,10 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional[str]]:
+    def _get_doc(
+        self, node: T_Doc
+    ) -> Tuple[T_Doc, Optional["ast.Constant | ast.Str"], Optional[str]]:
+        """Returns the docstring of a node both in str and nodes.Const."""
```
DanielNoord, line 121:
```suggestion
        """Returns the docstring of a node both in str and ast node form."""
```

astroid/rebuilder.py
```diff
@@ -204,14 +214,73 @@ def _get_position_info(
...
 
+    def _fix_doc_node_position(self, node: NodesWithDocsType) -> None:
+        """Fix start and end position of doc nodes for Python < 3.8."""
+        if not self._data or not node.doc_node or node.lineno is None:
+            return
+        if PY38_PLUS:
+            return
+
+        lineno = node.lineno or 1  # lineno of modules is 0
+        end_range: Optional[int] = node.doc_node.lineno
```
DanielNoord, line 225: Just to be sure, you added the typing for clarity right? This typing should be correct by default I think?

astroid/rebuilder.py
```diff
@@ -204,14 +214,73 @@ def _get_position_info(
...
+                continue
+            if t.type in skip_token:
+                continue
+            if t.type == token.STRING:
+                break
+            return
+        else:
+            return
+
+        node.doc_node.lineno = lineno - 1 + t.start[0]
```
DanielNoord, line 271: Why do we take `lineno - 1` here? Perhaps we should a comment explaining this? I can see us questioning this when we revisit this code at a later time.

astroid/rebuilder.py
```diff
@@ -220,7 +289,11 @@ def visit_module(
             package=package,
             parent=None,
         )
-        newnode.postinit([self.visit(child, newnode) for child in node.body])
+        newnode.postinit(
+            [self.visit(child, newnode) for child in node.body],
+            doc_node=self.visit(doc_ast_node, newnode),
+        )
+        self._fix_doc_node_position(newnode)
```
DanielNoord, line 296: Shouldn't the `_fix_doc_node_position` be called in the `postinit`? Do we ever want to call `postinit` and not call `_fix_doc_node_position`?

DanielNoord (changes requested):
@cdce8p Not sure if this was ready for review yet, but I thought it would be good to keep up with what you're doing so we can merge this as soon as it is done. Please dismiss if this is not ready yet.
I only have minor questions and comments though.

Could you also add a test case for return annotation in string format for a function definition? From my initial read of the code I wondered if this would be handled correctly. I tested, and it does (why would I think otherwise with one of your PR's 😅) but it would be good to have a test case as well.

astroid/rebuilder.py
cdce8p: Yeah, I wanted to be explicit here even if it can be inferred.

astroid/rebuilder.py
cdce8p: The token line number starts with 1 in the node line number. I.e. if we start tokenize in line 4, that's equivalent to `node.lineno`. However, for a match in line 6, the token position would be line 3, thus -1. `4 + 3 - 1 == 6`

I'll move the `-1` to the end, that should help.

astroid/rebuilder.py
cdce8p: `_fix_doc_node_position` depends on `self._data` from the `TreeRebuilder`.

astroid/rebuilder.py
DanielNoord: Ah understood! A comment might help still, but moving the `-1` will indeed help!

astroid/rebuilder.py
DanielNoord: Ah forgot about that. Hm yeah, passing `_data` to `postinit` also feels weird. Not really a big fan of having methods that are required to be run sequentially instead of at once, but perhaps here it is unavoidable without too convoluted solutions...

astroid/rebuilder.py
cdce8p: > A comment might help still, but moving the `-1` will indeed help!

I'm a bit hesitant about the comment. It's a basic pattern that repeats itself whenever we do use tokenize in the `TreeRebuilder`. Should we add it to every return value? That feels unnecessary. Furthermore, there are multiple tests in place. So if you do end up removing it, you would know very quickly why it's necessary.

DanielNoord: Fine with me!

astroid/rebuilder.py
DanielNoord: > When I added `self._data` initially, I also though if it would make sense to store it in the nodes itself. E.g. as attribute for `Module`. In the end I decided against it. These are are parsing steps which should just be done by the `TreeRebuilder`.

Yeah, this makes more sense.

> It might not be 100% perfect, but tbh I don't really mind it in this case. It's only for the tree visit functions anyway and similar to `_get_doc` which is also called sequentially. Any alternative would be worse.

It also helps that this is likely not going to change much in the future and will be unnecessary after 3.8+.

I'll do a final review but I think this is (almost) good to go!

ChangeLog
```diff
@@ -12,6 +12,8 @@ What's New in astroid 2.10.1?
 =============================
 Release date: TBA
 
+* Add new (optional) ``doc_node`` attribute to ``nodes.Module``, ``nodes.ClassDef``,
```
DanielNoord, line 15: Ah, that was one remaining question. Technically this should be in `2.11` I think as this can't really be considered a bug fix.
Do we want to release `2.11` with only this? Or do we keep it in `2.10.1`?

DanielNoord (approved):

cdce8p:
@kasium Sorry for taking over your PR. You already did an awesome job 🚀 Unfortunately, this PR blocks the next pylint release. Combined with the necessary changes, I though that it would be faster to work on it myself.

> @cdce8p Not sure if this was ready for review yet, but I thought it would be good to keep up with what you're doing so we can merge this as soon as it is done. Please dismiss if this is not ready yet. I only have minor questions and comments though.

It is ready now 😄

> Could you also add a test case for return annotation in string format for a function definition? From my initial read of the code I wondered if this would be handled correctly. I tested, and it does (why would I think otherwise with one of your PR's 😅) but it would be good to have a test case as well.

Done. Although string return annotations weren't an issue, I only accepted comments for Module docstrings. Noticed that this might be useful as I was writing the test case. It's fixed now.

ChangeLog
cdce8p: Though about this too. Let's move it to `2.11`. I've renamed to existing project milestones.
`2.10.1` -> `2.11.0` and `2.11.0` -> `2.12.0`.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -386,7 +386,7 @@ class Module(LocalsDictNodeNG):
     <Module l.0 at 0x7f23b2e4eda0>
     """
 
-    _astroid_fields = ("body",)
+    _astroid_fields = ("body", "doc_node")
```
DanielNoord, line 389: @cdce8p I just caught something.

https://github.com/PyCQA/astroid/blob/841401d48be94c1bec179864f36ac11beea61f97/astroid/nodes/node_ng.py#L254-L263

`last_child` iterates over this with `-1`. So I think the order of `astroid_fields` matters. I haven't looked at an actual edge case yet, but we might need to change this around?
```python
def func():
    """Test"""
    a = 1
```
I think the docstring will now be the last child in this example?

DanielNoord (changes requested):

astroid/nodes/scoped_nodes/scoped_nodes.py
cdce8p: Good catch!

astroid/rebuilder.py
```diff
@@ -225,11 +226,12 @@ def _fix_doc_node_position(self, node: NodesWithDocsType) -> None:
         end_range: Optional[int] = node.doc_node.lineno
         if IMPLEMENTATION_PYPY:
             end_range = None
+        # pylint: disable-next=unsubscriptable-object
         data = "\n".join(self._data[lineno - 1 : end_range])
 
         found_start, found_end = False, False
         open_brackets = 0
-        skip_token: set[int] = {token.NEWLINE, token.INDENT}
+        skip_token: Set[int] = {token.NEWLINE, token.INDENT}
```
DanielNoord, line 234: I actually saw this during review and thought I had learned something new from you (`set` being able to used on < 3.8/9) but alas... 😄

DanielNoord (approved):
I think we don't need a test for the order issue as it will be found in #1411

astroid/rebuilder.py
cdce8p: It can actually be used be used in 3.7 and 3.8 (with `from __future__ import annotations`), but not in 3.6.
Really time to remove support for 3.6 soon.

_Technically, you could use `"set[int]"` (note the quotation marks) in 3.6, but that isn't really worth it._

cdce8p:
> I think we don't need a test for the order issue as it will be found in #1411

Sounds good!

Have you been able to test the pylint issue? Will this PR resolve it?

DanielNoord:
> > I think we don't need a test for the order issue as it will be found in #1411
> 
> Sounds good!
> 
> Have you been able to test the pylint issue? Will this PR resolve it?

Not tested, but I'm fairly positive it should. See my comment in #1411. If those tests pass `pylint` will pass as well.
I'll do a final check just to be sure. Give me a minute.

cdce8p:
> Edit: @cdce8p Fixed 😄

I'll do one last check against Home Assistant, just to make sure and merge it afterwards. That way we can both still work on the two followups today and hopefully have the release ready tomorrow.

Around ~10min.

astroid/rebuilder.py
DanielNoord: Yeah, I have been looking forward to seeing you use your typing updater tool 😄

DanielNoord:
> > Edit: @cdce8p Fixed 😄
> 
> I'll do one last check against Home Assistant, just to make sure and merge it afterwards. That way we can both still work on the two followups today and hopefully have the release ready tomorrow.
> 
> Around ~10min.

@cdce8p Sounds good to me! Ping me when you need me 😄 

My PR is ready. You can review it if you want. It's just a bunch of tests now.

kasium:
Thanks for finishing it!

PR #1277, opened 2021-11-26
Upgrade to pylint 2.12 and enable ``for_any_all`` checker

nickdrozd:

### DIFF
--- a/astroid/mixins.py
+++ b/astroid/mixins.py
@@ -28,8 +28,6 @@ if TYPE_CHECKING:
 class BlockRangeMixIn:
     """override block range"""
 
-    # pylint: disable = no-member
-
     @decorators.cachedproperty
     def blockstart_tolineno(self):
         return self.lineno
--- a/astroid/mixins.py
+++ b/astroid/mixins.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
 
 class BlockRangeMixIn:
     """override block range"""
+
     # pylint: disable = no-member
 
     @decorators.cachedproperty
--- a/astroid/mixins.py
+++ b/astroid/mixins.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
 
 class BlockRangeMixIn:
     """override block range"""
+    # pylint: disable = no-member
 
     @decorators.cachedproperty
     def blockstart_tolineno(self):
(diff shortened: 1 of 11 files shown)


Pierre-Sassoulas (changes requested):
This is great, thank you ! Could you also fix the other issues newly detected by pylint 2.12 ? (Or disable them if you don't have the time right now)

nickdrozd:
Looks like the new version of Pylint triggers a few issues.

astroid/mixins.py
```diff
@@ -28,6 +28,8 @@
 class BlockRangeMixIn:
```
Pierre-Sassoulas, line 28: This mixin is not recognized as one, I'll fix this myself suggestions won't work here.

Pierre-Sassoulas (approved):
👍

astroid/nodes/node_ng.py
```diff
@@ -448,7 +445,7 @@ def _fixed_source_line(self) -> Optional[int]:
         We need this method since not all nodes have :attr:`lineno` set.
         """
         line = self.lineno
-        _node: Optional[NodeNG] = self
+        _node: Optional[NodeNG] = self  # pylint: disable = used-before-assignment
```
DanielNoord, line 448: Is this a new false positive?

cdce8p: https://github.com/PyCQA/pylint/issues/5342

pylintrc
```diff
@@ -301,6 +301,9 @@ ignored-modules=typed_ast.ast3
 # (useful for classes with attributes dynamically set).
 ignored-classes=SQLObject
 
+# Regex pattern to define which classes are considered mixins.
+mixin-class-rgx=.*Mix[i|I]n
```
cdce8p, line 305: @Pierre-Sassoulas This will also match `Mix|n`. It should just be `.*Mix[Ii]n`.

astroid/nodes/node_ng.py
DanielNoord: Ah! Thanks!

pylintrc
Pierre-Sassoulas: 😄 Obviously I'm just handing the case with really big using pipe, and did not make a regex mistake at all.

PR #1280, opened 2021-12-01
fix: fix as_string.py

WenJinfeng:
## Steps

## Type of Changes

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -182,8 +182,16 @@ class AsStringVisitor:
         """return an astroid.ClassDef node as string"""
         decorate = node.decorators.accept(self) if node.decorators else ""
         args = [n.accept(self) for n in node.bases]
+        # if node._metaclass and not node.has_metaclass_hack():
+        #     args.append("metaclass=" + node._metaclass.accept(self))
         if node._metaclass and not node.has_metaclass_hack():
-            args.append("metaclass=" + node._metaclass.accept(self))
+            # if class definition node has with_metaclass attribute, it does not need to add "args.append("metaclass=" + node._metaclass.accept(self))"
+            # for example, "class ClientCallDetails(six.with_metaclass(abc.ABCMeta))"
+            if (node.bases) and (node.bases[0]) and (type(node.bases[0]).__name__=="Call") and (node.bases[0].func) and (type(node.bases[0].func).__name__=="Attribute") and (node.bases[0].func.attrname) and (node.bases[0].func.attrname == "with_metaclass"):
+                #print("with_metaclass")
+                print("")
+            else:            
+                args.append("metaclass=" + node._metaclass.accept(self))
         args += [n.accept(self) for n in node.keywords]
         args = f"({', '.join(args)})" if args else ""
         docs = self._docs_dedent(node.doc) if node.doc else ""
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -58,6 +58,14 @@ class AsStringVisitor:
 
     def _docs_dedent(self, doc):
         """Stop newlines in docs being indented by self._stmt_list"""
+        # document string can not handle special character, such as \X, \x, \U, \u, \N and so on.
+        doc = doc.replace("\"\"\"", "\"")
+        doc = doc.replace('\\x', r'\-x')
+        doc = doc.replace('\\X', r'\-X')
+        doc = doc.replace('\\u', r'\-u')
+        doc = doc.replace('\\U', r'\-U')
+        doc = doc.replace('\\N', r'\-N')
+        doc.encode('utf-8').strip()
         return '\n{}"""{}"""'.format(self.indent, doc.replace("\n", DOC_NEWLINE))
 
     def _stmt_list(self, stmts, indent=True):
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -129,6 +129,10 @@ class AsStringVisitor:
     def visit_assign(self, node):
         """return an astroid.Assign node as string"""
         lhs = " = ".join(n.accept(self) for n in node.targets)
+        # handler the situation where "1e10000" is automatically converted to "inf"
+        # "inf" may influence the program execution
+        if f"{node.value.accept(self)}" == "inf":
+            return f"{lhs} = 1e10000"
         return f"{lhs} = {node.value.accept(self)}"
 
     def visit_augassign(self, node):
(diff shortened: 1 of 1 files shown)


WenJinfeng:
For my solutions, some may not work in certain situations. However, when I used the astroid package to transfer the corresponding source code, the conversion of some nodes contained in as_string.py may be problematic. Specific methods are about visit_assign, visit_classdef, _docs_dedent, visit_return, visit_boolop and so on. Thanks!

DanielNoord:
Hi @WenJinfeng 

While I have not looked at the specific code yet, the first step towards getting this merged is fixing the current failing tests and adding some tests for the code you have added. I don't completely understand the problem you're reporting, but some tests cases in which these special characters are included should be sufficient. If you need any help, please let me know!

CynthiaWendy:
Hello, @DanielNoord 
Thanks for your reply! For example, in as_string.py, 

original code:

def visit_assign(self, node):
&ensp;&ensp;&ensp;&ensp;"""return an astroid.Assign node as string"""
&ensp;&ensp;&ensp;&ensp;lhs = " = ".join(n.accept(self) for n in node.targets)
&ensp;&ensp;&ensp;&ensp;return f"{lhs} = {node.value.accept(self)}"

The problem is the situation where "a = 1e10000" is automatically converted to "a = inf". But, "inf" influences my program execution. Thus, I modify this conversation as the following example.

def visit_assign(self, node):
&ensp;&ensp;&ensp;&ensp; """return an astroid.Assign node as string"""
&ensp;&ensp;&ensp;&ensp;lhs = " = ".join(n.accept(self) for n in node.targets)
&ensp;&ensp;&ensp;&ensp;**if f"{node.value.accept(self)}" == "inf":
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;return f"{lhs} = 1e10000"**
&ensp;&ensp;&ensp;&ensp;return f"{lhs} = {node.value.accept(self)}"

I'm not sure if my code is reasonable, do you have any good suggestions? Looking forward to your reply. Thanks.

WenJinfeng:
Hello, @DanielNoord
Thanks for your reply! For example, in as_string.py,

original code:

def visit_assign(self, node):
    """return an astroid.Assign node as string"""
    lhs = " = ".join(n.accept(self) for n in node.targets)
    return f"{lhs} = {node.value.accept(self)}"

The problem is the situation where "a = 1e10000" is automatically converted to "a = inf". But, "inf" influences my program execution. Thus, I modify this conversation as the following example.

def visit_assign(self, node):
     """return an astroid.Assign node as string"""
    lhs = " = ".join(n.accept(self) for n in node.targets)
    **if f"{node.value.accept(self)}" == "inf":
      return f"{lhs} = 1e10000"**
    return f"{lhs} = {node.value.accept(self)}"

I'm not sure if my code is reasonable, do you have any good suggestions? Looking forward to your reply. Thanks.

DanielNoord:
I'm not quite sure where this automatic conversion takes place. Could you provide a specific code example that is creating this issue for you?

WenJinfeng:
My test library is about “google” under python3.6 as shown in the following figure.

image

google/protobuf/internal/encoder.py --> line 76 --> _POS_INF = 1e10000

Using astroid can transform automatically _POS_INF = 1e10000 of this file as _POS_INF = inf, it will cause the program execution.

DanielNoord:
Ah I think I'm starting to understand the problem here! Your commit does fix other issues as well though? Are those related to this issue, or are there other issues you're trying to solve?

In any case, I think some test cases for this would make sense. That would also help me and other contributors to fully understand the problem. You can look at `unittest_nodes` and `unittest_scoped_nodes` for how we currently these methods. One question I have though: what if the expected value is `"inf"`? Right now you're always converting "back" to `1e10000`, but that might not always be correct. How do we handle such situations?

astroid/nodes/as_string.py
```diff
@@ -156,7 +168,11 @@ def visit_binop(self, node):
     def visit_boolop(self, node):
         """return an astroid.BoolOp node as string"""
         values = [f"{self._precedence_parens(node, n)}" for n in node.values]
-        return (f" {node.op} ").join(values)
+        # return (f" {node.op} ").join(values)
+        tmp = (f" {node.op} ").join(values)
+        tmp = f"({tmp})"
+        # print("({})".format(tmp))
+        return tmp
```
cdce8p, line 175: What is the reason for this change?

astroid/nodes/as_string.py
```diff
@@ -449,7 +481,10 @@ def visit_return(self, node):
         """return an astroid.Return node as string"""
         if node.is_tuple_return() and len(node.value.elts) > 1:
             elts = [child.accept(self) for child in node.value.elts]
-            return f"return {', '.join(elts)}"
+            tmp = f"{', '.join(elts)}"
+            tmp = f"return ({tmp})"
+            return tmp
+            # return f"return {', '.join(elts)}"
```
cdce8p, line 487: Same here, why?

cdce8p (changes requested):
I'm not yet fully convince we should add specially case handling for corner cases.
Could you provide code examples that don't work currently?

astroid/nodes/as_string.py
WenJinfeng: I looked for the previous test, which was about the third-party library "Sklearn".
The code segment is about the modification of visit_return node.

For example, 
you can test the file --> Sklearn/discriminant_analysis.py/

def _decision_function(self, X):
        .....
&nbsp;&nbsp;&nbsp;&nbsp;return (-0.5 * (norm2 + u) + np.log(self.priors_))


as_string.py will transfer it as "return -0.5 * (norm2 + u) + np.log(self.priors_)" with missing the parentheses.

astroid/nodes/as_string.py
WenJinfeng: Thanks for your reply.
When I test the third-party library "sklearn"(scikit-learn, https://github.com/scikit-learn/scikit-learn/tree/main/sklearn), I encountered these problems and modifications. 
You can try to transfer the "sklearn" library, and then execute.

astroid/nodes/as_string.py
cdce8p: Even without the parentheses the result is valid. The goal of the `AsStringVisitor` isn't to create the exact input code, just a valid python code string. Some information just gets lost when the code is parsed to the AST. (There are libraries that handle this better though.)

Could you elaborate a bit more on the context you encountered the errors? I understand you mentioned `Sklearn`, but what were you trying to accomplish? Run pylint? A custom plugin? Something else entirely?

DanielNoord:
Hi @WenJinfeng,

Just checking in to see if you have given this PR any thought recently and if you managed to come up with some code examples that currently don't work. Those would definitely help move this PR along!

cdce8p:
Closing this one as there hasn't been any activity.

PR #1283, opened 2021-12-05
Fix typing of ``Arguments.args``

DanielNoord:
## Steps

## Description

Based on discussion in `pylint` here: https://github.com/PyCQA/pylint/pull/5409#discussion_r762497136

I have added changelog entry to increase visibility for this change.

## Type of Changes

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -860,9 +860,10 @@ class Arguments(mixins.AssignTypeMixin, NodeNG):
 
         self.args: typing.Optional[typing.List[AssignName]]
         """The names of the required arguments.
+
         Can be None if the assosciated function does not have a retrievable
-        signature and the arguments are therefore unknown. This happens with builtin
-        functions implemented in C.
+        signature and the arguments are therefore unknown.
+        This happens with builtin functions implemented in C.
         """
 
         self.defaults: typing.List[NodeNG]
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
👍 Justifying the necessary ``Optional`` is pretty important, because I'm sure there are a lot of "false optional" that we'll want to remove to simplify our live and have simpler typing/code.

DanielNoord:
> 👍 Justifying the necessary `Optional` is pretty important, because I'm sure there are a lot of "false optional" that we'll want to remove to simplify our live and have simpler typing/code.

If we could redesign `astroid` I think we would have solved this differently. However, I agree with @cdce8p that in the current situation this is the best way forward. We would need to deprecate etc. otherwise..

astroid/nodes/node_classes.py
```diff
@@ -858,8 +858,12 @@ def __init__(
         self.kwarg: Optional[str] = kwarg  # can be None
         """The name of the variable length keyword arguments."""
 
-        self.args: typing.List[AssignName]
-        """The names of the required arguments."""
+        self.args: typing.Optional[typing.List[AssignName]]
+        """The names of the required arguments.
+        Can be None if the assosciated function does not have a retrievable
+        signature and the arguments are therefore unknown. This happens with builtin
+        functions implemented in C.
```
cdce8p, line 865:
```suggestion
        """The names of the required arguments.

        Can be None if the assosciated function does not have a retrievable
        signature and the arguments are therefore unknown.
        This happens with builtin functions implemented in C.
```

cdce8p (approved):
Looks good! Thanks @DanielNoord 👍🏻

PR #1284, opened 2021-12-05
Fix crash in `LocalsDictNodeNG.qname` if name does not exist

jacobtylerwalls:
## Steps

## Description

https://github.com/PyCQA/pylint/issues/5461 reports an issue similar if not identical to #278.

Not all scoped nodes have a `name` attribute. The body of `LocalsDictNodeNG.qname()` is aware of this, disables `no-member` checks, and links to #278, but still depended on there being a `name`.

There was a commit (189b98bcdf8d2e5ec46d0dd3455adbaa516127ff) to remove `qname()` to a mixin, but I can't find that it ever made it into a release, so I'm not certain why it was closed.

Now, we have code depending on `qname()`, so I'm thinking it's safer to just work around the crash.

## Type of Changes

## Related Issue

Closes https://github.com/PyCQA/pylint/issues/5461

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,6 +20,11 @@ Release date: TBA
 
 * Fix typing and update explanation for ``Arguments.args`` being ``None``.
 
+* Fix crash in LocalsDictNodeNG.qname() if a variable named "type" is used in
+  a generator expression.
+
+  Closes PyCQA/pylint#5461
+
 
 What's New in astroid 2.9.0?
 ============================
--- a/astroid/nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes.py
@@ -243,7 +243,11 @@ class LocalsDictNodeNG(node_classes.LookupMixIn, node_classes.NodeNG):
         # pylint: disable=no-member; github.com/pycqa/astroid/issues/278
         if self.parent is None:
             return self.name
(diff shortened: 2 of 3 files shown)


cdce8p (changes requested):
Thank you for the PR, but I don't think it makes sense to allow `qname` `ListComp` and others nodes that don't have a qualified name. I believe that's also the reason it was removed from the mixin. https://github.com/PyCQA/astroid/commit/189b98bcdf8d2e5ec46d0dd3455adbaa516127ff
```txt
The new mixin is used by ClassDef, Lambda, FunctionDef and Module,
for defining a qname() method on them. This should fix a bug where
the qname was defined on any subclass of LocalsDictNode, even though
it wasn't possible to get a qualified name for the node in question.
```

The issue you linked can be fixed much more easily. https://github.com/PyCQA/pylint/issues/5461
https://github.com/PyCQA/astroid/blob/8ea31c8f197c3f4c1fefdbf861530c831d250a59/astroid/brain/brain_type.py#L50-L52
The condition should instead be
```py
    if (
        not isinstance(node_scope, nodes.Module)
        or node_scope.qname() != "builtins"
    ):
        raise UseInferenceDefault()
```

jacobtylerwalls:
Thanks for the feedback! Happy to get a new PR going.

PR #1294, opened 2021-12-15
Add stubs for ``setuptools`` and ``typed-ast`` to dependencies

DanielNoord:
## Steps

## Type of Changes

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -5,5 +5,8 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
+# Type packages for mypy
 types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.2
+types-setuptools==57.4.4
+types-typed-ast==1.5.0


Pierre-Sassoulas (changes requested):
Isn't it handled here : https://github.com/PyCQA/astroid/blob/main/.pre-commit-config.yaml#L77 ? We could also use ``--install-types`` with ``--non-interactive`` See https://mypy.readthedocs.io/en/stable/command_line.html?highlight=install-package#miscellaneous

Pierre-Sassoulas (approved):
I'm using pre-commit myself because it's just easier you have 2 commands, ``pre-commit install`` and ``pre-commit run [mypy] --all-files`` and everything is installed automatically and integrated in git. Keeping those requirements files produce a lot of requirements upgrade with dependabot and maintenance in general. Let's keep it as I don't want to force anyone to use pre-commit and re-open this topic with Marc 😄

DanielNoord:
The problem I have with `pre-commit` is that when I want to get a quick to-do list of `mypy` issues it runs all other hooks as well (or I just don't know the command to run a single hook). I don't want to wait on the `pylint` hook to finish before seeing the remaining `mypy` issues.

Pierre-Sassoulas:
It's ``pre-commit run mypy --all-files`` (Or  ``pre-commit run black --all-files``)

requirements_test.txt
```diff
@@ -5,5 +5,8 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
+# Type packages for mypy
 types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.2
+types-setuptools==57.4.4
+types-typed-ast==1.5.0
```
cdce8p, line 12: This is a duplicate. See line **9**:
```
types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
```

If you get type errors in Python >= 3.8, there is probably a `sys.version_info` check missing somewhere.

requirements_test.txt
```diff
@@ -5,5 +5,8 @@ coverage~=5.5
 pre-commit~=2.13
 pytest-cov~=2.11
 tbump~=6.3.2
+# Type packages for mypy
 types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.2
+types-setuptools==57.4.4
```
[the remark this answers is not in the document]
DanielNoord: `Mypy` complains about missing stubs on this line:
https://github.com/PyCQA/astroid/blob/37ce8bbf427b07b319a7a9d0289136579bf9f9c3/astroid/interpreter/_import/util.py#L6

I'll create a follow-up to add it to the CI and remove the duplicate entry.

requirements_test.txt
DanielNoord: Sorry this is a stupid mistake.. For some reason this wasn't installed locally..

DanielNoord: I just saw there is in, and I'm working on a fix in https://github.com/PyCQA/astroid/pull/1296.

I'll remove the line and see if this still throws errors after fixing #1296.

PR #1296, opened 2021-12-15
Add typing to ``_ast``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -1,21 +1,20 @@
 import ast
 import sys
+import types
 from collections import namedtuple
 from functools import partial
 from typing import Dict, Optional
 
 from astroid.const import PY38_PLUS, Context
 
-try:
-    import typed_ast.ast3 as _ast_py3
-except ImportError:
-    _ast_py3 = None
-
-
 if PY38_PLUS:
     # On Python 3.8, typed_ast was merged back into `ast`
-    _ast_py3 = ast
-
+    _ast_py3: Optional[types.ModuleType] = ast
+else:
+    try:
+        import typed_ast.ast3 as _ast_py3  # type: ignore[no-redef]
+    except ImportError:
+        _ast_py3 = None
 
 FunctionType = namedtuple("FunctionType", ["argtypes", "returns"])
 
@@ -51,16 +50,14 @@ def parse_function_type_comment(type_comment: str) -> Optional[FunctionType]:
     if _ast_py3 is None:
         return None
 
-    func_type = _ast_py3.parse(type_comment, "<type_comment>", "func_type")
+    func_type = _ast_py3.parse(type_comment, "<type_comment>", "func_type")  # type: ignore[attr-defined]
     return FunctionType(argtypes=func_type.argtypes, returns=func_type.returns)
 
 
 def get_parser_module(type_comments=True) -> ParserModule:
-    if not type_comments:
-        parser_module = ast
-    else:
+    parser_module = ast
+    if type_comments and _ast_py3:
         parser_module = _ast_py3
-    parser_module = parser_module or ast
 
     unary_op_classes = _unary_operators_from_module(parser_module)
     cmp_op_classes = _compare_operators_from_module(parser_module)
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -7,12 +7,12 @@ from typing import Dict, Optional
 
 from astroid.const import PY38_PLUS, Context
 
-if PY38_PLUS:
+if sys.version_info >= (3, 8):
     # On Python 3.8, typed_ast was merged back into `ast`
     _ast_py3: Optional[types.ModuleType] = ast
 else:
     try:
-        import typed_ast.ast3 as _ast_py3  # type: ignore[no-redef]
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👍

astroid/_ast.py
```diff
@@ -1,21 +1,20 @@
...
 
 from astroid.const import PY38_PLUS, Context
 
-try:
-    import typed_ast.ast3 as _ast_py3
-except ImportError:
-    _ast_py3 = None
-
-
 if PY38_PLUS:
```
cdce8p, line 10: Use `sys.version_info` here. `PY38_PLUS` can't be inferred by mypy and pyright.

astroid/_ast.py
```diff
@@ -1,21 +1,20 @@
...
     # On Python 3.8, typed_ast was merged back into `ast`
-    _ast_py3 = ast
-
+    _ast_py3: Optional[types.ModuleType] = ast
+else:
+    try:
+        import typed_ast.ast3 as _ast_py3  # type: ignore[no-redef]
```
cdce8p, line 15:
```suggestion
        import typed_ast.ast3 as _ast_py3
```
With `sys.version_info`, you don't need the `type: ignore` anymore.

astroid/_ast.py
```diff
@@ -51,16 +50,14 @@ def parse_function_type_comment(type_comment: str) -> Optional[FunctionType]:
     if _ast_py3 is None:
         return None
 
-    func_type = _ast_py3.parse(type_comment, "<type_comment>", "func_type")
+    func_type = _ast_py3.parse(type_comment, "<type_comment>", "func_type")  # type: ignore[attr-defined]
```
cdce8p, line 53:
```suggestion
    func_type = _ast_py3.parse(type_comment, "<type_comment>", "func_type")
```
Don't think that is necessary here, not even in strict mode.

cdce8p (changes requested):
I think this should probably be one of the last fully typed astroid modules.
A bit sketchy and difficult to get completely right.

astroid/_ast.py
DanielNoord: `Mypy` keeps complaining about it.. I think it might be because of the way we import and redefine `_ast_py3` and it can't follow that?

DanielNoord:
> I think this should probably be one of the last fully typed astroid modules.
> A bit sketchy and difficult to get completely right.

These are the only issues `mypy` reports, with the current PR this file passes (non-strict mode that is). Do you want to convert to draft and wait? It seemed to me to be easier than some of the other open issues.

astroid/_ast.py
cdce8p: I've been using a dev version of mypy. Seem like this false-positive will be fixed with the next release `0.930`. Let's leave the `type: ignore` in there then.

cdce8p (approved):

astroid/_ast.py
DanielNoord: Seems like a number of false positives will be fixed. Last release has been some time ago, I remember we encountered other issues as well in other PRs.

We'll update when they release I guess.

cdce8p:
> > I think this should probably be one of the last fully typed astroid modules.
> > A bit sketchy and difficult to get completely right.
> 
> These are the only issues `mypy` reports, with the current PR this file passes (non-strict mode that is). Do you want to convert to draft and wait? It seemed to me to be easier than some of the other open issues.

I was referring to the full typing not just fixing the errors. A few months ago, I looked at it while I was working on the TreeRebuilder. Especially the `ParserModule` might be difficult to get right, but that is a task for another day.

_Maybe it isn't as bad as I though it would be._

DanielNoord:
Ah, understood!

For me the current workflow/focus points work quite well:
1) Get `astroid` to a place where we can enable `mypy`
2) Add typing to `pylint` and fix issues in `astroid` as they arise (for example with `statement()`)

This is mainly because `astroid` has a much more difficult codebase (imo) so doing it based on issues encountered in `pylint` helps me understand what some of the code is trying to achieve.

Also, I think exposing all types in `pylint` should probably be a higher priority than those of `astroid` as the former will serve more people. That's why I'm focusing on `pylint` for now.

The biggest problem with 1) is #1250 I think. I encountered some issues in the last files I looked at which I could fix by simple `if not node.XXX: return` statements, but which I knew should never be `None` anyway. In terms of maintenance and typing for `astroid` I think that should be top priority.

Pierre-Sassoulas:
> Also, I think exposing all types in pylint should probably be a higher priority than those of astroid as the former will serve more people. That's why I'm focusing on pylint for now.

I supposed astroid was higher priority because pylint is based on it. So it can also helps ourselves for pylint development in particular. I also thought it's more likely that astroid is used as a library than pylint (pylint being extended with easy to understand new Checker more often imo ?) but maybe I'm wrong. 

Anyway, every typing is sacred, every typing is great, every  typing is wanted, every typing is good, every typing is needed, in our neighborhood.

> if not node.XXX: return statements, but which I knew should never be None anyway. In terms of maintenance and typing for astroid I think that should be top priority.

👍

PR #1298, opened 2021-12-15
Fix dataclass inference for marshmallow_dataclass

noirbee:
## Steps

## Description

This allows astroid/pylint to understand dataclasses as provided by [marshmallow-dataclass](https://github.com/lovasoa/marshmallow_dataclass), which are basically regular stdlib dataclasses with a few additional private attributes to generate marshmallow schemas from said dataclasses declaration.

Note that despite the fact that this affects inference for both `dataclass` and `field`, pylint still correctly detects that `marshmallow-dataclass` does not provide a `field` function.

## Type of Changes

(I'm putting this down as a bugfix because this was "broken" by the fact that recent astroid/pylint versions now correctly infers types from dataclasses, but I guess this could be considered a new feature)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,9 @@ Release date: TBA
 
   Closes PyCQA/pylint#5461
 
+* Enable inference of dataclass import from marshmallow_dataclass.
+  This allows the dataclasses brain to recognize dataclasses annotated by marshmallow_dataclass.
+
 
 What's New in astroid 2.9.0?
 ============================
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -3,9 +3,12 @@
 """
 Astroid hook for the dataclasses library
 
-Support both built-in dataclasses and pydantic.dataclasses. References:
+Support built-in dataclasses, pydantic.dataclasses, and marshmallow_dataclass-annotated
+dataclasses. References:
 - https://docs.python.org/3/library/dataclasses.html
 - https://pydantic-docs.helpmanual.io/usage/dataclasses/
+- https://lovasoa.github.io/marshmallow_dataclass/
+
 """
 from typing import FrozenSet, Generator, List, Optional, Tuple
 
@@ -35,7 +38,9 @@ from astroid.util import Uninferable
 
 DATACLASSES_DECORATORS = frozenset(("dataclass",))
 FIELD_NAME = "field"
-DATACLASS_MODULES = frozenset(("dataclasses", "pydantic.dataclasses"))
+DATACLASS_MODULES = frozenset(
+    ("dataclasses", "marshmallow_dataclass", "pydantic.dataclasses")
+)
 DEFAULT_FACTORY = "_HAS_DEFAULT_FACTORY"  # based on typing.py
 
 
--- a/tests/unittest_brain_dataclasses.py
+++ b/tests/unittest_brain_dataclasses.py
@@ -11,7 +11,7 @@ if not PY37_PLUS:
 
 
 parametrize_module = pytest.mark.parametrize(
-    ("module",), (["dataclasses"], ["pydantic.dataclasses"])
+    ("module",), (["dataclasses"], ["pydantic.dataclasses"], ["marshmallow_dataclass"])
 )
 
 
@@ -304,6 +304,8 @@ def test_inference_generic_collection_attribute(module: str):
         ("dataclasses", "typing"),
         ("pydantic.dataclasses", "typing"),
         ("pydantic.dataclasses", "collections.abc"),
+        ("marshmallow_dataclass", "typing"),
+        ("marshmallow_dataclass", "collections.abc"),
     ],
 )
 def test_inference_callable_attribute(module: str, typing_module: str):


Pierre-Sassoulas (approved):
Thank you for the MR. I discovered this package and it fix a problem I had with marshmallow's design 😄 

Regarding the change itself, I'm thinking there might be a problem with our design if we need to add the name of all dataclasses modules to this constant. I don't have a better solution as it's astroid and not pylint we can't add an option to let user add their own data-classes.

DanielNoord:
I took a quick look at how `marshmallow` defines their dataclasses but it will be difficult to do this with inference. Since functions don't inherit other functions like classes we'll need to look at their return values.
In the case of `marshmallow` this is difficult since we select their overloads instead of the actual function definitions and therefore don't actually see the return values of the functions that define the new dataclasses. I discuss this issue here: https://github.com/PyCQA/pylint/issues/5264#issuecomment-995591983

Even if we did find the correct function definition it would be difficult to fix this. We currently infer this as the return value from `marshmallow_dataclass.dataclass`:
`[<Lambda.<lambda> l.144 at 0x105951b50>, <Const.NoneType l.167 at 0x1059538e0>, Uninferable]`
[Code here](https://github.com/lovasoa/marshmallow_dataclass/blob/fa6c28980ccfe45742cdc9430bfb3b737690935f/marshmallow_dataclass/__init__.py#L106)

Making sure that all of these return values are dataclasses would require us to go several levels deeper and might be unfeasible. I think for now this might be the best thing we can do..

cdce8p:
> Thank you for the MR. I discovered this package and it fix a problem I had with marshmallow's design 😄
> 
> Regarding the change itself, I'm thinking there might be a problem with our design if we need to add the name of all dataclasses modules to this constant. I don't have a better solution as it's astroid and not pylint we can't add an option to let user add their own data-classes.

I think the way you're probably meant to deal with it would be trough a custom pylint / astroid plugin.
It's easy enough for us to add it now if it works good enough but in general such things are probably better done separately. Especially as we aren't able to support **all** packages. Someone has to maintain it in the end.

Pierre-Sassoulas:
> I think the way you're probably meant to deal with it would be trough a custom pylint / astroid plugin

Sounds good, but do we have astroid plugins ? If we go this route, each astroid brains could be an astroid plugins ? Seems to me a good idea as brains need to handle multiple versions of a package (?) and are rather naive in their current implementation. Having external astroid brains would permit to make them a lot more complicated than what they currently are. Then a pylint option could permits to add brain packages to your project. Thoughts @hippo91 ?

cdce8p:
> Sounds good, but do we have astroid plugins ?

Aren't those the [Transform plugins](https://pylint.pycqa.org/en/latest/how_tos/transform_plugins.html)? It looks like those should work with `inference_tip`. Although the documentations could definitely be updated.

> If we go this route, each astroid brains could be an astroid plugins ?

Most brain modules are actually for builtin functions and classes. Those should definitely stay within astroid. For the external once (like numpy) there could be an argument to move them to new projects. The issue then just becomes who will maintain them? For simple brains it's certainly much easier to add them to astroid directly.

In the end, I would think of it as _batteries included_. If it isn't too difficult or a very popular package, add it to astroid. For more complex stuff the might also require more maintenance than we can provide, a separate package is probably better.

hippo91:
@Pierre-Sassoulas i definitely agree with @cdce8p remarks!
By the way, i'm sorry to not have given news for a long period, but i have been very busy with work.
I hope i can be more present at the beginning of 2022.

Pierre-Sassoulas:
> i definitely agree with @cdce8p remarks!

Glad to see you're back and thanks for taking the time to check the problem ;)

> For more complex stuff that might also require more maintenance than we can provide, a separate package is probably better.

I wonder the criteria we could use. Maybe popularity of the package and the complexity of maintaining multiple versions together. Libraries like numpy and pandas are complex but also used a lot so keeping them in astroid makes some sense. The same question arise when we're asking ourselves about what dependency should be installed during tests (should we install numpy ? If we had an ``astroid-numpy-brain`` package the tests would not need to install numpy).

cdce8p:
> > For more complex stuff that might also require more maintenance than we can provide, a separate package is probably better.
> 
> I wonder the criteria we could use. Maybe popularity of the package and the complexity of maintaining multiple versions together. Libraries like numpy and pandas are complex but also used a lot so keeping them in astroid makes some sense. The same question arise when we're asking ourselves about what dependency should be installed during tests (should we install numpy ? If we had an `astroid-numpy-brain` package the tests would not need to install numpy).

The whole discussion should probably be moved to two separate issues:
* Improve pylint documentation around brain packages / plugins
* Which library brains should be included with astroid?

Some last comments, the package name should probably be something like `pylint-numpy` as it's a **pylint** plugin. Numpy is one of the things I also considered. As long as someone is willing to maintain it here, I think it's ok. Long term though, this would be a size for which a separate package could make sense.

--
As for this PR if you agree, I think it would be fine.

Pierre-Sassoulas (approved):
This is an external library but it's rather popular (350 star right now, but it's linked to marshmallow which is a lot more popular), and it's simple to maintain so following the rules we talked about, it can be included so astroid has "battery included".

Follow-up discussion in https://github.com/PyCQA/astroid/issues/1312

PR #1300, opened 2021-12-16
Upgrade mypy to 0.920

cdce8p:
## Description
http://mypy-lang.blogspot.com/2021/12/mypy-0920-released.html


DanielNoord:
Even quicker than Dependabot 😄 🎉

cdce8p:
> Even quicker than Dependabot 😄 🎉

Sometimes 😄

PR #1301, opened 2021-12-17
Treat NewTypes like normal subclasses

colatkinson:
## Steps

## Description

NewTypes are assumed not to inherit any members from their base classes.
This results in incorrect inference results. Avoid this by changing the
transformation for NewTypes to treat them like any other subclass.

## Type of Changes

|       | :sparkles: New feature |
|       | :hammer: Refactoring   |
|       | :scroll: Docs          |

## Related Issue

https://github.com/PyCQA/pylint/issues/3162
https://github.com/PyCQA/pylint/issues/2296


colatkinson:
Switched from checking for `bit_count()` to `bit_length()`, as I didn't realize the former was introduced in 3.10.

astroid/brain/brain_typing.py
```diff
@@ -52,6 +50,11 @@ def __args__(self):
 class {0}(metaclass=Meta):
     pass
 """
+# PEP484 suggests NewType is equivalent to this for typing purposes
```
DanielNoord, line 53: Could you add a link to section in the PEP? Saves a potential contributor from having to search for it.

tests/unittest_brain.py
```diff
@@ -1659,6 +1659,34 @@ def test_typing_types(self) -> None:
...
+        """
+        )
+        assert isinstance(ast_nodes, list)
+
+        # Base type given by reference
+        foo_node = ast_nodes[0]
+        foo_inferred = next(foo_node.infer())
+        self.assertIsInstance(foo_inferred, astroid.ClassDef)
+
+        # Check base type method is inferred
```
DanielNoord, line 1678:
```suggestion
        # Check base type method is inferred by accessing one of its methods
```

tests/unittest_brain.py
```diff
@@ -1659,6 +1659,34 @@ def test_typing_types(self) -> None:
...
+        foo_base_class_method = foo_inferred.getattr("endswith")[0]
+        self.assertIsInstance(foo_base_class_method, astroid.FunctionDef)
+
+        # Base type given by string
```
DanielNoord, line 1682:
```suggestion
        # Base type given by string (i.e., "int")
```

tests/unittest_brain.py
```diff
@@ -1659,6 +1659,34 @@ def test_typing_types(self) -> None:
...
+        assert isinstance(ast_nodes, list)
+
+        # Base type given by reference
+        foo_node = ast_nodes[0]
+        foo_inferred = next(foo_node.infer())
+        self.assertIsInstance(foo_inferred, astroid.ClassDef)
+
+        # Check base type method is inferred
+        foo_base_class_method = foo_inferred.getattr("endswith")[0]
+        self.assertIsInstance(foo_base_class_method, astroid.FunctionDef)
```
DanielNoord, line 1680: I'm not sure about this, but is there a way to infer that `endswith` comes from `str`? Some attribute within the `FunctionDef` node or the `qname()` of the node?
If there is I think it would be good to check this as well.

tests/unittest_brain.py
```diff
@@ -1659,6 +1659,34 @@ def test_typing_types(self) -> None:
...
+
+        # Base type given by string
+        bar_node = ast_nodes[1]
+        bar_inferred = next(bar_node.infer())
+        self.assertIsInstance(bar_inferred, astroid.ClassDef)
+
+        bar_base_class_method = bar_inferred.getattr("bit_length")[0]
+        self.assertIsInstance(bar_base_class_method, astroid.FunctionDef)
+
```
DanielNoord, line 1689: Could you also add a test for a user defined class/type?
The code in the original issue seems like it should be easily adaptable:
```python
class A:
    def __init__(self, value: int):
        self.value = value

a = A(5)
a.value  # no no-member error

B = NewType("B", A)

b = B(a)
b.value  # Instance of 'B' has no 'value' memberpylint(no-member)
```

astroid/brain/brain_typing.py
```diff
@@ -106,23 +109,32 @@ def __class_getitem__(cls, item):
 """
 
 
-def looks_like_typing_typevar_or_newtype(node):
+def looks_like_typing_typevar(node):
```
DanielNoord, line 112:
```suggestion
def looks_like_typing_typevar(node: nodes.Call) -> bool:
    """Checks whether a Call node is a TypeVar call"""
```

Could you add an `from astroid import nodes` line? I can't suggest it because it is outside of the original scope of the commits.

astroid/brain/brain_typing.py
```diff
@@ -106,23 +109,32 @@ def __class_getitem__(cls, item):
...
+    func = node.func
+    if isinstance(func, Attribute):
+        return func.attrname == "TypeVar"
+    if isinstance(func, Name):
+        return func.name == "TypeVar"
+    return False
+
+
+def looks_like_typing_newtype(node):
```
DanielNoord, line 121:
```suggestion
def looks_like_typing_newtype(node: nodes.Call) -> bool:
    """Checks whether a Call node is a NewType call"""
```

astroid/brain/brain_typing.py
```diff
@@ -132,6 +144,24 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
     return node.infer(context=context_itton)
 
 
+def infer_typing_newtype(node, context_itton=None):
```
DanielNoord, line 147:
```suggestion
def infer_typing_newtype(node: nodes.Call, context_itton: Optional[InferenceContext] =None):
```

I'm not sure about the return type here. Could you add it?
You'll likely also need to add `Optional` and `InferenceContext` imports.

astroid/brain/brain_typing.py
```diff
@@ -106,23 +109,32 @@ def __class_getitem__(cls, item):
...
+        return func.attrname == "NewType"
     if isinstance(func, Name):
-        return func.name in TYPING_TYPEVARS
+        return func.name == "NewType"
     return False
 
 
-def infer_typing_typevar_or_newtype(node, context_itton=None):
+def infer_typing_typevar(node, context_itton=None):
     """Infer a typing.TypeVar(...) or typing.NewType(...) call"""
```
DanielNoord, line 131:
```suggestion
    """Infer a typing.TypeVar(...) call"""
```

astroid/brain/brain_typing.py
```diff
@@ -106,23 +109,32 @@ def __class_getitem__(cls, item):
...
-        return func.attrname in TYPING_TYPEVARS
+        return func.attrname == "NewType"
     if isinstance(func, Name):
-        return func.name in TYPING_TYPEVARS
+        return func.name == "NewType"
     return False
 
 
-def infer_typing_typevar_or_newtype(node, context_itton=None):
+def infer_typing_typevar(node, context_itton=None):
```
DanielNoord, line 130: If you want we can also use this opportunity to update the typing here. However, (again) I'm not sure about the return type.

astroid/brain/brain_typing.py
```diff
@@ -132,6 +144,24 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
     return node.infer(context=context_itton)
 
 
+def infer_typing_newtype(node, context_itton=None):
+    """Infer a typing.TypeVar(...) or typing.NewType(...) call"""
```
DanielNoord, line 148:
```suggestion
    """Infer a typing.NewType(...) call"""
```

DanielNoord (changes requested):
Thank you @colatkinson!

This seems like a good change and closes two old `pylint` issues. Let me know if you have any questions about my comments.

colatkinson:
Made the requested changes -- let me know if I missed any! I also changed the transformation a bit, so that it takes in the correct number of arguments, and beefed up the tests a bit, both here and in the pylint PR (PyCQA/pylint#5542).

I also had to change the transformation to accurately resolve references to user-defined and imported types, so adding those tests was definitely important.

astroid/brain/brain_typing.py
```diff
@@ -132,6 +148,44 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+    )
+    new_node.parent = node.parent
+
+    # Base type arg is a normal reference, so no need to do special lookups
+    if not isinstance(base, nodes.Const):
+        new_node.bases = [base]
+
+    # If the base type is given as a string (e.g. for a forward reference),
+    # make a naive attempt to find the corresponding node.
+    # Note that this will not work with imported types.
```
colatkinson, line 180: This is the one remaining edge case I ran into. It comes up with code such as:

```python
from decimal import Decimal
A = NewType("A", "Decimal")
```

or

```python
import decimal
A = NewType("A", "decimal.Decimal")
```

The `node.frame().lookup()` won't find anything that can be used directly as a base. In the first case, I was able to hack something up by recursively resolving the `ImportFrom`'s module and doing a `lookup()` on that. In the latter, I wasn't able to figure out a clean solution.

Is there any comprehensive method to resolve these types of string references? If not, what would be the recommended course of action here? Worst case, I can also add a test and mark it as xfail to improve documentation of the limitation.

astroid/brain/brain_typing.py
```diff
@@ -132,6 +148,44 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+    base_name = base.as_string().strip("'")
+
+    new_node: ClassDef = extract_node(
+        TYPING_NEWTYPE_TEMPLATE.format(derived=derived_name, base=base_name)
+    )
+    new_node.parent = node.parent
+
+    # Base type arg is a normal reference, so no need to do special lookups
+    if not isinstance(base, nodes.Const):
+        new_node.bases = [base]
```
DanielNoord, line 176: Doesn't the `extract_node` already infer this? I'm not 100% proficient with astroid internals so this might not be the case, but this seems like it should.

astroid/brain/brain_typing.py
DanielNoord: I'm not that experienced with `astroid` brains, so I'm not sure honestly.

I think this is already seems like a good starting approach.

tests/unittest_brain.py
DanielNoord: @colatkinson Did you consider this comment?

DanielNoord (changes requested):
Thanks @colatkinson 

I'm a bit out of my depth with your last question so I hope another contributor with more `astroid` experience finds the time to review this soon. Just have some final questions myself.

tests/unittest_brain.py
colatkinson: Yup, I ended up just going with the `qname()` check.

astroid/brain/brain_typing.py
colatkinson: That was my initial thought (and implementation) as well. However, in that case the inference wasn't working, and any attrs were coming back as `Uninferable`.

If I had to take a guess, the issue is that the `extract_node()` above doesn't know the context of the surrounding module, and so the base class name isn't resolved properly.

Looking elsewhere in the code, the impression that I got was that constructing the `ClassDef` directly and/or using `ClassDef.postinit()` was more common. So if there is a way to make `extract_node()` behave as desired, I wasn't able to find it.

DanielNoord: Sorry for taking so long to respond to this.

Can we use `ClassDef.postinit` here? I have recently dabbled with the `NamedTuple` brain and also found that resolving base classes can be quite difficult, so I know the struggle now 😄

tests/unittest_brain.py
DanielNoord: Ah sorry! I missed that, thanks!

astroid/brain/brain_typing.py
DanielNoord: Having thought about this a bit more:
Would splitting on `"."`, looking op `decimal` and then looking up `Decimal` work? Perhaps you have already tried this. If so, I'll see if I can come up with something else.

colatkinson:
Hey sorry about the delay. I haven't had a whole lot of bandwidth to work on this, but I'm hoping to be able to get back to it this week. Just a heads up it hasn't totally fallen off my radar!

astroid/brain/brain_typing.py
colatkinson: Yeah I was playing with doing essentially that. There weren't any major problems, I just wasn't sure if there was a cleaner approach. I'll revive that code.

tests/unittest_brain.py
```diff
@@ -1659,6 +1659,128 @@ def test_typing_types(self) -> None:
...
+        NewType("Bar", "int") #@
+        NewType("Baz", Decimal) #@
+        NewType("Qux", decimal.Decimal) #@
+        """
+        )
+        assert isinstance(ast_nodes, list)
+
+        # Base type given by reference
+        foo_node = ast_nodes[0]
+        foo_inferred = next(foo_node.infer())
```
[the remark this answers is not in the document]
colatkinson: I was mostly just copying what some of the other tests do. But I agree, it makes sense to have the increased specificity for these test cases. Changed per your suggestion.

astroid/brain/brain_typing.py
colatkinson: Switched to `postinit`

astroid/brain/brain_typing.py
colatkinson: I took a shot at implementing this, let me know what you think. Definitely a bit hairy, but it seems to work for most cases.

astroid/brain/brain_typing.py
```diff
@@ -140,6 +157,150 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
     return node.infer(context=context_itton)
 
 
+def infer_typing_newtype(
+    node: nodes.Call, context_itton: typing.Optional[context.InferenceContext] = None
```
DanielNoord, line 161: Let's use `context` instead of `context_itton` here, that's the standard in most other similar functions. I think `context_itton` is a legacy name.

astroid/brain/brain_typing.py
```diff
@@ -140,6 +157,150 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+) -> typing.Iterator[nodes.ClassDef]:
+    """Infer a typing.NewType(...) call"""
+    try:
+        func = next(node.func.infer(context=context_itton))
```
DanielNoord, line 165: If there is any ambiguity in the inference then we fail to notice that here. Perhaps we should use `helpers.safe_infer`?

astroid/brain/brain_typing.py
```diff
@@ -140,6 +157,150 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+    derived, base = node.args
+    derived_name = derived.as_string().strip("'")
+    base_name = base.as_string().strip("'")
+
+    new_node: ClassDef = extract_node(
+        TYPING_NEWTYPE_TEMPLATE.format(derived=derived_name, base=base_name)
+    )
+    new_node.parent = node.parent
+
+    # Base type arg is a normal reference, so no need to do special lookups
```
DanielNoord, line 187: Just for the flow of code:
Let's first determine the `bases` and store it in a local variable and then call `postinit` with that local variable. That removes some duplication and (I believe) makes the code a little bit easier to follow.

tests/unittest_brain.py
```diff
@@ -1678,7 +1678,310 @@ def make_new_type(t):
...
+        foo_inferred_all = list(foo_node.infer())
+        assert len(foo_inferred_all) == 1
+
+        foo_inferred = foo_inferred_all[0]
+        self.assertIsInstance(foo_inferred, astroid.ClassDef)
+
+        # Check base type method is inferred by accessing one of its methods
+        foo_base_class_method = foo_inferred.getattr("endswith")[0]
+        self.assertIsInstance(foo_base_class_method, astroid.FunctionDef)
+        self.assertEqual("builtins.str.endswith", foo_base_class_method.qname())
```
DanielNoord, line 1711: Feel free to ignore this, but I would be okay with changing this to use `assert` instead of `self.assert...`. But that's quite a bit of work perhaps, so feel free to leave as is.

DanielNoord (changes requested):
I like the approach you took! I did a first review and left some comments. Let me know if you have any questions!

tests/unittest_brain.py
colatkinson: Done

astroid/brain/brain_typing.py
colatkinson: Done

astroid/brain/brain_typing.py
colatkinson: Ran into a small issue with this. Since `decimal.Decimal` has two implementations (`_pydecimal` vs `_decimal`) it always seems to end up ambiguous.

Is this the desired behavior for cases like this? If so I'll make the change and rework the tests, since I chose `Decimal` as a test type arbitrarily anyway, but I figured I'd confirm.

astroid/brain/brain_typing.py
colatkinson: Switched to `ctx`. Avoids a collision with the module-level `from astroid import context`, and seems to be the convention for the rest of the module.

astroid/brain/brain_typing.py
```diff
@@ -103,23 +109,34 @@ def __class_getitem__(cls, item):
...
+    """Infer a typing.TypeVar(...) call"""
     try:
-        func = next(node.func.infer(context=context_itton))
+        func = next(node.func.infer(context=ctx))
     except (InferenceError, StopIteration) as exc:
         raise UseInferenceDefault from exc
 
-    if func.qname() not in TYPING_TYPEVARS_QUALIFIED:
+    if func.qname() != "typing.TypeVar":
         raise UseInferenceDefault
```
DanielNoord, line 140: I believe this can be removed as it is non-sensical. This is tested by `looks_like_typing_typevar`.

astroid/brain/brain_typing.py
```diff
@@ -129,7 +146,137 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+def infer_typing_newtype(
+    node: nodes.Call, ctx: context.InferenceContext | None = None
+) -> Iterator[nodes.ClassDef]:
+    """Infer a typing.NewType(...) call"""
+    try:
+        func = next(node.func.infer(context=ctx))
+    except (InferenceError, StopIteration) as exc:
+        raise UseInferenceDefault from exc
+
+    if func.qname() != "typing.NewType":
```
DanielNoord, line 161: I believe this can be removed as it is non-sensical. This is tested by `looks_like_typing_newtype`.

astroid/brain/brain_typing.py
```diff
@@ -103,23 +109,34 @@ def __class_getitem__(cls, item):
...
-        func = next(node.func.infer(context=context_itton))
+        func = next(node.func.infer(context=ctx))
     except (InferenceError, StopIteration) as exc:
         raise UseInferenceDefault from exc
 
-    if func.qname() not in TYPING_TYPEVARS_QUALIFIED:
+    if func.qname() != "typing.TypeVar":
         raise UseInferenceDefault
     if not node.args:
         raise UseInferenceDefault
```
DanielNoord, line 142: L127-128 has a drop in coverage. Could you re-create a test for it?

astroid/brain/brain_typing.py
```diff
@@ -129,7 +146,137 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+) -> Iterator[nodes.ClassDef]:
+    """Infer a typing.NewType(...) call"""
+    try:
+        func = next(node.func.infer(context=ctx))
+    except (InferenceError, StopIteration) as exc:
+        raise UseInferenceDefault from exc
+
+    if func.qname() != "typing.NewType":
+        raise UseInferenceDefault
+    if len(node.args) != 2:
```
DanielNoord, line 163: Could you create a test for this? It is currently uncovered.

astroid/brain/brain_typing.py
```diff
@@ -129,7 +146,137 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+        raise UseInferenceDefault
+
+    # Cannot infer from a dynamic class name (f-string)
+    if isinstance(node.args[0], JoinedStr):
```
DanielNoord, line 167: You can probably re-use the test for this for the comment above.

DanielNoord (changes requested):
Some final comments.

I'd really like somebody with more experience on inference to take a look at this as I'm not sure the way we resolve the base names here is something we normally do in the `brains` or if we have `utils` for this.

@cdce8p Do you see chance to take a look at this PR perhaps?

astroid/brain/brain_typing.py
```diff
@@ -129,7 +144,135 @@ def infer_typing_typevar_or_newtype(node, context_itton=None):
...
+            new_bases = [base_node]
+        elif "." in base.value:
+            possible_base = _try_find_imported_object_from_str(node, base.value, ctx)
+            if possible_base:
+                new_bases = [possible_base]
+
+    if new_bases:
+        new_node.postinit(
+            bases=new_bases, body=new_node.body, decorators=new_node.decorators
+        )
```
cdce8p, line 202: `new_node` is fully constructed already. It's enough to set `bases` manually.
```suggestion
        new_node.bases = new_bases
```

cdce8p (changes requested):
Thanks for your patience. I finally got around to look it your PR. First of all, I'm quite impressed with the work you did here @colatkinson! You obviously have a pretty good understanding of the astroid internals already.

However, as impressive as the PR is, I'm not quite sure the current approach is the correct one. You've modeled the inference result after the typing implementation. Astroid (and pylint) aren't type checkers though. They use inference to determine the `runtime` results. For a `NewType` call, that's simply the argument itself which is returned and not an instance of a subclass.
```py
from typing import NewType

B = NewType("B", "A")

class A:
    def __init__(self, value: int):
        self.value = value

print(B(A(5)))
```
```
<__main__.A object at 0x100b1b160>
```

The inference result with the current implementation:
```
[<Instance of .B at 0...427248880>]
```

The CPython implementation: https://github.com/python/cpython/blob/v3.10.5/Lib/typing.py#L2482-L2483

--
Am I missing something / what do you think? In case we change the approach, I would like to suggest that you create a new PR for it. That way we can at least preserve the idea and implementation of resolving base classes should there ever be a need for it.

colatkinson:
@cdce8p Thanks for taking a look! The feedback from you and @DanielNoord has been super helpful throughout this process.

Your point about the divergence of the inferred and runtime types make sense to me. I modeled the implementation on the one currently in the main branch, which leaves things in roughly the same state (inferring a class that doesn't exist at runtime), but that's kind of the source of all this trouble in the first place :wink:.

Do you have any thoughts or pointers on a more correct implementation? My initial thought would be a simple identity function, e.g.

```python
def ChildType(x):
    return x
```

Do you think that would be sufficient? It could be a bit strange that the code would infer a function, which would then be referenced by type annotations elsewhere, but I'm not sure if this presents any problems in practice.

It might also make sense to have the inferred implementation be identical to the upstream CPython one you linked -- though in that case, is any transformation even necessary? Based on a quick test, the results come back as `Uninferable` without any transforms, but I'm not sure if there's a lighter-touch way around that.

Finally, with regards to creating a new PR: I was planning on keeping the current implementation around in the git history. Though if a new PR is easier for you and other maintainers, or if the project generally makes use of squash commits, I can definitely do that instead.

cdce8p:
> Do you have any thoughts or pointers on a more correct implementation? My initial thought would be a simple identity function, e.g.
> 
> ```python
> def ChildType(x):
>     return x
> ```
> 
> Do you think that would be sufficient? It could be a bit strange that the code would infer a function, which would then be referenced by type annotations elsewhere, but I'm not sure if this presents any problems in practice.

An identity function might work, but I would patch `ChildType`. Instead it might be better to replace the `NewType` inferred function itself. That could also solve the type annotation issue. Basically the inferred type should become `x` instead of `ChildType`. You just don't get the subclassing.

Tbh though, I haven't had time to investigate if it's at all possible.

> Finally, with regards to creating a new PR: I was planning on keeping the current implementation around in the git history. Though if a new PR is easier for you and other maintainers, or if the project generally makes use of squash commits, I can definitely do that instead.

We do use squash merge usually. However, my reasoning was more from a practical approach. No unrelated comments and if everything else fails, we can still come back to this one.

I'll convert the PR to a draft. We can close it once the alternative is merged.

PR #1302, opened 2021-12-19
Create ``scoped_nodes`` module

DanielNoord:
## Steps

## Description

I've been working on a fix for `LookupMixIn._filter_stmts`. However, for this I need to access `nodes.Module` which I currently can't because of the numerous circular imports and dependencies within the `nodes` module.
I would like to start fixing this by creating this `scoped_nodes` package > extracting `LocalsDictNodeNG` (which was already planned 12 (!) years ago https://github.com/PyCQA/astroid/commit/bec4ee12322d3f10df273e0764290e41cc6c701b > extracting `Module`.

Sadly this will impact the git blame view with the GitLens extension in VS Code again, but I think it is much better to do this in a separate module than to create `astroid/nodes/localsdictnodeng.py` and  `astroid/nodes/module.py`.

## Type of Changes


Pierre-Sassoulas (approved):
Wow, how did you find this commit talking about it from 12 years ago ? 😄 

I think bursting scoped_nodes makes a lot of sense, but it's not an easy task, the circular imports hell is real. Problems for another MR though.

DanielNoord:
> Wow, how did you find this commit talking about it from 12 years ago ? 😄

Just went back in time through the git blame 😄 

> I think bursting scoped_nodes makes a lot of sense, but it's not an easy task, the circular imports hell is real. Problems for another MR though.

I've been breaking my head for the last hour on this and still haven't found the correct pattern yet. It's a miracle it all works right now 😄 We do really need the fix that prompted my search into this though, so I'll keep trying.
For now: please don't merge this yet. I want to make sure that this can be the basis for an eventual fix.

Pierre-Sassoulas:
> It's a miracle it all works right now 😄 

After taking dozen of hours to try to break the circular import myself I must agree. It's really hard to know where to start. Some circular import cannot be changed without major redesign. Creating an API so we can move stuff around inside the scoped_nodes module is for sure the first step.

DanielNoord:
I'm going to close this PR as I found a better and less work-intensive solution to the problem I faced. I think there are more pressing matters than working on this circular import mess now.

Pierre-Sassoulas:
Are you sure ? This first step is pretty safe and would need to be done again later ? :)

DanielNoord:
I'm not against doing the change but was unsure if we were missing unforeseen backwards compatibility issues.. So instead of trying to figure those out I thought I would focus on actual issues.

I'm fine with restoring the branch and reopening. Do you want me to do that?

Pierre-Sassoulas:
The only problem that could happen is if the `__init__.py` lack some import in `__all__`. Maybe we could be super safe and only move `scoped_nodes.py` in `scoped_nodes/__init__.py`. Depends on how confident you are that the API you created is exhaustive.

DanielNoord:
I just copied the import from the ``__init__`` in nodes, so that should be relatively safe.
Obviously if we want to do this we should merge this sooner rather than later due to potential merge conflicts.

Pierre-Sassoulas (approved):
Right now it's relatively safe as git can detect a file move and a rebase would not have conflicts. It's when you start moving big chunk of code in new files that problem really arise. But there's no reason to not merge fast :)

PR #1303, opened 2021-12-20
Move ``_filter_stmts`` into its own file

DanielNoord:
## Steps

## Description

We need access to all types of nodes in ``_filter_stmts``. Not only those in ``node_classes`` (the current file in which it is defined).

This change shows that by moving `_filter_stmts` into its own file we can import the `nodes` module and access all of them. This will allow to fix some issue with it. I have tried to keep the change as minimal as possible and only changed how we refer to `nodes`.

I wonder if keeping the leading underscore is correct. It does show that this is not public API, but is this how you are supposed to do that when functions are in a separate file?

## Type of Changes


Pierre-Sassoulas (approved):
I think not changing the name make sense, especially if we want to be able to move it again or rename later. Good job identifying a separable part !

PR #1305, opened 2021-12-20
Add future=True to frame calls

tushar-deepsource:
This builds upon #1263, the diff should be smaller once that is merged.

## Type of Changes


astroid/nodes/node_ng.py
```diff
@@ -318,7 +317,19 @@ def frame(
...
         """
-        return self.parent.frame()
+        if self.parent is None:
+            if future:
+                raise ParentMissingError(target=self)
+            warnings.warn(
+                "In astroid 3.0.0 NodeNG.frame(future=True) will return either a Frame, "
+                "or raise ParentMissingError. AttributeError will no longer be raised. "
+                "This behaviour can already be triggered "
+                "by passing 'future=True' to a statement() call.",
```
jacobtylerwalls, line 327:
```suggestion
                "by passing 'future=True' to a frame() call.",
```

astroid/nodes/node_ng.py
```diff
@@ -318,7 +317,19 @@ def frame(
 
         :returns: The first parent frame node.
         """
-        return self.parent.frame()
+        if self.parent is None:
+            if future:
+                raise ParentMissingError(target=self)
+            warnings.warn(
+                "In astroid 3.0.0 NodeNG.frame(future=True) will return either a Frame, "
```
jacobtylerwalls, line 324:
```suggestion
                "In astroid 3.0.0 NodeNG.frame() will return either a Frame, "
```

jacobtylerwalls (commented):
Hi, just some comments on the deprecation warning itself.

astroid/nodes/node_ng.py
jacobtylerwalls: Pardon me I see you already have this feedback in #1263, I didn't see.

astroid/nodes/node_ng.py
jacobtylerwalls: Same as above, you already resolved this in #1263, never mind me!

tushar-deepsource:
This seems like an unrelated failure. Flaky test?

Pierre-Sassoulas:
> This seems like an unrelated failure. Flaky test?

Yes, we had the issue for other pipelines. I don't know what the problem is yet. Probably due to the new release of python 3.10.

tushar-deepsource:
Resolved merge conflict, should be good

Pierre-Sassoulas (approved):
LGTM. Can we merge this in 2.9.1 @DanielNoord ?

DanielNoord:
I haven't look at the other changes to actual code yet, but note that in https://github.com/PyCQA/astroid/pull/1217 and https://github.com/PyCQA/astroid/pull/1235 we did not change the existing tests but added additional tests with the `future` keyword. This guarantees that previous behaviour has not been changed.
See:
https://github.com/PyCQA/astroid/blob/7afd696520f70a46334adea6d39f7150379198f3/tests/unittest_scoped_nodes.py#L824-L825

It is a bit nitpicky, but might be good to do here as well. It also means we can remove the lines with `future` in `3.0` and the `diff` wouldn't change for the original tests.

Pierre-Sassoulas:
> It is a bit nitpicky, but might be good to do here as well. 

Yeah I hesitated doing that comment too. Having test for the old behavior is nice but I missed the fact that `` we can remove the lines with future in 3.0 and the diff wouldn't change for the original tests.`` and it's a very nice thing to be able to do. 👍

tusharsadhwani:
oh that's true. I'll fix that

tushar-deepsource:
All done

cdce8p (approved):
LGTM! Test failures are again unrelated

Pierre-Sassoulas (approved):
I'll let Daniel merge after his review :)

DanielNoord (approved):

PR #1306, opened 2021-12-20
Create a ``NamedTuple`` instance for NamedTuples

DanielNoord:
## Steps

## Description

This is the first step towards some improvements to the `NamedTuple` brain. One of the first things I thought we should fix is the fact that we currently lose the fact that something is a `NamedTuple`. We don't add it to the object's bases and they are undistinguishable from other instances.

I'm not sure if adding a new `Instance` is actually necessary. I thought it might make sense as `NamedTuple` behaves differently than `tuple` and can have default values for example. I have created the `T(opic): NamedTuple` label to identify some issues related to `NamedTuple`. I thought a new `Instance` type could be the basis for solving some of these.

Changes have been tested against current latest commit of `pytest`.

## Type of Changes

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -186,20 +186,29 @@ def infer_named_tuple(
     node: nodes.Call, context: InferenceContext | None = None
 ) -> Iterator[nodes.ClassDef]:
     """Specific inference function for namedtuple Call node"""
-    tuple_base_name = nodes.Name(name="tuple", parent=node.root())
-    class_node, name, attributes = infer_func_form(
-        node, tuple_base_name, context=context
-    )
-    call_site = arguments.CallSite.from_call(node, context=context)
-    node = extract_node("import collections; collections.namedtuple")
-    try:
+    # Infer which type of NamedTuple we're dealing with (typing or collections)
+    inferred_namedtuple_call = next(node.func.infer())
+    base_names = [
+        nodes.Name(name="tuple", parent=node.root()),
+        nodes.Name(
+            name=inferred_namedtuple_call.name,
+            parent=inferred_namedtuple_call.root(),
+        ),
+    ]
 
-        func = next(node.infer())
-    except StopIteration as e:
-        raise InferenceError(node=node) from e
+    class_node, name, attributes = infer_func_form(node, base_names, context=context)
+    call_site = arguments.CallSite.from_call(node, context=context)
+
     try:
-        rename = next(call_site.infer_argument(func, "rename", context)).bool_value()
+        rename = next(
+            call_site.infer_argument(inferred_namedtuple_call, "rename", context)
+        ).bool_value()
     except (InferenceError, StopIteration):
+        rename = False
+    # If inferred_namedtuple_call is the ClassDef of typing.NamedTuple
+    # infer_argument will raise AttributeError
+    # TODO: See if this exception can be prevented
+    except AttributeError:
         rename = False
 
     try:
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas:
@cdce8p do you think you'll have time to review this ? I see you self requested a review two months ago. I'm trying to release astroid 2.10 :)

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -182,23 +195,34 @@ def _looks_like(node, name):
...
+        rename = next(
+            call_site.infer_argument(inferred_namedtuple_call, "rename", context)
+        ).bool_value()
     except (InferenceError, StopIteration):
         rename = False
+    # If inferred_namedtuple_call is the ClassDef of typing.NamedTuple
+    # infer_argument will raise AttributeError
+    # TODO: See if this exception can be prevented
+    except AttributeError:
+        rename = False
```
areveny, line 226: Should we also support the parameters `defaults` and `module`? I can give this a shot. Do we need version guards?

https://docs.python.org/3/library/collections.html#collections.namedtuple

areveny (commented):
I'm beginning to understand Astroid brain a lot better after looking at this!

astroid/brain/brain_namedtuple_enum.py
DanielNoord: I think those are both out of scope for this PR, but I think at some point we will want to. Basically this PR is meant as the base for work on stuff like #242 and #182.

I think the biggest open question for this PR is whether creating a new `Instance` sub class is the right way to go here.  Or if there should be another mechanism for this.

cdce8p (commented):
Had finally time to look at it. Tbh I'm myself not exactly the "brain" expert. However from what I've gathered so far, I'm not sure this is the best solution to the problem.

I do understand your desire to have `NamedTuple` / `namedtuple` at least somewhere which isn't the case today, but it seems strange to add a new instance type just for that. Wouldn't it be much more intuitive to insert a new base which itself has `tuple` as base? That's at least roughly similar to the pattern we used in the typing brain.

Regarding the remaining changes, I would probably need more time but it looks like there are 3-4 others bundled here as well. That makes it more challenging to review.

Going forward, I would suggest we (1) move the PR back to `2.11`, (2) unbundle additional changes, and (3) explore if inserting a dummy base class would work.

DanielNoord:
> I do understand your desire to have `NamedTuple` / `namedtuple` at least somewhere which isn't the case today, but it seems strange to add a new instance type just for that. Wouldn't it be much more intuitive to insert a new base which itself has `tuple` as base? That's at least roughly similar to the pattern we used in the typing brain.

Hmm, I guess this depends on what `bases` and `Instances` are supposed to be. Personally I think it makes more sense to create a `NamedTuple` `Instance` that can be re-used rather than defining them through f-strings. In my mind it goes like `Instance` is an instances of a general `ClassDef`. A sub-class of `Instance` is thus an instance of a specific type of `ClassDef`.
But I can see why that might not always be logical.

> Regarding the remaining changes, I would probably need more time but it looks like there are 3-4 others bundled here as well. That makes it more challenging to review.
> 
> Going forward, I would suggest we (1) move the PR back to `2.11`, (2) unbundle additional changes, and (3) explore if inserting a dummy base class would work.

I'll try and separate some of the work here. Some of the typing should be fairly straightforward to extract and then review for example. I'll get back to you!

Pierre-Sassoulas:
@DanielNoord can we close this ?

DanielNoord:
I'll triage this somewhere next week to see if there is anything still relevant in here

DanielNoord:
Closing!

PR #1310, opened 2021-12-23
Upgrade mypy to 0.930

cdce8p:
## Description
http://mypy-lang.blogspot.com/2021/12/mypy-0930-released.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,7 +63,7 @@ repos:
           ]
         exclude: tests/testdata|conf.py
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.920
+    rev: v0.930
     hooks:
       - id: mypy
         name: mypy
--- a/requirements_test_pre_commit.txt
+++ b/requirements_test_pre_commit.txt
@@ -3,4 +3,4 @@ pylint==2.12.2
 isort==5.9.2
 flake8==4.0.1
 flake8-typing-imports==1.11.0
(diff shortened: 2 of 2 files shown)


cdce8p:
The failure is caused by Python 3.10.1, not the mypy upgrade. Needs to be fixed separately.

DanielNoord:
We postponed updating `tbump` in `pylint` because of that fail, see https://github.com/PyCQA/pylint/issues/5572

cdce8p:
> We postponed updating `tbump` in `pylint` because of that fail, see [PyCQA/pylint#5572](https://github.com/PyCQA/pylint/issues/5572)

@DanielNoord Are you sure those are related? The issue here looks like one with Python 3.10.1 on Windows.
https://github.com/PyCQA/astroid/runs/4618100769?check_suite_focus=true#step:7:70

DanielNoord:
Oh my bad! Didn't read the traceback fully. I saw `distutils` and assumed they were related. Seems like a strange coincidence then.

DanielNoord:
@cdce8p Are you looking into this failure or should somebody else do this? It's affecting tests of other PRs as well so we should probably try and fix this asap.

cdce8p:
> @cdce8p Are you looking into this failure or should somebody else do this? It's affecting tests of other PRs as well so we should probably try and fix this asap.

@DanielNoord I didn't had time for it yet. Do you by chance use Windows? I think that would definitely help when debugging it. I'm on macOS unfortunately for this one.

PR #1314, opened 2021-12-27
Bump CI cache version

cdce8p:
## Description
Preemptively bumping the cache version to prevent the issues seen in pylint.
https://github.com/PyCQA/pylint/pull/5602


Pierre-Sassoulas (approved):
Some job fails but we can still merge 👍

Issue #1103, opened 2021-07-19
Remove dependency to ``setuptools`` and ``distutil``


Pierre-Sassoulas:
As see in #1030  and #1100 there is a hidden dependency to setuptools (and probably distutils too), we made it explicit in #1100 but the better thing to do would be to remove it entirely if possible.

Keuricloud8866:
How can I learn as begineer

Pierre-Sassoulas:
Hi @Keuricloud8866, what do you want to learn ? In order to fix this ticket you should track the place where setuptools and distutils are used in pylint and find an alternative. For finding alternative we'll probably have to discuss how to do what we were doing before without setuptools.

Keuricloud8866:
I am new and begineer in software IT world i just jumped in and want to be in so

Pierre-Sassoulas:
You should probably check a tutorial about learning programming first. Another useful thing would be to know how to ask question:
- First read the error or documentation
- Second search Google or Stackoverflow
- Third ask for help in the right place

Did you search how to learn programming by yourself first and read a blog about it ? I think you jumped to three and this is not the right place to ask either. This is an issue about a particular issue in a particular software, you were very unlikely to get answers here. (in fact please do not ask other off-topic questions or I'll have to ban you).

DanielNoord:
Closed with #1323

DanielNoord:
Do we still need to dependency on `setuptools` now? It was added in https://github.com/PyCQA/astroid/commit/40629baba2de2c9eb5e11b65798b3fae79f7284a.
Now that we have removed `distutils` I don't think it is necessary, right?

Pierre-Sassoulas:
Probably ? I'm trying to think of a way to test that. Maybe launching pylint test suite in a venv after we deinstalled ``setuptools`` ?

Issue #1282, opened 2021-12-03
Python 3.10 deprecation warnings due to `distutils` import


The-Compiler:
### Steps to reproduce

- Run `python3.10 -Wall -c "import astroid"`

### Current behavior

```
.../astroid/interpreter/_import/spec.py:20: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils
.../astroid/modutils.py:51: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
  from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
```

### Expected behavior

No deprecation warnings.

### `python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"` output

2.9.1-dev0 (on 6691a9e97373b28e15a76b747f78c041f2064df4)

### Additional context

Might be related to #1103, but seems like a separate issue.

Pierre-Sassoulas:
Doing #1103 would work, but upgrading setuptools min version to one that do not use distutils would probably work (?). I think we won't change anything before 3.12 is available in github actions.

DanielNoord:
Closed with #1323

PR #1316, opened 2021-12-27
Fix distutils tests following setuptools 60.0.0 release

### LINKED ISSUE #1313: Fix Github Action for the new version of setuptools
Currently the pipeline is broken for the new version of python 3.10 on Windows:
```

================================== FAILURES ===================================
_________ ModuleFileTest.test_find_distutils_submodules_in_virtualenv _________

self = <tests.unittest_modutils.ModuleFileTest testMethod=test_find_distutils_submodules_in_virtualenv>

    def test_find_distutils_submodules_in_virtualenv(self) -> None:
        found_spec = spec.find_spec(["distutils", "version"])
>       self.assertEqual(found_spec.location, distutils.version.__file__)
E       AssertionError: 'C:\\hostedtoolcache\\windows\\Python\\3.1[32 chars]n.py' != 'D:\\a\\astroid\\astroid\\venv\\lib\\site-[40 chars]n.py'
E       - C:\hostedtoolcache\windows\Python\3.10.1\x64\lib\distutils\version.py
E       + D:\a\astroid\astroid\venv\lib\site-packages\setuptools\_distutils\version.py
```

cdce8p:
With v60.0.0 setuptools started monkeypatching the distuils module
to use their own vendored version. Longterm any uses of distuils
should be replaced or removed.

https://github.com/pypa/setuptools/pull/2896
https://setuptools.pypa.io/en/latest/history.html#v60-0-0

--
This change isn't ideal as it might hide an error, but for now it will unblock the CI pipeline.
Furthermore, at least for now, the setuptools copy should be identically with the stdlib one.

Closes #1313

References:
* #1282
* #1103

### DIFF
--- a/tests/unittest_modutils.py
+++ b/tests/unittest_modutils.py
@@ -33,6 +33,8 @@ import xml
 from xml import etree
 from xml.etree import ElementTree
 
+import pytest
+
 import astroid
 from astroid import modutils
 from astroid.interpreter._import import spec
@@ -72,6 +74,13 @@ class ModuleFileTest(unittest.TestCase):
             ["data", "MyPyPa-0.1.0-py2.5.egg", self.package],
         )
 
+    # TODO: Fix when removing distutils dependency
+    # https://github.com/pypa/setuptools/pull/2896
+    @pytest.mark.xfail(
+        sys.version_info > (3, 7),
+        sys.version_info < (3, 11),
+        reason="setuptools v60.0.0 starts using local copy of distutils",
+    )
     def test_find_distutils_submodules_in_virtualenv(self) -> None:
         found_spec = spec.find_spec(["distutils", "version"])
         self.assertEqual(found_spec.location, distutils.version.__file__)
--- a/tests/unittest_modutils.py
+++ b/tests/unittest_modutils.py
@@ -20,9 +20,8 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 
 """
-unit tests for module modutils (module manipulation utilities)
+Unit tests for module modutils (module manipulation utilities)
 """
-import distutils.version
 import email
 import os
 import shutil
@@ -33,7 +32,7 @@ import xml
 from xml import etree
 from xml.etree import ElementTree
 
-import pytest
+import wrapt
 
 import astroid
 from astroid import modutils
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (changes requested):
We're trying to test ``find_spec`` in this test, distutils was used because it was supposed to be always present i guess. We might as well use another dependency we'll keep, like ``wrapt``, ``lazy_object_proxy`` (or ``setuptools``, but it might become problematic when we adopt pyproject.toml and something else than setuptools or if we dump setuptools.).

DanielNoord:
I think changing the package used for this test makes sense!

tests/unittest_modutils.py
```diff
@@ -72,9 +73,9 @@ def test_find_egg_module(self) -> None:
             ["data", "MyPyPa-0.1.0-py2.5.egg", self.package],
         )
 
-    def test_find_distutils_submodules_in_virtualenv(self) -> None:
-        found_spec = spec.find_spec(["distutils", "version"])
-        self.assertEqual(found_spec.location, distutils.version.__file__)
+    def test_find_wrapt_submodules_in_virtualenv(self) -> None:
+        found_spec = spec.find_spec(["wrapt", "decorators"])
+        self.assertEqual(found_spec.location, wrapt.decorators.__file__)
```
cdce8p, line 78: The test was specifically designed to test the location of `distutils`. So this change wouldn't fix it

DanielNoord:
I retract my earlier statement, see https://github.com/PyCQA/astroid/pull/672/files.

We actually want to test `distutils` here, the test was added specifically for it.

cdce8p:
The idea behind this PR was to **temporarily** disable the test just until we fix / resolve #1103 and #1282. It doesn't look like it's causing any real world issue at the moment, so we might get away with it.

Pierre-Sassoulas:
Seeing the issue in https://github.com/PyCQA/pylint/issues/73 the result could be false positive for distutil in pylint ?

Pierre-Sassoulas (approved):
I think this is a good temporary solution even if we regress on pylint #73 as distutil is old and deprecated.

DanielNoord:
I think I might have found a fix. I'll create a new PR.

DanielNoord:
See #1321. I think we can close this in favour of that one, test seem to be passing there 😄

Pierre-Sassoulas:
Amazing fix @DanielNoord !

PR #1317, opened 2021-12-27
Fix node.statement overload typing - default argument

cdce8p:
## Description

As it turns out, the default arg is needed for astroid itself. We pass the future argument unconditionally to the parent call for which `future` can be `None`.
```py
...
return self.parent.statement(future=future)
```

/CC: @DanielNoord

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -278,7 +278,9 @@ class NodeNG:
         return any(self is parent for parent in node.node_ancestors())
 
     @overload
-    def statement(self) -> Union["nodes.Statement", "nodes.Module"]:
+    def statement(
+        self, *, future: Literal[None] = ...
+    ) -> Union["nodes.Statement", "nodes.Module"]:
         ...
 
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
It seems there's an issue with distutils. ,Maybe due to https://bugs.python.org/msg394277 ? It's surprising that it fail for old python version too now.

DanielNoord (approved):

cdce8p:
> It seems there's an issue with distutils. ,Maybe due to https://bugs.python.org/msg394277 ?

Not exactly, it's an issue with our hack around distutils and a new setuptools version which does some hacks itself.

> It's surprising that it fail for old python version too now.

Bumping the cache version caused all environments to be recreated. That installed the latest `setuptools` version in every one of them which subsequently fail. `3.6` is fine as setuptools `v60.0.0` requires Python 3.7.

Let's continue the discussion over in #1316

--
Since the test failures are unrelated, I'm going to merge this.

Issue #1320, opened 2021-12-29
Decrease the time required to import astroid


Pierre-Sassoulas:
### Steps to reproduce

``python3 -X importtime -c 'import astroid'``

### Current behavior

It takes a half second to import astroid. This account for 92% of pylint's startup time in 2.13.0-dev0. See https://github.com/PyCQA/pylint/issues/4814 for details.

```
import time: self [us] | cumulative | imported package
import time:       182 |        182 |   _io
import time:        59 |         59 |   marshal
import time:       213 |        213 |   posix
import time:       463 |        916 | _frozen_importlib_external
import time:       143 |        143 |   time
import time:       210 |        353 | zipimport
import time:        68 |         68 |     _codecs
import time:      8157 |       8225 |   codecs
import time:      4532 |       4532 |   encodings.aliases
import time:      2092 |      14848 | encodings
import time:       526 |        526 | encodings.utf_8
import time:       183 |        183 | _signal
import time:       910 |        910 | encodings.latin_1
import time:        34 |         34 |     _abc
import time:       527 |        560 |   abc
import time:       539 |       1099 | io
import time:        43 |         43 |       _stat
import time:       522 |        564 |     stat
import time:      1066 |       1066 |     _collections_abc
import time:       459 |        459 |       genericpath
import time:       481 |        939 |     posixpath
import time:       781 |       3349 |   os
import time:       451 |        451 |   _sitebuiltins
import time:        57 |         57 |     _locale
import time:       411 |        467 |   _bootlocale
import time:      1220 |       1220 |     apport_python_hook
import time:       396 |       1615 |   sitecustomize
import time:      4927 |      10808 | site
import time:       989 |        989 |     types
import time:       833 |        833 |     warnings
import time:      1833 |       3654 |   importlib
import time:       954 |        954 |         enum
import time:        58 |         58 |           _sre
import time:       560 |        560 |             sre_constants
import time:      1033 |       1593 |           sre_parse
import time:       788 |       2437 |         sre_compile
import time:        58 |         58 |               _operator
import time:      1129 |       1187 |             operator
import time:       424 |        424 |             keyword
import time:        34 |         34 |               _heapq
import time:       499 |        533 |             heapq
import time:        85 |         85 |             itertools
import time:       496 |        496 |             reprlib
import time:        47 |         47 |             _collections
import time:      1360 |       4129 |           collections
import time:        37 |         37 |           _functools
import time:      3949 |       8114 |         functools
[...]

PR #1321, opened 2021-12-29
Fix ``distutils`` path patching for setuptools >= 60.0.0

### LINKED ISSUE #1313: Fix Github Action for the new version of setuptools

DanielNoord:
## Steps

## Description

This passed locally. I wonder if this works here as well 😄 

## Type of Changes

## Related Issue

For future reference, see the original commit that added path patching for `distutils` here: https://github.com/PyCQA/astroid/commit/6ca01e0e1c691f08fbed9fab6c76339751c83b0b

Closes #1313

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,8 @@ Release date: TBA
 
 * Removed custom ``distutils`` handling for resolving paths to submodules.
 
+  Ref #1321
+
 * Fix ``deque.insert()`` signature in ``collections`` brain.
 
   Closes #1260
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,7 +24,7 @@ Release date: TBA
 
 * Require Python 3.6.2 to use astroid.
 
-* Removed custom ``distutils`` handling for resolving paths to submodules
+* Removed custom ``distutils`` handling for resolving paths to submodules.
 
 * Fix ``deque.insert()`` signature in ``collections`` brain.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,8 @@ Release date: TBA
 
 * Require Python 3.6.2 to use astroid.
 
+* Removed custom ``distutils`` handling for resolving paths to submodules
+
 * Fix ``deque.insert()`` signature in ``collections`` brain.
 
   Closes #1260
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -17,18 +17,12 @@
 
 import abc
 import collections
-import distutils
 import enum
 import importlib.machinery
 import os
 import sys
 import zipimport
 from functools import lru_cache
-
-try:
-    from setuptools import _distutils
-except ImportError:
-    _distutils = None
 
 from . import util
 
@@ -167,14 +161,6 @@ class ImportlibFinder(Finder):
                 if os.path.isdir(os.path.join(p, *processed))
             ]
         # We already import distutils elsewhere in astroid,
-        # so if it is the same module, we can use it directly.
-        elif spec.name == "distutils" and spec.location in distutils.__path__:
-            # distutils is patched inside virtualenvs to pick up submodules
-            # from the original Python, not from the virtualenv itself.
-            path = list(distutils.__path__)
-        elif spec.name == "distutils" and _distutils:
-            # distutils is patched in setuptools >= 60.0.0 to _distutils
-            path = list(_distutils.__path__)
         else:
             path = [spec.location]
         return path
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -25,6 +25,11 @@ import sys
 import zipimport
 from functools import lru_cache
 
+try:
+    from setuptools import _distutils
+except ImportError:
+    _distutils = None
+
 from . import util
 
 ModuleType = enum.Enum(
@@ -167,6 +172,9 @@ class ImportlibFinder(Finder):
             # distutils is patched inside virtualenvs to pick up submodules
             # from the original Python, not from the virtualenv itself.
             path = list(distutils.__path__)
+        elif spec.name == "distutils" and _distutils:
+            # distutils is patched in setuptools >= 60.0.0 to _distutils
+            path = list(_distutils.__path__)
         else:
             path = [spec.location]
         return path
--- a/tests/unittest_modutils.py
+++ b/tests/unittest_modutils.py
@@ -22,7 +22,6 @@
 """
 unit tests for module modutils (module manipulation utilities)
 """
-import distutils.version
 import email
 import os
 import shutil
@@ -71,10 +70,6 @@ class ModuleFileTest(unittest.TestCase):
             found_spec.location.split(os.sep)[-3:],
             ["data", "MyPyPa-0.1.0-py2.5.egg", self.package],
         )
-
-    def test_find_distutils_submodules_in_virtualenv(self) -> None:
-        found_spec = spec.find_spec(["distutils", "version"])
-        self.assertEqual(found_spec.location, distutils.version.__file__)
 
 
 class LoadModuleFromNameTest(unittest.TestCase):


Pierre-Sassoulas (approved):
🔥

Pierre-Sassoulas:
@cdce8p once we merge this we can release astroid 2.9.1, would you mind having a quick look ? :)

astroid/interpreter/_import/spec.py
```diff
@@ -167,6 +172,9 @@ def contribute_to_path(self, spec, processed):
             # distutils is patched inside virtualenvs to pick up submodules
             # from the original Python, not from the virtualenv itself.
             path = list(distutils.__path__)
+        elif spec.name == "distutils" and _distutils:
+            # distutils is patched in setuptools >= 60.0.0 to _distutils
+            path = list(_distutils.__path__)
```
cdce8p, line 177: Tbh I wonder if we should just always use `distutils.__path__`. We do import it earlier, so it's most likely the one we actually want. From what I can tell, the first condition is equal to the else case anyway.
```py
        elif spec.name == "distutils":
            path = list(distutils.__path__)
```

Defaulting to the `_distutils` path from setuptools alone can however be dangerous. It was added in `v48.0.0` but only became the default in `v60.0.0`. So just in theory, if the first condition wouldn't hit, it would pick the wrong path. In that case you probably should add an additional condition to make sure the setuptools distutils is actual the distutils we are using. With that we likely end up with the simplified case from above again.
```py
        elif spec.name == "distutils" and _distutils and _distutils.__path__ == distutils.__path__:
            path = list(distutils.__path__)
```

cdce8p (commented):
I tried to reproduce the issue mentioned in https://github.com/PyCQA/pylint/issues/73 and with the example from https://github.com/PyCQA/pylint/issues/2955 both with `v59` and `v60` of setuptools, but wasn't able too. With and without any changes.

It seems to have been fixed in `venv` in the meantime. We could consider adding a regression test in pylint though. This should be enough. If it doesn't raise an ImportError, everything is good.
```py
import distutils.util
```
_Example from https://github.com/PyCQA/pylint/issues/2955_

astroid/interpreter/_import/spec.py
cdce8p: Additionally, just doing `elif spec.name == "distutils"` would remove a `setuptools` dependency. -> #1320
Maybe we should just test it in production and see if we get any new bug reports 🤔

DanielNoord:
> I tried to reproduce the issue mentioned in [PyCQA/pylint#73](https://github.com/PyCQA/pylint/issues/73) and the example from [PyCQA/pylint#2955](https://github.com/PyCQA/pylint/issues/2955) both with `v59` and `v60` of setuptools, but wasn't able too. With and without any changes.

I didn't check but could the crash seen here be of any help? https://github.com/PyCQA/pylint/issues/5572
It seems to involve this same part of the code.

cdce8p:
> > I tried to reproduce the issue mentioned in [PyCQA/pylint#73](https://github.com/PyCQA/pylint/issues/73) and the example from [PyCQA/pylint#2955](https://github.com/PyCQA/pylint/issues/2955) both with `v59` and `v60` of setuptools, but wasn't able too. With and without any changes.
> 
> I didn't check but could the crash seen here be of any help? [PyCQA/pylint#5572](https://github.com/PyCQA/pylint/issues/5572) It seems to involve this same part of the code.

Not sure. Do you know if the issue still occurs with the latest setuptools version? There have been a lot of small (bugfix?) release lately. I would guess after they made their version the default, they noticed some more issues and fixed them in subsequent patches. We don't pin `setuptools` for our CI builts so it will only get updated if a requirement changes. You could bump the `CACHE_VERSION` again. That should install the latest version.

There is also https://github.com/PyCQA/pylint/pull/5600 which does look fine, aside from the fact that they dropped support for Python 3.6 which causes some tests to fail.

astroid/interpreter/_import/spec.py
DanielNoord: I'm not completely following you here. Especially with all the latest hot fixes and releases I'm not sure what actually needs to be fixed.

For `pylint` everything seems fine. We might be incompatible with `setuptools~=60.0` but due to all the recent releases it would be quite a big effort to find the specific release which fixed the bug that made the `tbump` upgrade fail. I feel like we can let that be, right?

For `astroid` this test is still failing for `setuptools==60.2.0`. I agree that the first `if` seems to do the same as the `else`. That can likely be removed, leaving only a `if` for `_distutils` to be necessary. Do you have any idea how to do that `if` without depending on `setuptools` but being compatible with both `v47` and `v60`?

Please correct me where I'm wrong. This has been a difficult issue to keep track off and its getting late 😄

cdce8p: > This has been a difficult issue to keep track off and its getting late 😄

Same 😴 I can try to summarize what I think is going on here.

**What do we know?**
Starting with `v60` setuptools patches (or better "does some sketchy things") to replace the stdlib version of distutils with their own copy. This is vendored from [pypa/distutils](https://github.com/pypa/distutils) and "should" be functionally identical with the stdlib version. If at all, it might include some more bugfixes that won't get included into stdlib.

**Why?**
`distuils` has been deprecated and is slated for removal in `3.12`. Everyone who is still using it should either switch to the `setuptools` version (or another similar one) or even better replace those function calls altogether. Most, but not all, functions already have equivalents in the stdlib. https://www.python.org/dev/peps/pep-0632/#migration-advice
Longterm, the maintainers of `setuptools` also plan to discontinue `distutils`, but for the meantime it will continue to work.

**What is actually failing?**
As best as I can tell, only the astroid test case. Since the included `distutils` copy is functional identical to the stdlib it doesn't really matter too much that the `spec` path points towards the `stdlib` while the module is coming from `setuptools._distutils`.

**What failed previously?**
I only have an educated guess here. There seems to be a habit among certain packages to ship with their own version of distutils. In particular a (presumably old) version of virtualenv seem to have done so. It seems `astroid` had inferred the virtualenv version as `spec.location` instead of the stdlib one. That caused `ImportErrors` as the virtualenv version wasn't a complete copy.

**Is that still an issue?**
I don't know 🤷🏻‍♂️

**What should the goal here be for `astroid`?**
Pylint shouldn't report a false-positive `ImportError` for things like `distutils.util`.

**What path for `distutils` should astroid return?**
I'm uncertain about this one. If the goal is to prevent the false-positive and enable inference, any **"good"** copy of `distutils` would probably work. It doesn't really matter too much if it's the one from the stdlib or setuptools, since I don't think the main interface is going to change.

**What do we do now?**
Some options
1. Disable the test. As explained in the section before it doesn't matter to much that it's inferred as the stdlib version although technically it's using the setuptools one.
2. Patch the test. Instead of comparing it with the actual `distutils.__path__`, we could use a good python file to find the stdlib location and built the expected stdlib distutils path from there.
```py
import os
[f"{os.__file__.rsplit('/', 1)[0]}/distutils"]
```
3. Use the actual setuptools path if it's imported from setuptools. Basically the proposed fix. We try to import distutils twice, once from `distutils` and another time from `setuptools._distutils`. If both match, we choose to use the setuptools one for the path. Although probably the "most correct one", it requires us to do an expensive import twice.
4. YOLO it. We just remove all custom handling for `distutils` and see if we get any bug reports. I suspect / hope not but if we do, we at least would know what needs fixing.

**Is the first conditional for distutils still needed?**
🤷🏻‍♂️

**My preference?**
I believe all of the 4 options above should work. If we don't want to risk it all with (4), option (2) might be an idea. We can always come back to it in case we start getting bug reports eventually.

**Did I miss something?**
Likely 😄 At this point I don't even know if I've fully understood what's going on. It's also much too late to be writing such long posts 😴

Pierre-Sassoulas:
> it requires us to do an expensive import twice.

Or does the first import fail very fast because there is nothing to import, and the second one execute normally ?  If not, this is another thing we could explore in #1320, it's often that we guard import to be able to handle multiple version of the same library. 

> YOLO it. We just remove all custom handling for distutils and see if we get any bug reports.

At this point, it seems like an acceptable solution seeing distutils is deprecated and also seeing how often we fix bug in pylint but just removing custom handling and simplifying the code (Daniel will know what I mean by that 😄).

I added the regression test in pylint (https://github.com/PyCQA/pylint/pull/5620) and I'll do some tests to see if things are green without specific handling with setuptools 47 or 60. (4 tests cases not the whole 47 to 60 versions one by one with each version of the code). It should be noted also that the error in 73 was a fatal (pylint itself can't import distutils), not a false positive ``no-member`` or ``no-name-in-module``.

DanielNoord:
@cdce8p Thanks for this good sum-up. It's quite in-depth for that time of day 😄 

I think we can try and go ahead with option 4: we know what to revert to when things do eventually go wrong and the three of us (and others) are quite active lately so we should be able to push out a hotfix quickly if things really go south. (I won't respond tonight though probably 😄 🎆 ). Saving the import and deprecation warnings are nice added benefits which we would have to solve at some point anyway.

DanielNoord:
@cdce8p Is the latest commit what you had in mind? Please make a suggestion so I can get you as co-author on this commit, you did most of the work 😄

astroid/interpreter/_import/spec.py
```diff
@@ -162,11 +161,6 @@ def contribute_to_path(self, spec, processed):
                 if os.path.isdir(os.path.join(p, *processed))
             ]
         # We already import distutils elsewhere in astroid,
```
DanielNoord, line 163: For example, to remove this comment 😅

Pierre-Sassoulas: You can add `Co-authored-by: Marc Mueller <30130371+cdce8p@users.noreply.github.com>` when cleaning the squash message up :)

cdce8p, line 163:
```suggestion
```

cdce8p (changes requested):
Could you add a short changelog entry? Something along the lines of "removed custom distutils handling". Just so we have something for reference later.

DanielNoord:
> Could you add a short changelog entry? Something along the lines of "removed custom distutils handling". Just so we have something for reference later.

@cdce8p See latest commits 😄

ChangeLog
```diff
@@ -24,6 +24,8 @@ Release date: TBA
 
 * Require Python 3.6.2 to use astroid.
 
+* Removed custom ``distutils`` handling for resolving paths to submodules.
+
```
cdce8p, line 28:
```suggestion
* Removed custom ``distutils`` handling for resolving paths to submodules.

  Ref #1321

```

cdce8p (approved):
Thanks @DanielNoord 👍🏻
Let's do it then!

DanielNoord:
For future reference:
This regressed sadly... See: https://github.com/PyCQA/pylint/issues/5645

PR #1322, opened 2021-12-31
Change site packages look up to use sysconfig instead of distutils

DanielNoord:
## Steps

## Description

While is was reading up on `distutils` etc I thought I would try and see if we can remove all calls to it.

There are three calls to `distutils`, which should all be done in separate PRs because of the complexity and risk of errors. If you agree I think we should try and put them in `2.9.1` as well (I know, another delay) so we can combine all fixes to the YOLO-option and removal of `distutils` at the same time. (YOLO-option from #1321, that is.)

For the change in this PR:
[`get_python_lib`](https://docs.python.org/3/distutils/apiref.html#distutils.sysconfig.get_python_lib) returns the directory of `site-packages`. "Return the directory for either the general or platform-dependent library installation. If plat_specific is true, the platform-dependent include directory is returned; if false or omitted, the platform-independent directory is returned."
This functionality is "replaced" by [`sysconfig.get_path`](https://docs.python.org/3/library/sysconfig.html#sysconfig.get_path). By using both `purelib` and `platlib` we got both of them again.

## Type of Changes

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -46,6 +46,7 @@ import itertools
 import os
 import platform
 import sys
+import sysconfig
 import types
 from distutils.errors import DistutilsPlatformError  # pylint: disable=import-error
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


cdce8p:
I haven't looked at it yet. Just one comment about the timeline. I would suggest / prefer to put this into `2.10`. Yeah it's awesome that we'll remove the `distutils` dependency, but it isn't that urgent. Better to do one step at a time. That would also make debugging any issues much easier.

DanielNoord:
> I haven't looked at it yet. Just one comment about the timeline. I would suggest / prefer to put this into `2.10`. Yeah it's awesome that we'll remove the `distutils` dependency, but it isn't that urgent. Better to do one step at a time. That would also make debugging any issues much easier.

Okay! I have added it as the milestone to all PRs.

Pierre-Sassoulas:
We could also create a 2.9.2 release, there's almost 50 issues in 2.9.1, it's a lot for astroid.

Pierre-Sassoulas (approved):
🎉 Small step toward removing distutils, great !

cdce8p (approved):
Looks good! Thanks @DanielNoord! I'll review the others too, but we could start with merging this one.

PR #1323, opened 2021-12-31
Use ``sysconfig.get_path`` instead of ``distutils`` to find ``stdlib``

DanielNoord:
## Steps

## Description

2/3 of PRs mentioned in #1322

For the change in this PR:
[`get_python_lib`](https://docs.python.org/3/distutils/apiref.html#distutils.sysconfig.get_python_lib) will "If standard_lib is true, the directory for the standard library is returned rather than the directory for the installation of third-party extensions."
This is thus the parent directory of `site-packages`.
This functionality is "replaced" by [`sysconfig.get_path`](https://docs.python.org/3/library/sysconfig.html#sysconfig.get_path) with `stdlib`.

The commit that added this particular fix was https://github.com/PyCQA/astroid/commit/70b7afc4e70de165b4770bd9abda6a9b5440fdef

I have looked around but the closest I got to finding any information on the presumed bug was https://bugs.python.org/issue11320. I also found a number of older Python 2 and 3.6 < issues about `virtualenv` and `distutils`. I couldn't find the actual issue here..
However, I think `sysconfig` in combination with newer versions of `virtualenv` have fixed these underlying issues. The tests seems to pass anyway, so I would argue to go for the YOLO-option here as well: we need to get rid of `distutils` someday anyway 😅 

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,6 +20,13 @@ Release date: TBA
 
   Closes #1330
 
+* Use ``sysconfig`` instead of ``distutils`` to determine the location of python files
+  and packages.
+
+  Related pull requests: #1322, #1323, #1324
+  Closes #1282
+  Ref #1103
+
 What's New in astroid 2.9.4?
 ============================
 Release date: TBA
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,8 +20,8 @@ Release date: TBA
 
   Closes #1330
 
-* Use ``sysconfig`` instead of ``distutils`` to determine the location of python files
-  and packages.
+* Use ``sysconfig`` instead of ``distutils`` to determine the location of
+  python stdlib files and packages.
 
   Related pull requests: #1322, #1323, #1324
   Closes #1282
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -46,8 +46,8 @@ import itertools
 import os
 import platform
 import sys
+import sysconfig
 import types
-from distutils.errors import DistutilsPlatformError  # pylint: disable=import-error
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
 from typing import Dict, Set
 
@@ -66,22 +66,7 @@ else:
     PY_COMPILED_EXTS = ("so",)
 
 
-try:
-    # The explicit sys.prefix is to work around a patch in virtualenv that
-    # replaces the 'real' sys.prefix (i.e. the location of the binary)
-    # with the prefix from which the virtualenv was created. This throws
-    # off the detection logic for standard library modules, thus the
-    # workaround.
-    STD_LIB_DIRS = {
-        get_python_lib(standard_lib=True, prefix=sys.prefix),
-        # Take care of installations where exec_prefix != prefix.
-        get_python_lib(standard_lib=True, prefix=sys.exec_prefix),
-        get_python_lib(standard_lib=True),
-    }
-# get_python_lib(standard_lib=1) is not available on pypy, set STD_LIB_DIR to
-# non-valid path, see https://bugs.pypy.org/issue1164
-except DistutilsPlatformError:
-    STD_LIB_DIRS = set()
+STD_LIB_DIRS = {sysconfig.get_path("stdlib")}
 
 if os.name == "nt":
     STD_LIB_DIRS.add(os.path.join(sys.prefix, "dlls"))
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -53,7 +53,6 @@ from typing import Dict, Set
 
 from astroid.interpreter._import import spec, util
 
-
 if sys.platform.startswith("win"):
     PY_SOURCE_EXTS = ("py", "pyw")
     PY_COMPILED_EXTS = ("dll", "pyd")
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -53,6 +53,7 @@ from typing import Dict, Set
 
 from astroid.interpreter._import import spec, util
 
+
 if sys.platform.startswith("win"):
     PY_SOURCE_EXTS = ("py", "pyw")
     PY_COMPILED_EXTS = ("dll", "pyd")
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -39,8 +39,6 @@
 :var BUILTIN_MODULES: dictionary with builtin module names has key
 """
 
-# We disable the import-error so pylint can work without distutils installed.
-# pylint: disable=no-name-in-module,useless-suppression
 
 import importlib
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (approved):
LGTM ! Thank you for opening small MR for this. I think sys config was not used before because it was not available in python 3.5 [citation required], and that it's pretty safe to use it now that we support only > 3.6.2.

astroid/modutils.py
```diff
@@ -66,22 +66,7 @@
...
-        get_python_lib(standard_lib=True, prefix=sys.prefix),
-        # Take care of installations where exec_prefix != prefix.
-        get_python_lib(standard_lib=True, prefix=sys.exec_prefix),
-        get_python_lib(standard_lib=True),
-    }
-# get_python_lib(standard_lib=1) is not available on pypy, set STD_LIB_DIR to
-# non-valid path, see https://bugs.pypy.org/issue1164
-except DistutilsPlatformError:
-    STD_LIB_DIRS = set()
+STD_LIB_DIRS = {sysconfig.get_path("stdlib")}
```
cdce8p, line 69: Should we include `sysconfig.get_path("platstdlib")` too?
That would match `get_python_lib(standard_lib=True, prefix=sys.prefix)`.

```py
{
 'platstdlib': '/home/runner/work/astroid/astroid/venv/lib/python3.10',
 'stdlib': '/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10'
}
```

Usually `stdlib` should be enough. I just don't know if `virtualenv` or others do something unexpected with the lib files.

DanielNoord: I'm not sure? It was added in:
https://github.com/PyCQA/astroid/commit/b8c48252e042659e009cb95b87c2db9bb225d4fb
But there was not test or related issue that prompted the change?

cdce8p: It would had to be a virtualenv issue from the looks of it. Tbh I don't know if it has been fixed in the meantime. Maybe we should include in for now, just to be safe. The additional path doesn't really hurt.

If we really want to, we could revisit this in the future and just remove it then to see if it causes any new regressions. I just wouldn't suggest to do it now. It would make it more difficult to see if the current changes were correct.

DanielNoord:
By the way, for some reason the previous crash no longer happens after merging `main`. I'd like to investigate this.

cdce8p:
> By the way, for some reason the previous crash no longer happens after merging `main`. I'd like to investigate this.

Which crash?

DanielNoord:
See failing test in the first commit and the issue I opened against `python`. This wasn't actually a python issue but rather something to do with `astroid` adding itself to PATH. That seems to have been resolved I guess? But I haven't looked into what actually changed.

cdce8p:
> See failing test in the first commit and the issue I opened against `python`. This wasn't actually a python issue but rather something to do with `astroid` adding itself to PATH. That seems to have been resolved I guess? But I haven't looked into what actually changed.

It looks like the setuptools / distutils.version issue we dealt we back in December. So all good I think.
https://github.com/PyCQA/astroid/runs/4673501905?check_suite_focus=true#step:6:67

DanielNoord:
> > See failing test in the first commit and the issue I opened against `python`. This wasn't actually a python issue but rather something to do with `astroid` adding itself to PATH. That seems to have been resolved I guess? But I haven't looked into what actually changed.
> 
> It looks like the setuptools / distutils.version issue we dealt we back in December. So all good I think. https://github.com/PyCQA/astroid/runs/4673501905?check_suite_focus=true#step:6:67

Well.. That actually regressed 😓 See https://github.com/PyCQA/pylint/issues/5645.

Hmm looking back the failing test is not actually related. Let me have another look.

cdce8p:
> Well.. That actually regressed 😓 See [PyCQA/pylint#5645](https://github.com/PyCQA/pylint/issues/5645).
> 
> Hmm looking back the failing test is not actually related. Let me have another look.
> 
> Edit: I'm mixing up my PRs. See #1326, that's where I opened an incorrect issue report against Python.

😞 We did however remove the failing test case with it and it wasn't added back. That at least explains why this one isn't failing anymore. https://github.com/PyCQA/astroid/pull/1321/files#diff-fce15b6315286261c040399638ba825e411e7c0284c1006a50aa271f405beef6L75

It should be safe to move forward with this PR. The last open items are probably
* Rebase / merge upstream
* Add `platstdlib`: https://github.com/PyCQA/astroid/pull/1323#discussion_r795627776
* Short changelog entry

ChangeLog
```diff
@@ -20,6 +20,13 @@ Release date: TBA
 
   Closes #1330
 
+* Use ``sysconfig`` instead of ``distutils`` to determine the location of python files
+  and packages.
```
cdce8p, line 24:
```suggestion
* Use ``sysconfig`` instead of ``distutils`` to determine the location of
  python stdlib files and packages.
```

astroid/modutils.py
```diff
@@ -39,8 +39,6 @@
 :var BUILTIN_MODULES: dictionary with builtin module names has key
 """
 
-# We disable the import-error so pylint can work without distutils installed.
-# pylint: disable=no-name-in-module,useless-suppression
```
cdce8p, line 42:
```suggestion
```
Empty line

astroid/modutils.py
```diff
@@ -51,42 +49,21 @@
...
-# weird path manipulations in order to get to the
-# real distutils module.
-
-
 if sys.platform.startswith("win"):
     PY_SOURCE_EXTS = ("py", "pyw")
     PY_COMPILED_EXTS = ("dll", "pyd")
 else:
     PY_SOURCE_EXTS = ("py",)
     PY_COMPILED_EXTS = ("so",)
```
cdce8p, line 63:
```suggestion


```
An additional empty line might be useful here.

cdce8p (approved):
Just some last formatting suggesting. Looks good otherwise!
Thank you @DanielNoord for driving the effort to replace distutils!

astroid/modutils.py
```diff
@@ -51,18 +48,11 @@
...
 from typing import Dict, Set
 
 from astroid.interpreter._import import spec, util
 
-# distutils is replaced by virtualenv with a module that does
-# weird path manipulations in order to get to the
-# real distutils module.
-
-
 if sys.platform.startswith("win"):
```
cdce8p, line 56:
```suggestion

if sys.platform.startswith("win"):
```
Let's also keep the two empty lines here.

cdce8p: Oh, well. Seems like black didn't like that 🤷🏻

cdce8p:
Think it's ready now 🥳

PR #1324, opened 2021-12-31
Use ``sysconfig`` to determine ``stdlib`` paths for ``PyPy``

DanielNoord:
## Steps

## Description

3/3 of PRs mentioned in #1322

This was a little more difficult to fix. See the structure of their repository [here](https://foss.heptapod.net/pypy/pypy).

Basically on `PyPy` `sysconfig.get_path` is having trouble getting to the correct folders. It returns `'{installed_base}/lib-{implementation_lower}'` (this can be seen by passing `expand=False`). However, on `PyPy` the `stdlib` is spread over two different folders.

It finds the first folder with `sysconfig.get_path`. I believe this is because `sys.prefix/lib_pypy` is symlinked to `sys.prefix/lib-pypy`. Note that I think we can use only `sys.prefix` (the standard for `sysconfig.get_path`)  as the issue with `sys.prefix` overwriting in `virtualenvs` is the same as mentioned in #1323 and seems to be fixed from my testing. This allows to remove the `try...except` block.

For the `lib-python/3` folder we can't seem to get `sysconfig.get_path` to work. I don't think the function is intended to be able to resolve a `stdlib` spread over two different directories. I think there was something strange going on in `distutils` that made `get_python_lib(standard_lib=True, prefix=sys.base_prefix)` return the `lib-python` one, as `sys.base_prefix` is still `~/.pyenv/versions/pypy3.7-7.3.7`. Not sure why `distutils` resolves this into the `lib-python` directory with `get_python_lib` but it worked... To fix this we can simply tell `sysconfig` that it should look for another `implementation_lower` and then it will find the correct directory.
Note that I have not added the `2.7` version as I think we don't support it?

I'm not sure what the order of review will be, but in the last PR of these three we should also add a changelog and remove all remaining imports. That doesn't necessarily need to be this PR though.

## Type of Changes

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -53,6 +53,7 @@ import sysconfig
 import types
 from distutils.errors import DistutilsPlatformError  # pylint: disable=import-error
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
+from pathlib import Path
 from typing import Dict, Set
 
 from astroid.interpreter._import import spec, util
@@ -106,10 +107,8 @@ if os.name == "nt":
 
 if platform.python_implementation() == "PyPy":
     # PyPy stores the stdlib in two places: sys.prefix/lib_pypy and sys.prefix/lib-python/3
-    # sysconfig.get_path on PyPy only returns the first so we patch this manually.
-    # Note that sysconfig.get_path expands: '{installed_base}/lib-{implementation_lower}'
-    # sys.prefix/lib-pypy resolves into sys.prefix/lib_pypy
-    STD_LIB_DIRS.add(sysconfig.get_path("stdlib"))
+    # sysconfig.get_path on PyPy returns the first, but without an underscore so we patch this manually.
+    STD_LIB_DIRS.add(Path(sysconfig.get_path("stdlib").parent.joinpath("lib_pypy")))
     STD_LIB_DIRS.add(
         sysconfig.get_path("stdlib", vars={"implementation_lower": "python/3"})
     )
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -46,6 +46,7 @@ import itertools
 import os
 import platform
 import sys
+import sysconfig
 import types
 from distutils.errors import DistutilsPlatformError  # pylint: disable=import-error
 from distutils.sysconfig import get_python_lib  # pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
We do not support python 2.7 anymore the last version supporting it is pylint 1.9.5 (I don't know for astroid).

astroid/modutils.py
```diff
@@ -101,26 +102,15 @@
...
-    try:
-        # real_prefix is defined when running inside virtualenv.
-        STD_LIB_DIRS.add(os.path.join(sys.base_prefix, "lib_pypy"))
-    except AttributeError:
-        pass
-    del _root
+    # PyPy stores the stdlib in two places: sys.prefix/lib_pypy and sys.prefix/lib-python/3
+    # sysconfig.get_path on PyPy only returns the first so we patch this manually.
+    # Note that sysconfig.get_path expands: '{installed_base}/lib-{implementation_lower}'
+    # sys.prefix/lib-pypy resolves into sys.prefix/lib_pypy
```
cdce8p, line 108: Are you sure about that? I couldn't find any symlink. To be safe it might be better to add `lib_pypy` as well.

astroid/modutils.py
```diff
@@ -101,26 +102,15 @@
...
+    STD_LIB_DIRS.add(sysconfig.get_path("stdlib"))
```
cdce8p, line 109: `sysconfig.get_path("stdlib")` is already part of the set -> #1323.
Instead we should include `lib_pypy` here. Maybe something like this?
```py
from pathlib import Path

Path(sysconfig.get_path("stdlib").parent.joinpath("lib_pypy"))
```

Furthermore, we should probably do the same for `platstdlib`.

astroid/modutils.py
```diff
@@ -101,26 +102,15 @@
...
+    STD_LIB_DIRS.add(
+        sysconfig.get_path("stdlib", vars={"implementation_lower": "python/3"})
```
cdce8p, line 111: Looks good, but we should also do that for `platstdlib`.

cdce8p (changes requested):
Left some comments. Although we might get away without adding `lib_pypy`, the safe thing to do would be to match all paths to the current implementation (even if some aren't strictly necessary). As a side note: It shouldn't be necessary to remove `lib-pypy`.

This is `STD_LIB_DIRS` on Github actions (for pypy):
```py
{'/home/runner/work/astroid/astroid/venv/lib-python/3',
'/opt/hostedtoolcache/PyPy/3.6.12/x64/lib-python/3',
'/home/runner/work/astroid/astroid/venv/lib_pypy',
'/opt/hostedtoolcache/PyPy/3.6.12/x64/lib_pypy',
'/home/runner/work/astroid/astroid/venv/lib/python3.6',
'/home/runner/work/astroid/astroid/venv/lib64/python3.6'}
```

--
Regarding the merge order. I think it would make sense to do this one next and close with #1323 where the changelog entry could be added.

astroid/modutils.py
DanielNoord: Hmm you're right. It's weird as this is what `sysconfig.get_path` returns and the test pass, so I naturally assumed it was a symlink.

I'll change the comment and change it to your suggestion in the other comment.

astroid/modutils.py
DanielNoord: Are there actually any differences? The example in the docs doesn't show a difference:
https://docs.python.org/3/library/sysconfig.html
In the removed code we also don't seem to get `platstdlib`?

DanielNoord: Added!

cdce8p (approved):
LGTM! Let's do it then!

Pierre-Sassoulas (approved):
Thank you @DanielNoord !

PR #1331, opened 2022-01-05
Add support for attrs v21.3.0+

### LINKED ISSUE #1330: `brain_attrs.py` does not support attrs v21.3.0+
### Steps to reproduce

The [latest release](https://github.com/python-attrs/attrs/releases/tag/21.3.0) of attrs added a new module `attrs` and it seems that pylint does not recognize it.

```py
from typing import List

import attr
import attrs  # new in 21.3.0


@attr.define
class Old:
    items: List[str] = attr.field(factory=list)


@attrs.define
class New:
    items: List[str] = attrs.field(factory=list)


old = Old(["one", "two"])
assert old.items[0] == "one"  # no error here

new = New(["three", "four"])
assert new.items[0] == "three"  # E1136: Value 'new.items' is unsubscriptable (unsubscriptable-object)
```

Running `pylint` on this file causes an `unsubscriptable-object` error for only the `New` class.

### Current behavior
`unsubscriptable-object` for `attrs.field` but NOT `attr.field`

### Expected behavior

`attrs.field` (and friends) should be treated like `attr.field`

2.9.2

Below is also the version of pylint used to reproduce the error above.
```py
> pylint --version
pylint 2.12.2
astroid 2.9.2
Python 3.6.12 (default, Dec  2 2020, 12:31:12) 
[GCC Apple LLVM 12.0.0 (clang-1200.0.32.27)]
```

jacobbogdanov:
## Description

Since version [21.3.0](https://github.com/python-attrs/attrs/releases/tag/21.3.0)
you can now `import attrs` instead of just `import attr`.

This patch adds support so that astroid doesn't barf on classes created using `@attrs.define`.

NOTE: Not not all names, most notably `attr.ib` and `attr.s` are available in `attrs`.

```py
>>> import attrs
>>> [name for name in dir(attrs) if not name.startswith("_")]
['Attribute', 'Factory', 'NOTHING', 'asdict', 'assoc', 'astuple', 'cmp_using', 'converters', 'defi
ne', 'evolve', 'exceptions', 'field', 'fields', 'fields_dict', 'filters', 'frozen', 'has', 'make_c
lass', 'mutable', 'resolve_types', 'setters', 'validate', 'validators']
```

The only decorators in that list are `@frozen`, `@mutable` and `@define`. The rest are classes or functions for interacting with attrs classes.

I added support for `@define` because the first example from the release notes has it,

```py
from attrs import define

@define
class HelloWorld:
    modern: bool = True
```

but was unsure about whether `@mutable` and `@frozen` should be done too.

## Type of Changes

## Related Issue

Closes #1330

### DIFF
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -10,16 +10,22 @@ from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib", "attr.field", "field"))
+ATTRIB_NAMES = frozenset(
+    ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field")
+)
 ATTRS_NAMES = frozenset(
     (
         "attr.s",
         "attrs",
+        "define",
         "attr.attrs",
         "attr.attributes",
         "attr.define",
         "attr.mutable",
         "attr.frozen",
+        "attrs.define",
+        "attrs.mutable",
+        "attrs.frozen",
     )
 )
 
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -11,7 +11,7 @@ from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unkn
 from astroid.nodes.scoped_nodes import ClassDef
 
 ATTRIB_NAMES = frozenset(
-    ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field")
+    ("attr.ib", "attr.attrib", "attr.field", "attrs.field")
 )
 ATTRS_NAMES = frozenset(
     (
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -10,9 +10,7 @@ from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(
-    ("attr.ib", "attr.attrib", "attr.field", "attrs.field")
-)
+ATTRIB_NAMES = frozenset(("attr.ib", "attr.attrib", "attr.field", "attrs.field"))
 ATTRS_NAMES = frozenset(
     (
         "attr.s",
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -10,7 +10,9 @@ from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset( ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field"))
+ATTRIB_NAMES = frozenset(
+    ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field")
+)
 ATTRS_NAMES = frozenset(
     (
         "attr.s",
--- a/astroid/brain/brain_attrs.py
+++ b/astroid/brain/brain_attrs.py
@@ -10,7 +10,7 @@ from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attr.attrib", "attr.field", "attrs.field"))
+ATTRIB_NAMES = frozenset( ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field"))
 ATTRS_NAMES = frozenset(
     (
         "attr.s",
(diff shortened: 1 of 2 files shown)


astroid/brain/brain_attrs.py
```diff
@@ -10,16 +10,22 @@
...
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib", "attr.field", "field"))
+ATTRIB_NAMES = frozenset(
+    ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field")
+)
 ATTRS_NAMES = frozenset(
     (
         "attr.s",
         "attrs",
+        "define",
```
DanielNoord, line 20: I wonder if we should add this.

Sadly this brain does not infer the actual function behind the decorator but only looks at the name. However, if anybody uses `define` as a decorator name in non-`attr` code they will now also get picked up by this brain. Fort some reason we allowed `attrib` and `field` here, see https://github.com/PyCQA/astroid/pull/1187#pullrequestreview-763579685.

Do we want to add `define` as well @Pierre-Sassoulas? I can't imagine it being a very often used name for decorators, but it might create troubles down the line..

Pierre-Sassoulas: Well I don't like assuming that something is part of a lib only because of its name especially since we have inference capability in astroid. I know we're doing it for ``np`` in some place. But inference would permit to infer the real name of the origin lib. I don't know enough about brains to suggest something proper instead. What do you think @hippo91 ?

hippo91: @Pierre-Sassoulas i agree with you and @DanielNoord as well.
IMHO `define` should not be part of this set.

jacobbogdanov: okay, I've removed `@define` from the list of supported decorator names.

Pierre-Sassoulas:
If you want black to format properly small line, you need to remove the ending comma (it's [black magic comma](https://github.com/psf/black/issues/1288))

jacobbogdanov:
> If you want black to format properly small line, you need to remove the ending comma (it's [black magic comma](https://github.com/psf/black/issues/1288))

Thanks for the pointer! I think the issue is that I have an older `black` installed on my system. I omitted the formatting changes in my first commit but on the second I forgot to only stage the relevant portions

astroid/brain/brain_attrs.py
```diff
@@ -10,7 +10,9 @@
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib", "attr.field", "field"))
+ATTRIB_NAMES = frozenset(
+    ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field")
```
DanielNoord, line 14: I think we could discuss if `attrib` and `field` should be in this list but that ship has sailed I think.

Thanks for removing `define`! 👍

tests/unittest_brain.py
```diff
@@ -2211,6 +2211,68 @@ class Eggs:
             should_be_unknown = next(module.getattr(name)[0].infer()).getattr("d")[0]
             self.assertIsInstance(should_be_unknown, astroid.Unknown)
 
+    def test_attrs_transform(self) -> None:
```
DanielNoord, line 2214:
```suggestion
    def test_attrs_transform(self) -> None:
        """Test brain for decorators of the 'attrs' package.
        
        Package added support for 'attrs' a long side 'attr' in v21.3.0.
        See: https://github.com/python-attrs/attrs/releases/tag/21.3.0
        """
```

DanielNoord (approved):
Thanks @jacobbogdanov!

One recommendation for a little test docstring but rest LGTM!

astroid/brain/brain_attrs.py
Pierre-Sassoulas: I think you're right, it's also a problem.
```suggestion
    ("attr.ib", "attr.attrib", "attr.field", "attrs.field")
```
Suggestion done  in order to check if it's causing test fails. Imo we can most probably not infer the value as aliasing of attrs must be rare considering it's a small name, but we also *should* infer.

jacobbogdanov: looks like it breaks the existing `test_attr_transform` as well as the newly added `test_attrs_transform`.
```py
        @attrs
        class Bah:
            d = attrib(attr.Factory(dict))
```

Also there's a regression test in `pylint` that also relies on the existing `attrib`; https://github.com/PyCQA/pylint/blob/main/tests/functional/r/regression/regression_4439.py

DanielNoord: @Pierre-Sassoulas Let's revert then. Thanks @jacobbogdanov for checking this out.

astroid/brain/brain_attrs.py
```diff
@@ -10,7 +10,7 @@
 from astroid.nodes.node_classes import AnnAssign, Assign, AssignName, Call, Unknown
 from astroid.nodes.scoped_nodes import ClassDef
 
-ATTRIB_NAMES = frozenset(("attr.ib", "attrib", "attr.attrib", "attr.field", "field"))
+ATTRIB_NAMES = frozenset(("attr.ib", "attr.attrib", "attr.field", "attrs.field"))
```
Pierre-Sassoulas, line 13:
```suggestion
ATTRIB_NAMES = frozenset( ("attr.ib", "attrib", "attr.attrib", "attr.field", "attrs.field", "field"))
```

Pierre-Sassoulas:
Thanks a lot for the fix @jacobbogdanov ! It will be available in astroid 2.10 !

DanielNoord:
We forgot to add a changelog entry 😅

PR #1347, opened 2022-01-13
Provide actual literal_eval() function. Adapted from standard library's.

tristanlatr:
Provide actual `literal_eval()` function. Adapted from standard library's.
Use that function in `inference.py` instead of the dummy `_to_literal()` function.

## Type of Changes

| ✓   | :hammer: Refactoring (tiny bit)  |

### DIFF
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -113,6 +113,7 @@ class TooManyLevelsError(AstroidImportError):
 
 class AstroidSyntaxError(AstroidBuildingError):
     """Exception class used when a module can't be parsed."""
+    # this should probably be a SyntaxError subclass.
 
 
 class NoDefault(AstroidError):
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -336,7 +336,8 @@ def literal_eval(node_or_string: Union[str, nodes.NodeNG]) -> Any:
         node_or_string = node_or_string.value
     def _raise_malformed_node(node):
         msg = "malformed node or string"
-        if lno := getattr(node, 'lineno', None):
+        lno = getattr(node, 'lineno', None)
+        if lno:
             msg += f' on line {lno}'
         raise ValueError(msg + f': {node!r}')
     def _convert_num(node):
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -339,10 +339,11 @@ def literal_eval(node_or_string: Union[str, nodes.NodeNG]) -> Any:
 
     def _raise_malformed_node(node):
         msg = "malformed node or string"
-        lno = getattr(node, 'lineno', None)
+        lno = getattr(node, "lineno", None)
         if lno:
-            msg += f' on line {lno}'
-        raise ValueError(msg + f': {node!r}')
+            msg += f" on line {lno}"
+        raise ValueError(msg + f": {node!r}")
+
     def _convert_num(node):
         if not isinstance(node, nodes.Const) or type(node.value) not in (
             int,
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -22,7 +22,6 @@ Various helper utilities.
 
 
 from typing import Any, Union
-
 from astroid import bases, builder, manager, nodes, raw_building, util
 from astroid.context import CallContext, InferenceContext
 from astroid.exceptions import (
@@ -332,14 +331,14 @@ def literal_eval(node_or_string: Union[str, nodes.NodeNG]) -> Any:
     if isinstance(node_or_string, str):
         _node = builder.parse(node_or_string.lstrip(" \t")).body
         if len(_node) != 1:
-            raise ValueError(f"expected only one expression, found {len(_node)}")
+            raise ValueError(f"Expected only one expression, got {len(_node)}")
         node_or_string = _node[0]
     if isinstance(node_or_string, nodes.Expr):
         node_or_string = node_or_string.value
 
     def _raise_malformed_node(node):
         msg = "malformed node or string"
-        lno = getattr(node, "lineno", None)
+        lno = node.lineno
         if lno:
             msg += f" on line {lno}"
         raise ValueError(msg + f": {node!r}")
@@ -374,10 +373,16 @@ def literal_eval(node_or_string: Union[str, nodes.NodeNG]) -> Any:
         elif (
             isinstance(node, nodes.Call)
             and isinstance(node.func, nodes.Name)
-            and node.func.name == "set"
             and node.args == node.keywords == []
         ):
-            return set()
+            if node.func.name == "set":
+                return set()
+            if node.func.name == "tuple":
+                return tuple()
+            if node.func.name == "list":
+                return list()
+            if node.func.name == "dict":
+                return dict()
         elif isinstance(node, nodes.Dict):
             return {_convert(k): _convert(v) for k, v in node.items}
         elif isinstance(node, nodes.BinOp) and node.op in ("+", "-"):
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -21,7 +21,8 @@ Various helper utilities.
 """
 
 
-from astroid import bases, manager, nodes, raw_building, util
+from typing import Any, Union
+from astroid import bases, manager, nodes, raw_building, util, builder
 from astroid.context import CallContext, InferenceContext
 from astroid.exceptions import (
     AstroidTypeError,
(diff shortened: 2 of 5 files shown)


astroid/helpers.py
```diff
@@ -314,3 +316,77 @@ def object_len(node, context=None):
...
+        _node = builder.parse(node_or_string.lstrip(" \t")).body
+        if len(_node) != 1:
+            raise ValueError(f"expected only one expression, found {len(_node)}")
+        node_or_string = _node[0]
+    if isinstance(node_or_string, nodes.Expr):
+        node_or_string = node_or_string.value
+
+    def _raise_malformed_node(node):
+        msg = "malformed node or string"
+        if lno := getattr(node, "lineno", None):
```
Pierre-Sassoulas, line 342: We're going to end python 3.6 support soon, but 3.7 will still be used for quite some time.

tristanlatr: Fixed

astroid/exceptions.py
```diff
@@ -114,6 +114,8 @@ def __init__(
 class AstroidSyntaxError(AstroidBuildingError):
     """Exception class used when a module can't be parsed."""
 
+    # this should probably be a SyntaxError subclass.
```
DanielNoord, line 117: We also re-raise it from `TypeError`, `ValueError` and `LookupError`. So it doesn't seem to only be a `SyntaxError`?

astroid/helpers.py
```diff
@@ -314,3 +316,78 @@ def object_len(node, context=None):
...
+    sets, booleans, and None.
+
+    Can raise `AstroidSyntaxError` or `ValueError`.
+
+    Adapted from `ast.literal_eval`.
+    """
+    if isinstance(node_or_string, str):
+        _node = builder.parse(node_or_string.lstrip(" \t")).body
+        if len(_node) != 1:
+            raise ValueError(f"expected only one expression, found {len(_node)}")
```
DanielNoord, line 335:
```suggestion
            raise ValueError(f"Expected only one expression, got {len(_node)}")
```

astroid/helpers.py
```diff
@@ -314,3 +316,78 @@ def object_len(node, context=None):
...
+        node_or_string = _node[0]
+    if isinstance(node_or_string, nodes.Expr):
+        node_or_string = node_or_string.value
+
+    def _raise_malformed_node(node):
+        msg = "malformed node or string"
+        lno = getattr(node, "lineno", None)
```
DanielNoord, line 342: Would you be willing to check if this is indeed necessary for `astroid` nodes? I would assume we initialise `lineno` on every node, even on `Module` after #1262.

DanielNoord (changes requested):
Thanks @tristanlatr. However, I don't really see why it is necessary to maintain our own helper function? If I replace `helpers.literal_eval` in the tests for `ast.literal_eval` it works as well?

Wouldn't it also work if we accepted `str` for `_to_literal` as well and just passed the string instead of `node.as_string()` to `ast.literal_eval`?

astroid/helpers.py
```diff
@@ -314,3 +316,78 @@ def object_len(node, context=None):
...
+            return list(map(_convert, node.elts))
+        elif isinstance(node, nodes.Set):
+            return set(map(_convert, node.elts))
+        elif (
+            isinstance(node, nodes.Call)
+            and isinstance(node.func, nodes.Name)
+            and node.func.name == "set"
+            and node.args == node.keywords == []
+        ):
+            return set()
```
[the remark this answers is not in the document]
tristanlatr: `set()` is a special case because we can't create empty set other wise. 
I don;t think we should add this kind of support for `dict()` or `list()` since we can use `{}` and `[]`.

astroid/helpers.py
tristanlatr: For nodes created from scratch it might be necessary.

tests/unittest_helpers.py
```diff
@@ -259,6 +259,69 @@ class A(type): #@
...
+        self.assertEqual(helpers.literal_eval("3.25"), 3.25)
+        self.assertEqual(helpers.literal_eval("+3.25"), 3.25)
+        self.assertEqual(helpers.literal_eval("-3.25"), -3.25)
+        self.assertEqual(repr(helpers.literal_eval("-0.0")), "-0.0")
+        self.assertRaises(ValueError, helpers.literal_eval, "++6")
+        self.assertRaises(ValueError, helpers.literal_eval, "+True")
+        self.assertRaises(ValueError, helpers.literal_eval, "2+3")
+
+    def test_literal_eval_complex(self):
+        # Issue #4907
```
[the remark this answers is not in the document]
tristanlatr: Python's

astroid/exceptions.py
tristanlatr: I don't know, this is out of the scope if this PR anyway.

astroid/helpers.py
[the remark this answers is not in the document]
tristanlatr: I'll add a test

DanielNoord:
> > If I replace helpers.literal_eval in the tests for ast.literal_eval it works as well?
> 
> No, astroid nodes and ast nodes are not 100% compatible.

The tests you added do work with `ast.literal_eval`.

> > Wouldn't it also work if we accepted str for _to_literal as well and just passed the string instead of node.as_string() to ast.literal_eval?
> 
> Yes but the cool thing with `helpers.literal_eval` is that we can evaluate `NodeNG` without converting them to string before (which is hightly innefficient)!
>
> I'ts necessary if we don't want to convert back to string all `NodeNG` we might want to evaluate...

Hmm, that's indeed a benefit. However, I wonder if this warrants having to maintain this ourselves. We only use `_to_literal` in the inference of `nodes.Compare`. I'll leave this decision to others though. Perhaps it would be good if we had some actual data just how inefficient `_to_literal` is right now.

astroid/helpers.py
DanielNoord: That shouldn't be the case. The `init` of each node should handle setting `lineno`. If they do not so this would be a bug and should be fixed.

tests/unittest_helpers.py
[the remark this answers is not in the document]
DanielNoord: Perhaps add a reference to it then. Referring to issues from other projects/repo without any context can be confusing I think.

astroid/exceptions.py
DanielNoord: Wouldn't the comment then not also be out of scope?

astroid/helpers.py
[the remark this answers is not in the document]
DanielNoord: Ah yeah. Still, if we're going to maintain our own function shouldn't we add some obvious extra functionality while we're at it?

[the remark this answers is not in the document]
tristanlatr: I get your point, we could handle empty calls to `list`, `dict`, `tuple`, `frozenset`, if it's only that, I'm ok implementing it. 

But in general, I think we should stay the closest possible from `ast` module, and let users call this function with a possible inferred node.

tristanlatr:
I agree it could work by doing something like:
```python
def _dummy_literal_eval(node: Union[str, NodeNG]) -> Any:
    return ast.literal_eval(node if isinstance(node, str) else node.as_string())
```

But maybe some speed data comparaison will make you prefer my proposition: 
```python
>>> import timeit
>>> timeit.timeit('literal_eval(node)', setup='import json;from pathlib import Path;from astroid.helpers import literal_eval;from astroid.builder import extract_node;node = extract_node(repr(json.load(Path("./10kSamplejson.json").open("r"))))', number=20)
0.9220764310000007
>>> timeit.timeit('_dummy_literal_eval(node)', setup='import json;from pathlib import Path;from astroid.helpers import _dummy_literal_eval;from astroid.builder import extract_node;node = extract_node(repr(json.load(Path("./10kSamplejson.json").open("r"))))', number=20)
7.558820483999995
```
I'v used this big testing json: the JSON file is here https://docs.google.com/uc?export=download&id=1RAp9w12hZkmLt2vJcAjdT-UtC3KpFyq4

astroid/helpers.py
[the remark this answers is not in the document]
DanielNoord: Yeah, I think it makes sense to add those three as well. There is a big chance that we will run out of sync with the "upstream" version of this due to limited maintenance time. Makes sense imo to add some slight improvements while we're looking at it.

DanielNoord:
I'd like someone else's opinion on this. I'm -0.5 to it.

Obviously this would be a performance improvement and it might make sense to add this. However, since we're using this only in one place ourselves and we're basically copying the function from `ast` I keep wondering whether the improvement outweighs potentially missing out on any "upstream" fixes. Especially considering the clear lack of active (and more importantly experienced) maintainers we're currently experiencing.
I'm certainly not going to block this as I do see the benefit, but I'd like someone else's opinion on this matter before merging/finalising this!

(Thanks @tristanlatr for the ping, I should have made this clearer 😄)

Pierre-Sassoulas:
Opened #1361 to discuss this.

astroid/helpers.py
```diff
@@ -314,3 +315,84 @@ def object_len(node, context=None):
...
+            and node.args == node.keywords == []
+        ):
+            if node.func.name == "set":
+                return set()
+            if node.func.name == "tuple":
+                return tuple()
+            if node.func.name == "list":
+                return list()
+            if node.func.name == "dict":
+                return dict()
```
tristanlatr, line 385:
```suggestion
            if node.func.name == "set":
                return set()
            elif node.func.name == "tuple":
                return tuple()
            elif node.func.name == "list":
                return list()
            elif node.func.name == "dict":
                return dict()
            else:
                _raise_malformed_node(node)
```

cdce8p (changes requested):
I'm with @DanielNoord on this on. It isn't clear to me why this would be useful. Even if it provides a small performance benefit, the cost of maintaining it is higher IMO. Especially considering `literal_eval` is only used in the inference of `Compare` nodes. With that kind of limited scope, it's unlikely that we would see any performance improvements.

If you do have a use-case for it, I would be interested to hear it. Maybe as response here? https://github.com/PyCQA/astroid/issues/1361#issuecomment-1025275715

cdce8p:
I would like to close this one. As I explained in https://github.com/PyCQA/astroid/issues/1361#issuecomment-1025815781, I don't think the helper would provide value for astroid at the current time. Furthermore, it would increase the maintenance burden if a new Python version is released and something changes.

IMO every time there is an alternative in the stdlib, it's better to use that than to implement your own version for it. In this case the workaround with `ast.literal_eval(node.as_string())` is good enough.

DanielNoord:
Seeing as this discussion seems to have died down I think I'm going to close this. There seems to be somewhat of a consensus (from the maintainer's side) that this is not something we would want to add right now.

PR #1348, opened 2022-01-13
Visitor

### LINKED ISSUE #1342: Provide base `NodeVisitor` class
### Current behavior

The is no equivalent for `ast.NodeVisitor` and the `ast.NodeVisitor` class simply does not work with astroid classes (it relies on the `_fields` class varaible, which is named `_astroid_fields` in astroid). 

### Expected behavior

Provide a base visitor to walk an astroid tree.

Document the fact that visitor methods are lower case by convention. 

astroid version: 2.9.3

tristanlatr:
## Description

Provide `astroid.helpers.NodeVisitor` and `astroid.helpers.NodeTransformer` 
  adapted from `ast.NodeVisitor` and `ast.NodeTransformer`.

## Type of Changes

## Next step

The next step would be to use the base visitor class in the `AsStringVisitor` and maybe in `TransformVisitor` too. 

## Related Issue

Closes #1342

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,8 @@ What's New in astroid 2.10.0?
 =============================
 Release date: TBA
 
+* Provide `astroid.helpers.NodeVisitor` and `astroid.helpers.NodeTransformer` 
+  adapted from `ast.NodeVisitor` and `ast.NodeTransformer`.
 
 What's New in astroid 2.9.4?
 ============================
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -21,6 +21,7 @@ Various helper utilities.
 """
 
 
+from typing import Any, Iterable, Union, Tuple
 from astroid import bases, manager, nodes, raw_building, util
 from astroid.context import CallContext, InferenceContext
 from astroid.exceptions import (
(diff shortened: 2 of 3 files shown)


cdce8p (commented):
I'm not sure that it would make sense to add those. `astroid` already implements a way to transform nodes with `register_transform`. The `NodeTransformer` would essentially be a duplicate with less specific features.

I mentioned in [already](https://github.com/PyCQA/astroid/issues/1361#issuecomment-1025815781), but IMO astroid isn't really designed for that kind of operations. At least, I don't think we should support them officially. The main focus is still to provide an augmented AST for pylint in particular. For use cases other than inference, it's thus probably better to use the builtin `ast` module as it's more versatile.

As for the general `NodeVisitor`, although it might be useful in some limited cases, it's completely isolated from the rest. For pylint we already have something similar with the `ASTWalker` but more customized so we are unlikely to use it there.

Unless there is a really compelling reason for them, I would suggest to close the PR.

cdce8p (changes requested):
See comment above 🔝

tristanlatr:
Please look a real use use case of node transformer here: https://github.com/tristanlatr/pydocspec/blob/65a3cfa1564af8e0bee2a097d76b11c72ff557cc/pydocspec/astroidutils.py#L495

PR #785, opened 2020-05-07
handle deprecation warnings when walking module members

CrepeGoat:
## Steps

- [ ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [ ] Write a good description on what the PR does.

## Description
This change prevents `DeprecationWarning`s that arise from dynamically walking a module's members from propagating. Specifically, it changes the member walking portion to treat warnings as errors, and specifically handles `DeprecationWarning`s to halt their propagation.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #770

PR #1352, opened 2022-01-16
Use ``warnings.simplefilter`` instead of ``filterwarnings``

DanielNoord:
## Steps

## Description

See documentation for [warnings](https://docs.python.org/3/library/warnings.html).
This was added in https://github.com/PyCQA/astroid/pull/785/files. However, a little profiling of `pylint` shows that this function is actually quite "expensive", accounting for 1.75% of runtime on a empty file.
You can that a smaller share of `object_build` is spent in `warnings` in the screenshot:
`(1,74/36,77) > (1,119/35,27)`.

The gain is minor but when I found out it didn't make sense not to make a little PR.

Schermafbeelding 2022-01-16 om 17 07 10
Schermafbeelding 2022-01-16 om 17 10 12

## Type of Changes

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -340,7 +340,7 @@ class InspectBuilder:
         for name in dir(obj):
             try:
                 with warnings.catch_warnings():
-                    warnings.filterwarnings("error")
+                    warnings.simplefilter("error")
                     member = getattr(obj, name)
             except (AttributeError, DeprecationWarning):
                 # damned ExtensionClass.Base, I know you're there !


Pierre-Sassoulas (approved):
For reference, the doc:

> warnings.filterwarnings(action, message='', category=Warning, module='', lineno=0, append=False)

> Insert an entry into the list of warnings filter specifications. The entry is inserted at the front by default; if append is true, it is inserted at the end. This checks the types of the arguments, compiles the message and module regular expressions, and inserts them as a tuple in the list of warnings filters. Entries closer to the front of the list override entries later in the list, if both match a particular warning. Omitted arguments default to a value that matches everything.

> warnings.simplefilter(action, category=Warning, lineno=0, append=False)

> Insert a simple entry into the list of warnings filter specifications. The meaning of the function parameters is as for filterwarnings(), but regular expressions are not needed as the filter inserted always matches any message in any module as long as the category and line number match.

We don't need regex, so we can safely switch to simple filter. Nice catch !

PR #1353, opened 2022-01-17
Add typing to ``__str__``, ``__repr__`` and ``_repr_name``

DanielNoord:
## Steps

## Description

Puts `mypy astroid` under 90 errors.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -174,7 +174,7 @@ class NodeNG:
         context.inferred[key] = tuple(results)
         return
 
-    def _repr_name(self):
+    def _repr_name(self) -> str:
         """Get a name for nice representation.
 
         This is either :attr:`name`, :attr:`attrname`, or the empty string.
@@ -186,7 +186,7 @@ class NodeNG:
             return getattr(self, "name", "") or getattr(self, "attrname", "")
         return ""
 
-    def __str__(self):
+    def __str__(self) -> str:
         rname = self._repr_name()
         cname = type(self).__name__
         if rname:
@@ -212,7 +212,7 @@ class NodeNG:
             "fields": (",\n" + " " * alignment).join(result),
         }
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         rname = self._repr_name()
         if rname:
             string = "<%(cname)s.%(rname)s l.%(lineno)s at 0x%(id)x>"


Pierre-Sassoulas (approved):
90 errors only ? Is the goal of fully typed astroid for 2.10 is reachable according to you ?

DanielNoord:
> 90 errors only ? Is the goal of fully typed astroid for 2.10 is reachable according to you ?

No! These are just the errors you get when trying to run `mypy` without any arguments. `mypy --strict` returns 2437 errors.
If we solve these 90 we can turn `mypy` on in `pre-commit` though, which should help!
However, I'd be surprised if we would get these 90 fixed in `2.11` honestly. There are some pretty big issues still open, most of which have to do with non-optional attributes being typed as optional.

Pierre-Sassoulas:
> mypy --strict returns 2437 errors.

Approximately the same than for pylint :smile:  I hope that doing astroid first will make pylint easier though. 

It will help to be able to activate mypy if only so we don't introduce knew problem unknowingly. Maybe we should add type ignores on existing issues so we can activate mypy right now anyway ? With less than a hundred error left it seems feasible.

DanielNoord:
I don't think we should do that. We would ignore important issues that might require refactoring. If we then start increasing typing build on incorrect code we can create problems for our future selves.

To give a rough summary I think of the 90 error still open:
20 are easy and straightforward issues where sets or dicts don't have any typing.
30 are caused by 2/3 main issues where a method is defined outside of the class file and mypy warns that the method is not recognised on the class. See #1249. It takes quite a bit of effort to get that right, but should be doable even by less experienced `astroid` developers.
20 are caused by attributes being defined as `None` while they should never be `None`. See #1250.
20 are caused by inheritance and design problems that will require some refactoring.

It are especially those last two categories that are difficult to tackle.

Pierre-Sassoulas:
Thanks for the recap !

> We would ignore important issues that might require refactoring. If we then start increasing typing build on incorrect code we can create problems for our future selves.

I'm not advocating for ignoring them ad infinitam, just right now and until we fix them all, so we don't introduce new issue unknowingly because mypy does not check anything unless you launch it yourself :)

Inheritance and design problem in particular sounds like it will take some time to get fixed, so a solution to be able to partially run mypy in the meantime would be beneficial imo.

DanielNoord:
We could write a little script which compares `mypy` output on `main` with the commit in a PR. That might solve that issue in the meantime.

I understand why we would want to disable for now, but the current errors point to actual problems/issues with the current typing. If we disable them people might start adding typing that builds upon these ignored errors. If we then try to resolve these temporary disables, the problem has only become larger. That would only increase the effort to completely type `astroid`. I'd be very hesitant to add ignores for that reason.

Pierre-Sassoulas:
> We could write a little script which compares mypy output on main with the commit in a PR.

This makes sense and shouldn't be too hard to implement (not much more than adding ignores). But how does it prevent that "people might start adding typing that builds upon these ignored errors." ?

DanielNoord:
It doesn't really, but it prevents the issue where the ignore seems to indicate "we acknowledge this is wrong" rather than "we haven't gotten too this yet".

DanielNoord:
In any case, `astroid` development is rather slow so I don't think we'll see much benefit from enabling `mypy` now and going through the effort to add ignores or this trick.

Pierre-Sassoulas:
>  the ignore seems to indicate "we acknowledge this is wrong" rather than "we haven't gotten too this yet".

This argument make a lot of sense, I'm convinced.

PR #1355, opened 2022-01-17
Rearrange and type attribute definition of ``nodes.Module``

DanielNoord:
## Steps

## Description

The way these were defined and documented didn't make a lot of sense.
A lot of attributes were initialised as `None` only to then be set by the `__init__`. The following comment therefore made little sense: `# attributes below are set by the builder module or by raw factories`.

The change is quite large, sorry about that, but it's basically just a lot moving docstrings around. The real change is the added typing for `globals` and removing all the unnecessary `Nones`.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -476,7 +476,6 @@ class Module(LocalsDictNodeNG):
         """A map of the name of a global variable to the node defining the global."""
 
         self.locals = self.globals = {}
-        """A map of the name of a local variable to the node defining the local."""
 
         self.body: Optional[List[node_classes.NodeNG]] = []
         """The contents of the module."""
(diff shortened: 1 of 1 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -475,53 +428,61 @@ class Module(LocalsDictNodeNG):
...
+
         self.pure_python = pure_python
+        """Whether the ast was built from source."""
+
+        self.globals: Dict[str, List[node_classes.NodeNG]]
+        """A map of the name of a global variable to the node defining the global."""
+
         self.locals = self.globals = {}
-        """A map of the name of a local variable to the node defining the local.
+        """A map of the name of a local variable to the node defining the local."""
```
DanielNoord, line 479:
```suggestion
```

Unnecessary as defined by parent class

Pierre-Sassoulas (approved):
LGTM. There were some typo in the old docstring, like ``str(str)```. It's nice to make it explicit and enforceable by mypy.

PR #1357, opened 2022-01-17
Fix incorrect typing of ``doc`` attribute of ``Module``

DanielNoord:
## Steps

## Description

This was shown by #1355. Made no sense to change how the function itself works, so this seems best way to fix this.

## Type of Changes

## Related Issue


DanielNoord:
Ah, turns out #1355 was wrong. Typing was incorrect. I'll change this PR to become a PR that fixes the mistake.

Pierre-Sassoulas (approved):

PR #1359, opened 2022-01-17
Use ``_extract_single_node`` in ``brain_re``

DanielNoord:
## Steps

## Description

Should probably merge #1358 first as one of its commit introduce the new method.

## Type of Changes

## Related Issue


Pierre-Sassoulas (changes requested):
Could you rebase, please ? It seems that even if we merged the function before, but the diff still show it, I'm afraid it could be duplicated.

Pierre-Sassoulas (approved):

PR #1360, opened 2022-01-18
Fix inference of ``self`` in a list or tuple within bin. ops.

DanielNoord:
## Steps

## Description

Will close this issue PyCQA/pylint#4826.

The main change is that the inference is no longer bound to the instance node on which the binary operation is performed.
Without this change it kept inferring `a` in `print(a)` as being as `list` because that was the bound node as set by the binary operation inference method.

I've also added typing which should help troubleshoot future problems and explained what this method is actually referring to.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,10 @@ What's New in astroid 2.10.0?
 =============================
 Release date: TBA
 
+* Fixed inference of ``self`` in binary operations in which ``self``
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Very elegant one line fix !

DanielNoord:
Tests in `pylint` have been added as well!

Issue #1365, opened 2022-01-20
`inference_tip` suppresses transformation of built-in `property`


areveny:
### Steps to reproduce

1. Run Astroid on a file with a call to `property`
2. The replacement/transformation does not seem to occur. I've tried putting a breakpoint in `brain_builtin_inference.infer_property` but it never gets hit.

```
class Something:
    def getter():
        return 5
    asd = property(getter)
```

### Current behavior
This follows from a flaky error we were seeing on https://github.com/PyCQA/pylint/pull/5610 where we would semi-randomly get a `Property` node without `args`.

In `brain_builtin_inference.py`, all the transforms are wrapped in `inference_tip`. `inference_tip` does not replace `node` with the outcome of the `transform` function, it just stores an inference in `node._explicit_inference`. There is a `@wrapt.decorator` annotation on`inference_tip_cached` that seems important for grasping this issue but I don't understand it well enough yet.

The returned node from `inference_tip.inference_tip.transform` replaces the original in `transforms._transform`, but we return the original node so there is no replacement.
 
### Expected behavior
The function call to the builtin `property` should be replaced with an `objects.Property` node.

The flaky test in the linked Pylint issue suggests that sometimes we do transform into `Property` but I can't figure out what triggers those cases. Possibly something to do with the inference tip cache.

### `python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"` output
2.10.0-dev0

DanielNoord:
@areveny What platform are you developing on? For me `infer_property` gets called every time when I run `pylint` over you code example.. I'm on Mac.

areveny:
I'm on Mac M1.

DanielNoord:
```diff
diff --git a/astroid/brain/brain_builtin_inference.py b/astroid/brain/brain_builtin_inference.py
index 0bf352607..3ea683a12 100644
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -553,6 +553,7 @@ def infer_property(node, context=None):
     This only infers the output of `property`
     call, not the arguments themselves.
     """
+    print("CALLED")
     if len(node.args) < 1:
         # Invalid property call.
         raise UseInferenceDefault
```

```console
❯ cat test.py
class Something:
    def getter():
        return 5

    asd = property(getter)
❯ pylint test.py
CALLED
************* Module test
test.py:2:4: E0211: Method has no argument (no-method-argument)

[...]

PR #1366, opened 2022-01-20
Fix builtin inference on property not including function arguments

areveny:
## Steps

## Description

This ensures that when we transform a call to the builtin `property` into an `objects.Property`, we correctly postinitialize its function arguments.

Not doing so leads to a crash in https://github.com/PyCQA/pylint/pull/5610. 

This PR can be merged, but testing it is dependent on #1365 because this code is not being called consistently so we can't write a test for it.

But sometimes the code is called, which is why it is needed for the Pylint issue above.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,8 @@ What's New in astroid 2.10.0?
 =============================
 Release date: TBA
 
+* Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which 
+  resulted in instance arguments missing on these nodes.
 
 What's New in astroid 2.9.4?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,7 +6,7 @@ What's New in astroid 2.10.0?
 =============================
 Release date: TBA
 
-* Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which 
+* Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Great !

ChangeLog
```diff
@@ -6,6 +6,7 @@ What's New in astroid 2.10.0?
 =============================
 Release date: TBA
 
+* Fixed builtin inferenence on `property` calls not including function arguments
```
DanielNoord, line 9:
```suggestion
* Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which
   resulted in instance arguments missing on these nodes.
```

DanielNoord (changes requested):

areveny:
Added a unit test to validate postinit on the inferred `Property` by checking that the `args` attribute that is giving us trouble is initialized.

tests/testdata/python3/data/logging/__init__.py
```diff
@@ -0,0 +1,2261 @@
+# Copyright 2001-2019 by Vinay Sajip. All Rights Reserved.
```
DanielNoord, line 1: Should this be in here?

DanielNoord (approved):
Looks good, besides the file that I think was added by mistake.

Thanks @areveny!

areveny:
Ah, dang. I merged without squashing. I'll check that it is squashed next time.

PR #1368, opened 2022-01-24
Get argument infos from method descriptors

tristanlatr:
## Steps

## Description

This PR adds argument infos from method descriptors, meaning from functions defined in python C-extensions. 

## Type of Changes


astroid/raw_building.py
```diff
@@ -186,14 +186,27 @@ def object_build_class(node, member, localname):
...
     kwonlyargs = []
+
+    try:
+        signature = inspect.signature(member)
+    except ValueError:
+        # function either has an invalid text signature or no signature
+        # at all. We distinguish between the two by looking at the
+        # __text_signature__ attribute
+        if getattr(member, "__text_signature__", None) is not None:
+            f"Cannot parse __text_signature__ signature of {member}"
```
tristanlatr, line 207: Can we log a warning from astroid?

tests/unittest_builder.py
```diff
@@ -790,5 +797,53 @@ def test_parse_module_with_invalid_type_comments_does_not_crash():
...
+        spec = importlib.util.spec_from_file_location(module_full_name, path)
+        if spec is None:
+            raise RuntimeError(
+                f"Cannot find spec for module {module_full_name} at {path}"
+            )
+        py_mod = importlib.util.module_from_spec(spec)
+        loader = spec.loader
+        assert isinstance(loader, importlib.abc.Loader), loader
+        loader.exec_module(py_mod)
+        return py_mod
```
tristanlatr, line 816: There is probably a utility in astroid to do this...

astroid/raw_building.py
Pierre-Sassoulas: We could but do not, generally we raise ``InferenceError`` see this issue in flake8 where a log take 6% of the exec time even when not displayed : https://github.com/PyCQA/flake8/pull/1544

tests/unittest_builder.py
```diff
@@ -790,5 +797,53 @@ def test_parse_module_with_invalid_type_comments_does_not_crash():
     assert isinstance(node, nodes.Module)
 
 
+def test_c_module_text_signature():
```
tristanlatr, line 800: I think this test cannot pass on pypy, since they do not support CPython extensions natively.

astroid/raw_building.py
```diff
@@ -224,13 +250,18 @@ def object_build_datadescriptor(node, member, name):
...
-        getattr(member, "__name__", None) or localname, doc=member.__doc__
+        name=getattr(member, "__name__", None) or localname,
+        args=args,
+        posonlyargs=posonlyargs,
+        defaults=defaults,
+        doc=member.__doc__,
+        kwonlyargs=kwonlyargs,
     )
-    # set node's arguments to None to notice that we have no information, not
-    # and empty argument list
```
tristanlatr, line 232: To still be inline with this, I should set `None` to args if we get a `ValueError` calling `signature.signature()`.

astroid/raw_building.py
tristanlatr: Should we raise `InferenceError` here though? 
It's not technically inference what we're doing here.

tristanlatr:
Also it seems that the python 3.10 tests are consistently failling the following three tests:
```
tests/unittest_brain.py::SixBrainTest::test_attribute_access 
tests/unittest_brain.py::test_http_client_brain
tests/unittest_regrtest.py::NonRegressionTests::test_ssl_protocol
```

I'll looks into that, but I'de like to hear opinions in favour of such features before I spend too much time on it and at the end it does not fit in your agenda..

Thanks.

Still Todo:
- Change the Note in the docstring of Arguments.args: https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.Arguments.args saying that we now get infos from function defined in c-extensions. 
- Skip c-module test on pypy
- `args` should be None if we get a ValueError calling signature.signature()

tristanlatr:
Up!

DanielNoord:
@tristanlatr Could you resolve the merge conflicts?

tristanlatr:
@DanielNoord,

Conflicts are resolved, but I've not yet lint the code. There are still a couple points to discuss first. 

Thanks

astroid/raw_building.py
```diff
@@ -179,13 +179,29 @@ def object_build_class(node, member, localname):
     return _base_class_object_build(node, member, basenames, localname=localname)
 
 
-def object_build_function(node, member, localname):
-    """create astroid for a living function object"""
-    signature = inspect.signature(member)
+def _get_args_info_from_callable(
+    member: Callable,
```
DanielNoord, line 183: Do we know more about this `Callable`?

astroid/raw_building.py
DanielNoord: Yeah, setting back to `None` seems the best option here. That mimics what we currently do.

astroid/raw_building.py
DanielNoord: Yes! Let's do that!

tests/testdata/python3/data/c_module_invalid_text_signature/setup.py
```diff
@@ -0,0 +1,8 @@
+from setuptools import setup, Extension
+
+cmodule = Extension("mymod.base", sources=["mymod/base.c"],)
+setup(
+    name="mymod",
+    ext_modules=[cmodule],
+    packages=['mymod'],
+)
```
DanielNoord, line 8:
```suggestion
)

```

tests/unittest_builder.py
DanielNoord: We should be able to tell pytest to skip it right?

Although I must confess I don't know if this is actually true. Did you find any documentation about this?

DanielNoord (commented):
I find it difficult to review this as I have absolutely zero knowledge of C-extensions or C in general. I'm not sure we actually have anybody in the maintainers team who does..

At least I answered one question 😅

astroid/raw_building.py
tristanlatr: No, it can be any python callable.

tests/unittest_builder.py
tristanlatr: > We should be able to tell pytest to skip it right?

Yes we can

> Although I must confess I don't know if this is actually true. Did you find any documentation about this?

Yes I've read that from pypy's documentation, but I can't find the link now. But here's what the [pypy wiki](https://en.wikipedia.org/wiki/PyPy) says: "There is a compatibility layer for CPython C API extensions called CPyExt, but it is incomplete and experimental.". So I'm positive that, by default, pypy has no support for c-modules built for cpython.

DanielNoord: Let's skip then!

tristanlatr:
There is still the issue of the failing tests...
I'll work on this when I have some time. 

This is definitely a required feature for pydoctor to adopt astroid, because we offer support for c-modules.
I hope in can be merged once I've fixed the issues.

Talk to you later,

Tristan

tests/unittest_builder.py
tristanlatr: I assume it's OK to include these private helpers, it's only going to be used in the tests. And I didn't see anything the does such imports in astroid, but maybe I did not look at the right place.

DanielNoord:
@tristanlatr I have done some exploring in the module and am now more confident in merging changes here. Please let me know if you have any further questions. We should be cautious about changing stuff here, but I'm willing to help get this merged.

One thing I believe would help is if you separated the creation of the new method from also using it in `object_build_methoddescriptor`. That makes this the diff smaller and more reviewable.

astroid/raw_building.py
tristanlatr: Can you more explicit and maybe suggest another typing for the function ?

astroid/raw_building.py
```diff
@@ -179,13 +179,29 @@ def object_build_class(node, member, localname):
...
+
+    try:
+        signature = inspect.signature(member)
+    except ValueError:
+        # function either has an invalid text signature or no signature
+        # at all. We distinguish between the two by looking at the
+        # __text_signature__ attribute
+        if getattr(member, "__text_signature__", None) is not None:
+            f"Cannot parse __text_signature__ signature of {member}"
+        signature = inspect.Signature()
```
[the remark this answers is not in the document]
tristanlatr: Yes you are right, this was part of the stuff that I needed to change.

tristanlatr:
Hi @DanielNoord,

I don’t think we’re going to use astroid with pydoctor in the end, turns out to be much more complicated than I thought because of divergences compared to the standard library.

So I probably won’t update this PR anytime soon. If one likes this feature please fell free to build off my changes and open another PR.

Cheers,

DanielNoord:
> Hi @DanielNoord,
> 
> I don’t think we’re going to use astroid with pydoctor in the end, turns out to be much more complicated than I thought because of divergences compared to the standard library.
> 
> So I probably won’t update this PR anytime soon. If one likes this feature please fell free to build off my changes and open another PR.
> 
> Cheers,

Thanks for the headups! I'll likely cherry pick some changes from this PR and see what I can merge without too much trouble.

DanielNoord:
@tristanlatr I have opened https://github.com/PyCQA/astroid/pull/1593 based on your work and some previous cherry picking. I'm going to close this PR as the discussion is quite long and most issues have been resolved in that PR.

Would you be willing to do a review of that PR? I think most issue have been addressed, but considering I have no experience with C I'd like some other reviews before undrafting it.

PR #1370, opened 2022-01-26
Fix crash on ``Super.getattr`` for previously uninferable attributes

DanielNoord:
## Steps

## Description

Will close when merged: #PyCQA/pylint#5679

Pretty difficult to replicate this, but whenever the result of the inference was `Uninferable` we did not add that to the cache. Instead we added `[]` to the cache as this is how `results` is initialised. On a next try of inference of that node we see that the node is in the cache and yield from the empty list instead of yielding `Uninferable`. This seems only relevant whenever we get more than 100+ inference tries for a single node, which I think was happening with the `sqlalchemy` node.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,9 +10,8 @@ Release date: TBA
   resulted in instance arguments missing on these nodes.
 
 * Fixed a crash on ``Super.getattr`` when the attribute was previously uninferable due to a cache
-  limit size.
-  This limit can be hit when the inheritance pattern of a class (and therefore of the ``__init__``
-  attribute) is very large.
+  limit size. This limit can be hit when the inheritance pattern of a class (and therefore of the
+  ``__init__`` attribute) is very large.
 
   Closes PyCQA/pylint#5679
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,7 +14,7 @@ Release date: TBA
   This limit can be hit when the inheritance pattern of a class (and therefore of the ``__init__``
   attribute) is very large.
 
-  Closes #PyCQA/pylint#5679
+  Closes PyCQA/pylint#5679
 
 What's New in astroid 2.9.4?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,13 @@ Release date: TBA
 * Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which
   resulted in instance arguments missing on these nodes.
 
+* Fixed a crash on ``Super.getattr`` when the attribute was previously uninferable due to a cache
+  limit size.
+  This limit can be hit when the inheritance pattern of a class (and therefore of the ``__init__``
+  attribute) is very large.
+
+  Closes #PyCQA/pylint#5679
+
 What's New in astroid 2.9.4?
 ============================
 Release date: TBA
--- a/tests/unittest_regrtest.py
+++ b/tests/unittest_regrtest.py
@@ -24,11 +24,13 @@ import unittest
 
 import pytest
 
-from astroid import MANAGER, Instance, nodes, parse, test_utils
+from astroid import MANAGER, Instance, bases, nodes, parse, test_utils
 from astroid.builder import AstroidBuilder, extract_node
 from astroid.const import PY38_PLUS
+from astroid.context import InferenceContext
 from astroid.exceptions import InferenceError
 from astroid.raw_building import build_module
+from astroid.util import Uninferable
 
 from . import resources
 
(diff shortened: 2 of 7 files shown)


tests/unittest_regrtest.py
```diff
@@ -399,5 +401,33 @@ class Another(subclass):
...
+    super_node = next(init_attr_node.expr.infer())
+
+    # Arbitrarily limit the max number of infered nodes per context
+    InferenceContext.max_inferred = -1
+    context = InferenceContext()
+
+    # Try to infer 'object.__init__' > because of limit is impossible
+    for _ in bases._infer_stmts([init_object_node], context, frame=super):
+        pass
+
```
Pierre-Sassoulas, line 425: Shouldn't there also be an assert here ?

Pierre-Sassoulas (commented):
Looks very good already !

tests/unittest_regrtest.py
DanielNoord: We could, it's just here to populate the `context` like it would normally. You can make a suggestion if you want to add it. I think `assert _ == Uninferable` should pass.

Pierre-Sassoulas: I thought it was a test in itself because the  ``max_inferred`` value is changed before ? Why is there a particular value for max_inferred in the context if we only want to populate context ?

DanielNoord: When `max_inferred` is lower than the current number of `inferred` values in a context when return `Uninferable`: if we need to infer 101 nodes to understand what node A is then we let node A be `Uninferable`. This limit is reached in `pylint` for this particular inheritance pattern, but I couldn't get it to hit for `astroid` when only inferring the base files. (When running `pylint` other values get stored in the cache as well which is why we hit it sooner).
To make this bug consistent I set the limit to -1 so that we will always hit it, thereby replicating the behaviour in `pylint`.

Because we hit the limit we store yield `Uninferable` for node A and store this in the cache. However, without this commit we wouldn't store `Uninferable` for node A, but rather `[]`. Thus when we did another lookup for node A we would yield from `[]` instead of from `[Uninferable,]`. This made it so that while we had node A as a key in the inference cache we didn't actually have any values for that key, which caused the exception to be raised.

Pierre-Sassoulas: Ok, this is what I understood as well. So would it make sense that the max_inferred limit worked as we thought it would and that we do yield an ``Uninferable`` ?

DanielNoord: Yes! `for _ in bases._infer_stmts([init_object_node], context, frame=super):` will work irrespective of this change. It is the `getattr` call that is/was causing problems. Because `getattr` also calls `_infer_stmts` it tries to use the cache and then yields from `[]`.

We could add the `assert _ == Uninferable` test, but it is not necessarily related to the actual bug here, which is why I originally excluded it. `_infer_stmts` on a "101-inference node" called once works fine, `_infer_stmts` called twice on a "101-inference node" is broken.

tests/unittest_regrtest.py
```diff
@@ -399,5 +401,33 @@ class Another(subclass):
...
+    init_object_node = module.body[-1].mro()[-1]["__init__"]
+    super_node = next(init_attr_node.expr.infer())
+
+    # Arbitrarily limit the max number of infered nodes per context
+    InferenceContext.max_inferred = -1
+    context = InferenceContext()
+
+    # Try to infer 'object.__init__' > because of limit is impossible
+    for _ in bases._infer_stmts([init_object_node], context, frame=super):
+        pass
```
Pierre-Sassoulas, line 424:
```suggestion
        assert _ == Uninferable
```
```suggestion
        pass
    assert _ == Uninferable
```
I'll let you apply the one you wanted :)

DanielNoord, line 424:
```suggestion
    for inferred in bases._infer_stmts([init_object_node], context, frame=super):
        assert inferred == Uninferable
```

DanielNoord: Did another change, where `_` is also renamed. If this passes the CI this should be ready for review again!

ChangeLog
```diff
@@ -9,6 +9,13 @@ Release date: TBA
 * Fixed builtin inferenence on `property` calls not calling the `postinit` of the new node, which
   resulted in instance arguments missing on these nodes.
 
+* Fixed a crash on ``Super.getattr`` when the attribute was previously uninferable due to a cache
+  limit size.
+  This limit can be hit when the inheritance pattern of a class (and therefore of the ``__init__``
+  attribute) is very large.
```
Pierre-Sassoulas, line 15:
```suggestion
* Fixed a crash on ``Super.getattr`` when the attribute was previously uninferable due to a cache
  limit size. This limit can be hit when the inheritance pattern of a class (and therefore of the
  ``__init__`` attribute) is very large.
```

Pierre-Sassoulas (approved):
👍

drothlis:
This also fixes pylint no-member false positives: no-member doesn't warn if the owner is Uninferable, but because of this bug the Uninferable value was missing from `infer()`.

PR #1371, opened 2022-01-27
Add a changelog entry for attrs support

### LINKED ISSUE #1330: `brain_attrs.py` does not support attrs v21.3.0+

jacobbogdanov:
The original PR #1331 was missing a changelog entry

Closes #1330

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,11 @@ Release date: TBA
 
   Closes PyCQA/pylint#5679
 
+* Add support for [attrs v21.3.0](https://github.com/python-attrs/attrs/releases/tag/21.3.0) which
+  added a new `attrs` module alongside the existing `attr`.
+
+  Closes #1330
+
 What's New in astroid 2.9.4?
 ============================
 Release date: TBA


Pierre-Sassoulas (approved):
Thanks a lot for the follow-up I did not expect it and I appreciate it ;)

PR #1372, opened 2022-01-27
Use constants defined in ``astroid.const`` where applicable

DanielNoord:
## Steps

## Description

Something I had on my mind for some time.

## Type of Changes

### DIFF
--- a/astroid/brain/brain_signal.py
+++ b/astroid/brain/brain_signal.py
@@ -27,6 +27,7 @@ import sys
 
 from astroid.brain.helpers import register_module_extender
 from astroid.builder import parse
+from astroid.const import WIN32
 from astroid.manager import AstroidManager
 
 
@@ -49,7 +50,7 @@ def _signals_enum():
         SIGSEGV   = enum.auto()
         SIGTERM   = enum.auto()
     """
-    if sys.platform != "win32":
+    if not WIN32:
         signals_enum += """
         SIGALRM   = enum.auto()
         SIGBUS    = enum.auto()
@@ -76,7 +77,7 @@ def _signals_enum():
         SIGXCPU   = enum.auto()
         SIGXFSZ   = enum.auto()
         """
-    if sys.platform == "win32":
+    if WIN32:
         signals_enum += """
         SIGBREAK  = enum.auto()
         """
@@ -103,7 +104,7 @@ def _handlers_enum():
 
 def _sigmasks_enum():
     """Generates the source code for the Sigmasks int enum."""
-    if sys.platform != "win32":
+    if not WIN32:
         return """
     import enum
     class Sigmasks(enum.IntEnum):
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
I think this is nice change and also a small perf improvement but also that @cdce8p was against it because it makes it harder to detect sys guard. (We reverted some of it before). I don't have strong opinion about this but I also don't want to enter an edit war.

DanielNoord:
I left occurrences for imports and redefinition of methods for specific versions, because of that issue. Of course we could still decide against it. I don't mind leaving as is.

Pierre-Sassoulas:
> I left occurrences for imports and redefinition of methods for specific versions, because of that issue.

Then I think we could merge if Marc do not react before 2.10.

cdce8p:
> > I left occurrences for imports and redefinition of methods for specific versions, because of that issue.
> 
> Then I think we could merge if Marc do not react before 2.10.

Sorry guys! I needed some time to work on different things and just didn't look into any pylint or astroid issues as my backlog was getting larger. Am slowly catching up now, although I probably won't have much free time the next few months. If I'm unresponsive on Github you can usually still reach me on Discord, fyi.

> I think this is nice change and also a small perf improvement but also that @cdce8p was against it because it makes it harder to detect sys guard.

Both mypy and pyright know how to handle `sys` guards. They don't have any logic for arbitrary constant guards. @DanielNoord You mentioned it yourself that you left some sys guards in-place because of it. Thus, I would still recommend against this change.

If performance is that important to you, I would suggest moving the sys guards one level up. E.g. for the `TreeRebuilder` instead of doing the checks inside the individual methods, define the methods itself conditionally. That way the check would only be run once on load, instead of for each method call.

There are some huge drawbacks though. By definition, you'll need to copy a lot of code. With that it will be much more difficult to maintain. That's why I didn't do it when I added the guards for `end_lineno` and `end_col_offset`. IMO it's just not worth it. At some point even Python 3.7 is deprecated and the problem resolves itself.

DanielNoord:
> Sorry guys! I needed some time to work on different things and just didn't look into any pylint or astroid issues as my backlog was getting larger. Am slowly catching up now, although I probably won't have much free time the next few months. If I'm unresponsive on Github you can usually still reach me on Discord, fyi.

As always, no worries! If you have any time to spare on `astroid` or `pylint` issues I think your input would be most valuable on the currently opened `astroid` PRs.  There are both some crucial changes to `distutils` handling as well as the `NamedTuple` instance and mirroring `ast` APIs which would benefit from additional reviews.  

>  At some point even Python 3.7 is deprecated and the problem resolves itself.

Fine with me. I just thought I would propose the change and start the discussion. Let's not do this! Thanks @cdce8p for the input! 😄

cdce8p:
The `distutils` PRs are still on my list. I've also seen a notification about `ast`. I don't want to make any promises, but just maybe I'll get to them.

PR #1377, opened 2022-01-30
Fix typing for `assigned_stmts`

cdce8p:
## Description
Methods are attributes of the class and should therefore be typed as ClassVar.
https://docs.python.org/3/library/typing.html#typing.ClassVar

This isn't recognized by mypy yet. I had a similar issue with Home Assistant when someone mentioned `ClassVar`.

/CC @DanielNoord

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -44,7 +44,16 @@ import sys
 import typing
 import warnings
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Callable, Generator, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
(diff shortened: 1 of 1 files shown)


DanielNoord (approved):

astroid/nodes/node_classes.py
```diff
@@ -478,7 +487,7 @@ def __init__(
             parent=parent,
         )
 
-    assigned_stmts: AssignedStmtsCall["AssignName"]
+    assigned_stmts: ClassVar[AssignedStmtsCall["AssignName"]]
```
cdce8p, line 490: Fyi, I just thought that we could included `ClassVar` in the type alias, too. However, that is a type error. `ClassVar` is only accepted on class attributes.

PR #1381, opened 2022-02-04
Add missing ssl enums to astroid/brain

cipherboy:
## Steps

## Description

`astroid`'s `brian_ssl` wrapper lacks several enums constructed in the core
`ssl` module. This eventually affects `pylint`'s ability to parse code which
makes use of these enums:

> ```
> example.py:9:6: E1101: Module 'ssl' has no 'VerifyMode' member (no-member)
> ```

Adding these enums to `brain_ssl` fixes the reported issue against `pylint`.

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

## Type of Changes

## Related Issue

Closes: https://github.com/PyCQA/pylint/issues/3691


DanielNoord (commented):
I took a quick look at the source code of `ssl` [here](https://github.com/python/cpython/blob/3.10/Lib/ssl.py).
Do you know if there is any logic to these imports? Are they grouped together based on type, or is it completely arbitrary?

DanielNoord:
No I think the order is fine as you proposed it. I don't have any knowledge of the `ssl` module so I was wondering if the current list resembles any logical order, but I agree that it doesn't seem that way.
Thus, adding the additional enums in this order is also correct.

Thanks for this!

DanielNoord (approved):

DanielNoord:
@Pierre-Sassoulas Do we want a test for this? If so, I'm retracting my approval 😄

DanielNoord:
Yes! If you could add a similar test that would be highly appreciate!

Pierre-Sassoulas (commented):
A test is always nice ! 😄 Did you manage to find example of how we test brain @cipherboy ?

cipherboy:
If I add something simliar to `test_ssl_protocol` it says the type of `ssl.VerifyMode` is `Uninferable` which seems less than ideal? Besides `Expr` tho, I'm not really sure what type `ssl.VerifyMode` should be? I'm thinking `ClassDef` isn't right because its not a _definition_ but maybe that's right? It doesn't seem right that it'd be a constant either. 

@Pierre-Sassoulas Is that the right type of test to be adding? :-)

DanielNoord:
Could you add the test and push the commit? I can probably take a look and see why the inference fails. There might be something complicated going on in the definition of the `Enums` in `ssl` that is unrelated to this PR but which trips astroid.

cipherboy:
@DanielNoord Pushed, please see the top commit. I tried `ssl.VerifyMode.CERT_REQUIRED` this morning (as it is a specific enum value, rather than the class of all values -- so I would've expected it to be a CONST or EXPR or something) and it has the same `Uninferable` type.

Pierre-Sassoulas:
> If I add something simliar to test_ssl_protocol it says the type of ssl.VerifyMode is Uninferable which seems less than ideal?

As ssl is bultin it's probably an issue with astroid itself (otherwise we have test that require a lib to be installed, for numpy for example).

DanielNoord:
@cipherboy I finally got time to look at this. I have changed the brain to better reflect what CPython actually does in https://github.com/python/cpython/blob/3.10/Lib/ssl.py. I have also updated the tests to test all three forms of definitions that are now available in the module + brain.

@Pierre-Sassoulas @jacobtylerwalls I think this is ready for review!

cipherboy (commented):
Changes look good, thanks @DanielNoord for finishing this up for me!

It looks like [because `IntEnum._convert_` was used](https://github.com/python/cpython/blob/7b3cf3a2884b130c519fbe271e4d2a762248c2a0/Lib/enum.py#L538), we've gotta create our own class wrappers for everything rather than inferring it from the IntEnum usage? That makes sense, its probably hard to detect in general when a new class object is returned.

DanielNoord:
> Changes look good, thanks @DanielNoord for finishing this up for me!
> 
> It looks like [because `IntEnum._convert_` was used](https://github.com/python/cpython/blob/7b3cf3a2884b130c519fbe271e4d2a762248c2a0/Lib/enum.py#L538), we've gotta create our own class wrappers for everything rather than inferring it from the IntEnum usage? That makes sense, its probably hard to detect in general when a new class object is returned.

Yeah, I don't think we are able to infer `_convert_` correctly due to the use of the `if ... else` although I'm not sure. This fixes the issues anyway.

cdce8p (commented):
The changes look fine. However, I'm wondering if `ssl.VerifyMode.CERT_REQUIRED` would be inferable now? That was one of the initial test cases.

DanielNoord:
> ssl.VerifyMode.CERT_REQUIRED

Returns `Uninferable`. But that might be due to general lack of `Enum` support. Would that be a blocker?

cdce8p:
> > ssl.VerifyMode.CERT_REQUIRED
> 
> Returns `Uninferable`. But that might be due to general lack of `Enum` support. Would that be a blocker?

It would only resolve some of the `no-member` errors. Better if we can do all.

It's quite cumbersome for all constants, but would something like this work?
```py
    class VerifyMode(_IntEnum):
        CERT_NONE = 0
        CERT_OPTIONAL = 1
        CERT_REQUIRED = 2
```

That's comparable to what [typeshed](https://github.com/python/typeshed/blob/a2ba0c8a00385e62655c7fc714b814eeb2533b5a/stdlib/ssl.pyi#L132-L139) does for types.

DanielNoord:
> > > ssl.VerifyMode.CERT_REQUIRED
> > 
> > 
> > Returns `Uninferable`. But that might be due to general lack of `Enum` support. Would that be a blocker?
> 
> It would only resolve some of the `no-member` errors. Better if we can do all.
> 
> It's quite cumbersome for all constants, but would something like this work?
> 
> ```python
>     class VerifyMode(_IntEnum):
>         CERT_NONE = 0
>         CERT_OPTIONAL = 1
>         CERT_REQUIRED = 2
> ```
> 
> That's comparable to what [typeshed](https://github.com/python/typeshed/blob/a2ba0c8a00385e62655c7fc714b814eeb2533b5a/stdlib/ssl.pyi#L132-L139) does for types.

Added the `Enums` based on `typeshed`. I don't think the `if sys.version` works, right? So that should probably be its own string?

cdce8p:
> I don't think the `if sys.version` works, right? So that should probably be its own string?

Yeah, take a look at `brain_collections.py` for it.
https://github.com/PyCQA/astroid/blob/c4b59dde8e91fd342fa28485953baa59aa82d65b/astroid/brain/brain_collections.py#L69-L78

DanielNoord:
> > I don't think the `if sys.version` works, right? So that should probably be its own string?
> 
> Yeah, take a look at `brain_collections.py` for it.
> 
> https://github.com/PyCQA/astroid/blob/c4b59dde8e91fd342fa28485953baa59aa82d65b/astroid/brain/brain_collections.py#L69-L78

Fixed!

astroid/brain/brain_ssl.py
```diff
@@ -4,14 +4,18 @@
 
 """Astroid hooks for the ssl library."""
 
+
```
cdce8p, line 7:
```suggestion
```

astroid/brain/brain_ssl.py
```diff
@@ -64,8 +68,101 @@ def ssl_transform():
...
+        CERT_OPTIONAL = 1
+        CERT_REQUIRED = 2
     """
-    )
+
+    def options_enum() -> str:
+        enum = """
+    class Options(_IntFlag):
+        OP_ALL = 1
+        OP_NO_SSLv2 = 2
```
cdce8p, line 122: It would probably be better if the function wasn't nested. I would stick with the pattern here:
https://github.com/PyCQA/astroid/blob/20adb5420d29b7ad9528941c1eb391e58e004db3/astroid/brain/brain_collections.py#L13-L24

astroid/brain/brain_ssl.py
```diff
@@ -64,8 +68,101 @@ def ssl_transform():
...
+        OP_NO_TICKET = 12
+        OP_NO_RENEGOTIATION = 13
+        """
+
+        if PY38_PLUS:
+            enum += """
+        OP_ENABLE_MIDDLEBOX_COMPAT = 14
+        """
+
+        return enum
```
cdce8p, line 141:
```suggestion
        OP_NO_RENEGOTIATION = 13"""
        if PY38_PLUS:
            enum += """
        OP_ENABLE_MIDDLEBOX_COMPAT = 14"""
        return enum
```

This can be a bit more concise, similar to:

https://github.com/PyCQA/astroid/blob/20adb5420d29b7ad9528941c1eb391e58e004db3/astroid/brain/brain_collections.py#L61-L66

astroid/brain/brain_ssl.py
```diff
@@ -64,8 +68,101 @@ def ssl_transform():
...
+        VERIFY_X509_TRUSTED_FIRST = 4
+        """
+
+        if PY310_PLUS:
+            enum += """
+        VERIFY_ALLOW_PROXY_CERTS = 5
+        VERIFY_X509_PARTIAL_CHAIN = 6
+        """
+
+        return enum
```
cdce8p, line 161: Same here, both with nesting and whitespaces.

cdce8p (approved):
Looks good!

DanielNoord:
Thanks @cipherboy for the initial PR and thanks @cdce8p for the reviews! We're nearing a release I think 🎉

PR #1382, opened 2022-02-04
Include names of keyword-only arguments in `argnames()`

jacobtylerwalls:
## Steps

## Description
Before, `astroid.scoped_nodes.Lambda.argnames()` left out the names of keyword-only arguments. This caused a false positive in a Pylint checker.

Of course, it's possible confusion could continue in the future if we continue to let `Arguments.arguments()` skip these. 🤔 

## Type of Changes

## Related Issue

Ref. PyCQA/pylint#5771

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1376,6 +1376,7 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
             names = _rec_get_names(self.args.arguments)
         else:
             names = []
+        names += [elt.name for elt in self.args.kwonlyargs]
         if self.args.vararg:
             names.append(self.args.vararg)
         if self.args.kwarg:
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1366,8 +1366,10 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
         """
         return True
 
-    def argnames(self):
-        """Get the names of each of the arguments.
+    def argnames(self) -> List[str]:
+        """Get the names of each of the arguments, including that
+        of the collections of variable-length arguments ("args", "kwargs",
+        etc.), as well as keyword-only arguments.
 
         :returns: The names of the arguments.
         :rtype: list(str)
@@ -1992,7 +1994,7 @@ class AsyncFunctionDef(FunctionDef):
     """
 
 
-def _rec_get_names(args, names=None):
+def _rec_get_names(args, names: Optional[List[str]] = None) -> List[str]:
     """return a list of all argument names"""
     if names is None:
         names = []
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
I think it's a desirable change. Thank you @jacobtylerwalls

DanielNoord (changes requested):
Could you add the typing to the function as well (I know... 😅). Perhaps we could also update the docstring.

And have you tested this against `pylint` to see if any other checkers depend on the current behaviour?

jacobtylerwalls:
Yep, pylint test suite passes.

Just browsing other uses of `argnames()` in pylint, I see we can subsequently clean up this special case (similar issue, similar fix):
https://github.com/PyCQA/pylint/commit/01cc02d5e

The other uses in Pylint looked like `argnames()[0]` uses to get `self`, `cls`, etc, and in one case were already guarded by `if node.args.args` in order to guarantee that the first arg is not keyword-only. I haven't done a thorough analysis of the other uses, but at a glance I didn't see red flags.

DanielNoord:
I wonder if we should add a new attribute for `self` and `cls`. Seems tricky for that to rely on the order of `argnames`, but maybe not, hm I don't know 😄

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1991,7 +1994,7 @@ async def func(things):
     """
 
 
-def _rec_get_names(args, names=None):
+def _rec_get_names(args, names: Optional[List[str]] = None) -> List[str]:
```
DanielNoord, line 1997: Thanks for this!

DanielNoord (approved):

DanielNoord:
Btw, we tend to close `pylint` issues with `astroid` related fixes only after we added a regression test in `pylint` itself. I changed the PR description to not use the closes keyword so the issue won't be closed automatically.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1376,6 +1378,7 @@ def argnames(self):
             names = _rec_get_names(self.args.arguments)
         else:
             names = []
+        names += [elt.name for elt in self.args.kwonlyargs]
```
cdce8p, line 1381: Didn't saw it earlier. Wouldn't it make sense to add them **after** `vararg`? That would keep the correct order.

tests/unittest_scoped_nodes.py
```diff
@@ -473,6 +473,10 @@ def test_argnames(self) -> None:
         astroid = builder.parse(code, __name__)
         self.assertEqual(astroid["f"].argnames(), ["a", "b", "c", "args", "kwargs"])
 
+        code_with_kwonly_args = "def f(a, b, *, c=None, d=None): pass"
```
cdce8p, line 476: If you change `*` into `*args`, that could test the correct order. Maybe add `**kwargs` as well.

Furthermore, maybe a bit outside the scope of the PR, it could make sense to test positional only arguments too.

DanielNoord:
@jacobtylerwalls  I think @cdce8p's suggestions make sense. Would you want to do a follow-up PR?

jacobtylerwalls:
Sure thing. I hope to do it this week.

PR #1384, opened 2022-02-08
Remove unused tryfinally builder function

DanielNoord:
## Steps

## Description

This is no longer used I think.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -2168,24 +2168,6 @@ class TreeRebuilder:
             return self.visit_tryexcept(node, parent)
         return None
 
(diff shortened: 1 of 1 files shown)


DanielNoord:
Test on `pylint` passed.

I wonder if this should be deprecated? This method won't be called in Python 3.3+ and is only called (normally) via the automatic visiting process. So I don't expect anybody on any of the supported python versions is actually using this..

cdce8p (approved):
Great catch! It should indeed be safe to remove. Seems like I missed that during my refactoring 🙈 Tbh the whole Try block parsing should be fixed at some point. The separation into `TryExcept` and `TryFinally` nodes doesn't represent the current AST anymore. I'm working on a PR to make that possible (in a backwards compatible manner) at the moment.

PR #1385, opened 2022-02-08
Add typing to the parent attribute of ExceptHandler

DanielNoord:
## Steps

## Description

Inspired by the work of @jacobtylerwalls in https://github.com/PyCQA/pylint/pull/5764.

Is there a good way to make `parent` non-optional? 

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -2523,11 +2523,16 @@ class ExceptHandler(mixins.MultiLineBlockMixin, mixins.AssignTypeMixin, Statemen
     _astroid_fields = ("type", "name", "body")
     _multi_line_block_fields = ("body",)
 
+    # TODO: Make parent non-optional
+    parent: Optional["nodes.TryExcept"]
+    """The parent node in the syntax tree."""
+
     def __init__(
         self,
         lineno: Optional[int] = None,
         col_offset: Optional[int] = None,
-        parent: Optional[NodeNG] = None,
+        # TODO: Make parent non-optional
+        parent: Optional["nodes.TryExcept"] = None,
         *,
         end_lineno: Optional[int] = None,
         end_col_offset: Optional[int] = None,
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -253,7 +253,7 @@ class TreeRebuilder:
 
         @overload
         def visit(
-            self, node: "ast.ExceptHandler", parent: NodeNG
+            self, node: "ast.ExceptHandler", parent: nodes.TryExcept
         ) -> nodes.ExceptHandler:
             ...
 
@@ -599,7 +599,7 @@ class TreeRebuilder:
 
         @overload
         def visit(
-            self, node: "ast.ExceptHandler", parent: NodeNG
+            self, node: "ast.ExceptHandler", parent: nodes.TryExcept
         ) -> nodes.ExceptHandler:
             ...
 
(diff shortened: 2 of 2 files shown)


astroid/nodes/node_classes.py
```diff
@@ -2523,11 +2523,16 @@ class ExceptHandler(mixins.MultiLineBlockMixin, mixins.AssignTypeMixin, Statemen
...
+    # TODO: Make parent non-optional
+    parent: Optional["nodes.TryExcept"]
+    """The parent node in the syntax tree."""
+
     def __init__(
         self,
         lineno: Optional[int] = None,
         col_offset: Optional[int] = None,
-        parent: Optional[NodeNG] = None,
+        # TODO: Make parent non-optional
```
cdce8p, line 2534: This wouldn't really work for plugins / brain modules. They might want to construct a Node outside the tree first before inserting it in the end.

astroid/nodes/node_classes.py
```diff
@@ -2523,11 +2523,16 @@ class ExceptHandler(mixins.MultiLineBlockMixin, mixins.AssignTypeMixin, Statemen
     _astroid_fields = ("type", "name", "body")
     _multi_line_block_fields = ("body",)
 
+    # TODO: Make parent non-optional
+    parent: Optional["nodes.TryExcept"]
```
cdce8p, line 2527: Saying the `parent` attribute is required (except for Module nodes), might work with a decorator.
```py
    def __init__(..., parent: Optional["NodeNG"], ...):
        self._parent = parent  # this needs to stay optional

    @property
    def parent(self) -> "NodeNG":
        assert self._parent  # here, we make sure it has actually been assigned
        return self._parent

    @parent.setter
    def parent(self, value: "NodeNG") -> None:
        self._parent = value
```

If you would like to explore that further in a new PR, an additional tip. Since `Module` inherits from `NodeNG` and doesn't have a parent, the decorator can't be added there. You'd probably need to create a new subclass which all nodes (except for `Module`) inherit from to implement it.

cdce8p (changes requested):
I recommend not to move forward with this exact PR **at this time**.

It would be better to implement the `parent` decorator first (if it works as expected). Furthermore, I'm working on some extensive `rebuilder` changes. It's probably better to consider it after those are done. There is also the rewrite for the Try nodes I mentioned in https://github.com/PyCQA/astroid/pull/1384#pullrequestreview-876744418. Lastly, assuming the parent should be a `TryExcept` node might make sense here. That doesn't mean though that similar cases also do. In general we should be cautious about such assumptions, not all astroid plugins might adhere to the official Python AST. Adding an additional `isinstance` check in `pylint` certainly doesn't hurt.

DanielNoord:
Good points @cdce8p! I'll wait on the rebuilder refactor before doing anything like this. Let me know if I can be of any help anywhere!

cdce8p:
> Good points @cdce8p! I'll wait on the rebuilder refactor before doing anything like this. Let me know if I can be of any help anywhere!

I'll probably need to write some kind of design doc (either in the PR or in a separate issue). Will make sure to let you know when I do, feedback should definitely help!

The `parent` decorator concept shouldn't interfere with the refactor if you want to look into that.

cdce8p:
> The `parent` decorator concept shouldn't interfere with the refactor if you want to look into that.

It's also something I hope to use for attributes that are currently defined as optional but never are. Like the `name` and `attrname` attributes on `AssignName`, `DelName`, `Name`, `AssignAttr`, and `DelAttr`.
If you have some time, implementing the decorator concept for those would probably be quite useful.

DanielNoord:
I can certainly take a look at the decorator concept when I get some time. Probably after `pylint` `2.13` and `astroid` `2.10` though 😄

PR #1386, opened 2022-02-09
Restore custom `distutils` handling for resolving paths to submodules.

jacobtylerwalls:
## Steps

## Description
I understand @DanielNoord wants to bisect before evaluating a path forward. But this test case fails without these changes, so it could be a sufficient way forward.

## Type of Changes

## Related Issue

Refs https://github.com/PyCQA/pylint/issues/5645

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -434,5 +434,6 @@ jobs:
           python -m pip install -e .
       - name: Test distutils edge case
         run: |
+          . venv2\scripts\activate
           echo "import distutils.util  # pylint: disable=unused-import" > test.py
-          python -m pylint test.py
+          pylint test.py
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -406,38 +406,3 @@ jobs:
         run: |
           . venv/bin/activate
           pytest tests/
-
-  specific-virtualenv-test:
-    name: Regression test for virtualenv==15.1.0 on Windows
-    runs-on: windows-latest
-    timeout-minutes: 5
-    steps:
-      - name: Check out code from GitHub
-        uses: actions/checkout@v2.3.4
-      - name: Set up Python ${{ matrix.python-version }}
-        id: python
-        uses: actions/setup-python@v2.2.1
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Restore Python virtual environment
-        id: cache-virtualenv15
-        uses: actions/cache@v2.1.4
-        with:
-          path: venv2
-          key: >-
-            ${{ runner.os }}-virtualenv15
-          restore-keys: |
-            ${{ runner.os }}-virtualenv15-${{ env.CACHE_VERSION }}
-      - name: Create Python virtual environment with virtualenv==15.1.0
-        if: steps.cache-virtualenv15_1_0.outputs.cache-hit != 'true'
-        run: |
-          python -m pip install virtualenv==15.1.0
-          python -m virtualenv venv2
-          . venv2\scripts\activate
-          python -m pip install pylint
-          python -m pip install -e .
-      - name: Test distutils edge case
-        run: |
-          . venv2\scripts\activate
-          echo "import distutils.util  # pylint: disable=unused-import" > test.py
-          pylint test.py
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -406,3 +406,29 @@ jobs:
         run: |
           . venv/bin/activate
           pytest tests/
+
+  extra-special-test-case:
+    name: Test case that requires specific virtualenv
```
jacobtylerwalls, line 411: Obviously we're not going to merge it like this.

DanielNoord:
I wonder how we can fully test this..
This does seem like it would work though.

@cdce8p What do you think? Try this and see if the problems go away?

jacobtylerwalls:
> I wonder how we can fully test this..

I mean, we _could_ add the CI step in this PR if we really want to have an extra workflow that only runs one corner case!

.github/workflows/ci.yaml
jacobtylerwalls: I improved this CI step to give us the option of merging it.

Pierre-Sassoulas (approved):
Thank you @jacobtylerwalls. We just need to remove the temporary CI job before merging, right ?

DanielNoord:
Do we though? Won't the CI job be a nice test to see this actually works?

jacobtylerwalls:
Fine with me. I added caching to make it more efficient.

cdce8p (approved):
The change looks like it should work! I'm just not sure that we should run the CI job for every time. It's excellent to have but maybe better as part of an optional CI suite?

What do you guys think about moving it to a separate workflow file and using the `workflow_dispatch` event to trigger it whenever we need to? For that case, it would make sense to add a small comment in the code mentioning the workflow and that it can be run manually. Lastly, the caching could probably be removed. Although slightly more efficient, it can also create unexpected issues. If only run on request, there likely won't be any cache which can be restored to begin with.

An alternative would be to add `if: false` to the job. It would also be trivial to change it to `true` whenever we need to work on it (which itself shouldn't be all too often).

After writing this, I'm just slightly leaning towards the latter option.

jacobtylerwalls:
I'm thinking along the lines of the former option: what about creating a "release testing" CI workflow with `on: workflow_dispatch`, and then adding a reminder to the release.md checklist to remember to run it before releases?

cdce8p:
> The issue is then that we can tag buggy commit and have nothing released.

The idea was to have run automatically for every release. I still wanted to keep the manual trigger.
You do have a point though that deleting a tag is quite some work. Let's do just the `workflow_dispatch` trigger then with a note in `release.md` as suggested.

.github/workflows/release-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: Release tests
+
+on: workflow_dispatch
+
+env:
+  DEFAULT_PYTHON: 3.8
+
+jobs:
+  virtualenv-15-windows-test:
+    name: Regression test for virtualenv==15.1.0 on Windows
```
cdce8p, line 10: Last two 🙂

I believe it's possible to add comments in workflow files too. It would be great if we could link to the PR here just to have a reference later why it was created.

astroid/interpreter/_import/spec.py
```diff
@@ -160,6 +161,20 @@ def contribute_to_path(self, spec, processed):
                 for p in sys.path
                 if os.path.isdir(os.path.join(p, *processed))
             ]
+        elif spec.name == "distutils":
+            # virtualenv below 20.0 patches distutils in an unexpected way
+            # so we just find the location of distutils that will be
+            # imported to avoid spurious import-error messages
```
cdce8p, line 167: A PR (or issue) link could also be helpful here.

cdce8p:
Thanks @jacobtylerwalls 🐬

DanielNoord:
@ everybody, do we want to test this as well in `pylint`? I don't think so right? (Seeing the difficulty of doing so in `astroid`). If not we can also close the `pylint` issue I think.

cdce8p:
> @ everybody, do we want to test this as well in `pylint`? I don't think so right? (Seeing the difficulty of doing so in `astroid`). If not we can also close the `pylint` issue I think.

Yeah, let's close the `pylint` issue. The test here should be enough.

PR #1387, opened 2022-02-09
Improve order of keyword-only args in argnames()

jacobtylerwalls:
## Description
Follow up to https://github.com/PyCQA/astroid/pull/1382. The order of results is implied that it will be correct, so provide a correct order of argnames if there is a variadic argument name and some keyword-only arguments.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,10 @@ Release date: TBA
 
   Closes PyCQA/pylint#5771
 
+* Improve location of keyword-only arguments in ``astroid.scoped_nodes.Lambda.argnames``.
+
+  Refs PyCQA/pylint#5771
+
 * Add support for [attrs v21.3.0](https://github.com/python-attrs/attrs/releases/tag/21.3.0) which
   added a new `attrs` module alongside the existing `attr`.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,10 +25,6 @@ Release date: TBA
 
   Closes PyCQA/pylint#5771
 
-* Improve location of keyword-only arguments in ``astroid.scoped_nodes.Lambda.argnames``.
-
-  Refs PyCQA/pylint#5771
-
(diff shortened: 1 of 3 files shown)


tests/unittest_scoped_nodes.py
```diff
@@ -473,9 +473,19 @@ def test_argnames(self) -> None:
...
+            astroid["f"].argnames(), ["a", "b", "args", "c", "d", "kwargs"]
+        )
+
+    @unittest.skipUnless(PY38_PLUS, "positional-only argument syntax")
+    def test_positional_only_argnames(self) -> None:
+        code = "def f(a, b, /, c=None, *args, d, **kwargs): pass"
+        astroid = builder.parse(code, __name__)
+        self.assertEqual(
+            astroid["f"].argnames(), ["a", "b", "c", "args", "d", "kwargs"]
+        )
```
cdce8p, line 488: Interesting that positional-only arguments are included 🤔
Wasn't sure about that, since the method uses `args.arguments` when there is `args.posonlyargs` as well. It does make sense however to include all names here. A check if an arg is positional-only should be done separately (in pylint).

Maybe it would be good to update the docstring a bit?
```
[...] as well as positional-only and keyword-only arguments.
```

ChangeLog
```diff
@@ -25,6 +25,10 @@ Release date: TBA
 
   Closes PyCQA/pylint#5771
 
+* Improve location of keyword-only arguments in ``astroid.scoped_nodes.Lambda.argnames``.
+
+  Refs PyCQA/pylint#5771
+
```
cdce8p, line 31: That is not necessary. We haven't release a new version since, so the entry for the initial change is enough.

cdce8p (commented):
Thanks for the quick follow-up @jacobtylerwalls!

cdce8p (approved):
Thanks @jacobtylerwalls 🐬

PR #1391, opened 2022-02-11
Add ``is_dataclass`` attribute to ``ClassDef``

DanielNoord:
## Steps

## Description

This would solve the issue in https://github.com/PyCQA/pylint/pull/5544.

I explored three options for this:
1) My first idea was to create a new `Instance` for dataclasses. However, this is incorrect.
```python
@dataclass
class A(): ...
```
If you instantiate `A` it is not a `Dataclass`, it is an instance of `A` which happens to be a dataclass. Thus, a new `Instance` is incorrect.
2) I explored if there is any merit in using `is_decorated_with_dataclass` or exposing it as util. However, it uses `nodes` to do som stuff which became difficult because of cyclic imports really fast. I don't think there is a good way to extract it from the brain (I tried) so if we want to make a similar util we will need to maintain the same code in two places.
3) The current solution. Actually quite simple. If we make any changes to dataclass inference it will also be easily accessible to `pylint` without any changes.

Let me know what you think!

_One remaining issue is that this doesn't allow setting the `decorators` attribute to be non-optional (which it is for dataclasses). However, creating a new `DataclassDef` node seemed a bit overkill (for now)._

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2212,6 +2212,9 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
         :type doc: str or None
         """
 
+        self.is_dataclass: bool = False
+        """Whether this class is a dataclass."""
+
         super().__init__(
             lineno=lineno,
             col_offset=col_offset,
@@ -2224,9 +2227,6 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
 
         for local_name, node in self.implicit_locals():
             self.add_local_node(node, local_name)
-
-        self.is_dataclass: bool = False
-        """Whether this class is a dataclass."""
 
     def implicit_parameters(self):
         return 1
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2138,6 +2138,9 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
     _other_other_fields = ("locals", "_newstyle")
     _newstyle = None
 
+    is_dataclass = False
+    """Whether this class is a dataclass."""
+
     def __init__(
         self,
         name=None,
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2225,7 +2225,7 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
         for local_name, node in self.implicit_locals():
             self.add_local_node(node, local_name)
 
-        self.is_dataclass = False
+        self.is_dataclass: bool = False
         """Whether this class is a dataclass."""
 
     def implicit_parameters(self):
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2134,12 +2134,9 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
             ":type: str"
         ),
     )
-    _other_fields = ("name", "doc")
+    _other_fields = ("name", "doc", "is_dataclass")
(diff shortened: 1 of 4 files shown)


cdce8p:
> I chose option three. I think option two might work as well, but to me it seemed more "natural" to let `astroid` do the inference and "type determination" of the class and then let `pylint` access that, rather than `pylint` having to rely on an `util` from `astroid`.

My main concern is that it's not well defined what constitutes a dataclass.  Of course there is the stdlib module but aside from that there aren't any real guidelines. Adding an attribute for it thus might be a bit confusing. We could fall back to say it should behave similarly to the stdlib implementation. That might work. I.e. what you've done here, everything that we can handle with the dataclass brain is fine, `attrs` is not (?).

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2138,6 +2138,9 @@ def my_meth(self, arg):
     _other_other_fields = ("locals", "_newstyle")
     _newstyle = None
 
+    is_dataclass = False
+    """Whether this class is a dataclass."""
```
cdce8p, line 2142: This should be an instance attribute, i.e. set in `__init__`.
Could you add a type annotation, too?

Furthermore, I believe the attribute name should also be added to `_other_fields`.

cdce8p (changes requested):

DanielNoord:
> That might work. I.e. what you've done here, everything that we can handle with the dataclass brain is fine, `attrs` is not (?).

I think that makes the most sense, although it could be improved a little. See #1394.

> This should be an instance attribute, i.e. set in `__init__`. Could you add a type annotation, too?

Is that necessary? Won't it be inferred automatically from `False`?

tests/unittest_brain_dataclasses.py
```diff
@@ -683,3 +687,29 @@ class A:
     inferred = code.inferred()
     assert len(inferred) == 1
     assert isinstance(inferred[0], nodes.ClassDef)
+    assert inferred[0].is_dataclass
+
+
+def test_non_dataclass_is_not_dataclass() -> None:
+    """Test that something that isn't a dataclass has the correct attribute."""
+    module = astroid.parse(
```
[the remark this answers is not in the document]
DanielNoord: I'll change it. I always use `parse`, but don't have a preference.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2225,6 +2225,9 @@ def __init__(
         for local_name, node in self.implicit_locals():
             self.add_local_node(node, local_name)
 
+        self.is_dataclass: bool = False
+        """Whether this class is a dataclass."""
```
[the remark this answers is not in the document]
DanielNoord: :+1

tests/unittest_brain_dataclasses.py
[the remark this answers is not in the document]
cdce8p: Both work fine. `extract_node` might just be a bit easier to annotate if you only selected one node type.

DanielNoord:
Pff.. What's wrong on `3.7`? 😢

tests/unittest_brain_dataclasses.py
```diff
@@ -683,3 +687,25 @@ class A:
     inferred = code.inferred()
     assert len(inferred) == 1
     assert isinstance(inferred[0], nodes.ClassDef)
+    assert inferred[0].is_dataclass
+
+
+def test_non_dataclass_is_not_dataclass() -> None:
+    """Test that something that isn't a dataclass has the correct attribute."""
+    ast_nodes = astroid.extract_node(
```
cdce8p, line 695:
```suggestion
    ast_nodes: List[nodes.NodeNG] = astroid.extract_node(  # type: ignore[assignment]
```
`extract_node` can return either a list or an individual node. (Not the best API design 😕) Typing it as `List[nodes.NodeNG]` should help.

DanielNoord: That's why I added `_extract_single_node` in https://github.com/PyCQA/astroid/pull/1358/files. However, now that I think about it it should probably use an `assert` instead of returning the first element..
We could create `extract_nodes` as well?

cdce8p: Something for another day. We have took many open problems already 😅

cdce8p (approved):
Looks good! Btw. I had the same issue with `extract_node` and `ClassDef` 😄 Opened #1395

PR #1392, opened 2022-02-11
Fix recursion error for inference of self-referencing class attribute

DanielNoord:
## Steps

## Description

This will close when merged https://github.com/PyCQA/pylint/issues/5408.

## Type of Changes

### DIFF
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -7,16 +7,17 @@ import typing
 
 import wrapt
 
-from astroid import bases, nodes, util
+from astroid import bases, util
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
+from astroid.nodes import NodeNG
 
 InferFn = typing.Callable[..., typing.Any]
 InferOptions = typing.Union[
-    nodes.NodeNG, bases.Instance, bases.UnboundMethod, typing.Type[util.Uninferable]
+    NodeNG, bases.Instance, bases.UnboundMethod, typing.Type[util.Uninferable]
 ]
 
 _cache: typing.Dict[
-    typing.Tuple[InferFn, nodes.NodeNG], typing.Optional[typing.List[InferOptions]]
+    typing.Tuple[InferFn, NodeNG], typing.Optional[typing.List[InferOptions]]
 ] = {}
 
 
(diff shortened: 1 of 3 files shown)


astroid/inference_tip.py
```diff
@@ -3,16 +3,20 @@
 
 """Transform utilities (filters and decorator)"""
 
-import typing
+
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union
```
Pierre-Sassoulas, line 7: It's imported this way in astroid because List, Dict, and Tuple are also astroid.nodes so it can become confusing. It's probably worth it to stay consistent accross the code base.

astroid/inference_tip.py
```diff
@@ -3,16 +3,20 @@
...
 
 import wrapt
 
-from astroid.exceptions import InferenceOverwriteError
-from astroid.nodes import NodeNG
+from astroid import bases, nodes, util
```
Pierre-Sassoulas, line 11: There are some circular import in astroid, between files and between packages, I think importing only what's necessary and not whole package will help to understand what is happening in the code if we want to one day handle the issue.

Pierre-Sassoulas (changes requested):
The actual change looks great, but I don't think we should change the way we currentely import typing and nodes.

astroid/inference_tip.py
DanielNoord: In this file we only use 2 nodes and a lot of different typing stuff. That's why I opted to do it the other way around in this file.

I'm fine with reverting!

astroid/inference_tip.py
DanielNoord: On the other hand, importing submodules avoid such import errors as we will get notified of those patterns sooner.

But I'll revert if you want me to!

Pierre-Sassoulas: @cdce8p do you have an opinion ? It's been a long time since I tried to fix those circular imports, all I remember is the pain 😄

astroid/inference_tip.py
DanielNoord: Reverted to only import `typing`.

Pierre-Sassoulas (approved):
Let's worry about circular import when they happen or when we fix them, thank you @DanielNoord

PR #1397, opened 2022-02-15
Add regression test for changes to support hermetic interpreters

jacobtylerwalls:
## Steps

## Description
Add a regression test for #1207.

## Type of Changes

| ✓   | ⏪ Regression Test

## Related Issue
Closes https://github.com/PyCQA/pylint/issues/5341
Alternative to https://github.com/PyCQA/pylint/pull/5210

### DIFF
--- a/tests/unittest_builder.py
+++ b/tests/unittest_builder.py
@@ -804,6 +804,11 @@ class HermeticInterpreterTest(unittest.TestCase):
             sys.path.append(tmp_dir)
 
             # Write a python file and compile it to .pyc
+            # To make this test have even more value, we would need to come up with some 
+            # code that gets inferred differently when we get its "partial representation".
+            # This code is too simple for that. But we can't use builtins either, because we would 
+            # have to delete builtins from the filesystem.  But even if we engineered that, 
+            # the difference might evaporate over time as inference changes.
             cls.code_snippet = "def func():  return 42"
             with tempfile.NamedTemporaryFile(
                 mode="w", dir=tmp_dir, suffix=".py", delete=False
@@ -823,7 +828,7 @@ class HermeticInterpreterTest(unittest.TestCase):
             os.remove(cls.imported_module_path)
             sys.path.remove(tmp_dir)
 
-    def test_build_from_live_module_without_source_file(self):
+    def test_build_from_live_module_without_source_file(self) -> None:
         """Assert that inspect_build() is not called.
         See comment in module_build() before the call to inspect_build():
             "get a partial representation by introspection"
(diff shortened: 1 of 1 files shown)


tests/unittest_builder.py
```diff
@@ -790,5 +794,62 @@ def test_parse_module_with_invalid_type_comments_does_not_crash():
...
+    """Modeled on https://github.com/PyCQA/astroid/pull/1207#issuecomment-951455588"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Simulate a hermetic interpreter environment having no code on the filesystem."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            sys.path.append(tmp_dir)
+
+            # Write a python file and compile it to .pyc
+            cls.code_snippet = "def func():  return 42"
```
jacobtylerwalls, line 807: To make this test have even more value, we would need to come up with some code that gets inferred differently when we get its "partial representation". This code is too simple for that. But we can't use builtins either, because we would have to delete builtins from the filesystem. This is why the author speculated about writing a "littletest" engineered to have inference ambiguity. But even if we engineered that, the _difference_ might evaporate over time as inference changes.

DanielNoord (approved):
Thanks a lot @jacobtylerwalls. This has been a very long outstanding issue.

tests/unittest_builder.py
Pierre-Sassoulas:
```suggestion
            # To make this test have even more value, we would need to come up with some 
            # code that gets inferred differently when we get its "partial representation".
            # This code is too simple for that. But we can't use builtins either, because we would 
            # have to delete builtins from the filesystem.  But even if we engineered that, 
            # the difference might evaporate over time as inference changes.
            cls.code_snippet = "def func():  return 42"
```

tests/unittest_builder.py
```diff
@@ -790,5 +794,62 @@ def test_parse_module_with_invalid_type_comments_does_not_crash():
...
+            # Import the module
+            cls.imported_module_path = pathlib.Path(cls.pyc_name)
+            cls.imported_module = importlib.import_module(cls.imported_module_path.stem)
+
+            # Delete source code from module object, filesystem, and path
+            del cls.imported_module.__file__
+            os.remove(cls.imported_module_path)
+            sys.path.remove(tmp_dir)
+
+    def test_build_from_live_module_without_source_file(self):
```
Pierre-Sassoulas, line 826:
```suggestion
    def test_build_from_live_module_without_source_file(self) -> None:
```

Pierre-Sassoulas (approved):
Thanks a lot @jacobtylerwalls !

PR #1399, opened 2022-02-16
More tests for imported modules

### LINKED ISSUE #1398: Cannot infer assignment with operator for a simple list of strings
### Steps to reproduce

```python
def test_imported_module_var_inferable():
    mod1 = parse(("from top.mod import v as z\n"
                 "w = [1] + z"),
                 module_name='top')
    mod2 = parse("v = [2]", module_name='top.mod')
    w_val = mod1.body[-1].value
    i_w_val = next(w_val.infer())
    assert i_w_val != util.Uninferable
    assert i_w_val.as_string() == "[1, 2]" # Works

def test_imported_module_var_inferable2():
    mod1 = parse(("from top.mod import v as z\n"
                 "w = ['1'] + z"),
                 module_name='top')
    mod2 = parse("v = ['2']", module_name='top.mod')
    w_val = mod1.body[-1].value
    i_w_val = next(w_val.infer())
    assert i_w_val != util.Uninferable
    assert i_w_val.as_string() == "['1', '2']" # Fails with result ['1', 2] instead

def test_imported_module_var_inferable3():
    mod1 = parse(("from top.mod import __dunder_var__ as v\n"
                 "__dunder_var__ = ['w'] + v"),
                 module_name='top')
    mod2 = parse("__dunder_var__ = ['v']", module_name='top.mod')
    w_val = mod1.body[-1].value
    i_w_val = next(w_val.infer())
    assert i_w_val != util.Uninferable # Retuns Uninferable
    assert i_w_val.as_string() == "['w', 'v']"
```

### Current behavior

It seems that operators are fully supported for lists of ints but not for list of strings, I'm unsure why, it seems the same level of complexity. Also it's very weird that it gives the `Uninferable` result for dunder variables. 

### Expected behavior

All tests should pass IMO. 

Astroid version 2.9.3

tristanlatr:
## Type of Changes

This PR adds more tests, see https://github.com/PyCQA/astroid/issues/1398

## Related Issue

Resolves #1398


tristanlatr:
The way the tests are faillinbg is very weird and I'm unsure if it's linked to my changes. 

The following test fails:
`tests/unittest_scoped_nodes.py::ClassNodeTest::test_mro_with_factories` with a difference in the computer MRO: 

```
E   - ['FinalClass', 'ClassB', 'MixinB', 'ClassA', 'MixinA', 'Base', 'object']
E   + ['FinalClass', 'ClassB', 'MixinB', '', 'ClassA', 'MixinA', '', 'Base', 'object']
```

I wonder if this is a side effect of calling `clear_cache()` or it's totally unrelated? Any clue?

Pierre-Sassoulas:
Ugh, I think you're right the tests might not be independent and rely on the caching between tests. Do you have an idea how to fix the failing test ?

tristanlatr:
I could restore the cache after the tests ?

tristanlatr:
Looks like it doesn’t work :/ I must be missing something…

tristanlatr:
Hello @jacobtylerwalls,

Thanks a lot for your input here. If I understand correctly, AstroidCacheSetupMixin clear the cache between each test method, right ?

If all tests passes, I would be happy that these test land in the main branch.

jacobtylerwalls:
Not between each test method but at the beginning and end of the class. (The reason your test was failing was the same thing I saw and fixed in #1528.) Do we need to clear the cache with every test method? If so we can adjust the behavior of the mixin to use `setUp` and `tearDown` instead (probably in another PR).

The failure on 3.11 is unrelated and tracked in #1551

jacobtylerwalls (approved):
I think we can merge this; if we want to increase the test isolation to clear caches every test method, we can do that in another PR anyway.

PR #1400, opened 2022-02-19
Fix crash inferring on NewType named with f-string

jacobtylerwalls:
## Steps

## Description
The f-string here:
```python
new_type = NewType(f'IntRange_{t}', t)
```
was altered by astroid here to remove its final `'`:

```python
typename = node.args[0].as_string().strip("'")
```
That caused a string parsing error, since there was an unterminated f-string.
Solution in this PR is to short circuit, rather than cope with a string formatted at runtime.

## Type of Changes

## Related Issue
Refs [PyCQA/pylint/#5770](https://github.com/PyCQA/pylint/issues/5770)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,10 @@ Release date: TBA
 
   Closes PyCQA/pylint#5771
 
+* Fixed a crash inferring on a ``NewType`` named with an f-string.
+
+  Closes PyCQA/pylint#5770
+
 * Add support for [attrs v21.3.0](https://github.com/python-attrs/attrs/releases/tag/21.3.0) which
   added a new `attrs` module alongside the existing `attr`.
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Thank you for finding this crash and fixing it ! 

Did you check that pylint's tests are still passing with this change ?

DanielNoord (approved):

jacobtylerwalls:
> Did you check that pylint's tests are still passing with this change ?

They are!

***
Will a regression test also be needed in Pylint or is this test sufficient?

Pierre-Sassoulas:
> Will a regression test also be needed in Pylint or is this test sufficient?

We generally close the pylint issue when the regression test is merged.

PR #1401, opened 2022-02-19
Fix incorrect test skip message

jacobtylerwalls:
## Steps

I think this can skip the changelog.

## Description
Fixes an incorrect skip message found while running the test suite on 3.9.10:
```
Skipped: Needs Python > 3.7. Current version is 3.9.10.final.0.
```

## Type of Changes


Pierre-Sassoulas (approved):
😄 Nice catch !

PR #1407, opened 2022-02-27
Limit expensive decorator function

### LINKED ISSUE #1383: Fix performance issue with ``inspect`` calls in deprecation warnings decorators
### Context

[See discussion here](https://github.com/PyCQA/astroid/pull/1157#issuecomment-1013744113)

cdce8p:
## Description
Modeled after the idea [here](https://github.com/PyCQA/astroid/pull/1157#issuecomment-1025269781), replace the `deprecate_default_argument_values` decorator with a default one if `DeprecationWarnings` aren't enabled. I.e. these will still show during Pytest runs or if a user chooses to enable warnings (with `-Wd` or similar).

Github Actions isn't the most reliable / reproducible, but I think it's about a 10-30% improvement.

@DanielNoord You did the initial performance analysis in https://github.com/PyCQA/astroid/pull/1157#issuecomment-1013744113. Would you be able to check again with these changes? Just a heads up, I don't know if the profiler enables `DeprecationWarnings`. Just for testing, you could add a simple `print` inside the `if` block to make sure.

https://docs.python.org/3/library/warnings.html#default-warning-filter
https://docs.python.org/3/using/cmdline.html#cmdoption-W

Closes #1383

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -152,58 +152,79 @@ def raise_if_nothing_inferred(func, instance, args, kwargs):
     yield from generator
 
 
-def deprecate_default_argument_values(
-    astroid_version: str = "3.0", **arguments: str
-) -> Callable[[Callable[P, R]], Callable[P, R]]:
-    """Decorator which emitts a DeprecationWarning if any arguments specified
-    are None or not passed at all.
+# Expensive decorators only used to emit Deprecation warnings.
+# If no other than the default DeprecationWarning are enabled,
+# fall back to passthrough implementations.
+if util.check_warnings_filter():
 
-    Arguments should be a key-value mapping, with the key being the argument to check
-    and the value being a type annotation as string for the value of the argument.
-    """
-    # Helpful links
-    # Decorator for DeprecationWarning: https://stackoverflow.com/a/49802489
-    # Typing of stacked decorators: https://stackoverflow.com/a/68290080
+    def deprecate_default_argument_values(
+        astroid_version: str = "3.0", **arguments: str
+    ) -> Callable[[Callable[P, R]], Callable[P, R]]:
+        """Decorator which emits a DeprecationWarning if any arguments specified
+        are None or not passed at all.
 
-    def deco(func: Callable[P, R]) -> Callable[P, R]:
-        """Decorator function."""
+        Arguments should be a key-value mapping, with the key being the argument to check
+        and the value being a type annotation as string for the value of the argument.
 
-        @functools.wraps(func)
-        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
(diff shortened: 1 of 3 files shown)


astroid/decorators.py
```diff
@@ -152,58 +152,79 @@ def raise_if_nothing_inferred(func, instance, args, kwargs):
...
-                    )
-            return func(*args, **kwargs)
-
-        return wrapper
-
-    return deco
+# Expensive decorators only used to emit Deprecation warnings.
+# If no other than the default DeprecationWarning are enabled,
+# fall back to passthrough implementations.
+if util.check_warnings_filter():
```
Pierre-Sassoulas, line 158: Nitpick: I think we could remove some duplication by putting this if inside ``deprecate_default_argument_values``

Pierre-Sassoulas (commented):
LGTM, nice way to keep the deprecation warning when it matters.

But it means when deprecation warnings are activated in user code astroid will still be slow right ? Maybe we should add a warning about that in the changelog ?

DanielNoord:
> @DanielNoord You did the initial performance analysis in [#1157 (comment)](https://github.com/PyCQA/astroid/pull/1157#issuecomment-1013744113). Would you be able to check again with these changes? Just a heads up, I don't know if the profiler enables `DeprecationWarnings`. Just for testing, you could add a simple `print` inside the `if` block to make sure.

FYI, I checked and with `python -m cProfile` the profiler does not enable `DeprectationWarnings`. I'm obviously unsure if there are specifics about other ways of invoking the profiler.

This is indeed much better! I did a rough profile and on a successive run over `pylint` itself I went from 64 to 53 seconds. I'm always unsure how to calculate this by 53/64 = 0,82. So I guess you could say almost a 20% performance increase?
Of course this is all very basic evidence, but this definitely solves the underlying issue. Thanks for looking into this @cdce8p!

astroid/decorators.py
cdce8p: I would suggest not to do that. The way it's written currently, the check is only done once when the module is loaded initially. If we move it inside the function, it's executed once per call.

cdce8p:
> But it means when deprecation warnings are activated in user code astroid will still be slow right ? Maybe we should add a warning about that in the changelog ?

I've already mentioned it in the Changelog message.
```
Only run expensive decorator functions when
non-default Deprecation warnings are enabled, eg. during a Pytest run.
```
The change itself doesn't make astroid slower than it's already. It just is an improvement if Deprecation warnings are disabled.

astroid/decorators.py
Pierre-Sassoulas: Right, I did not realize that ! Resolving.

Pierre-Sassoulas (approved):

PR #1408, opened 2022-02-27
Add ``Position`` to the ``__init__`` of ``nodes``

DanielNoord:
## Steps

## Description

See my comment here https://github.com/PyCQA/astroid/pull/1393#issuecomment-1053405431.

Think it will be useful to access this in `pylint`.

## Type of Changes

### DIFF
--- a/astroid/nodes/__init__.py
+++ b/astroid/nodes/__init__.py
@@ -121,6 +121,7 @@ from astroid.nodes.scoped_nodes import (
     function_to_method,
     get_wrapping_class,
 )
+from astroid.nodes.utils import Position
 
 _BaseContainer = BaseContainer  # TODO Remove for astroid 3.0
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👍

astroid/nodes/__init__.py
```diff
@@ -306,4 +307,5 @@
     "With",
     "Yield",
     "YieldFrom",
+    "Position",
```
cdce8p, line 310: I think, we tried to keep `__all__` sorted. I.e. it should probably be between `Pass` and `Raise`.

cdce8p (approved):

astroid/nodes/__init__.py
DanielNoord: I'll change it!

PR #1406, opened 2022-02-26
Partially revert #1326

DanielNoord:
## Steps

- [x] Write a good description on what the PR does.

## Description

This fixes the issues @cdce8p identified in #1326.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

PR #1409, opened 2022-02-27
Revert "Use ``importlib`` instead of ``pkg_resources`` for determining namespace packages"

DanielNoord:
Reverts PyCQA/astroid#1326

See discussion in https://github.com/PyCQA/astroid/pull/1406.

Sad 😢

### DIFF
--- a/astroid/const.py
+++ b/astroid/const.py
@@ -1,6 +1,7 @@
 import enum
 import sys
 
+PY36 = sys.version_info[:2] == (3, 6)
 PY38 = sys.version_info[:2] == (3, 8)
 PY37_PLUS = sys.version_info >= (3, 7)
 PY38_PLUS = sys.version_info >= (3, 8)
--- a/astroid/const.py
+++ b/astroid/const.py
@@ -1,7 +1,6 @@
 import enum
 import sys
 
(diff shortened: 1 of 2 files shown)


astroid/const.py
```diff
@@ -1,7 +1,6 @@
 import enum
 import sys
```
DanielNoord, line 3:
```suggestion

PY36 = sys.version_info[:2] == (3, 6)
```

Pierre-Sassoulas (approved):
Getting rid of ``pkg_resources``won't be easy, but a lot of preliminary work was already done and the situation is still better than in 2.9, thank you @DanielNoord 👍

DanielNoord:
> Getting rid of `pkg_resources`won't be easy, but a lot of preliminary work was already done and the situation is still better than in 2.9, thank you @DanielNoord 👍

With some debug tips from @cdce8p I'm certain we can figure this out. I know much more about `importlib` now so we're probably on the right track already 😄

cdce8p (approved):
LGTM. I left some debug tips [here](https://github.com/PyCQA/astroid/pull/1406#issuecomment-1053537537).

PR #1411, opened 2022-02-27
Add tests for position attributes of definitions with only docstrings

DanielNoord:
## Steps

## Description

I think this should fix the issue identified in https://github.com/PyCQA/pylint/pull/5802

If not, this is a nice change either way 😄 

## Type of Changes

### DIFF
--- a/tests/unittest_builder.py
+++ b/tests/unittest_builder.py
@@ -221,28 +221,28 @@ class FromToLineNoTest(unittest.TestCase):
 
         a = ast_module.body[0]
         assert isinstance(a, nodes.ClassDef)
-        assert a.fromlineno == 2
-        assert a.tolineno == 4
+        assert a.fromlineno == 1
+        assert a.tolineno == 3
 
         b = ast_module.body[1]
         assert isinstance(b, nodes.ClassDef)
-        assert b.fromlineno == 6
-        assert b.tolineno == 7
+        assert b.fromlineno == 5
+        assert b.tolineno == 6
 
         c = ast_module.body[2]
         assert isinstance(c, nodes.ClassDef)
-        assert c.fromlineno == 9
-        assert c.tolineno == 11
+        assert c.fromlineno == 8
+        assert c.tolineno == 10
 
         d = ast_module.body[3]
         assert isinstance(d, nodes.ClassDef)
-        assert d.fromlineno == 13
-        assert d.tolineno == 16
+        assert d.fromlineno == 12
+        assert d.tolineno == 15
 
         e = ast_module.body[4]
         assert isinstance(d, nodes.ClassDef)
-        assert e.fromlineno == 18
-        assert e.tolineno == 19
+        assert e.fromlineno == 17
+        assert e.tolineno == 18
 
     @staticmethod
     def test_function_with_docstring() -> None:
@@ -276,28 +276,28 @@ class FromToLineNoTest(unittest.TestCase):
 
         a = ast_module.body[0]
         assert isinstance(a, nodes.FunctionDef)
-        assert a.fromlineno == 2
-        assert a.tolineno == 4
+        assert a.fromlineno == 1
+        assert a.tolineno == 3
 
         b = ast_module.body[1]
         assert isinstance(b, nodes.FunctionDef)
-        assert b.fromlineno == 6
-        assert b.tolineno == 7
+        assert b.fromlineno == 5
+        assert b.tolineno == 6
 
         c = ast_module.body[2]
         assert isinstance(c, nodes.FunctionDef)
-        assert c.fromlineno == 9
-        assert c.tolineno == 11
+        assert c.fromlineno == 8
+        assert c.tolineno == 10
 
         d = ast_module.body[3]
         assert isinstance(d, nodes.FunctionDef)
-        assert d.fromlineno == 13
-        assert d.tolineno == 16
+        assert d.fromlineno == 12
+        assert d.tolineno == 15
 
         e = ast_module.body[4]
         assert isinstance(e, nodes.FunctionDef)
-        assert e.fromlineno == 18
-        assert e.tolineno == 21
+        assert e.fromlineno == 17
+        assert e.tolineno == 20
 
     def test_class_lineno(self) -> None:
         stmts = self.astroid.body
--- a/tests/unittest_builder.py
+++ b/tests/unittest_builder.py
@@ -191,6 +191,114 @@ class FromToLineNoTest(unittest.TestCase):
             assert c.fromlineno == 13
         assert c.tolineno == 14
 
+    @staticmethod
+    def test_class_with_docstring() -> None:
+        """Test class nodes which only have docstrings."""
+        code = textwrap.dedent(
+            '''
+        class A:
+            """My docstring"""
+            var = 1
+
+        class B:
+            """My docstring"""
+
+        class C:
+            """My docstring
+            is long."""
+
+        class D:
+            """My docstring
+            is long.
+            """
+
+        class E:
+            ...
+        '''
+        )
+
+        ast_module: nodes.Module = builder.parse(code)  # type: ignore[assignment]
+
+        a = ast_module.body[0]
+        assert isinstance(a, nodes.ClassDef)
+        assert a.fromlineno == 2
+        assert a.tolineno == 4
+
+        b = ast_module.body[1]
+        assert isinstance(b, nodes.ClassDef)
+        assert b.fromlineno == 6
+        assert b.tolineno == 7
+
+        c = ast_module.body[2]
+        assert isinstance(c, nodes.ClassDef)
+        assert c.fromlineno == 9
+        assert c.tolineno == 11
+
+        d = ast_module.body[3]
+        assert isinstance(d, nodes.ClassDef)
+        assert d.fromlineno == 13
+        assert d.tolineno == 16
+
+        e = ast_module.body[4]
+        assert isinstance(d, nodes.ClassDef)
+        assert e.fromlineno == 18
+        assert e.tolineno == 19
+
+    @staticmethod
+    def test_function_with_docstring() -> None:
+        """Test function defintions with only docstrings."""
+        code = textwrap.dedent(
+            '''
+        def a():
+            """My docstring"""
+            var = 1
+
+        def b():
+            """My docstring"""
+
+        def c():
+            """My docstring
+            is long."""
+
+        def d():
+            """My docstring
+            is long.
+            """
+
+        def e(a, b):
+            """My docstring
+            is long.
+            """
+        '''
+        )
+
+        ast_module: nodes.Module = builder.parse(code)  # type: ignore[assignment]
+
+        a = ast_module.body[0]
+        assert isinstance(a, nodes.FunctionDef)
+        assert a.fromlineno == 2
+        assert a.tolineno == 4
+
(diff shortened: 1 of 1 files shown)


DanielNoord:
Oops, incorrect copy paste 😅

DanielNoord:
One example is still inferred incorrectly:
```
def func(a,
    b,
    c,
):
    """Test."""
```

But is due to the line numbering of `Arguments` not being correct. Therefore that is out of the scope of this PR (I think).

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1979,6 +1979,19 @@ def frame(self: T, *, future: Literal[None, True] = None) -> T:
...
 
+    @decorators_mod.cachedproperty
+    def tolineno(self) -> Optional[int]:
+        """The last line that this node appears on in the source code."""
+        if self.end_lineno is not None:
+            return self.end_lineno
+        last_child = self.last_child()
+        # This is a hack to approximate the final line of a node
+        # with only a docstring, as those are not represented as child.
+        if isinstance(last_child, Arguments):
```
cdce8p, line 1990: I do wonder how resilient that will be. `last_child` can be any astroid node. For `FunctionDef` those fields are `("decorators", "bases", "keywords", "body")`.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -3254,3 +3267,17 @@ def frame(self: T, *, future: Literal[None, True] = None) -> T:
...
         """
         return self
+
+    @decorators_mod.cachedproperty
+    def tolineno(self) -> Optional[int]:
+        """The last line that this node appears on in the source code."""
+        if self.end_lineno is not None:
+            return self.end_lineno
+        last_child = self.last_child()
+        if last_child is None:
```
cdce8p, line 3277: Same question for `ClassDef`, the possible fields `("decorators", "bases", "keywords", "body")`.

cdce8p (changes requested):
I'm not really sure this is the best solution here. The code you wrote can break just too easily. Won't #1276 help with this issue?

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: https://github.com/PyCQA/astroid/blob/56f5f055e4847e9fc2b74162ccad8f2a11db3bfc/astroid/nodes/node_ng.py#L445-L457

This is copied from the original implementation. However, since `FunctionDef` always has an `Arguments` `last_child is None` doesn't work. This tries to replicate that.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: https://github.com/PyCQA/astroid/blob/56f5f055e4847e9fc2b74162ccad8f2a11db3bfc/astroid/nodes/node_ng.py#L445-L457

This is just copied from the original implementation. Only thing that changes here is the `if self.doc` line.

DanielNoord:
> I'm not really sure this is the best solution here. The code you wrote can break just too easily. Won't #1276 help with this issue?

Yes and no. If there is an actual node `pylint` can use it to infer the length of a definition to determine the "scope" of a disable. However, I don't think that that should be `pylint's` task. We should probably do this in `astroid`.

When #1276 lands this portion of code could be updated to use the `end_lineno` of the `doc_node` instead of counting `"\n"`. If we ever improve on `doc_node` then it will trickle down. For now, I think this is actually okay as it is not quite that different from the standard `tolineno`.

astroid/nodes/scoped_nodes/scoped_nodes.py
cdce8p: What if the function has a return type annotation?
```py
def func() -> int:
    """Hello World"""
```
`if isinstance(last_child, Arguments)` won't work then and it would default to `last_child.tolineno`.

astroid/nodes/scoped_nodes/scoped_nodes.py
cdce8p: What about this?
```py
class A(B, var=42):
    """Hello World"""
```

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Hmm, I guess this is correct.
```python
def func() -> None:
    """Hello World"""
```

But I mean how often do people actually write function and classes with only a `docstring`. This problem existed in previous versions as well, as `tolineno` was broken for those versions as well. Only because we are now using `end_lineno` we noticed this. I wonder how big of a problem this actually is.

That said, I'm also fine with putting more effort into #1276 as finding all edge cases here will probably take considerable time. We will need to take either approach before we can accept `2.1x` in `pylint` though.

cdce8p: This might be relative strait forward to solve (at least compared to the issue with ClassDef` below).

We know `last_child` will at least be `Arguments`, so we don't need to care about decorators anymore. Something like this could work
```py
if not body and last_child:
    ...
```

astroid/nodes/scoped_nodes/scoped_nodes.py
cdce8p: Again, there are ways to work around it, although slightly more complicated once here.
We need:
* Some way to filter `decorators`
* Make sure `last_child` is not a `body` node
* And if `last_child` is not None but we don't have a `doc`, we probably shouldn't fall back to `self.fromlineno`.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord:
```python
def func() -> None:
    """Hello World"""
```

I don't think that will always work. `last_child` is a `Const` `None` for this example. We could some comparison with `last_child.tolineno` and the `end_lineno` of the `FunctionDef` and see if they are similar. But then when the return type annotation is on a new line we would still wrongly infer the line number.
I think this is getting unnecessarily complicated knowing that #1276 is on the horizon. I think it makes more sense to help land that quickly and then revisit this.

In the meantime we could revert the commit that changed the `tolineno` behaviour to make `astroid` usable for `pylint`, but that might also be a little too much.

Pierre-Sassoulas (commented):
Thank you @DanielNoord ! If I understand correctly we'll be able to remove this when we drop python 3.7 support ?

DanielNoord:
> Thank you @DanielNoord ! If I understand correctly we'll be able to remove this when we drop python 3.7 support ?

Yes! From 3.8 we will always have access to `end_lineno` (and could possible consider deprecating `tolineno`).

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1979,6 +1979,19 @@ def frame(self: T, *, future: Literal[None, True] = None) -> T:
         """
         return self
 
+    @decorators_mod.cachedproperty
+    def tolineno(self) -> Optional[int]:
+        """The last line that this node appears on in the source code."""
```
Pierre-Sassoulas, line 1984:
```suggestion
    def tolineno(self) -> Optional[int]:
        """The last line that this node appears on in the source code.
        
        TODO: Remove when we do not support python 3.7 anymore.
        """
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -3254,3 +3267,17 @@ def frame(self: T, *, future: Literal[None, True] = None) -> T:
         :returns: The node itself.
         """
         return self
+
+    @decorators_mod.cachedproperty
+    def tolineno(self) -> Optional[int]:
+        """The last line that this node appears on in the source code."""
```
Pierre-Sassoulas, line 3273:
```suggestion
    def tolineno(self) -> Optional[int]:
        """The last line that this node appears on in the source code.
          
        TODO: Remove when we do not support python 3.7 anymore.
        """
```

astroid/nodes/scoped_nodes/scoped_nodes.py
cdce8p: Not sure that comment makes much sense. The whole `tolineno` logic can be updated / deprecated then.

DanielNoord:
Since `tolineno` uses `last_child` the addition of `doc_node` to `astroid_fields` should actually fix this without any issues.
I will wait for the PR to land but then convert this to a "add tests" PR (and possible remove the changelog entry).

DanielNoord:
Something is up with the cache, I'm seeing multiple failed "cache restorations" across PRs. For now, rerunning seems to work. Might just be GitHub Servers struggling with the end of day CI runs from the US that typically occur on Monday.

Pierre-Sassoulas (approved):
LGTM !

tests/unittest_builder.py
```diff
@@ -191,6 +191,114 @@ class C:
             assert c.fromlineno == 13
         assert c.tolineno == 14
 
+    @staticmethod
+    def test_class_with_docstring() -> None:
+        """Test class nodes which only have docstrings."""
+        code = textwrap.dedent(
+            '''
```
cdce8p, line 198:
```suggestion
            '''\
```
I've started adding backslashes to the first line. That way it starts counting with `1` like you would expect, instead of `2`. It's not something I would update all tests for, but useful for new ones.

tests/unittest_builder.py
```diff
@@ -191,6 +191,114 @@ class C:
...
+            """My docstring
+            is long.
+            """
+
+        class E:
+            ...
+        '''
+        )
+
+        ast_module: nodes.Module = builder.parse(code)  # type: ignore[assignment]
```
cdce8p, line 220:
```suggestion
        ast_module = builder.parse(code)
```
Why is this necessary? `builder.parse(...)` is annotated to return a `nodes.Module` instance.

tests/unittest_builder.py
```diff
@@ -191,6 +191,114 @@ class C:
...
+        e = ast_module.body[4]
+        assert isinstance(d, nodes.ClassDef)
+        assert e.fromlineno == 18
+        assert e.tolineno == 19
+
+    @staticmethod
+    def test_function_with_docstring() -> None:
+        """Test function defintions with only docstrings."""
+        code = textwrap.dedent(
+            '''
```
cdce8p, line 251:
```suggestion
            '''\
```

tests/unittest_builder.py
```diff
@@ -191,6 +191,114 @@ class C:
...
+            """
+
+        def e(a, b):
+            """My docstring
+            is long.
+            """
+        '''
+        )
+
+        ast_module: nodes.Module = builder.parse(code)  # type: ignore[assignment]
```
cdce8p, line 275:
```suggestion
        ast_module = builder.parse(code)
```

cdce8p (changes requested):

tests/unittest_builder.py
DanielNoord: Copied from the test above 😉

tests/unittest_builder.py
Pierre-Sassoulas: Very clever !

cdce8p (approved):

PR #1415, opened 2022-03-01
Add some typing to ``astroid.inference``

DanielNoord:
## Steps

## Description

The typing on this is incorrect I believe.

https://github.com/PyCQA/astroid/blob/c0d2e5c89a1525e9f500c43b04529628a70e070f/astroid/util.py#L33-L34

I don't think `mypy` recognises that the decorator on `Uninferable` makes `util.Uninferable` an instance instead of the type. I checked the issue tracker and there is no issue about this yet, so I am slightly doubting whether I am in correct in this, but I think this is the correct fix.

Edit: Seeing as `pyright` does get this right I opened https://github.com/python/mypy/issues/12265.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -34,7 +34,7 @@ import ast
 import functools
 import itertools
 import operator
-from typing import Any, Callable, Dict, Iterable, Optional
+from typing import Any, Callable, Dict, Iterable, Iterator, Optional, Union
 
 import wrapt
 
@@ -824,7 +824,7 @@ def _to_literal(node: nodes.NodeNG) -> Any:
 
 def _do_compare(
     left_iter: Iterable[nodes.NodeNG], op: str, right_iter: Iterable[nodes.NodeNG]
-) -> "bool | type[util.Uninferable]":
+) -> Union[bool, util.Uninferable]:
     """
     If all possible combinations are either True or False, return that:
     >>> _do_compare([1, 2], '<=', [3, 4])
@@ -839,17 +839,17 @@ def _do_compare(
     """
     retval = None
     if op in UNINFERABLE_OPS:
-        return util.Uninferable
+        return util.Uninferable  # type: ignore[return-value]
     op_func = COMPARE_OPS[op]
 
     for left, right in itertools.product(left_iter, right_iter):
         if left is util.Uninferable or right is util.Uninferable:
-            return util.Uninferable
+            return util.Uninferable  # type: ignore[return-value]
 
         try:
             left, right = _to_literal(left), _to_literal(right)
         except (SyntaxError, ValueError, AttributeError):
-            return util.Uninferable
+            return util.Uninferable  # type: ignore[return-value]
 
         try:
             expr = op_func(left, right)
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
LGTM, thank you @DanielNoord

astroid/inference.py
```diff
@@ -862,14 +862,15 @@ def _do_compare(
             return util.Uninferable
             # (or both, but "True | False" is basically the same)
 
+    assert retval is not None
     return retval  # it was all the same value
```
cdce8p, line 866: Can you add a type hint to the definition of `retval` in L840?

cdce8p (commented):
It might make sense to update the PR title, I think.

astroid/inference.py
```diff
@@ -824,7 +824,7 @@ def _to_literal(node: nodes.NodeNG) -> Any:
 
 def _do_compare(
     left_iter: Iterable[nodes.NodeNG], op: str, right_iter: Iterable[nodes.NodeNG]
-) -> "bool | type[util.Uninferable]":
+) -> Union[bool, util.Uninferable]:
```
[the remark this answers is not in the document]
cdce8p: > Ah yeah, I didn't think about the implications of this. Well, then we also don't need the ignores because since `mypy` infers `Uninferable` incorrectly the typing is actually correct in its eyes.

They are offset by 1 (i.e. from the right column to the middle one). Fortunately for us, all are so it still works (at least here).

astroid/inference.py
```diff
@@ -837,7 +843,7 @@ def _do_compare(
     >>> _do_compare([1, 3], '<=', [2, 4])
     util.Uninferable
     """
-    retval = None
+    retval: Union[Literal[None], bool, Type[util.Uninferable]] = None
```
cdce8p, line 846:
```suggestion
    retval: Union[None, bool] = None
```
`Literal[None]` is not necessary. Both mypy and pyright understand `None`.

Additionally, `Uninferable` is never assigned to `retval`, only returned directly.

astroid/inference.py
```diff
@@ -58,6 +59,11 @@
 from astroid.interpreter import dunder_lookup
 from astroid.manager import AstroidManager
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
```
cdce8p, line 66:
```suggestion
```

astroid/inference.py
```diff
@@ -34,7 +34,8 @@
 import functools
 import itertools
 import operator
-from typing import Any, Callable, Dict, Iterable, Optional
+import sys
```
cdce8p, line 37:
```suggestion
```

astroid/inference.py
```diff
@@ -862,14 +868,15 @@ def _do_compare(
...
     return retval  # it was all the same value
 
 
 def _infer_compare(
     self: nodes.Compare, context: Optional[InferenceContext] = None
-) -> Any:
+) -> Iterator[Union[nodes.Const, Type[util.Uninferable]]]:
     """Chained comparison inference logic."""
-    retval = True
+    retval: Union[bool, Type[util.Uninferable]] = True
```
cdce8p, line 879: This however is correct. `Uninferable` can be assigned in the for loop.

cdce8p (changes requested):

astroid/inference.py
DanielNoord: Hmm there a quite a bit of `Literal[None]`'s in the codebase. Perhaps we should update those at some point.

cdce8p (approved):

PR #1418, opened 2022-03-01
Add dependabot configuration

DanielNoord:
## Steps

- [x] Write a good description on what the PR does.

## Description

@Pierre-Sassoulas Is there a reason this was not set up for `astroid` that I should know about?

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

PR #1419, opened 2022-03-01
Create dependabot configuration

Pierre-Sassoulas:
## Type of Changes

## Related Issue

Closes #1418

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependency"
+    open-pull-requests-limit: 10
+    rebase-strategy: "disabled"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependency"
+    open-pull-requests-limit: 10
+    rebase-strategy: "disabled"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
     labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
DanielNoord, line 9:
```suggestion
      interval: "daily"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+    labels:
+      - "dependency"
+    open-pull-requests-limit: 10
+    rebase-strategy: "disabled"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
DanielNoord, line 18:
```suggestion
      interval: "daily"
```

DanielNoord (changes requested):
🎉

Pierre-Sassoulas:
Outch everything updated at once kinda hurt github actions 😄 I think we'll have to relaunch one by one and take it slow

DanielNoord:
We're just failing on concurrency for `readthedocs`. I would be okay with merging if all else passes and finally making sure that `main` builds the docs (perhaps with a manual relaunch).

Pierre-Sassoulas:
Let's do that, except for the sphinx update

Pierre-Sassoulas:
The coverage is also blocking, let's merge without the coverage, except for the upgrade of coverall, coverage, and pytest-cov.

DanielNoord:
I've been cancelling some workflows on `main` (to avoid duplication). I'll let some pass when we near the end.

Pierre-Sassoulas:
Nice I did not think of that. Maybe we could cancel everything and then rebase the last one instead ?

DanielNoord:
Rest of the PRs can't be merged.

Btw, I don't see a `readthedocs` action on `main`. Are we sure it builds correctly from `main`?

DanielNoord:
Never mind, apparently there is no `action` on `main`. Only on PRs. I double-checked in other repo's.

Guess we're done here 😄

PR #1432, opened 2022-03-02
Deprecate ``doc`` attribute

### LINKED ISSUE #1414: Deprecate `doc` attribute?
With the new addition of `doc_node` (#1276), should the "old" `doc` attribute be deprecated and later removed?

/CC: @DanielNoord

DanielNoord:
## Steps

## Description

Closes #1414.

/CC @cdce8p I just saw I took a completely different approach than what you had in mind. Is this something we should pursue or do you want to use your method? (I didn't look completely in to how that works)

## Type of Changes

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -570,12 +570,16 @@ def infer_property(node, context=None):
     prop_func = objects.Property(
         function=inferred,
         name=inferred.name,
-        doc=getattr(inferred, "doc", None),
+        doc=getattr(inferred, "doc_node.value", None),
         lineno=node.lineno,
         parent=node,
         col_offset=node.col_offset,
     )
-    prop_func.postinit(body=[], args=inferred.args)
+    prop_func.postinit(
+        body=[],
+        args=inferred.args,
+        doc_node=getattr(inferred, "doc_node", None),
+    )
     return prop_func
 
 
--- a/astroid/brain/brain_functools.py
+++ b/astroid/brain/brain_functools.py
@@ -101,7 +101,9 @@ def _functools_partial_inference(node, context=None):
     partial_function = objects.PartialFunction(
         call,
         name=inferred_wrapped_function.name,
-        doc=inferred_wrapped_function.doc,
+        doc=inferred_wrapped_function.doc_node.value
+        if inferred_wrapped_function.doc_node
+        else None,
         lineno=inferred_wrapped_function.lineno,
         col_offset=inferred_wrapped_function.col_offset,
         parent=node.parent,
@@ -113,6 +115,7 @@ def _functools_partial_inference(node, context=None):
         returns=inferred_wrapped_function.returns,
         type_comment_returns=inferred_wrapped_function.type_comment_returns,
         type_comment_args=inferred_wrapped_function.type_comment_args,
+        doc_node=inferred_wrapped_function.doc_node,
     )
     return iter((partial_function,))
 
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -598,9 +598,7 @@ class GeneratorModel(FunctionModel):
     @property
     def attr___doc__(self):
         return node_classes.Const(
-            value=self._instance.parent.doc_node.value
-            if self._instance.parent.doc_node
-            else None,
+            value=getattr(self._instance.parent.doc_node, "value", None),
             parent=self._instance,
         )
 
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -163,7 +163,10 @@ class ModuleModel(ObjectModel):
 
     @property
     def attr___doc__(self):
-        return node_classes.Const(value=self._instance.doc, parent=self._instance)
+        return node_classes.Const(
+            value=self._instance.doc_node.value if self._instance.doc_node else None,
+            parent=self._instance,
+        )
 
     @property
     def attr___file__(self):
@@ -209,7 +212,10 @@ class FunctionModel(ObjectModel):
 
     @property
     def attr___doc__(self):
-        return node_classes.Const(value=self._instance.doc, parent=self._instance)
+        return node_classes.Const(
+            value=self._instance.doc_node.value if self._instance.doc_node else None,
+            parent=self._instance,
+        )
 
     @property
     def attr___qualname__(self):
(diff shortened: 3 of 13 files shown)


astroid/raw_building.py
```diff
@@ -119,6 +119,13 @@ def build_module(name: str, doc: Optional[str] = None) -> nodes.Module:
 def build_class(name, basenames=(), doc=None):
     """create and initialize an astroid ClassDef node"""
     node = nodes.ClassDef(name, doc)
```
Pierre-Sassoulas, line 121: I don't understand why we have to use the postinit here ?

astroid/brain/brain_builtin_inference.py
```diff
@@ -570,12 +570,16 @@ def infer_property(node, context=None):
...
+        doc=getattr(inferred, "doc_node.value", None),
         lineno=node.lineno,
         parent=node,
         col_offset=node.col_offset,
     )
-    prop_func.postinit(body=[], args=inferred.args)
+    prop_func.postinit(
+        body=[],
+        args=inferred.args,
+        doc_node=getattr(inferred, "doc_node", None),
```
Pierre-Sassoulas, line 581: Isn't it set line 573 in the constructor of Property ?

DanielNoord: That's the `doc` for `Property.doc`, this is for `doc_node`. I don't know why exactly but not using `getattr` crashes here, so we need to use `getattr`.

astroid/brain/brain_functools.py
```diff
@@ -101,7 +101,9 @@ def _functools_partial_inference(node, context=None):
     partial_function = objects.PartialFunction(
         call,
         name=inferred_wrapped_function.name,
-        doc=inferred_wrapped_function.doc,
+        doc=inferred_wrapped_function.doc_node.value
```
cdce8p, line 104: It seems like both `PartialFunction` and `Property` should accept `doc_node` instead.

astroid/interpreter/objectmodel.py
```diff
@@ -584,7 +598,10 @@ def attr___name__(self):
     @property
     def attr___doc__(self):
         return node_classes.Const(
-            value=self._instance.parent.doc, parent=self._instance
+            value=self._instance.parent.doc_node.value
+            if self._instance.parent.doc_node
+            else None,
```
cdce8p, line 603:
```suggestion
            value=getattr(self._instance.parent.doc_node, "value", None),
```
Not the best solution but a bit less ugly.

astroid/nodes/as_string.py
```diff
@@ -183,7 +183,7 @@ def visit_classdef(self, node):
             args.append("metaclass=" + node._metaclass.accept(self))
         args += [n.accept(self) for n in node.keywords]
         args = f"({', '.join(args)})" if args else ""
-        docs = self._docs_dedent(node.doc) if node.doc else ""
+        docs = self._docs_dedent(node.doc_node.value) if node.doc_node else ""
```
cdce8p, line 186: Should `self._docs_dedent` accept `nodes.Const | None` instead?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -457,8 +457,8 @@ def __init__(
         self.name = name
         """The name of the module."""
 
-        self.doc = doc
-        """The module docstring."""
+        # TODO: Remove this attribute
+        self._doc = doc
```
cdce8p, line 461: At least for `Module.__init__` `doc` is still a required argument. I think we can add a default value.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -513,6 +513,16 @@ def _get_stream(self):
             return stream
         return None
 
+    @property
+    def doc(self) -> Optional[str]:
+        """The module docstring."""
```
cdce8p, line 518: For full backwards compatibility, we also need a `@doc.setter` method.

cdce8p (changes requested):
This PR contains a lot of different things at the moment which make it difficult to review. I left some early comments, however we should try to get some order into these changes by splitting them up. #1433 doesn't help yet, as it too contains multiple things.

What I have mind
1. Update `Module`, `ClassDef`, `FunctionDef`.
  1.1 The `doc` attribute in `__init__` should be optional with default. Required if we later want to remove it from existing calls.
  1.2 Property and property setter method for `doc` with DeprecationWarning
2. Update other (similar) classes, mainly `object.Property` and `PartialFunction`, with similar changes if necessary.
3. Update code base, to remove the `doc` attribute where it's passed to `__init__` and replace it with `doc_node` to `postinit`. Also update functions which could handle an optional doc_node instead of a string -> `AsStringVisitor`.
4. Add decorator to deprecate `doc` attribute from `__init__` calls.
5. ... Did I miss something 🤔  ... _probably_
6. Update all attribute calls to `.doc` if not already done. `ClassModel.attr___doc__`, `AsStringVisitor._docs_dedent`, ...

DanielNoord:
> This PR contains a lot of different things at the moment which make it difficult to review. I left some early comments, however we should try to get some order into these changes by splitting them up. #1433 doesn't help yet, as it too contains multiple things.
> 
> What I have mind
> 
> 1. Update `Module`, `ClassDef`, `FunctionDef`.
>    1.1 The `doc` attribute in `__init__` should be optional with default. Required if we later want to remove it from existing calls.
>    1.2 Property and property setter method for `doc` with DeprecationWarning
> 2. Update other (similar) classes, mainly `object.Property` and `PartialFunction`, with similar changes if necessary.
> 3. Update code base, to remove the `doc` attribute where it's passed to `__init__` and replace it with `doc_node` to `postinit`. Also update functions which could handle an optional doc_node instead of a string -> `AsStringVisitor`.
> 4. Add decorator to deprecate `doc` attribute from `__init__` calls.
> 5. ... Did I miss something 🤔  ... _probably_

Hmm, that seems like quite a long list. I didn't consider removing `doc` from the `__init__` calls initially and only considered accessing the `doc` attribute.
Although I agree this is something we should do, might it make sense to release `2.11` without the deprecation? That way `pylint` can move on as well.

I'll make a PR that does 1 first.

astroid/brain/brain_functools.py
DanielNoord: Not in their `init` right? They share the `postinit` from `FunctionDef` and don't redefine it, so I don't think we actually need to change those classes? We just need to pass different arguments to their `postinit`.

astroid/nodes/as_string.py
DanielNoord: See https://github.com/PyCQA/astroid/pull/1448

DanielNoord:
Closing this in favour of various other PRs.

PR #1433, opened 2022-03-02
Do not use ``doc`` attribute internally

DanielNoord:
## Steps

## Type of Changes

## Related Issue


cdce8p (changes requested):
See https://github.com/PyCQA/astroid/pull/1432#pullrequestreview-898319366

astroid/interpreter/objectmodel.py
```diff
@@ -163,7 +163,10 @@ def attr___name__(self):
 
     @property
     def attr___doc__(self):
-        return node_classes.Const(value=self._instance.doc, parent=self._instance)
+        return node_classes.Const(
+            value=self._instance.doc_node.value if self._instance.doc_node else None,
```
cdce8p, line 167:
```suggestion
            value=getattr(self._instance.doc_node, "value", None),
```
Let's try to replace the ifs with `getattr`. The should simplify that.

astroid/raw_building.py
```diff
@@ -119,6 +119,13 @@ def build_module(name: str, doc: Optional[str] = None) -> nodes.Module:
 def build_class(name, basenames=(), doc=None):
     """create and initialize an astroid ClassDef node"""
     node = nodes.ClassDef(name, doc)
```
cdce8p, line 121:
```suggestion
    node = nodes.ClassDef(name)
```

astroid/raw_building.py
```diff
@@ -119,6 +119,13 @@ def build_module(name: str, doc: Optional[str] = None) -> nodes.Module:
...
+    node.postinit(
+        [],
+        [],
+        None,
```
cdce8p, line 125: Can you add keyword arguments here?

cdce8p (changes requested):

cdce8p (approved):

PR #1434, opened 2022-03-03
Deprecate accessing the ``doc`` attribute on nodes

DanielNoord:
## Steps

## Description

Step 1 as discussed by @cdce8p in https://github.com/PyCQA/astroid/pull/1432#pullrequestreview-898319366.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,7 +11,7 @@ Release date: TBA
 
 * Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
   ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node`` attribute.
-  Note: ``doc_node`` is ``nodes.Const | None`` whereas ``doc`` was ``str | None``.
+  Note: ``doc_node`` is an (optional) ``nodes.Const`` whereas ``doc`` was an (optional) ``str``.
 
 * Replace custom ``cachedproperty`` with ``functools.cached_property`` and deprecate it
   for Python 3.8+.
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,9 @@ Release date: TBA
 * Add new (optional) ``doc_node`` attribute to ``nodes.Module``, ``nodes.ClassDef``,
   and ``nodes.FunctionDef``.
 
+* Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
+  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node.value`` attribute.
+
 * Replace custom ``cachedproperty`` with ``functools.cached_property`` and deprecate it
   for Python 3.8+.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,7 +10,8 @@ Release date: TBA
   and ``nodes.FunctionDef``.
 
 * Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
-  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node.value`` attribute.
+  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node`` attribute.
+  Note: ``doc_node`` is ``nodes.Const | None`` whereas ``doc`` was ``str | None``.
 
 * Replace custom ``cachedproperty`` with ``functools.cached_property`` and deprecate it
   for Python 3.8+.
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -438,7 +438,7 @@ class Module(LocalsDictNodeNG):
     def __init__(
         self,
         name: str,
-        doc: Optional[str],
+        doc: Optional[str] = None,
         file: Optional[str] = None,
         path: Optional[List[str]] = None,
         package: Optional[bool] = None,
@@ -463,7 +463,7 @@ class Module(LocalsDictNodeNG):
         self.name = name
         """The name of the module."""
 
-        self.doc = doc
+        self._doc = doc
         """The module docstring."""
 
         self.file = file
@@ -509,6 +509,25 @@ class Module(LocalsDictNodeNG):
         """
         self.body = body
         self.doc_node = doc_node
+
+    @property
+    def doc(self) -> Optional[str]:
+        """The module's docstring."""
+        warnings.warn(
+            "The 'Module.doc' attribute is deprecated, "
+            "use 'Module.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        return self._doc
+
+    @doc.setter
+    def doc(self, value: Optional[str]) -> None:
+        warnings.warn(
+            "The 'Module.doc' attribute is deprecated, "
+            "use 'Module.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        self._doc = value
 
     def _get_stream(self):
         if self.file_bytes is not None:
@@ -1517,7 +1536,7 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
     def __init__(
         self,
         name=None,
-        doc=None,
+        doc: Optional[str] = None,
         lineno=None,
         col_offset=None,
         parent=None,
@@ -1530,7 +1549,6 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
         :type name: str or None
 
         :param doc: The function's docstring.
-        :type doc: str or None
 
         :param lineno: The line that this node appears on in the source code.
         :type lineno: int or None
@@ -1555,11 +1573,8 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
         :type name: str or None
         """
 
-        self.doc = doc
-        """The function's docstring.
-
-        :type doc: str or None
-        """
+        self._doc = doc
+        """The function's docstring."""
 
         self.doc_node: Optional[Const] = None
         """The doc node associated with this node."""
@@ -1616,6 +1631,25 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
         self.type_comment_args = type_comment_args
         self.position = position
         self.doc_node = doc_node
+
+    @property
+    def doc(self) -> Optional[str]:
+        """The function's docstring."""
+        warnings.warn(
+            "The 'FunctionDef.doc' attribute is deprecated, "
+            "use 'FunctionDef.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        return self._doc
+
+    @doc.setter
+    def doc(self, value: Optional[str]) -> None:
+        warnings.warn(
+            "The 'FunctionDef.doc' attribute is deprecated, "
+            "use 'FunctionDef.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        self._doc = value
 
     @cached_property
     def extra_decorators(self) -> List[node_classes.Call]:
@@ -2166,7 +2200,7 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
     def __init__(
         self,
         name=None,
-        doc=None,
+        doc: Optional[str] = None,
         lineno=None,
         col_offset=None,
         parent=None,
@@ -2178,8 +2212,7 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
         :param name: The name of the class.
         :type name: str or None
 
-        :param doc: The function's docstring.
-        :type doc: str or None
+        :param doc: The class' docstring.
 
         :param lineno: The line that this node appears on in the source code.
         :type lineno: int or None
@@ -2231,11 +2264,8 @@ class ClassDef(mixins.FilterStmtsMixin, LocalsDictNodeNG, node_classes.Statement
         :type name: str or None
         """
 
-        self.doc = doc
-        """The class' docstring.
-
-        :type doc: str or None
-        """
+        self._doc = doc
+        """The class' docstring."""
 
         self.doc_node: Optional[Const] = None
         """The doc node associated with this node."""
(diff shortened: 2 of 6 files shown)


tests/unittest_scoped_nodes.py
```diff
@@ -956,7 +965,12 @@ def test_cls_special_attributes_1(self) -> None:
             self.assertEqual(
                 len(cls.getattr("__doc__")), 1, (cls, cls.getattr("__doc__"))
             )
-            self.assertEqual(cls.getattr("__doc__")[0].value, cls.doc)
+            with pytest.warns(DeprecationWarning) as records:
+                self.assertEqual(cls.getattr("__doc__")[0].value, cls.doc)
+                # TODO: This should be 1 once we update the use of 'doc' internally
+                assert len(records) == 2
+            # TODO: The doc_node attribute needs to be set in the raw_building process
```
DanielNoord, line 972: I'm not sure if this should be in here, but I think we should test this and with this TODO we will stumble upon this when we change L969-971.

DanielNoord:
2022 seems to be a rough year for GitHub Actions, something going on with the downloading of `actions/checkout` across multiple repositories. Tests pass on Linux and PyPy though, so don't let that fool you 😄

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1555,11 +1573,8 @@ def __init__(
         :type name: str or None
         """
 
-        self.doc = doc
-        """The function's docstring.
-
-        :type doc: str or None
-        """
+        self._doc = doc
```
Pierre-Sassoulas, line 1576:
```suggestion
        self._doc = None
        if doc is not None:
            warnings.warn("Specific warning for the constructor ?") 
            self.doc = doc
```
Would be the occasion to tell the user what he should do instead (use postinit ?). We can also use the property directly instead of assigning to ``_doc``.

DanielNoord: I think that is step 4 of @cdce8p's deprecation plan. For now we only want to warn when users access `node.doc`.

Pierre-Sassoulas (approved):

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -510,6 +510,25 @@ def postinit(self, body=None, *, doc_node: Optional[Const] = None):
...
+            "use 'Module.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        return self._doc
+
+    @doc.setter
+    def doc(self, value: Optional[str]) -> None:
+        warnings.warn(
+            "The 'Module.doc' attribute is deprecated, "
+            "use 'Module.doc_node.value' instead.",
```
cdce8p, line 527:
```suggestion
            "Setting the 'Module.doc' attribute is deprecated, "
            "use 'Module.doc_node' instead.",
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1617,6 +1632,25 @@ def postinit(
...
+            "use 'FunctionDef.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        return self._doc
+
+    @doc.setter
+    def doc(self, value: Optional[str]) -> None:
+        warnings.warn(
+            "The 'FunctionDef.doc' attribute is deprecated, "
+            "use 'FunctionDef.doc_node.value' instead.",
```
cdce8p, line 1649:
```suggestion
            "Setting the 'FunctionDef.doc' attribute is deprecated, "
            "use 'FunctionDef.doc_node' instead.",
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2256,6 +2286,25 @@ def __init__(
...
+            "use 'ClassDef.doc_node.value' instead.",
+            DeprecationWarning,
+        )
+        return self._doc
+
+    @doc.setter
+    def doc(self, value: Optional[str]) -> None:
+        warnings.warn(
+            "The 'ClassDef.doc' attribute is deprecated, "
+            "use 'ClassDef.doc_node.value' instead.",
```
cdce8p, line 2303:
```suggestion
            "Setting the 'ClassDef.doc' attribute is deprecated, "
            "use 'ClassDef.doc_node.value' instead.",
```

tests/unittest_builder.py
```diff
@@ -768,7 +768,10 @@ def test_module_base_props(self) -> None:
         """test base properties and method of an astroid module"""
         module = self.module
         self.assertEqual(module.name, "data.module")
-        self.assertEqual(module.doc, "test module for astroid\n")
+        with pytest.warns(DeprecationWarning) as records:
+            self.assertEqual(module.doc, "test module for astroid\n")
+            assert len(records) == 1
+        self.assertEqual(module.doc_node.value, "test module for astroid\n")
```
cdce8p, line 774: Should we make these test type safe by adding `isinstance(module.doc_node, nodes.Const)`?

tests/unittest_scoped_nodes.py
```diff
@@ -2637,5 +2651,49 @@ def test_non_frame_node():
         assert module.body[1].value.locals["x"][0].frame(future=True) == module
 
 
+def test_deprecation_of_doc_attribute() -> None:
```
cdce8p, line 2654: We should add some more tests here. One block to make sure we don't unintentionally raise `DeprecationWarning`.
```py
    doc_node = nodes.Const("Docstring")
        # If 'doc' isn't passed to Module, ClassDef, FunctionDef,
        # no DeprecationWarning should be raised
        # Modify warnings filter to raise error for DeprecationWarning
        warnings.simplefilter("error", DeprecationWarning)
        node_module = nodes.Module(name="MyModule")
        node_module.postinit(body=[], doc_node=doc_node)
        assert node_module.doc_node == doc_node
        node_class = nodes.ClassDef(name="MyClass")
        node_class.postinit(bases=[], body=[], decorators=[], doc_node=doc_node)
        assert node_class.doc_node == doc_node
        node_func = nodes.FunctionDef(name="MyFunction")
        node_func.postinit(args=nodes.Arguments(), body=[], doc_node=doc_node)
        assert node_func.doc_node == doc_node
```

And another to make sure, the `doc` node is assigned even if only `doc_node` is passed to `postinit`. Otherwise we can't remove the old usage without breaking backwards compatibility. _This currently fails._
```py
    # Test 'doc' attribute if only 'doc_node' is passed
        assert node_module.doc == "Docstring"
        assert len(records) == 1
        assert node_class.doc == "Docstring"
        assert len(records) == 1
        assert node_func.doc == "Docstring"
        assert len(records) == 1
```

cdce8p (changes requested):
We should also update the `postinit` methods to allow setting the `doc` attribute by passing `doc_node`.
```py
        if doc_node:
            self._doc = doc_node.value
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -510,6 +510,25 @@ def postinit(self, body=None, *, doc_node: Optional[Const] = None):
         self.body = body
         self.doc_node = doc_node
 
+    @property
+    def doc(self) -> Optional[str]:
+        """The module's docstring."""
+        warnings.warn(
+            "The 'Module.doc' attribute is deprecated, "
+            "use 'Module.doc_node.value' instead.",
```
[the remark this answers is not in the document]
DanielNoord: Are you sure? If you want to directly replicate `.doc` you should use `doc_node.value`. That's why I changed this to mention `.value`.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: I intended this PR to only deprecate `accessing` and doing `setting` in a later PR. See (future) other comment.

tests/unittest_builder.py
DanielNoord: Hm, we could but aren't we testing this in other places as well? Feels a bit like redundant double testing, but I'll happily change!

DanielNoord:
> We should also update the `postinit` methods to allow setting the `doc` attribute by passing `doc_node`.
> 
> ```python
>         if doc_node:
>             self._doc = doc_node.value
> ```

I intended to only tackle accessing in this PR (trying to split this job in as a small as possible steps as possible).
What I thought was 1) deprecate accessing, 2) pass `doc_node` in all cases where necessary, 3) deprecate setting and passing `doc`.
I'll happily change things if necessary, but I think if we only want to deprecate accessing in this PR it should be good (of course, with the textual changes).

astroid/nodes/scoped_nodes/scoped_nodes.py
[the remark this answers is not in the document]
cdce8p: `doc_node.value` can be misleading as it suggests a 1 to 1 replacement is possible. That's not the case. If `doc_node` is `None` an exception would be raised. That's why I think we should instead mention just `doc_node`. It's not ideal, but we have to trust that the name change is enough of a difference to notice the change.

In the end, it's more for use anyway than anyone else.

tests/unittest_builder.py
cdce8p: I'm split on this one. It wouldn't be strictly necessary, but maybe a bit better for type checking.

DanielNoord: I'll change. Give me a coupe of minutes!

DanielNoord: See latest commit.

cdce8p:
> I intended to only tackle accessing in this PR (trying to split this job in as a small as possible steps as possible).
> What I thought was 1) deprecate accessing, 2) pass `doc_node` in all cases where necessary, 3) deprecate setting and passing `doc`.
> I'll happily change things if necessary, but I think if we only want to deprecate accessing in this PR it should be good (of course, with the textual changes).

I think we have a slightly different view here. My intention for this first PR would be to modify `Module`, `ClassDef`, and `FunctionDef` such that we can later just replace `doc_node` with `doc`. The only thing missing would be the assignment in `postinit`.

Another aspect, when I view a PR that introduces a Deprecation warning, either an alternative should already be in place or should be added in the same PR. Otherwise we create a state that will always emit a warning with no option to turn it off. Yeah, we would do it in a followup, but I still think it's a good practice to follow nevertheless. Especially if the last required changes are minor once like these here.

DanielNoord:
> > I intended to only tackle accessing in this PR (trying to split this job in as a small as possible steps as possible).
> > What I thought was 1) deprecate accessing, 2) pass `doc_node` in all cases where necessary, 3) deprecate setting and passing `doc`.
> > I'll happily change things if necessary, but I think if we only want to deprecate accessing in this PR it should be good (of course, with the textual changes).
> 
> I think we have a slightly different view here. My intention for this first PR would be to modify `Module`, `ClassDef`, and `FunctionDef` such that we can later just replace `doc_node` with `doc`. The only thing missing would be the assignment in `postinit`.
> 
> Another aspect, when I view a PR that introduces a Deprecation warning, either an alternative should already be in place or should be added in the same PR. Otherwise we create a state that will always emit a warning with no option to turn it off. Yeah, we would do it in a followup, but I still think it's a good practice to follow nevertheless. Especially if the last required changes are minor once like these here.

Yeah, fine with me! I just tried to keep it as small as possible so review would be easier, but since you already reviewed most of this, these changes should be fine.

I'll try and get this done asap.

tests/unittest_scoped_nodes.py
DanielNoord: Tests added!

ChangeLog
```diff
@@ -9,6 +9,9 @@ Release date: TBA
 * Add new (optional) ``doc_node`` attribute to ``nodes.Module``, ``nodes.ClassDef``,
   and ``nodes.FunctionDef``.
 
+* Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
+  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node.value`` attribute.
```
cdce8p, line 13:
```suggestion
* Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node`` attribute.
```
Maybe we should add an additional note here that captures the concerns you have raised? Something like this?
```
Note: ``doc_node`` is ``nodes.Const | None`` whereas ``doc`` was ``str | None``.
```

DanielNoord:
```suggestion
* Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node`` attribute.
  Note: ``doc_node`` is ``nodes.Const | None`` whereas ``doc`` was ``str | None``.
```

DanielNoord: I like the note (committed). Should we also mention `.value` or should people (we and who else...?) figure that out themselves?

cdce8p: If you have a good idea, go for it! We just need to be cautious that `doc_node` can be `None` and thus `doc_node.value` won't always work.

DanielNoord: Let's leave as is. Not really helpful to create a very long changelog entry which is likely only useful to us and current `pylint` contributors.

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 * Add new (optional) ``doc_node`` attribute to ``nodes.Module``, ``nodes.ClassDef``,
   and ``nodes.FunctionDef``.
 
+* Accessing the ``doc`` attribute of ``nodes.Module``, ``nodes.ClassDef``, and
+  ``nodes.FunctionDef`` has been deprecated in favour of the ``doc_node`` attribute.
+  Note: ``doc_node`` is ``nodes.Const | None`` whereas ``doc`` was ``str | None``.
```
cdce8p, line 14:
```suggestion
  Note: ``doc_node`` is an (optional) ``nodes.Const`` whereas ``doc`` was an (optional) ``str``.
```
That might look slightly better.

DanielNoord:
Tests show failing due to `venv`. The `pull_request` CI did pass though, so we should be fine.

cdce8p (approved):
Last comment, can be merged once resolved!
Thanks @DanielNoord 🚀

PR #1436, opened 2022-03-03
Pass ``doc_node`` to postinit of all ``Property`` constructors

DanielNoord:
## Steps

## Description

Partial step 2 as discussed in https://github.com/PyCQA/astroid/pull/1432#pullrequestreview-898319366.

Blocked by #1434 as there is a test that depends on the deprecation.

## Type of Changes

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -23,9 +23,11 @@
 """Astroid hooks for various builtins."""
 
 from functools import partial
+from typing import Optional
 
 from astroid import arguments, helpers, inference_tip, nodes, objects, util
 from astroid.builder import AstroidBuilder
+from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidTypeError,
     AttributeInferenceError,
@@ -548,7 +550,9 @@ def infer_callable(node, context=None):
     return nodes.Const(inferred.callable())
 
 
-def infer_property(node, context=None):
+def infer_property(
+    node: nodes.Call, context: Optional[InferenceContext] = None
+) -> objects.Property:
     """Understand `property` class
 
     This only infers the output of `property`
@@ -570,12 +574,16 @@ def infer_property(node, context=None):
     prop_func = objects.Property(
         function=inferred,
         name=inferred.name,
-        doc=getattr(inferred, "doc", None),
+        doc=getattr(inferred, "doc_node.value", None),
         lineno=node.lineno,
         parent=node,
         col_offset=node.col_offset,
     )
-    prop_func.postinit(body=[], args=inferred.args)
+    prop_func.postinit(
+        body=[],
+        args=inferred.args,
+        doc_node=getattr(inferred, "doc_node", None),
+    )
     return prop_func
 
 
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -34,7 +34,18 @@ import ast
 import functools
 import itertools
 import operator
-from typing import Any, Callable, Dict, Iterable, Iterator, Optional, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    Iterator,
+    Optional,
+    Type,
+    Union,
+)
 
 import wrapt
 
@@ -57,6 +68,9 @@ from astroid.exceptions import (
 )
 from astroid.interpreter import dunder_lookup
 from astroid.manager import AstroidManager
+
+if TYPE_CHECKING:
+    from astroid import objects as astroid_objects
 
 # Prevents circular imports
 objects = util.lazy_import("objects")
@@ -1063,7 +1077,13 @@ def _cached_generator(func, instance, args, kwargs, _cache={}):  # noqa: B006
 # are mutated with a new instance of the property. This is why we cache the result
 # of the function's inference.
 @_cached_generator
-def infer_functiondef(self, context=None):
+def infer_functiondef(
+    self: nodes.FunctionDef, context: Optional[InferenceContext] = None
+) -> Generator[
+    Union["astroid_objects.Property", nodes.FunctionDef],
+    None,
+    Dict[str, Union[nodes.FunctionDef, InferenceContext, None]],
+]:
     if not self.decorators or not bases._is_property(self):
         yield self
         return dict(node=self, context=context)
@@ -1071,12 +1091,12 @@ def infer_functiondef(self, context=None):
     prop_func = objects.Property(
         function=self,
         name=self.name,
-        doc=self.doc,
+        doc=self.doc_node.value if self.doc_node else None,
         lineno=self.lineno,
         parent=self.parent,
         col_offset=self.col_offset,
     )
-    prop_func.postinit(body=[], args=self.args)
+    prop_func.postinit(body=[], args=self.args, doc_node=self.doc_node)
     yield prop_func
     return dict(node=self, context=context)
 
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -6146,6 +6146,26 @@ def test_property_callable_inference() -> None:
     assert inferred.value == 42
 
 
+def test_property_docstring() -> None:
+    code = """
+    class A:
+        @property
+        def test(self):
+            '''Docstring'''
+            return 42
+
+    A.test #@
+    """
+    node = extract_node(code)
+    inferred = next(node.infer())
+    assert isinstance(inferred, objects.Property)
+    assert inferred.doc_node
+    assert inferred.doc_node.value == "Docstring"
+    with pytest.warns(DeprecationWarning) as records:
+        assert inferred.doc == "Docstring"
+        assert len(records) == 1
+
+
 def test_recursion_error_inferring_builtin_containers() -> None:
     node = extract_node(
         """


astroid/inference.py
```diff
@@ -58,6 +69,9 @@
 from astroid.interpreter import dunder_lookup
 from astroid.manager import AstroidManager
 
+if TYPE_CHECKING:
+    from astroid import objects as astroid_objects
```
cdce8p, line 73: Let's import `Property` here instead.

astroid/inference.py
```diff
@@ -1063,20 +1077,26 @@ def _cached_generator(func, instance, args, kwargs, _cache={}):  # noqa: B006
 # are mutated with a new instance of the property. This is why we cache the result
 # of the function's inference.
 @_cached_generator
-def infer_functiondef(self, context=None):
+def infer_functiondef(
+    self: nodes.FunctionDef, context: Optional[InferenceContext] = None
```
cdce8p, line 1081: `nodes.FunctionDef` is not quite enough since we also yield the same type. We need to use a bound TypeVar here.

astroid/inference.py
```diff
@@ -1063,20 +1077,26 @@ def _cached_generator(func, instance, args, kwargs, _cache={}):  # noqa: B006
...
+) -> Generator[
+    Union["astroid_objects.Property", nodes.FunctionDef],
+    None,
+    Dict[str, Union[nodes.FunctionDef, InferenceContext, None]],
```
cdce8p, line 1085: A `Union` value type for dicts is not really helpful. We should either use `Dict[str, Any]` or a `TypedDict`.

astroid/inference.py
```diff
@@ -1063,20 +1077,26 @@ def _cached_generator(func, instance, args, kwargs, _cache={}):  # noqa: B006
...
+]:
     if not self.decorators or not bases._is_property(self):
         yield self
         return dict(node=self, context=context)
 
     prop_func = objects.Property(
         function=self,
         name=self.name,
-        doc=self.doc,
+        doc=self.doc_node.value if self.doc_node else None,
```
cdce8p, line 1094:
```suggestion
```
If `postinit` sets `doc`, this shouldn't be necessary. See https://github.com/PyCQA/astroid/pull/1434#pullrequestreview-899198368

astroid/brain/brain_builtin_inference.py
```diff
@@ -570,12 +574,16 @@ def infer_property(node, context=None):
...
+        doc=getattr(inferred, "doc_node.value", None),
         lineno=node.lineno,
         parent=node,
         col_offset=node.col_offset,
     )
-    prop_func.postinit(body=[], args=inferred.args)
+    prop_func.postinit(
+        body=[],
+        args=inferred.args,
+        doc_node=getattr(inferred, "doc_node", None),
```
cdce8p, line 585:
```suggestion
        doc_node=inferred.doc_node
```
Both `FunctionDef` and `Lambda` should always have a `doc_node` attribute.

tests/unittest_inference.py
```diff
@@ -6146,6 +6146,26 @@ class A:
...
+        def test(self):
+            '''Docstring'''
+            return 42
+
+    A.test #@
+    """
+    node = extract_node(code)
+    inferred = next(node.infer())
+    assert isinstance(inferred, objects.Property)
+    assert inferred.doc_node
```
cdce8p, line 6162:
```suggestion
    assert isinstance(inferred.doc_node, nodes.Const)
```

cdce8p (changes requested):

astroid/inference.py
DanielNoord: https://github.com/PyCQA/astroid/blob/393b27660f63c5c93d6e3bdb23da3534065564b2/astroid/inference.py#L1084

This is set to be a method on `FunctionDef`. It doesn't get used anywhere else. Can `self` be anything else than `FunctionDef`?

astroid/inference.py
DanielNoord: Doesn't this narrow down the type from `Any`?  It's the results of these return values:

https://github.com/PyCQA/astroid/blob/393b27660f63c5c93d6e3bdb23da3534065564b2/astroid/inference.py#L1069
https://github.com/PyCQA/astroid/blob/393b27660f63c5c93d6e3bdb23da3534065564b2/astroid/inference.py#L1081

How would you type those?

astroid/inference.py
DanielNoord: See my comment there.

astroid/brain/brain_builtin_inference.py
DanielNoord: This crashes currently as well. For some reason accessing `doc_node` or `doc` directly raises `StopIteration`. I felt like narrowing that issue down was out of scope. See also L573 on the old code. There is an pre-existing issue here.

astroid/inference.py
cdce8p: What about `AsyncFunctionDef`? That inherits from `FunctionDef`.

cdce8p: If it would only be used to annotate `self`, that wouldn't be an issue. However, here we also yield `self`. The result would be that `AsyncFunctionDef._infer` would yield just a `FunctionDef` which isn't correct.

DanielNoord: Pff, you must get tired of me making these easily avoidable mistakes. For real though, I truly appreciated the patience you're always showing with me. I'm learning so much from all the feedback you're giving me on my PRs. Very, very much appreciated! 😃 

Changed!

astroid/inference.py
cdce8p: The type itself is correct, it's just not very useful.

Think of it as a consumer of this function. What would you do if you know the expected return type is `Dict[str, Union[...]]`? You would need to combine every attribute access to say `d.node` with an isinstance call to narrow down the type. That is just impractical. If it's instead typed as `Any`, mypy will just skip it.

The alternative would be a TypeDict. Maybe that's even more useful since other inference functions seem to return the same dict. Something like this, also I'm not sure about the name itself. Would need to check if it's actual an `InferenceError`.
```py
class InferenceError(TypedDict):
    node: nodes.NodeNG
    context: Optional[InferenceContext]
```

DanielNoord: A little breakpointing showed that it is used here:
https://github.com/PyCQA/astroid/blob/393b27660f63c5c93d6e3bdb23da3534065564b2/astroid/inference.py#L1047-L1057
(Might be used elsewhere of course).

So here the return value of `infer_functiondef` is part of the key in cache.
Perhaps `class NodeContextPair`?

cdce8p: Let's use that for now. I'll come back to it once I'm done reviewing the first doc PR.

astroid/inference.py
```diff
@@ -58,9 +71,24 @@
...
+
 # Prevents circular imports
 objects = util.lazy_import("objects")
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+T_FunctionDefSelf = TypeVar("T_FunctionDefSelf", bound=nodes.FunctionDef)
```
cdce8p, line 85:
```suggestion
FunctionDefT = TypeVar("FunctionDefT", bound=nodes.FunctionDef)
```
Let's use the "right" naming style. At some point I should probably look at that pylint PR again 🙈

astroid/brain/brain_builtin_inference.py
cdce8p: Just checked, `inferred` is narrowed to `FunctionDef | Lambda` and only `FunctionDef` has a `doc` / `doc_node` attribute. So `getattr` is correct here.

astroid/inference.py
DanielNoord: You'll be happy to know I split it up into two smaller, but still hellish, PRs 😄

astroid/brain/brain_builtin_inference.py
```diff
@@ -570,12 +574,16 @@ def infer_property(node, context=None):
     prop_func = objects.Property(
         function=inferred,
         name=inferred.name,
-        doc=getattr(inferred, "doc", None),
+        doc=getattr(inferred, "doc_node.value", None),
```
cdce8p, line 577:
```suggestion
```
`Property` inherits from `FunctionDef` and the `doc` attribute is only passed to `super().__init__` so it should be safe to remove it now.

astroid/inference.py
```diff
@@ -58,9 +71,24 @@
...
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+FunctionDefT = TypeVar("FunctionDefT", bound=nodes.FunctionDef)
+
+
+class NodeContextPair(TypedDict):
```
cdce8p, line 88: I wasn't so far off with my first guess. Take a look at the `raise_if_nothing_inferred` decorator.
https://github.com/PyCQA/astroid/blob/40e1a64ad67d4613b0bef461b0227653c8f085a1/astroid/decorators.py#L149-L155
It assigns the `StopIteration` error args to the `InferenceError` exception class.
(A `StopIteration` exception is raised with a return inside a generator function.)

We shouldn't use `InferenceError` for the dict itself. Maybe `InferenceErrorTD`? Don't really know, maybe there is an even better name.

DanielNoord: Hm, but what about the `_cache` then? That does not raise an `InferenceError` right? 
I tested with the `test_extension_modules` test and a breakpoint just before the second return in `infer_functiondef`.

cdce8p: Ah, did confuse that with the `_cached_generator` with the `cached` decorator. Just too many decorators 😄

That a look at what is actually stored in the `_cache` dict. Only `[<Property.name ...>]` not the return value. `list(copy)` does iterate over the Generator but implicitly handles `StopIteration` exceptions (with the `return` statements basically are). https://stackoverflow.com/a/16780113

cdce8p: > Hm, so what do we want to name this? `InferenceErrorTD`?

As we both have no other idea, I guess it's good enough 🤷🏻‍♂️

DanielNoord: I mean, `NodeContextPair` makes this somewhat re-usable if we ever need it in another context, but both are fine.

I'll change to `InferenceErrorTD`

cdce8p: What about `InferenceErrorInfoTD`? From what I can tell, it's only ever used for the `raise_if_nothing_inferred` decorator and the fallback case even mentions "without any error information".
https://github.com/PyCQA/astroid/blob/2a064bac3430daead3a0c564fb5abf82257654f1/astroid/decorators.py#L147-L158

I would also suggest, to move it to `decorators.py` directly. We might later need to use if in `node_classes.py` for `unpack_infer` and since it's always used in combination with `raise_if_nothing_inferred` the `decorators` file is already imported. So we don't risk any new circular dependencies.

_As we'll be importing it in other files, it should also be public (without the leading `_`)._

DanielNoord: > _As we'll be importing it in other files, it should also be public (without the leading `_`)._

Doesn't the leading underscore mean it is private for "external" users? As in, to be used internally within this package?
If not, how do we signal that something is used internally but should not be considered public?

Pierre-Sassoulas: > If not, how do we signal that something is used internally but should not be considered public?

There's definitely some clarifying to do about that. It's a little fuzzy. I think right now we consider everything "private" unless it's documented (how to run pylint programmatically ?) or widely used by downstream libraries. But we had a lot of regressions downstream with this this attitude (we don't know what is used by downstream libraries, they are a lot of them).

cdce8p: > Doesn't the leading underscore mean it is private for "external" users? As in, to be used internally within this package?

Yes, but "external" users in this case are also other modules. Usually, it's better to avoid importing anything with a leading underscore (even from the same package).

> If not, how do we signal that something is used internally but should not be considered public?

Python doesn't really have a good concept for that, at least non that I know of. In general everything should be considered public. Some pypi packages try to split public and private functions and create entirely separate packages. `pytest` is an example here with `_pytest` for internal code. However that is a lot of work, which is not worth it IMO.

> There's definitely some clarifying to do about that. It's a little fuzzy. I think right now we consider everything "private" unless it's documented (how to run pylint programmatically ?) or widely used by downstream libraries. But we had a lot of regressions downstream with this this attitude (we don't know what is used by downstream libraries, they are a lot of them).

Yeah there is a risk of accidental regressions but as far as I remember, those where mostly caused by large scale refactorings rather than actual code changes.

In the end, it's a tradeoff. For methods and attributes I usually add an underscore (when it should be private). Just because it's easy to do and the code is already scoped. For functions only if they are true helper functions used exclusively in that module. Similar for classes. A good analogy here are the TypeVar and TypeAliases. There it does make sense to add one.

Those "soft" rules seem to have worked quite good so far.

Pierre-Sassoulas: It's definitely a trade off. Should we document the current way of doing thing it in the dev guide ? Saying that everything is public unless prefixed by an underscore will increase the burden of handling retro-compatibility by a lot.

DanielNoord: > Yes, but "external" users in this case are also other modules. Usually, it's better to avoid importing anything with a leading underscore (even from the same package).

> In the end, it's a tradeoff. For methods and attributes I usually add an underscore (when it should be private). Just because it's easy to do and the code is already scoped. For functions only if they are true helper functions used exclusively in that module. Similar for classes. A good analogy here are the TypeVar and TypeAliases. There it does make sense to add one.

The best "authoritative" piece in PEP8 on this topic I could find is [this](https://www.python.org/dev/peps/pep-0008/#id46).

For me the benefit of underscoring `astroid` "internals", comes from avoiding downstream regressions. I understand the rule is that you also shouldn't import underscored interfaces from modules within the same package, but when you do and a regression occurs you'll at least be able to fix that yourself.
For me that's a better trade-off than having to tread carefully everywhere because downstream packages might depend on something that is not underscored.

cdce8p: > It's definitely a trade off. Should we document the current way of doing thing it in the dev guide ? Saying that everything is public unless prefixed by an underscore will increase the burden of handling retro-compatibility by a lot.

Not adding an underscore doesn't necessarily mean that it's intended to be public. IMO it's better to leave the situation as is and decide from case to case if necessary. That did work so far with only a few exceptions.

The only things I consider to be **truly** public are those in `astroid.__init__` and `astroid.nodes.__init__` (in `__all__`).

To close this, I'm not sure the discussion here is worth it. It's a TypedDict used only for astroid after all. Even without an underscore, external projects are unlikely to use it. They simply don't need to.

astroid/inference.py
```diff
@@ -58,9 +71,24 @@
...
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+FunctionDefT = TypeVar("FunctionDefT", bound=nodes.FunctionDef)
+
+
+class InferenceErrorTD(TypedDict):
+    node: nodes.NodeNG
+    context: Optional[InferenceContext]
```
cdce8p, line 90: What about `InferenceErrorInfo`? Can you also move it to `decorators.py` or maybe a new `typing.py` file? See also: https://github.com/PyCQA/astroid/pull/1436#discussion_r821198275

cdce8p (approved):
Did some last updates. Looks good now!

* In `typing.py` we should try to use `if TYPE_CHECKING` whenever possible. Since it's likely to be imported all over the place, it could create circular dependencies otherwise.
* Added a docstring for the TypedDict
* The `FunctionDefT` TypeVar is only used in `inference.py`, so it can actually be "private".

astroid/inference.py
Pierre-Sassoulas: > The only things I consider to be truly public are those in astroid.__init__ and astroid.nodes.__init__ (in __all__).

Agreed, also numversion in __pkginfo__.

PR #1438, opened 2022-03-03
New release process

Pierre-Sassoulas:
## Description

Similar to https://github.com/PyCQA/pylint/pull/5858

## Type of Changes

## Related Issue

This document the new git flow.

### DIFF
--- a/doc/release.md
+++ b/doc/release.md
@@ -15,8 +15,8 @@ So, you want to release the `X.Y.Z` version of astroid ?
 6. Check the result and push the tag.
 7. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
-8. (If you just released a minor (`X.Y.0`), also release what was not yet released on
-   the `X.Y-1` branch then delete it.)
+8. (If you're going to release a minor (`X.Y.0`), first release what was not yet
+   released on the `X.Y-1` branch then delete it.)
 
 ## Post release
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -17,9 +17,6 @@ So, you want to release the `X.Y.Z` version of astroid ?
    appropriate changelog in the description. This trigger the pypi release.
 8. (If you just released a minor (`X.Y.0`), also release what was not yet released on
    the `X.Y-1` branch then delete it.)
-9. (If you just released a patch from a maintenance branch, cherry-pick the version
-   bumping commit on the main branch so the main branch has the changelog for
-   `X.Y.Z+1`.)
 
 ## Post release
 
@@ -31,6 +28,8 @@ When a crash or a bug is fixed on the main branch, and it needs backport:
 - (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
 - After the PR is merged cherry-pick the commit on the `X.Y-1` branch
 - Do a release for `X.Y-1.Z`.
+- Cherry-pick the version bumping commit on the main branch so the main branch has the
+  changelog for `X.Y-1.Z+1`.
 
 ### Back to a dev version
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -37,17 +37,3 @@ Check the result and then upgrade the main branch
 
 We move issue that were not done in the next milestone and block release only if it's an
 issue labelled as blocker.
-
-## Post release
-
-### Merge tags in main for pre-commit
-
-If the tag you just made is not part of the main branch, merge the tag `vX.Y.Z` in the
-main branch by doing a history only merge. It's done in order to signal that this is an
-official release tag, and for `pre-commit autoupdate` to works.
-
-```bash
-git checkout main
-git merge --no-edit --strategy=ours vX.Y.Z
-git push
-```
--- a/doc/release.md
+++ b/doc/release.md
@@ -22,7 +22,7 @@ So, you want to release the `X.Y.Z` version of astroid ?
 
 ### Backport fixes from main
 
-When a crash or a bug is fixed on the main branch, and it needs backport:
+When a crash or a bug is fixed on the main branch, and it needs to be backported:
 
 - Make sure that the changelog is for the patch version `X.Y-1.Z'`.
 - (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
--- a/doc/release.md
+++ b/doc/release.md
@@ -10,11 +10,17 @@ So, you want to release the `X.Y.Z` version of astroid ?
 2. Check the result (Do `git diff vX.Y.Z-1 ChangeLog` in particular).
 3. Install the release dependencies `pip3 install -r requirements_test.txt`
 4. Bump the version and release by using `tbump X.Y.Z --no-push`.
-5. Push the tag.
+5. Push the tag, push a `X.Y` branch.
 6. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
 
 ## Post release
+
+### Backport fixes from main
+
+When a crash or a bug is fixed on the main branch, and it needs backport, make sure that
+the changelog is for the patch version `X.Y-1.Z'` then after the PR is merged
+cherry-pick the commit on the `X.Y-1` branch and do a release for `X.Y-1.Z`.
 
 ### Back to a dev version
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -73,8 +73,10 @@ tbump X.Y.Z+1-dev0 --no-tag --no-push  # You can interrupt after the first step
 git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 ```
 
+For example:
+
 ```bash
-tbump X.3.6-dev0 --no-tag --no-push
+tbump 2.3.6-dev0 --no-tag --no-push
 git commit -am "Upgrade the version to 2.3.6-dev0 following 2.3.5 release"
 ```
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -8,19 +8,19 @@ So, you want to release the `X.Y.Z` version of astroid ?
 the maintenance branch. If so, release a last patch release first. See
 `Release a patch version`.**
 
-1. Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`. (For
-   example: `v2.3.5`)
-2. Check the result of `git diff vX.Y-1.Z' ChangeLog`. (For example:
-   `git diff v2.3.4 ChangeLog`)
-3. Install the release dependencies: `pip3 install -r requirements_test.txt`
-4. Bump the version and release by using `tbump X.Y.0 --no-push`. (For example:
-   `tbump 2.4.0 --no-push`)
-5. Check the result visually and then by triggering the "release tests" workflow in
-   GitHub Actions first.
-6. Push the tag.
-7. Release the version on GitHub with the same name as the tag and copy and paste the
-   appropriate changelog in the description. This trigger the pypi release.
-8. Move the `main` branch up to a dev version with `tbump`:
+- Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`. (For
+  example: `v2.3.5`)
+- Check the result of `git diff vX.Y-1.Z' ChangeLog`. (For example:
+  `git diff v2.3.4 ChangeLog`)
+- Install the release dependencies: `pip3 install -r requirements_test.txt`
+- Bump the version and release by using `tbump X.Y.0 --no-push`. (For example:
+  `tbump 2.4.0 --no-push`)
+- Check the result visually and then by triggering the "release tests" workflow in
+  GitHub Actions first.
+- Push the tag.
+- Release the version on GitHub with the same name as the tag and copy and paste the
+  appropriate changelog in the description. This trigger the pypi release.
+- Move the `main` branch up to a dev version with `tbump`:
 
 ```bash
 tbump X.Y+1.0-dev0 --no-tag --no-push  # You can interrupt after the first step
@@ -36,37 +36,37 @@ git commit -am "Upgrade the version to 2.5.0-dev0 following 2.4.0 release"
 
 Check the result and then upgrade the main branch
 
-11. Delete the `maintenance/X.Y-1.x` branch. (For example: `maintenance/2.3.x`)
+- Delete the `maintenance/X.Y-1.x` branch. (For example: `maintenance/2.3.x`)
 
 ## Backporting a fix from `main` to the maintenance branch
 
 Whenever a commit on `main` should be released in a patch release on the current
 maintenance branch we cherry-pick the commit from `main`.
 
-1. During the merge request on `main`, make sure that the changelog is for the patch
-   version `X.Y-1.Z'`. (For example: `v2.3.5`)
-2. Create a `maintenance/X.Y.x` branch on Github --if it does not already exist-- from
-   the `X.Y-1.0` tag . (For example: `maintenance/2.3.x` from the `v2.3.0` tag.)
-3. After the PR is merged on `main` cherry-pick the commits on the `maintenance/X.Y.x`
-   branch (For example: from `maintenance/2.4.x` cherry-pick a commit from `main`)
-4. Release a patch version
+- During the merge request on `main`, make sure that the changelog is for the patch
+  version `X.Y-1.Z'`. (For example: `v2.3.5`)
+- Create a `maintenance/X.Y.x` branch on Github --if it does not already exist-- from
+  the `X.Y-1.0` tag . (For example: `maintenance/2.3.x` from the `v2.3.0` tag.)
+- After the PR is merged on `main` cherry-pick the commits on the `maintenance/X.Y.x`
+  branch (For example: from `maintenance/2.4.x` cherry-pick a commit from `main`)
+- Release a patch version
 
 ## Releasing a patch version
 
 We release patch version when a crash or a bug is fixed on the main branch and has been
 cherry-picked on the maintenance branch.
 
-4. Check the result of `git diff vX.Y-1.Z-1 ChangeLog`. (For example:
-   `git diff v2.3.4 ChangeLog`)
-5. Install the release dependencies: `pip3 install -r requirements_test.txt`
-6. Bump the version and release by using `tbump X.Y-1.Z --no-push`. (For example:
-   `tbump 2.3.5 --no-push`)
-7. Check the result visually and then by triggering the "release tests" workflow in
-   GitHub Actions first.
-8. Push the tag.
-9. Release the version on GitHub with the same name as the tag and copy and paste the
-   appropriate changelog in the description. This trigger the pypi release.
-10. Move the `main` branch up to a dev version with `tbump`:
+- Check the result of `git diff vX.Y-1.Z-1 ChangeLog`. (For example:
+  `git diff v2.3.4 ChangeLog`)
+- Install the release dependencies: `pip3 install -r requirements_test.txt`
+- Bump the version and release by using `tbump X.Y-1.Z --no-push`. (For example:
+  `tbump 2.3.5 --no-push`)
+- Check the result visually and then by triggering the "release tests" workflow in
+  GitHub Actions first.
+- Push the tag.
+- Release the version on GitHub with the same name as the tag and copy and paste the
+  appropriate changelog in the description. This trigger the pypi release.
+- Move the `main` branch up to a dev version with `tbump`:
 
 ```bash
 tbump X.Y.Z+1-dev0 --no-tag --no-push  # You can interrupt after the first step
@@ -82,11 +82,11 @@ git commit -am "Upgrade the version to 2.3.6-dev0 following 2.3.5 release"
 
 Check the result and then upgrade the `maintenance/X.Y.x` branch
 
-13. Merge the `maintenance/X.Y.x` branch on the main branch. The main branch should have
-    the changelog for `X.Y-1.Z+1` (For example `v2.3.6`). This merge is required so
-    `pre-commit autoupdate` works for pylint.
-14. Fix version conflicts properly, or bump the version to `X.Y.0-devZ` (For example:
-    `2.4.0-dev6`) before pushing on the main branch
+- Merge the `maintenance/X.Y.x` branch on the main branch. The main branch should have
+  the changelog for `X.Y-1.Z+1` (For example `v2.3.6`). This merge is required so
+  `pre-commit autoupdate` works for pylint.
+- Fix version conflicts properly, or bump the version to `X.Y.0-devZ` (For example:
+  `2.4.0-dev6`) before pushing on the main branch
 
 ## Milestone handling
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -2,50 +2,62 @@
 
 So, you want to release the `X.Y.Z` version of astroid ?
 
-## Process
+## Release a minor or major version
 
-(Consider triggering the "release tests" workflow in GitHub Actions first.)
-
-1. Check if the dependencies of the package are correct
-2. (If you're releasing a minor (`X.Y.0`), remove the empty changelog for the last
-   unreleased patch version `X.Y-1.Z'`.)
-3. Check the result of `git diff vX.Y.Z-1 ChangeLog` or `git diff vX.Y-1.Z' ChangeLog`.
+1. Release what was not yet released on the `X.Y-1` branch.
+2. Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`.
+3. Check the result of `git diff vX.Y-1.Z' ChangeLog`.
 4. Install the release dependencies `pip3 install -r requirements_test.txt`
 5. Bump the version and release by using `tbump X.Y.Z --no-push`.
-6. Check the result and push the tag.
-7. Release the version on GitHub with the same name as the tag and copy and paste the
+6. Check the result visually and then by triggering the "release tests" workflow in
+   GitHub Actions first.
+7. Push the tag.
+8. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
-8. (If you're going to release a minor (`X.Y.0`), first release what was not yet
-   released on the `X.Y-1` branch then delete it.)
-
-## Post release
-
-### Backport fixes from main
-
-When a crash or a bug is fixed on the main branch, and it needs to be backported:
-
-- Make sure that the changelog is for the patch version `X.Y-1.Z'`.
-- (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
-- After the PR is merged cherry-pick the commits on the `X.Y-1` branch
-- Do a release for `X.Y-1.Z`.
-- Bump `X.Y-1.Z` branch to `X.Y-1.Z+1-dev0`
-- Merge the `X.Y-1` branch on the main branch. The main branch should have the changelog
-  for `X.Y-1.Z+1`. (You need to merge so `pre-commit autoupdate` works.)
-- Fix version conflicts properly, or bump the version to `X.Y.0-devZ` before pushing on
-  the main branch
-
-### Back to a dev version
-
-Move back to a dev version with `tbump`:
+9. Move back to a dev version with `tbump`:
 
 ```bash
-tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt after the first step
+tbump X.Y.Z+1-dev0 --no-tag --no-push  # You can interrupt after the first step
 git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 ```
 
 Check the result and then upgrade the main branch
 
-### Milestone handling
+9. Delete the `X.Y-1` branch.
+
+## Release a patch version
+
+We release patch version when a crash or a bug is fixed on the main branch, and it needs
+to be backported.
+
+1. During the merge request on main, make sure that the changelog is for the patch
+   version `X.Y-1.Z'`.
+2. (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
+3. After the PR is merged cherry-pick the commits on the `X.Y-1` branch
+4. Check the result of `git diff vX.Y-1.Z-1 ChangeLog`.
+5. Install the release dependencies `pip3 install pre-commit tbump`
+6. Bump the version and release by using `tbump X.Y-1.Z --no-push`.
+7. Check the result visually and then by triggering the "release tests" workflow in
+   GitHub Actions first.
+8. Push the tag.
+9. Release the version on GitHub with the same name as the tag and copy and paste the
+   appropriate changelog in the description. This trigger the pypi release.
+10. Move back to a dev version with `tbump`:
+
+```bash
+tbump X.Y-1.Z+1-dev0 --no-tag --no-push  # You can interrupt after the first step
+git commit -am "Upgrade the version to x.y-1.z+1-dev0 following x.y-1.z release"
+```
+
+Check the result and then upgrade the `X.Y-1` branch
+
+13. Merge the `X.Y-1` branch on the main branch. The main branch should have the
+    changelog for `X.Y-1.Z+1`. This merge is required so `pre-commit autoupdate` works
+    for pylint.
+14. Fix version conflicts properly, or bump the version to `X.Y.0-devZ` before pushing
+    on the main branch
+
+## Milestone handling
 
 We move issue that were not done in the next milestone and block release only if it's an
 issue labelled as blocker.
--- a/doc/release.md
+++ b/doc/release.md
@@ -10,7 +10,7 @@ So, you want to release the `X.Y.Z` version of astroid ?
 2. Check the result (Do `git diff vX.Y.Z-1 ChangeLog` in particular).
 3. Install the release dependencies `pip3 install -r requirements_test.txt`
 4. Bump the version and release by using `tbump X.Y.Z --no-push`.
-5. Push the tag, push a `X.Y` branch.
+5. Push the tag.
 6. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
 
@@ -18,9 +18,12 @@ So, you want to release the `X.Y.Z` version of astroid ?
 
 ### Backport fixes from main
 
-When a crash or a bug is fixed on the main branch, and it needs backport, make sure that
-the changelog is for the patch version `X.Y-1.Z'` then after the PR is merged
-cherry-pick the commit on the `X.Y-1` branch and do a release for `X.Y-1.Z`.
+When a crash or a bug is fixed on the main branch, and it needs backport:
+
+- Make sure that the changelog is for the patch version `X.Y-1.Z'`.
+- (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
+- After the PR is merged cherry-pick the commit on the `X.Y-1` branch
+- Do a release for `X.Y-1.Z`.
 
 ### Back to a dev version
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -8,7 +8,7 @@ So, you want to release the `X.Y.Z` version of astroid ?
 
 1. Check if the dependencies of the package are correct
 2. (If you're releasing a minor (`X.Y.0`), remove the empty changelog for the last
-   unreleased `X.Y-1.Z'` version.)
+   unreleased patch version `X.Y-1.Z'`.)
 3. Check the result of `git diff vX.Y.Z-1 ChangeLog` or `git diff vX.Y-1.Z' ChangeLog`.
 4. Install the release dependencies `pip3 install -r requirements_test.txt`
 5. Bump the version and release by using `tbump X.Y.Z --no-push`.
@@ -26,10 +26,13 @@ When a crash or a bug is fixed on the main branch, and it needs backport:
 
 - Make sure that the changelog is for the patch version `X.Y-1.Z'`.
 - (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
-- After the PR is merged cherry-pick the commit on the `X.Y-1` branch
+- After the PR is merged cherry-pick the commits on the `X.Y-1` branch
 - Do a release for `X.Y-1.Z`.
-- Cherry-pick the version bumping commit on the main branch so the main branch has the
-  changelog for `X.Y-1.Z+1`.
+- Bump `X.Y-1.Z` branch to `X.Y-1.Z+1-dev0`
+- Merge the `X.Y-1` branch on the main branch. The main branch should have the changelog
+  for `X.Y-1.Z+1`. (You need to merge so `pre-commit autoupdate` works.)
+- Fix version conflicts properly, or bump the version to `X.Y.0-devZ` before pushing on
+  the main branch
 
 ### Back to a dev version
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -10,9 +10,10 @@ So, you want to release the `X.Y.Z` version of astroid ?
 2. Check the result (Do `git diff vX.Y.Z-1 ChangeLog` in particular).
 3. Install the release dependencies `pip3 install -r requirements_test.txt`
 4. Bump the version and release by using `tbump X.Y.Z --no-push`.
-5. Push the tag.
+5. Check the result and push the tag.
 6. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
+7. Release what was not yet released on the `X.Y-1` branch then delete it.
 
 ## Post release
 
--- a/doc/release.md
+++ b/doc/release.md
@@ -2,62 +2,91 @@
 
 So, you want to release the `X.Y.Z` version of astroid ?
 
-## Release a minor or major version
+## Releasing a major or minor version
 
-1. Release what was not yet released on the `X.Y-1` branch.
-2. Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`.
-3. Check the result of `git diff vX.Y-1.Z' ChangeLog`.
-4. Install the release dependencies `pip3 install -r requirements_test.txt`
-5. Bump the version and release by using `tbump X.Y.Z --no-push`.
-6. Check the result visually and then by triggering the "release tests" workflow in
+**Before releasing a major or minor version check if there are any unreleased commits on
+the maintenance branch. If so, release a last patch release first. See
+`Release a patch version`.**
+
+1. Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`. (For
+   example: `v2.3.5`)
+2. Check the result of `git diff vX.Y-1.Z' ChangeLog`. (For example:
+   `git diff v2.3.4 ChangeLog`)
+3. Install the release dependencies: `pip3 install -r requirements_test.txt`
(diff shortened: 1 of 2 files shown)


doc/release.md
```diff
@@ -10,12 +10,20 @@ So, you want to release the `X.Y.Z` version of astroid ?
 2. Check the result (Do `git diff vX.Y.Z-1 ChangeLog` in particular).
 3. Install the release dependencies `pip3 install pre-commit tbump`
 4. Bump the version and release by using `tbump X.Y.Z --no-push`.
-5. Push the tag.
+5. Push the tag, push a `X.Y` branch
 6. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
 
 ## Post release
```
Pierre-Sassoulas, line 17:
```suggestion
```

DanielNoord:
@Pierre-Sassoulas Can I recommend not to create a branch every time but just create a branch from the release commit when needed? This is probably personal preference, but I don't really see the benefit in keeping 10+ unused branches around.

You can just 1) create branch from release commit, 2) cherry-pick relevant hot fixes, 3) release, 4) merge hotfix branch into `main` (to merge commit history and changelog), 5) drop hot-fix branch.
You could wait with 4 and 5 for a couple of days to wait for any potential new hot fixes of course.

Pierre-Sassoulas:
Thanks for the suggestion I upgraded. 

Regarding 4) and 5):

> 4) merge hotfix branch into main (to merge commit history and changelog),

I think the changelog will already be in the main branch as all MR are for the main branch. In pylint we'll need to do that because of pre-commit but maybe not for astroid ?

5) drop hot-fix branch.

I don't know about this one, it's better to keep the historic of release branch if we mess up and need to investigate. Also probably easier than having to recreate branch from tag. (I don't want to have to maintain the n-2 release to be clear, but keeping the branch for a year does not hurt that much.)

DanielNoord:
> > 4. merge hotfix branch into main (to merge commit history and changelog),
> 
> I think the changelog will already be in the main branch as all MR are for the main branch. In pylint we'll need to do that because of pre-commit but maybe not for astroid ?

If we have commits on `main` for `2.20` there won't be a changelog entry/list for `2.19.1` anymore. If we then need a hotfix for `2.19` and create `2.19.1` we will lose that information if we don't merge the `2.19.1` branch into `main`.

> I don't know about this one, it's better to keep the historic of release branch if we mess up and need to investigate. Also probably easier than having to recreate branch from tag. (I don't want to have to maintain the n-2 release to be clear, but keeping the branch for a year does not hurt that much.)

Hm, okay but perhaps we should then mention that they should be removed after a year? Or after `2.20` is released?

Pierre-Sassoulas:
> If we have commits on main for 2.20 there won't be a changelog entry/list for 2.19.1 anymore.

Hmm, right, but it should, so the merge goes well later on. Also because of " Make sure that the changelog is for the patch version `X.Y-1.Z'`". Let me check the update script then.

>  Or after 2.20 is released?

Yes it seem sensible. Indicate that we're not maintaining it anymore. At this point the latest tag is keeping the history accessible.

Pierre-Sassoulas:
I think the current script would work, right now for example we have a changelog for 2.10.1 : https://github.com/PyCQA/astroid/blob/main/ChangeLog#L18

I'm going to remove it before releasing for 2.11.0 as it's never going to be released. But it could be used to do hotfix the 2.10 branch. What we also need to do is launch the script to get the changelog for 2.10.2 on the main branch when/if we release 2.10.1.

doc/release.md
```diff
@@ -7,15 +7,30 @@ So, you want to release the `X.Y.Z` version of astroid ?
...
+### Backport fixes from main
+
+When a crash or a bug is fixed on the main branch, and it needs backport:
+
+- Make sure that the changelog is for the patch version `X.Y-1.Z'`.
+- (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
+- After the PR is merged cherry-pick the commit on the `X.Y-1` branch
+- Do a release for `X.Y-1.Z`.
+- Cherry-pick the version bumping commit on the main branch so the main branch has the
+  changelog for `X.Y-1.Z+1`.
```
cdce8p, line 32: I think it would be better to merge the bug fix branch into main, instead of cherry-picking a single commit.

cdce8p (changes requested):
Some questions
1. If we cherry-pick a commit, the changelog entry will still be for the next minor release. During the release process, I think the maintainer would need to manually copy the entries to the bug fix release. Maybe with a prefix `Backport: `?
2. When should we stop to backport bug fixes? Once the next minor version is released?
3. Depending on (2): What should the changelog order be? Currently it's both sorted chronological and by version. If we support multiple minor versions in parallel, it might make sense to drop the version sorting. E.g. `2.10.0` -> `2.11.0` -> `2.10.1` -> `2.10.2`
4. Should a new bug fix release be done for every single cherry-picked commit? Grouping would make sense IMO. Of course only if there are multiple issues that need to be fixed.

doc/release.md
Pierre-Sassoulas: Hmm, you're right, it also fix the pre-commit issue we have with pylint.

Pierre-Sassoulas:
Right now in astroid we have 2.11.0 and 2.10.1 in the changelog, I'm going to use that example for clarity.

> If we cherry-pick a commit, the changelog entry will still be for the next minor release. During the release process, I think the maintainer would need to manually copy the entries to the bug fix release. Maybe with a prefix Backport: ?

If this is something we want to backport in 2.10, we would need to put the changelog in 2.10.1 when we merge the PR in ``main`` (and always squash ?). Then we cherry-pick the (squashed) commit on the ``2.10`` branch. I think it work because if it's in 2.10.1, it's natural that it's in 2.11.0 too and we don't have anything more to do?

> When should we stop to backport bug fixes? Once the next minor version is released?

I think it make sense that's what I was thinking of (we also delete the 2.10 branch which makes our intent clear).

> Depending on (2): What should the changelog order be? Currently it's both sorted chronological and by version. If we support multiple minor versions in parallel, it might make sense to drop the version sorting. E.g. 2.10.0 -> 2.11.0 -> 2.10.1 -> 2.10.2

I was thinking about semver order, once 2.11 is out we'll never release a 2.10.x again. I did not thing about major version yet. How long are we going to support 2.X in parallel when 3.0.0 is out ?

> Should a new bug fix release be done for every single cherry-picked commit? Grouping would make sense IMO. Of course only if there are multiple issues that need to be fixed.

I was thinking of releasing frequently (maybe not each bug fix, but at least each week if something was fixed during the previous week). This is why I did #1441 because it's really taking a long time in the release process and was not acceptable anymore if we release often.

cdce8p:
> If this is something we want to backport in 2.10, we would need to put the changelog in 2.10.1 when we merge the PR in `main` (and always squash ?). Then we cherry-pick the (squashed) commit on the `2.10` branch. I think it work because if it's in 2.10.1, it's natural that it's in 2.11.0 too and we don't have anything more to do?

🤔 So you're basically suggestion adding two changelog entries one for `2.10.1` and one for `2.11.0`?
Although that might work, the issue then becomes that the `2.10` branch will contain the `2.11.0` changelog entries too which would need be to removed manually I think.

What if we instead remove the sections for unreleased versions all together and only add them during the release. New changelog entries would only be added once. Cherry-picking a commit with a subsequent release + merge back into main would "double" it. (The last step might create a small merge conflict every time. However it should be fairly easy to resolve as the new release should be below new changes.

> > When should we stop to backport bug fixes? Once the next minor version is released?
> 
> I think it make sense that's what I was thinking of (we also delete the 2.10 branch which makes our intent clear).

👍🏻 Sounds good to me.

> > Depending on (2): What should the changelog order be? Currently it's both sorted chronological and by version. If we support multiple minor versions in parallel, it might make sense to drop the version sorting. E.g. 2.10.0 -> 2.11.0 -> 2.10.1 -> 2.10.2
> 
> I was thinking about semver order, once 2.11 is out we'll never release a 2.10.x again. I did not thing about major version yet. How long are we going to support 2.X in parallel when 3.0.0 is out ?

If we drop the last bug fix branch once the next minor version is release, I don't think this will be an issue anymore.
The move to `3.0.0` is something different, but personally I would strongly prefer **not** to support multiple major versions at the same time. On the off-chance we do need to do some last fix, we would probably insert it in version order anyway, since the `2.x.x` branch wouldn't contain the changelog for `3.x`.

> > Should a new bug fix release be done for every single cherry-picked commit? Grouping would make sense IMO. Of course only if there are multiple issues that need to be fixed.
> 
> I was thinking of releasing frequently (maybe not each bug fix, but at least each week if something was fixed during the previous week). This is why I did #1441 because it's really taking a long time in the release process and was not acceptable anymore if we release often.

👍🏻

--
A few additional things.
1. Do we what to add change categories to the changelog? E.g. `Feature`, `Fix`, `Deprecation`, `Breaking`. Another example here is could be black: https://github.com/psf/black/blob/main/CHANGES.md
2. Looking at the black changelog, they do have the `Unreleased` heading. Maybe that would fit for what I mentioned above.

DanielNoord:
I think one issue we need to tackle is whether we want to merge `2.10.1` into the `2.11` branch before releasing `2.11`. Initially I suggested to do so, but I'm thinking that this might be complicate things (a lot).
`cpython` does not (as far as I can see) and whenever there is a security fix a commit is made on all supported branches. This makes it much easier as you don't have the issue of what to do with the changelog entries of `2.10.1` on the `2.11` branch.

I would suggest the following workflow (but I'd really like input here):
1) Release `2.10`
2) Bump `main` to `2.11.0+dev`
3) Create `2.10` branch with bump to `2.10.1+dev`
4) Commit fix to `main`
5) Cherry-pick fix commit to `2.10` branch
6) Release `2.10.1` from `2.10` branch > bump to `2.10.2+dev`
7) ...
8) Drop `2.10` branch
9) Release `2.11`
Repeat step 1.

Basically that would mean that we support every minor version with necessary hot fixes that get cherry picked from `main` until we release a new minor version, which then becomes the new branch to cherry-pick to.

Pierre-Sassoulas:
>  So you're basically suggestion adding two changelog entries one for 2.10.1 and one for 2.11.0?

No, only one for 2.10.1. On the main branch we have the changelog for 2.10.1 https://github.com/PyCQA/astroid/blob/main/ChangeLog#L18, we just need to decide if we're going to backport before merging into main  so the changelog is in the proper section.

> What if we instead remove the sections for unreleased versions all together and only add them during the release. 

I think it's going to make a lot of work and potential mistake reviewing everything at release time and classifying what goes where. I think it's better if it's done little by little during each merge request.

>  I would strongly prefer not to support multiple major versions at the same time.

👍 Then it's easier than a minor release version. It make sense too, because there will be breaking change, backporting would be harder.


> Do we what to add change categories to the changelog? E.g. Feature, Fix, Deprecation, Breaking. Another example here is could be black: https://github.com/psf/black/blob/main/CHANGES.md
> Looking at the black changelog, they do have the Unreleased heading. Maybe that would fit for what I mentioned above.

I like what black is doing. We have another issue we might want to fix at the same time (it's only in pylint but let's discuss it here). We're duplicating the changelog entries between the whatsnew and the changelog file https://github.com/PyCQA/pylint/issues/5728. This might be the occasion to stop doing that. Maybe creating the whatsnew programmatically from a well organized changelog ? Maybe taking all the handling code from black directly ? (I'd like to do that only after 2.13 is released though, it's a lot of work.)

> This makes it much easier as you don't have the issue of what to do with the changelog entries of 2.10.1 on the 2.11 branch.

I think there won't be too much issue as we have only the 2.10 branch (we delete the 2.9 branch) we just need to [make sure the changelog entry for the 2.10.2 version end on the main branch when we release 2.10.1](https://github.com/PyCQA/astroid/pull/1438/files#diff-ab2a79e380c3f0c3fd87abb91dc5cae3aac0a02f84a1ac39b4a3e75af9890a5aR31).

I adapted a little what you proposed @DanielNoord :

1) Release `2.10`
2) Bump `main` to `2.11.0+dev0`
3) Commit fix to `main` # We create the branch only if there's a fix to backport
4) Create `2.10` branch with bump to `2.10.1+dev`
5) Cherry-pick fix commit to `2.10` branch
6) Release `2.10.1` from `2.10` branch > bump to `2.10.2+dev`
7) Bump 2.10 branch to `2.10.2+dev`
8) Cherry-pick the  bump commit of the 2.10.1 release from the 2.10 branch on main
9) bump the version to ``2.11.0-dev1`` on main (as the cherry-picked commit above will be versioned 2.10.2-dev0)
10) amend the cherry-picked commit, and push
...
42) Drop `2.10` branch
43) Release `2.11`
Repeat step 1.

cdce8p:
Is cherry-picking the bump commits enough for pre-commit autoupdate?

jacobtylerwalls:
> Cherry-pick the bump commits from the 2.10 branch on main

Just a question: why would do the bump commits need to be forward-ported? Do other projects do that?

Pierre-Sassoulas:
> Just a question: why do the bump commits need to be forward-ported? Do other projects do that?

The bump commit contain a change to the release date in the changelog as well as the creation of the new version (and also the change to copyright in files but I expect to remove that in #1441) see here : https://github.com/PyCQA/astroid/commit/07c0f60ffc1017d0a9a2bb605a5c645781a8c088

So we need to cherry-pick it on main in order to have the changelog header for 2.10.2 on the main branch for easy cherry-picking of fix commit on 2.10 later.

Pierre-Sassoulas:
> Is cherry-picking the bump commits enough for pre-commit autoupdate?

Sorry I did not answer earlier because I did not understand the question. ``pre-commit autoupdate`` is handled by the pre-commit ci bot every monday. Are you asking if ``pre-commit run --all-files`` need to be relaunched after the cherry-pick ? I don't think so but I could check.

jacobtylerwalls:
> So we need to cherry-pick it on main in order to have the changelog header for 2.10.2 on the main branch for easy cherry-picking of fix commit on 2.10 later.

I was thinking we could just always have the 2.10.2 header on `main`. (That's what Django does, for instance, after any release, the next commit on main has stub headers for new versions, including bugfix branches, and when finalizing a release date, that's done on main also.) Then if there's some header that never gets filled in with contents because there's a perfect release, you do a commit to delete it and throw a huge party.

> as well as the creation of the new version

I guess I don't understand that part, wouldn't we want to _avoid_ forward-porting changes to `__VERSION__`?

cdce8p:
> > Is cherry-picking the bump commits enough for pre-commit autoupdate?
> 
> Sorry I did not answer earlier because I did not understand the question. `pre-commit autoupdate` is handled by the pre-commit ci bot every monday. Are you asking if `pre-commit run --all-files` need to be relaunched after the cherry-pick ? I don't think so but I could check.

Not exactly. If I remember correctly, we had an issue in pylint a while back that pre-commit autoupdate wasn't working with our "old" release process. https://github.com/PyCQA/pylint/issues/4245

Pierre-Sassoulas:
> Not exactly. If I remember correctly, we had an issue in pylint a while back that pre-commit autoupdate wasn't working with our "old" release process

You're right, we need to merge in pylint for this reason. So we might as well merge for astroid too so process are closer.

> I was thinking we could just always have the 2.10.2 header on main.

Nice, that would reduce the merge / rebase conflicts drastically and the number of cherry-pick we need to do. Let's do that. How many patch version should we add ? 5 ?

jacobtylerwalls:
> How many patch version should we add ? 5 ?

How about just one at a time?
- do backports of hotfixes over a week
- finalize release date on main and backport it
- publish from the hotfix branch
- add a (single) new stub on main and backport it

cdce8p:
> > How many patch version should we add ? 5 ?
> 
> How about just one at a time?

I agree. Anything else would just be confusing.

> So we might as well merge for astroid too so process are closer.

I'm not an expert in branching strategies, but wouldn't `git merge --no-edit --strategy=ours` mean that any changes from the bugfix branch would be ignore. Like the final commit release with changelog change?

Pierre-Sassoulas:
> I'm not an expert in branching strategies, but wouldn't git merge --no-edit --strategy=ours mean that any changes from the bugfix branch would be ignore. Like the final commit release with changelog change?

Yes and @jacobtylerwalls suggested to modify the changelog on the main branch so then we cherry-pick on 2.10 and we don't need to cancel the version change so it remove one step 👍 .

Pierre-Sassoulas:
So to sumpup:

1) Release `2.10`
2) Bump `main` to `2.11.0+dev0`
3) Commit fix to `main`
4) Create `2.10` branch with bump to `2.10.1+dev`
5) Cherry-pick fix commit to `2.10` branch
6) Apply the changelog script as if we were releasing `2.10.1` on main branch (after a week)
8) Cherry-pick the changelog commits from the main branch on 2.10
9) Release 2.10.1 from the 2.10 branch
7) Bump 2.10 branch to `2.10.2+dev`
10) ~~bump the version to ``2.11.0-dev1`` on main (as the cherry-picked commit above will be versioned 2.10.2-dev0)~~
11) ~~amend the cherry-picked commit, and push~~ 
12)  ``git merge --no-edit --strategy=ours 2.10`` from the main branch
...
42) Release `2.11`
43) Drop `2.10` branch

Repeat step 1.

After writing it, I realize it's one step less but it's also not using tbump normally (we have to launch a changelog script without modifying the version when it was encapsulated with tbump previousely). Doing it from the main branch is not the "natural" place to tbump 2.10.1.

I think this would make sense:

1) Release `2.10`
2) Bump `main` to `2.11.0+dev0`
3) Commit fix to `main` # We create the branch only if there's a fix to backport
4) Create `2.10` branch with bump to `2.10.1+dev`
5) Cherry-pick fix commit to `2.10` branch
6) Release `2.10.1` from `2.10` branch
7) Bump 2.10 branch to `2.10.2+dev`
8) ``git merge 2.10``, fix conflict (only on version change imo ?) bump the version to ``2.11.0-dev1``
9) push the merge commit
...
42) Release `2.11`
43) Drop `2.10` branch

DanielNoord:
I think it might be worth testing this on [Test pypi](https://test.pypi.org). We might even want to test how `pre-commit` autoupdate respond to this.

Does `git merge --no-edit --strategy=ours 2.10.1` allow checking out `2.10.1` from the `main` branch? Or can you only check out to the `2.10.1` tag from the `2.10` branch?
Given:
`main` on:
`prev_main` > commit A > commit B > commit C
`2.10` on:
`prev_main` > commit B 

Release `2.10.1`.

`git merge --no-edit --strategy=ours 2.10.1` into `main`.

When I'm on `main` can I check out the `2.10.1` branch, basically telling `git` to drop the A and C commit?

Pierre-Sassoulas:
``git merge --no-edit --strategy=ours 2.10.1`` works well for pre-commit see https://github.com/PyCQA/pylint/issues/4245#issuecomment-803616793  (Anthony Sottile is pre-commit's maintener)

doc/release.md
```diff
@@ -7,15 +7,33 @@ So, you want to release the `X.Y.Z` version of astroid ?
...
-6. Release the version on GitHub with the same name as the tag and copy and paste the
+2. (If you're releasing a minor (`X.Y.0`), remove the empty changelog for the last
+   unreleased patch version `X.Y-1.Z'`.)
+3. Check the result of `git diff vX.Y.Z-1 ChangeLog` or `git diff vX.Y-1.Z' ChangeLog`.
+4. Install the release dependencies `pip3 install pre-commit tbump`
+5. Bump the version and release by using `tbump X.Y.Z --no-push`.
+6. Check the result and push the tag.
+7. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
+8. (If you just released a minor (`X.Y.0`), also release what was not yet released on
```
DanielNoord, line 18: This should probably be done before the minor is released.
So: `2.9.0` > `2.9.1` > `2.10.0`.

doc/release.md
```diff
@@ -7,15 +7,33 @@ So, you want to release the `X.Y.Z` version of astroid ?
...
+   the `X.Y-1` branch then delete it.)
 
 ## Post release
 
+### Backport fixes from main
+
+When a crash or a bug is fixed on the main branch, and it needs backport:
```
DanielNoord, line 25:
```suggestion
When a crash or a bug is fixed on the main branch, and it needs to be backported:
```

doc/release.md
```diff
@@ -7,15 +7,33 @@ So, you want to release the `X.Y.Z` version of astroid ?
...
+
+- Make sure that the changelog is for the patch version `X.Y-1.Z'`.
+- (Create a `X.Y-1` branch from the `X.Y-1.0` tag if it does not already exist.)
+- After the PR is merged cherry-pick the commits on the `X.Y-1` branch
+- Do a release for `X.Y-1.Z`.
+- Bump `X.Y-1.Z` branch to `X.Y-1.Z+1-dev0`
+- Merge the `X.Y-1` branch on the main branch. The main branch should have the changelog
```
DanielNoord, line 32: We should make the merge strategy explicit here.

DanielNoord (changes requested):
I'm still a bit unsure about merging the `fix` branch into the `minor` branch, but if others think that is the right way forward I guess we should go ahead.

doc/release.md
Pierre-Sassoulas: I'm not absolutely certain that there won't be conflict in the changelog that should be done manually (there's date that could be added on one side and not the other). I think it's better if the releaser think about it in case of conflict.

Pierre-Sassoulas: More precisely I think about the new changelog for 2.10.2 added on the fix branch which add text very similar to the changelog for 2.11.0 and there will be new entry for 2.11 feature on the main branch.

doc/release.md
Pierre-Sassoulas: [90d3373](https://github.com/PyCQA/astroid/pull/1438/commits/90d3373a1a6a8692896a494c8a28f5811b02e433)

doc/release.md
DanielNoord: Yes, but these are (again) issues that stem from whether we merge `2.10.2` into `main` anyway. You could also say that when you release `2.11` the releaser makes a manual commit which copies the `2.10.x` changelogs to `main` and then drops the `2.10` branch without any merging.
If we do want to merge there should at least be something about `git merge --no-edit --strategy=ours` in the documentation.

DanielNoord:
I like how [numpy](https://github.com/numpy/numpy) does this. They have extensive release documentation [here](https://numpy.org/devdocs/dev/releasing.html).

Basically, after a maintenance release they commit changes to documentation, changelogs and contributors to `main`. See [the commits](https://github.com/numpy/numpy/tree/main/doc/source/release) here.
I wonder if this is compatible with an automated contributor/license list like we are working in, as I am not sure that will pick up on contributors that only contribute to maintenance branches.

Schermafbeelding 2022-03-10 om 08 55 08

To make this actionable, I would propose the following:
1) Clearly separate the `minor` and `fix` release instructions (nothing like `(If you're going to release a minor (X.Y.0),` in the instructions, just separate the two cases).
2) If `main` is on `2.11.0+dev` do not include a `2.10.1` changelog stub. If there needs to be one it will be pushed from the `maintenance/2.10.x` branch.
3) Include instructions on what to merge from `maintenance/2.10.x` into `main`.

After we did this a first time we should check whether the `2.10.1` tag is indeed available on `main` after step 3. I tested with `numpy` and it seems to work, but it doesn't hurt to try this again.

Pierre-Sassoulas:
I've read the doc for numpy release and it's a little different than what we need. They have a lot of maintenance branches when we only support the latest minor. Let's talk about the release note automated generation in more detail for https://github.com/PyCQA/pylint/issues/5728

DanielNoord:
> I've read the doc for numpy release and it's a little different than what we need. They have a lot of maintenance branches when we only support the latest minor. Let's talk about the release note automated generation in more detail for [PyCQA/pylint#5728](https://github.com/PyCQA/pylint/issues/5728)

The only difference (I think) is that we drop the maintenance branch once we release the next minor version. All the stuff about preparing the maintenance branch for future development and merging the changelog changes into `main` is relevant for us I think.

doc/release.md
Pierre-Sassoulas: > If we do want to merge there should at least be something about git merge --no-edit --strategy=ours in the documentation.

We're forced to merge so that ``pre-commit autoupdate`` works. The question is whether or not the merge is ``no edit`` or the one where we make sure the changelog modification are coherent.

> Yes, but these are (again) issues that stem from whether we merge 2.10.2 into main anyway. You could also say that when you release 2.11 the releaser makes a manual commit which copies the 2.10.x changelogs to main and then drops the 2.10 branch without any merging.

I think both option work and that we're well past the point of reaching this https://xkcd.com/1445/ 😄

Pierre-Sassoulas:
> All the stuff about preparing the maintenance branch for future development and merging the changelog changes into main is relevant for us I think.

Could you copy paste the relevant part, please ? There's a lot of content, I think I missed it :)

doc/release.md
```diff
@@ -2,23 +2,19 @@
...
+1. Release what was not yet released on the `X.Y-1` branch.
+2. Remove the empty changelog for the last unreleased patch version `X.Y-1.Z'`.
+3. Check the result of `git diff vX.Y-1.Z' ChangeLog`.
+4. Install the release dependencies `pip3 install pre-commit tbump`
+5. Bump the version and release by using `tbump X.Y.Z --no-push`.
+6. Check the result visually and then by triggering the "release tests" workflow in
+   GitHub Actions first.
+7. Push the tag.
+8. Release the version on GitHub with the same name as the tag and copy and paste the
    appropriate changelog in the description. This trigger the pypi release.
```
DanielNoord, line 16:
```suggestion
## Releasing a major or minor version

**Before releasing a major or minor version check if there are any unreleased commits on the maintenance branch. If so, release a last patch release first. See `Releasing a patch version`.**

1. Install the release dependencies `pip3 install pre-commit tbump`
2. Bump the version and release by using `tbump X.Y.Z --no-push`. (For example: `tbump 2.4.0 --no-push`)
3. Check the result visually and then by triggering the "release tests" workflow in
   GitHub Actions first.
3. Push the tag.
4. Release the version on GitHub with the same name as the tag and copy and paste the
   appropriate changelog in the description. This triggers the PyPI release.
5. Create a `maintenance/X.Y.x` branch on Github. (For example: `maintenance/2.4.x`)
```

doc/release.md
```diff
@@ -2,23 +2,19 @@
...
-
-## Post release
-
-### Back to a dev version
-
-Move back to a dev version with `tbump`:
+9. Move back to a dev version with `tbump`:
```
DanielNoord, line 17:
```suggestion
6. Move the `main` branch up to a dev version with `tbump`:
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 
 Check the result and then upgrade the main branch
 
-### Milestone handling
-
-We move issue that were not done in the next milestone and block release only if it's an
-issue labelled as blocker.
+9. Delete the `X.Y-1` branch.
```
DanielNoord, line 26:
```suggestion
```

doc/release.md
```diff
@@ -2,23 +2,19 @@
...
-
-## Post release
-
-### Back to a dev version
-
-Move back to a dev version with `tbump`:
+9. Move back to a dev version with `tbump`:
 
 ```bash
 tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt during copyrite
```
DanielNoord, line 20:
```suggestion
tbump X.Y+1.0-dev0 --no-tag --no-push # You can interrupt during copyrite
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
```
DanielNoord, line 23:
```suggestion
For example:
``bash
 tbump 2.5.0-dev0 --no-tag --no-push # You can interrupt during copyrite
 git commit -am "Upgrade the version to 2.5.0-dev0 following 2.4.0 release"
``
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
 Check the result and then upgrade the main branch
 
-### Milestone handling
-
-We move issue that were not done in the next milestone and block release only if it's an
-issue labelled as blocker.
+9. Delete the `X.Y-1` branch.
 
-## Post release
+## Release a patch version
```
DanielNoord, line 28:
```suggestion
## Releasing a patch version
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
 
-### Merge tags in main for pre-commit
+We release patch version when a crash or a bug is fixed on the main branch, and it needs
+to be backported.
```
DanielNoord, line 31:
```suggestion
We release patch version when a crash or a bug is fixed on the main branch and has been cherry-picked on the maintenance branch.
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 
 Check the result and then upgrade the main branch
 
-### Milestone handling
-
-We move issue that were not done in the next milestone and block release only if it's an
-issue labelled as blocker.
+9. Delete the `X.Y-1` branch.
```
DanielNoord, line 27:
```suggestion
## Backporting a fix from `main` to the maintenance branch

Whenever a commit on `main` should be released in a patch release on the current maintenance branch we cherry-pick the commit from `main`.

1. After the PR is merged on `main` cherry-pick the commits on the `maintenance/X.Y.x` branch
2. Add a commit that moves the changelog entry from the version being developed on `main` to the changelog for the patch release. (For example, from `2.5.0` to `2.4.1`)
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 
 Check the result and then upgrade the main branch
```
DanielNoord, line 24:
```suggestion
7. Move the maintenance branch up to a dev version with `tbump`:

``bash
 tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt during copyrite
 git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
``

For example:
``bash
 tbump 2.4.1-dev0 --no-tag --no-push # You can interrupt during copyrite
 git commit -am "Upgrade the version to 2.4.1-dev0 following 2.4.0 release"
``
```

DanielNoord: Using "``" for code block because otherwise the suggestion breaks. This should be fixed manually.

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
+3. After the PR is merged cherry-pick the commits on the `X.Y-1` branch
+4. Check the result of `git diff vX.Y-1.Z-1 ChangeLog`.
+5. Install the release dependencies `pip3 install pre-commit tbump`
+6. Bump the version and release by using `tbump X.Y-1.Z --no-push`.
+7. Check the result visually and then by triggering the "release tests" workflow in
+   GitHub Actions first.
+8. Push the tag.
+9. Release the version on GitHub with the same name as the tag and copy and paste the
+   appropriate changelog in the description. This trigger the pypi release.
+10. Move back to a dev version with `tbump`:
```
DanielNoord, line 45:
```suggestion
1. Install the release dependencies `pip3 install pre-commit tbump`
2. Bump the version and release by using `tbump X.Y-1.Z --no-push`. (For example, `tbump 2.4.1 --no-push`
3. Check the result visually and then by triggering the "release tests" workflow in
   GitHub Actions first.
4. Push the tag.
5. Release the version on GitHub with the same name as the tag and copy and paste the
   appropriate changelog in the description. This trigger the pypi release.
6. Move up to a dev version with `tbump`:
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
 
 ```bash
-git checkout main
-git merge --no-edit --strategy=ours vX.Y.Z
-git push
+tbump X.Y-1.Z+1-dev0 --no-tag --no-push # You can interrupt during copyrite
+git commit -am "Upgrade the version to x.y-1.z+1-dev0 following x.y-1.z release"
 ```
+
```
DanielNoord, line 51:
```suggestion
For example:
``bash
tbump 2.4.2-dev0 --no-tag --no-push # You can interrupt during copyrite
git commit -am "Upgrade the version to 2.4.2-dev0 following 2.4.1 release"
``
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
+Check the result and then upgrade the `X.Y-1` branch
```
DanielNoord, line 52:
```suggestion
Check the result and then push to the the `maintenance/X.Y.x` branch
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
+
+13. Merge the `X.Y-1` branch on the main branch. The main branch should have the
+    changelog for `X.Y-1.Z+1`. This merge is required so `pre-commit autoupdate` works
+    for pylint.
+14. Fix version conflicts properly, or bump the version to `X.Y.0-devZ` before pushing
+    on the main branch
```
DanielNoord, line 58:
```suggestion
7. Create a new branch based on the `main` branch.
8. Merge the newly created tag into the branch: `git merge --no-edit --strategy=ours vX.Y-1.Z`. (For example: `git merge --no-edit --strategy=ours v2.4.1`.
9. Merge the changes to changelog, contributors and documentation from the maintenance branch into the branch. **Note: make sure that all changelog entries included in the patch release are removed from the changelog for the in-development version on `main`.**
10. Make a PR from the new branch on `main`.

The last steps are necessary so that `main` has access to the correct tags as well as a consolidated changelog history.
```

doc/release.md
```diff
@@ -27,21 +23,41 @@ git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
...
+
+## Milestone handling
+
+We move issue that were not done in the next milestone and block release only if it's an
+issue labelled as blocker.
```
DanielNoord, line 63:
```suggestion
We move issues that were not done to the next milestone and block releases only if there are any open issues labelled as `blocker`.
```

DanielNoord (commented):
😅 

Perhaps I should just have made a counter-PR.
I think my proposed changes address all points raised so far and are explicit in how to handle everything. Let me know what you think!

doc/release.md
Pierre-Sassoulas: Ha I understand why you don't want to merge the fix branch. You want to remove the patch release note from the main branch ? But we have only one documentation online, we want to have all the release notes on it for all versions in the right version. If we backport fix in the 2.10.x branch the first time something is released will be 2.10.1 or something, but it will also be included in 2.11.0. If we remove those release note it won't be in the release notes at all (the main branch being the source of truth for the online doc). Thus we can simply merge.

DanielNoord: I think it will be on `main`.

What I propose is:
1) Initial PR targets `2.5.0` changelog 
_When you open the PR you cannot determine which patch release it will go in, as we are now probably releasing those more often. Therefore, you should just target the in-development version and let the patch version releaser deal with moving the changelog entry_
2) After the cherry-pick a commit is added that moves the changelog entry from `2.5.0` to `2.4.1`
3) After `2.4.1` is released a post-release PR is made to `main` which updates changelog, contributors and docs. Thereby also moving the changelog entry from `2.5.0` to `2.4.1` on `main`.

At the end the changelog entry will thus be in the version that first released it. Doing it this way makes it so that updating the changelog position becomes the responsibility of the patch version releaser (who only has to do it once) instead of the PR opener (who will have to update his PR every time a new patch release is released).
For `astroid` the difference might be quite low as we don't release patches very often, but for `pylint` not having to update all PRs after every patch release is quite useful I think.

Pierre-Sassoulas: I see !

I think patch releases would still be "blocked" until we have ~~all~~ most the flying bug/crash PR done so it should not happen very often that you need to rebase and change the location in the changelog. But if it does then we don't need to do it, I'll fix it at release time like I currently do (this is what the ``git diff changelog v2.4.0`` is about, I check that all modification to the changelog are in the proper section). I'd rather have some release note to fix than all release note to fix. Also the decision to backport should be taken in the MR.  When we merge the fix branch in main, it will also fix the ``main`` branch. 

What we can do to make this process easier is to create a "To be backported" section that we then copy pasted later when releasing (Marc talked about that being done in black)

DanielNoord: >  Also the decision to backport should be taken in the MR. When we merge the fix branch in main, it will also fix the `main` branch.

This is also fine with me, but note that this is a different process than the one we have been discussing so far.

As I interpreted it, we either:
Commit to `main`, cherry-pick to `maintenance`, push changelog/doc and the tag to `main` from `maintenance`
Or:
Commit to `maintenance`, merge `maintenance` in `main`. 

I thought the whole idea was not to merge `maintenance` into `main` and only push the tags into `main`. Otherwise any fix-PR can just target the `maintenance` branch, as it will eventually end up in `main`. There is no reason to cherry-pick in that case.

But okay, we have been spending way too much time discussing this and you're likely the one who has to fix any issues that come from this process. I just wanted to highlight (and potentially solve) some future issues that might arise from this not being crystal-clear. As long as you have a clear idea of how you want to approach things I guess it should be fine.

Pierre-Sassoulas: I did not apply the suggestion directly in github because there was a lot to change but I applied it manually.

doc/release.md
```diff
@@ -2,46 +2,93 @@
...
-tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt after the first step
-git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
+tbump 2.5.0-dev0 --no-tag --no-push
+git commit -am "Upgrade the version to 2.5.0-dev0 following 2.4.0 release"
 ```
 
 Check the result and then upgrade the main branch
 
-### Milestone handling
+11. Delete the `maintenance/X.Y-1.x` branch. (For example: `maintenance/2.3.x`)
```
DanielNoord, line 39: And create a new ......

doc/release.md
```diff
@@ -2,46 +2,93 @@
...
+
+## Backporting a fix from `main` to the maintenance branch
+
+Whenever a commit on `main` should be released in a patch release on the current
+maintenance branch we cherry-pick the commit from `main`.
+
+1. During the merge request on `main`, make sure that the changelog is for the patch
+   version `X.Y-1.Z'`. (For example: `v2.3.5`)
+2. Create a `maintenance/X.Y.x` branch on Github --if it does not already exist-- from
```
DanielNoord, line 48: I think it's good to make the `maintenance` branch always as it is also a visual pointer about which versions are still supported.

doc/release.md
```diff
@@ -2,46 +2,93 @@
...
+   the `X.Y-1.0` tag . (For example: `maintenance/2.3.x` from the `v2.3.0` tag.)
+3. After the PR is merged on `main` cherry-pick the commits on the `maintenance/X.Y.x`
+   branch (For example: from `maintenance/2.4.x` cherry-pick a commit from `main`)
+4. Release a patch version
+
+## Releasing a patch version
 
-We move issue that were not done in the next milestone and block release only if it's an
-issue labelled as blocker.
+We release patch version when a crash or a bug is fixed on the main branch and has been
```
DanielNoord, line 56:
```suggestion
We release patch versions when a crash or a bug is fixed on the main branch and has been
```

doc/release.md
```diff
@@ -2,46 +2,93 @@
...
+cherry-picked on the maintenance branch.
 
-## Post release
+4. Check the result of `git diff vX.Y-1.Z-1 ChangeLog`. (For example:
```
DanielNoord, line 59: Numbering is incorrect here.

doc/release.md
```diff
@@ -2,46 +2,93 @@
...
+   `git diff v2.3.4 ChangeLog`)
+5. Install the release dependencies: `pip3 install -r requirements_test.txt`
+6. Bump the version and release by using `tbump X.Y-1.Z --no-push`. (For example:
+   `tbump 2.3.5 --no-push`)
+7. Check the result visually and then by triggering the "release tests" workflow in
+   GitHub Actions first.
+8. Push the tag.
+9. Release the version on GitHub with the same name as the tag and copy and paste the
+   appropriate changelog in the description. This trigger the pypi release.
+10. Move the `main` branch up to a dev version with `tbump`:
```
DanielNoord, line 69:
```suggestion
10. Move the `maintenance` branch up to a dev version with `tbump`:
```

DanielNoord (changes requested):
Thanks @Pierre-Sassoulas! This is becoming much clearer and should be followable for anybody not directly associated with the project.

Some last spelling comments from my end 😄

doc/release.md
Pierre-Sassoulas: I removed the numbering, it's hard to update and it's ruining the diff for each change :)

doc/release.md
Pierre-Sassoulas: [f594919](https://github.com/PyCQA/astroid/pull/1438/commits/f59491952604e9690699c74035b917e445025d09)

doc/release.md
Pierre-Sassoulas: [f594919](https://github.com/PyCQA/astroid/pull/1438/commits/f59491952604e9690699c74035b917e445025d09)

doc/release.md
Pierre-Sassoulas: Just realized that we definitely don't want to make our life harder by doing that !

doc/release.md
```diff
@@ -2,46 +2,76 @@
 
 So, you want to release the `X.Y.Z` version of astroid ?
 
-## Process
+## Releasing a major or minor version
 
-(Consider triggering the "release tests" workflow in GitHub Actions first.)
+**Before releasing a major or minor version check if there are any unreleased commits on
+the maintenance branch. If so, release a last patch release first. See
+`Release a patch version`.**
```
DanielNoord, line 9:
```suggestion
`Releasing a patch version`.**
```

doc/release.md
```diff
@@ -2,46 +2,76 @@
...
+- Check the result of `git diff vX.Y-1.Z' ChangeLog`. (For example:
+  `git diff v2.3.4 ChangeLog`)
+- Install the release dependencies: `pip3 install -r requirements_test.txt`
+- Bump the version and release by using `tbump X.Y.0 --no-push`. (For example:
+  `tbump 2.4.0 --no-push`)
+- Check the result visually and then by triggering the "release tests" workflow in
+  GitHub Actions first.
+- Push the tag.
+- Release the version on GitHub with the same name as the tag and copy and paste the
+  appropriate changelog in the description. This trigger the pypi release.
```
DanielNoord, line 22:
```suggestion
  appropriate changelog in the description. This triggers the PyPI release.
```

doc/release.md
```diff
@@ -2,46 +2,76 @@
...
+- Check the result of `git diff vX.Y-1.Z-1 ChangeLog`. (For example:
+  `git diff v2.3.4 ChangeLog`)
+- Install the release dependencies: `pip3 install -r requirements_test.txt`
+- Bump the version and release by using `tbump X.Y-1.Z --no-push`. (For example:
+  `tbump 2.3.5 --no-push`)
+- Check the result visually and then by triggering the "release tests" workflow in
+  GitHub Actions first.
+- Push the tag.
+- Release the version on GitHub with the same name as the tag and copy and paste the
+  appropriate changelog in the description. This trigger the pypi release.
```
DanielNoord, line 67:
```suggestion
  appropriate changelog in the description. This triggers the PyPI release.
```

DanielNoord (approved):

PR #1441, opened 2022-03-03
Simplify hard to maintain copyright notice

Pierre-Sassoulas:
## Description

git is the source of truth for the copyright, copyrite (the tool) was taking exponentially longer with each release, and it's
polluting the code with sometime as much as 50 lines of names. I replaced it with a single line linking to the contributors shown by github.

Also added a pre-commit hook to check and fix the copyright notice, and fixed the existing files so they have a notice

## Type of Changes


tbump.toml
```diff
@@ -28,10 +28,6 @@ src = "astroid/__pkginfo__.py"
 name = "Upgrade changelog changelog"
 cmd = "python3 script/bump_changelog.py {new_version}"
 
-[[before_commit]]
-name = "Upgrade copyrights"
-cmd = "pip3 install copyrite;copyrite --contribution-threshold 1 --change-threshold 3 --backend-type git --aliases=.copyrite_aliases . --jobs=8"
-
```
Pierre-Sassoulas, line 34: This is not automated.

doc/release.md
```diff
@@ -21,7 +21,7 @@ So, you want to release the `X.Y.Z` version of astroid ?
 Move back to a dev version with `tbump`:
 
 ```bash
-tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt during copyrite
+tbump X.Y.Z+1-dev0 --no-tag --no-push # You can interrupt after the first step
 git commit -am "Upgrade the version to x.y.z+1-dev0 following x.y.z release"
 ```
```
Pierre-Sassoulas, line 26: Another change that is not automated

.pre-commit-config.yaml
```diff
@@ -20,6 +20,13 @@ repos:
           - --expand-star-imports
           - --remove-duplicate-keys
           - --remove-unused-variables
+  - repo: https://github.com/Pierre-Sassoulas/copyright_notice_precommit
+    rev: d9215b6b2a028d1614c92cf43a9fcff3b1dd889e
+    hooks:
+      - id: copyright-notice
+        args: ["--notice=copyright.txt", "--enforce-all"]
+        exclude: tests/testdata
+        types: [python]
```
Pierre-Sassoulas, line 29: Last change that is not automated.

Pierre-Sassoulas (commented):
Approving without

DanielNoord:
@Pierre-Sassoulas Could you fix the `pre-commit`?

astroid/brain/brain_scipy_signal.py
```diff
@@ -1,14 +1,9 @@
...
-# Copyright (c) 2020 Claudiu Popa <pcmanticore@gmail.com>
-# Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
-# Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
-
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
-
+# Copyright (c) https://github.com/PyCQA/astroid/graphs/contributors
 
 """Astroid hooks for scipy.signal module."""
```
Pierre-Sassoulas, line 5: @DanielNoord I can be removing the line break here. But it's hard to understand why it fail, I'm investigating. Maybe because of Windows line ending in this file. I thougth the line ending were normalized.

Pierre-Sassoulas: (Ideally I'd like to have a line break between copyright and docstring, and I don't want to have a strange error reappear if someone add a line break later)

Pierre-Sassoulas: Yeap, that was the reason, see https://github.com/PyCQA/astroid/pull/1443

Pierre-Sassoulas:
Blocked because we need to merge #1443 first.

astroid/_ast.py
```diff
@@ -1,3 +1,6 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/graphs/contributors
```
DanielNoord, line 3: Can we make the `pre-commit` to add a new line here?

Pierre-Sassoulas: It's pretty complicated, we're checking that the text exists, and if it does not then we add it. If we change the copyright.txt with a newline, it would "work" but removing a line before the docstring would make the script add everything again not replace what's existing. Imo this is "good enough" ™️ but we could modify the pre-commit hook too, what do you think ?

DanielNoord: I'm not sure what you mean "a line before the docstring". There is not docstring here right?

Pierre-Sassoulas: Say we have this file:
```python
# Copyright
from a import b
```
And the copyright.txt is this:
```shell
# Copyright


```
Then the result will be this:
```python
# Copyright

# Copyright
from a import b
```
Unless we change the way the pre-commit hook works this is not ideal.

DanielNoord: Hmm yeah, we should probably try and change the hook then. Should be fairly trivial to add a `--new-line` flag.

```python
if not token[i + 1] == token.NL:
    # add token
```

Or something 😄

Pierre-Sassoulas: It would be trivial, if the code was working with tokens like a very good docstring linter I know 😄 But it's working using bytes directly (thus the issue with line ending) https://github.com/Pierre-Sassoulas/copyright_notice_precommit/blob/main/scripts/copyright_notice.py#L71

DanielNoord: Hmm, but in your `--auto-fix` PR can't you do something like:
```python
if config.new_line:
    lines_to_write += "\n"
file.write(lines_to_write)
```

Pierre-Sassoulas: I think it's more complicated than that, you'd have to detect that some line already exists to add only a line break instead of adding everything again, you'd also have to handle the line ending for linux/windows with input copyright.txt containing line break for linux/windows (?)

DanielNoord: Hm, time for `pycopyrightformatter` I guess? 😄 

But I do think we shouldn't add a `pre-commit` hook if it doesn't really satisfy our needs. In its current form this will always require us to do manual input after `pre-commit` has been ran against a new file. Can we do this change without the `pre-commit` hook?

Pierre-Sassoulas: I think it's good enough because we have no reason to change the beginning of files and once it's done it's done. I think I can do it properly now and make sure new files are done properly later.

Pierre-Sassoulas: ie. the important thing is the copyright notice, not the line break after it

cdce8p (commented):
I think it's ok to remove the list of contributors from each file. Just not sure a link to Github is all that useful. The contributor graph probably doesn't show everyone.

In pylint, we have a `CONTRIBUTORS` file. I think that would make sense a bit more sense. Tbh I don't know how much work it would be to gather all individual contributors and generate that file once. After that we could ask new ones to add their names to it themselves.

In contrast to pylint, I do however think that only name / username + email (optional) should be stored. Not necessary to mention individual contributions there.

If we do decide to go that route, we should make sure to add the filename to `license_files` in `setup.cfg` so it can be included in the distribution package.

Pierre-Sassoulas:
> The contributor graph probably doesn't show everyone.

True. I just checked. That's a problem.

>  Tbh I don't know how much work it would be to gather all individual contributors and generate that file once.

Not much see ``git shortlog --summary --numbered --email``. We can use the copyrite config for aliases and we're good to go. I'll do that.

Pierre-Sassoulas:
> Not much

I can't believe how much in underestimated the number of person who did 2 commits total and managed to still use a different email each time 😄

.pre-commit-config.yaml
```diff
@@ -20,6 +20,13 @@ repos:
           - --expand-star-imports
           - --remove-duplicate-keys
           - --remove-unused-variables
+  - repo: https://github.com/Pierre-Sassoulas/copyright_notice_precommit
+    rev: 6f5a59c3e1cb0f20731eb1ff32622c9a2fc72d9a
+    hooks:
+      - id: copyright-notice
+        args: ["--notice=copyright.txt", "--enforce-all", "--autofix"]
+        exclude: tests/testdata
+        types: [python]
```
DanielNoord, line 29: Shouldn't this be all? Do we only add copyright to Python files?

astroid/nodes/const.py
```diff
@@ -1,6 +1,6 @@
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
-# For details: https://github.com/PyCQA/astroid/blob/master/LICENSE
-
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
```
DanielNoord, line 2: Nice 😄

script/.copyrigth_aliases.json
```diff
@@ -0,0 +1,133 @@
+[
```
DanielNoord, line 1: To make this more "parseable" (if for whatever reason we ever want to) shouldn't we make this a object where the `name` is the key for an object with two "sub-keys": `mails` and `authorative_mail`?

script/.copyrigth_aliases.json
```diff
@@ -0,0 +1,133 @@
...
+  {
```
DanielNoord, line 2: Should this file be onder `script`? Feels a bit weird.

script/create_contributor_list.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class Alias(NamedTuple):
+    mails: List[str]
+    authoritative_mail: Optional[str]
+    name: str
+
+
+class Person(NamedTuple):
+    number_of_commit: int
```
DanielNoord, line 27:
```suggestion
    number_of_commits: int
```

script/create_contributor_list.py
```diff
@@ -0,0 +1,115 @@
...
+    name: str
+    mail: str
+
+    def __gt__(self, other: "Person") -> bool:
+        """Permit sorting contributors by number of commit."""
```
DanielNoord, line 32:
```suggestion
        """Permit sorting contributors by number of commits."""
```

script/create_contributor_list.py
```diff
@@ -0,0 +1,115 @@
...
+        return self.number_of_commit.__gt__(other.number_of_commit)
```
DanielNoord, line 33:
```suggestion
        return self.number_of_commits.__gt__(other.number_of_commits)
```

script/create_contributor_list.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __add__(self, other: "Person") -> "Person":
+        assert self.name == other.name, f"{self.name} != {other.name}"
+        assert self.mail == other.mail, f"{self.mail} != {other.mail}"
+        return Person(
+            self.number_of_commit + other.number_of_commit, self.name, self.mail
```
DanielNoord, line 39:
```suggestion
            self.number_of_commits + other.number_of_commits, self.name, self.mail
```

script/create_contributor_list.py
```diff
@@ -0,0 +1,115 @@
...
+def main() -> None:
+    parser = argparse.ArgumentParser(__doc__)
+    parser.add_argument(
+        "-o",
+        "--output",
+        default=str(DEFAULT_CONTRIBUTOR_PATH),
+        help="Where to output the contributor list",
+    )
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", default=False, help="Logging or not"
```
DanielNoord, line 55:
```suggestion
        "-v", "--verbose", action="store_true", default=False, help="Display logging messages"
```

script/.copyrigth_aliases.json
Pierre-Sassoulas: I moved it in the base dire like before.

script/.copyrigth_aliases.json
Pierre-Sassoulas: I think it's a good idea, but I don't want to change it, ever 😄 (We'd have to change the existing copyrite alias in astroid/pylint, it's not worth it imo).

.pre-commit-config.yaml
Pierre-Sassoulas: The issue is that if the comment (meaning ``#``) do not work the same it would be an issue. We only have copyright on the code right now.

cdce8p, line 29: What do you guys think about adding it only to Python files inside `astroid/*`?
IMO there isn't much value in adding it to test files or custom scripts.

CONTRIBUTORS.txt
```diff
@@ -0,0 +1,160 @@
+# This file is autogenerated by 'contributors-txt',
+# please do not modify manually
+
+Contributors
+------------
+- Claudiu Popa (pcmanticore@gmail.com)
```
cdce8p, line 6:
```suggestion
- Claudiu Popa <pcmanticore@gmail.com>
```
I think it would make sense if `contributors-txt` would use the default git style for author information.

copyright.txt
```diff
@@ -0,0 +1,3 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
```
cdce8p, line 3: I'm not a huge fan of having a dedicated `copyright.txt` file besides `LICENSE` just for that.
Would it be possible to pass it as string? Or maybe simpler, we could move it into the `script` subfolder.

setup.py
```diff
@@ -1,3 +1,7 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
```
cdce8p, line 3: Case in point, this is pretty useless here.

tests/resources.py
```diff
@@ -1,16 +1,6 @@
...
-# Copyright (c) 2019 Ashley Whetter <ashley@awhetter.co.uk>
-# Copyright (c) 2020-2021 hippo91 <guillaume.peillex@gmail.com>
-# Copyright (c) 2020 David Cain <davidjosephcain@gmail.com>
-# Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
-# Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
-# Copyright (c) 2022 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
-
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
```
cdce8p, line 3: I would suggest we remove the license header from all test files.

cdce8p (commented):
Please add `CONTRIBUTORS.txt` to `setup.cfg` so it's included in the distribution package.
This is how pylint does it. [pylint -> setup.cfg](https://github.com/PyCQA/pylint/blob/4ca885fd8c35e6781a359cda0e4de2c0910973b6/setup.cfg#L10-L12)

script/.copyrigth_aliases.json
Pierre-Sassoulas: Backtracking on that : the current format permit inconsistencies in the json, (see the "bot" it has two entries with 2 sets of ``mails``). I'm going to make a migration tool.

copyright.txt
Pierre-Sassoulas: Let's move it to the subfolder, the pre-commit hook is using bytes which make passing a string with line break complicated.

tests/resources.py
Pierre-Sassoulas: It send the message that the test files are not copyrighted, no ? And they were originally. How about ``astroid.*`` and ``tests*`` but not custom scripts ?

cdce8p: I just wonder what the benefit of having it in test files would be. IMO it's more or less useless. Those files aren't part of the distribution package and mostly boilerplate. In the end, there is still the LICENSE file which applies to all files, regardless if a file has a license header or not.

_Disclaimer: I'm not a lawyer, this is not legal advice._

Pierre-Sassoulas: Also not a lawyer 😄 To be frank I also wondered what is the use of having it in any files at all. Apparently the need to do that was removed in march 1989 when the US joined the Berne Convention (before that anything not expressively copyrighted could be stolen !). When pylint was created the remnant of the practice of putting copyright everywhere were still very much alive, probably the reason why it's done that way. That being said, I'd like to be pretty conservative and careful here as I'd prefer to be safe than sorry in this case.  (Fun fact an actual law person in at my job told us developper to do that for internal unpublished code in 2021. Guess no lawyers ever got sacked for asking to put copyright and trademark everywhere)

.pre-commit-config.yaml
DanielNoord: Fine with me!

tests/resources.py
cdce8p: You can't add too much usually 😄 Some projects include the whole license text in every file. Others, like Python, don't add any headers.

In our case it's much more a question what is practical. My argument would be that it's quite unlikely anyone will ever pursue the copyright or GPL license for pylint / astroid code. Much less so for test code. It's already incredibly difficult to do these things for linux, a much bigger project.

Anyway, I guess a few lines more or less don't really hurt.
_To get back to pylint. There we have added header to general test files, but not the individual functional tests. I think we should keep it that way._

Pierre-Sassoulas: I agree we should not touch pylint's functional files especially as it would modify all the result txt files and do a massive diff for no reason.

script/.copyrigth_aliases.json
Pierre-Sassoulas: This is blocking the release of astroid 2.11, so it's not perfect but I'm going to do that in https://github.com/Pierre-Sassoulas/contributors-txt/issues/3... later.

DanielNoord: Are you saying you're going to merge this and do a follow-up after `2.11`, or are you moving this from `2.11`?

Either way, I'd say let's move to `2.12`. Doesn't really make sense to do two commits of 100+ lines in quick succession. This is not impacting the performance of `astroid` for now.

Pierre-Sassoulas: I was thinking about merging before 2.11 without changing the data structure of the json. It's not really problematic that there are duplication in the json. The fix will be a version upgrade + a migration of the json.

setup.py
Pierre-Sassoulas: [62e01c1](https://github.com/PyCQA/astroid/pull/1441/commits/62e01c19ae4a90503d1f3212c062a44ac347cd35)

copyright.txt
Pierre-Sassoulas: https://github.com/PyCQA/astroid/pull/1441/commits/c8e8991853ab217d87f3ea4ea80e25fdd8720b30

.pre-commit-config.yaml
Pierre-Sassoulas: Discussed [here](https://github.com/PyCQA/astroid/pull/1441#discussion_r820155488)

Pierre-Sassoulas:
> Please add CONTRIBUTORS.txt to setup.cfg so it's included in the distribution package.

👍 https://github.com/PyCQA/astroid/pull/1441/commits/db0bff2886dd5c176b93d208f81d64adf13e1441

script/.copyrigth_aliases.json
DanielNoord: That does feel kind of rushed: why would we commit code/data that we know is faulty. As I said, since there is no reason to rush this in the next 24 hours and we'll likely have some more review time while we wait on the last `pylint` `2.13` PRs to be adopted/reviewed I think it's best to make this PR the _big_ copyright PR. That would also help with porting this PR to `pylint` as all open discussions would be solved.

Pierre-Sassoulas: Fair enough.

Pierre-Sassoulas: Done, and upgraded :)

CONTRIBUTORS.txt
Pierre-Sassoulas: Done [f77b0c3](https://github.com/PyCQA/astroid/pull/1441/commits/f77b0c3d44944ab13d80bb76c8c3d1e0bb6aa0da)

CONTRIBUTORS.txt
```diff
@@ -0,0 +1,161 @@
...
+Contributors
+------------
+- Claudiu Popa <pcmanticore@gmail.com>
+- LOGILAB S.A. (Paris, FRANCE) <contact@logilab.fr>
+- Sylvain Thénault <thenault@gmail.com>
+- Pierre Sassoulas <pierre.sassoulas@gmail.com>
+- hippo91 <guillaume.peillex@gmail.com>
+- Marc Mueller <30130371+cdce8p@users.noreply.github.com>
+- Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
+- Google, Inc. <None>
```
cdce8p, line 13: Should we keep `<None>` or remove it in this case?

Pierre-Sassoulas: Nice catch !

CONTRIBUTORS.txt
```diff
@@ -0,0 +1,161 @@
...
+
+Contributors
+------------
+- Claudiu Popa <pcmanticore@gmail.com>
+- LOGILAB S.A. (Paris, FRANCE) <contact@logilab.fr>
+- Sylvain Thénault <thenault@gmail.com>
+- Pierre Sassoulas <pierre.sassoulas@gmail.com>
+- hippo91 <guillaume.peillex@gmail.com>
+- Marc Mueller <30130371+cdce8p@users.noreply.github.com>
+- Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
```
cdce8p, line 12: I was wondering about the sorting. Does it make sense to sort by number of commits? That might cause a more extensive diff.

Maybe just alphabetical would be better? Or two different sections `Team` and `Contributors`? _Section names can be improved probably._

Pierre-Sassoulas: I don't mind the little diff that will happen when Daniel surpass us all 😄 

> two different sections Team and Contributors? 

This is what is done for pylint more or less (just really out of date), but we'd need to have some configuration with the list of admin though. Is it worth it ?

cdce8p: If you're ok with the diff, then it's fine by me 🤷🏻‍♂️

Pierre-Sassoulas: Just FYI I might try to do something a little fancier for pylint to preserves the comments about what was done by whom but I did not do anything yet. I'm trying to release astroid first 😄

script/.contributors_aliases.json
```diff
@@ -0,0 +1,108 @@
...
+  },
+  "Stefan Scherfke": {
+    "authoritative_mail": "stefan@sofa-rockers.org",
+    "mails": ["stefan.scherfke@energymeteo.de", "stefan@sofa-rockers.org"]
+  },
+  "Sylvain Th\u00e9nault": {
+    "authoritative_mail": "thenault@gmail.com",
+    "mails": ["thenault@gmail.com", "sylvain.thenault@logilab.fr"]
+  },
+  "Ville Skytt\u00e4": {
```
DanielNoord, line 97: Should this be normalised? It is in `CONTRIBUTORS.txt`

Pierre-Sassoulas: [dc10b01](https://github.com/PyCQA/astroid/pull/1441/commits/dc10b0184d1565273e4c65389e7fb1b09f4bf003)

CONTRIBUTORS.txt
```diff
@@ -1,24 +1,30 @@
 # This file is autogenerated by 'contributors-txt',
+# using the configuration in 'script/.contributors_aliases.json'
 # please do not modify manually
 
-Contributors
-------------
+Mainteners
```
DanielNoord, line 5:
```suggestion
Maintainers
```

Pierre-Sassoulas: Urgh, costly typo. 2 commits a full release and a regeneration 😄

DanielNoord (approved):

script/.contributors_aliases.json
```diff
@@ -0,0 +1,153 @@
+{
+  "Areveny": {
+    "authoritative_mail": "areveny@protonmail.com",
+    "mails": ["areveny@protonmail.com", "self@areveny.com"],
+    "team": "true"
```
cdce8p, line 5: A true / false value here is quite inflexible. What about the actual team / group name? That would allow for more than two groups if we ever want that.

Pierre-Sassoulas: [c8f4f3b](https://github.com/PyCQA/astroid/pull/1441/commits/c8f4f3b5ab2004b03a0dd2a23c1626b791cc8029)

sandrotosi:
> git is the source of truth for the copyright, 

I'm sorry i'm late to the party, and please note i'm coming from Debian's perspective, but this statement is not correct.

copyright is the list of people who can apply a license to their work (ie they will need to agree to change the license)

Now it looks like every single person in https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt bears equal copyright claims to every single file in the astroid project (ie, *all* of them would need to agree to change the license of a single file, even tho they never contributed to it)

this is a worst situation than before and, while IANAL it may not even be legally appropriate.

Take my notes with caution, i just package astroid  for debian and i have no standing on how astroid is developed and maintained.

Pierre-Sassoulas:
Thank you for the insight @sandrotosi, I'm not a lawyer either and I know this subject can be touchy ! (But a list of 20 peoples on top of file that we had to update in quadrinomial time was a lot to deal with). Would you think it's better if we specify the actual git command to get the contributor for a file for each file, and remove the misleading link to the full list of contributor ? 

I.e. a light version of that would be simply the github link to the file as in https://github.com/PyCQA/astroid/blob/main/astroid/_ast.py you have a list of 4 contributors in interface, Maybe there's an API I can find for this so we only get the actual list.

As an aside I think changing the license is never going to happen (even if I'd prefer a MIT license myself) as asking everyone is near impossible, but there's probably other use cases too ?

sandrotosi:
sorry i didnt reply sooner

> Thank you for the insight @sandrotosi, I'm not a lawyer either and I know this subject can be touchy !

IANAL either

> (But a list of 20 peoples on top of file that we had to update in quadrinomial time was a lot to deal with).

i simply cant gauge the level of pain this caused, from an outside POV, this list could be updated just once before a release: would that help?

> Would you think it's better if we specify the actual git command to get the contributor for a file for each file, and remove the misleading link to the full list of contributor ?

i'm afraid that's not enough (the git repo is not what get's distributed, which is what matters here): the way i use to clarify copyright&license is that the copyright holders are the only legal entity that can set a given license for the piece of code they have copyright rights over.

now, if you simply state

`# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt`

that has simply no legal value, because it's an ever changing file, and it does not specify a legal entity able to hold a copyright over code (a URL in this case).

But even if we would consider that file a list of contributors and ship it with the code tarball (which we do for debian, since we fetch it from the github tags, and they contain all the files in the repo), do really all the people listed in that file have legal claim to change the license of every single line of code in astroid? i think not.

Other projects i think use CLA, and/or other mechanisms to handle copyright claims, and from a personal perspective i always liked the complete list of copyright holders for each files (it makes my debian work so much easier), so i may be more than biased in this discussion.

> As an aside I think changing the license is never going to happen (even if I'd prefer a MIT license myself) as asking everyone is near impossible, but there's probably other use cases too ?

a given set of copyright holders can decided to change the license of their code, as long as it remains compatible with LGPL2.1+ (which is the global license for the whole project). It's a bit of a stretch, but legally possible

Pierre-Sassoulas:
> that has simply no legal value, because it's an ever changing file, and it does not specify a legal entity able to hold a copyright over code (a URL in this case).

Say we specify the commit like this for each file so it's not changing anymore (https://github.com/PyCQA/astroid/blob/e5f7488f9370b8f69c3605aa1be8650c7a3d76e4/astroid/_ast.py). Would it work for you ?

Result at url:
copyrightholfder

We could maybe find an API to have a json directly.

>  CLA, and/or other mechanisms to handle copyright claims,

This could work, but for me working on the copyright does not make pylint better so I really don't want to spend time on this (unless paid).

The license has been LGPL for years, it's contaminating, so whatever is done is under LGPL. Having very detailed and exact copyright feel like law theater to be honest. And I have also a feeling it's only benefiting corporation, more precisely some over-cautious corporate lawyer that do not pay for pylint because everyone working on it do it for free anyway. I'm not going to sue some multinational company if they make a private fork and do not share the result, I won't even know about it.  This is not going to be prevented by me painfully upgrading every file with a 2 screen long list of copyright holder that everyone will then have to scroll past every time they open a file to edit.

PR #1442, opened 2022-03-03
Add `pathlib` brain to handle `.parents` inference

deepyaman:
## Steps

## Description

The `PurePath.parents` sequence supports slices since Python 3.10. This brain allows pylint to understand that passing an index (as opposed to a slice) to `PurePath.parents` will still yield a path, not a tuple.

Given `test_index.py`:

```python
from pathlib import Path


def _main() -> None:
    current_path = Path().resolve()
    parent_path = current_path.parents[2].resolve()
    print(f"{current_path=}")
    print(f"{parent_path=}")


if __name__ == "__main__":
    _main()
```

Running pylint yields:

```
(astroid-3.10) WASH-178551-C02X31K9JHD4:customerone deepyaman_datta$ pylint test_index.py --disable missing-module-docstring

Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

Given `test_slice.py`:

```python
from pathlib import Path


def _main() -> None:
    current_path = Path().resolve()
    parent_paths = [p.resolve() for p in current_path.parents[:2]]
    print(f"{current_path=}")
    print(f"{parent_paths=}")


if __name__ == "__main__":
    _main()
```

Running pylint yields:

```
(astroid-3.10) WASH-178551-C02X31K9JHD4:customerone deepyaman_datta$ pylint test_slice.py --disable missing-module-docstring

Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

## Type of Changes

## Related Issue

Closes with a test case https://github.com/PyCQA/pylint/issues/5783
Partially addresses https://github.com/PyCQA/pylint/issues/5832


astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+
+from typing import Iterator, Optional
+
+from astroid import bases, context, inference_tip
+from astroid.const import PY310_PLUS
+from astroid.exceptions import UseInferenceDefault
+from astroid.manager import AstroidManager
+from astroid.nodes.node_classes import Attribute, Slice, Subscript
```
DanielNoord, line 12: Could you import `from astroid import nodes` here? The `nodes` module shouldn't depend on this brain so we should be fine 😄

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+from astroid.nodes.node_ng import NodeNG
+
+
+def _looks_like_parents_subscript(node: NodeNG) -> bool:
```
DanielNoord, line 16: It has been quite some time that I looked into this myself, but I believe `register_transform` makes it so that only the class passed as first argument is pass to this function.

Perhaps you could do this and run pytest:
```python
if not isinstance(node, nodes.Subscript):
    breakpoint()
```

If there is no break then we can update this to be `nodes.Subscript` instead.

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+
+import astroid
+from astroid import bases
+from astroid.const import PY310_PLUS
+
+if not PY310_PLUS:
+    pytest.skip(
```
DanielNoord, line 8: Can you add tests for <3.10 as well? So just mark the tests for 3.10+ as such and add some tests for the old behaviour.

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
...
+        "The parents sequence supports slices since 3.10", allow_module_level=True
+    )
+
+
+def test_inference_parents_subscript_index():
```
DanielNoord, line 13:
```suggestion
def test_inference_parents_subscript_index() -> None:
```

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
...
+    parent_path = current_path.parents[2]
+    parent_path
+    """
+    )
+    inferred = next(name_node.infer())
+    assert isinstance(inferred, bases.Instance)
+    assert inferred.qname() == "pathlib.Path"
+
+
+def test_inference_parents_subscript_slice():
```
DanielNoord, line 29:
```suggestion
def test_inference_parents_subscript_slice() -> None:
```

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
...
+    name_node = astroid.extract_node(
+        """
+    from pathlib import Path
+
+    current_path = Path().resolve()
+    parent_path = current_path.parents[2]
+    parent_path
+    """
+    )
+    inferred = next(name_node.infer())
```
DanielNoord, line 24: Could you use `list` here and assert that `len(list) == 1`. We want to make sure that there is no ambiguity in the inference here and `next()` does not tell us that.

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
...
+    name_node = astroid.extract_node(
+        """
+    from pathlib import Path
+
+    current_path = Path().resolve()
+    parent_path = current_path.parents[:2]
+    parent_path
+    """
+    )
+    inferred = next(name_node.infer())
```
DanielNoord, line 40: Could you use `list` here and assert that `len(list) == 1`. We want to make sure that there is no ambiguity in the inference here and `next()` does not tell us that.

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
+import pytest
```
DanielNoord, line 1: Just for "totality" could you also add a test for the inference of `parents` itself?

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+    return (
+        isinstance(node, Subscript)
+        and isinstance(node.value, Attribute)
+        and node.value.attrname == "parents"
+    )
+
+
+def infer_parents_subscript(
+    subscript_node: Subscript, ctx: Optional[context.InferenceContext] = None
+) -> Iterator[bases.Instance]:
```
DanielNoord, line 26: Perhaps we should add a docstring to explain why this is needed, or a comment between L33 and L34.

DanielNoord (commented):
Thanks @deepyaman, left some initial comments!

I also updated your OP not to directly close the `pylint` issue. We tend to fix issues in `astroid` and then add a regression test to `pylint` to fully close the issue. If you want you can make a PR with a regression test for `pylint` as well!

deepyaman:
@DanielNoord Thank you for the detailed feedback! I will try and implement the requested changes/feedback over the next few days and get back to you. :)

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+from astroid.nodes.node_classes import Attribute, Slice, Subscript
+from astroid.nodes.node_ng import NodeNG
+
+
+def _looks_like_parents_subscript(node: NodeNG) -> bool:
+    return (
+        isinstance(node, Subscript)
+        and isinstance(node.value, Attribute)
+        and node.value.attrname == "parents"
+    )
```
cdce8p, line 21: This isn't enough, it will also match unintended code. Just an example. I would suggest adding something like it as a test.
```py
name_node = astroid.extract_node(
    """\
class A:
    parents = 42

c = A()
error = c.parents[2]
error
"""
)
inferred = next(name_node.infer())
print(inferred)
```

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def infer_parents_subscript(
+    subscript_node: Subscript, ctx: Optional[context.InferenceContext] = None
+) -> Iterator[bases.Instance]:
+    if isinstance(subscript_node.slice, Slice):
+        raise UseInferenceDefault
+
+    yield from subscript_node.value.expr.infer(context=ctx)
```
cdce8p, line 30: As mentioned earlier, it's not quite enough to check that `attrname == "parents"`.
A rough idea what might work instead
```py
    inferred = next(subscript_node.value.expr.infer(context=ctx))
    if not (
        isinstance(inferred, bases.Instance)
        and isinstance(inferred._proxied, nodes.ClassDef)
        and inferred.qname().startswith("pathlib.")
        and inferred.qname().endswith("Path")
    ):
        raise UseInferenceDefault
    if isinstance(subscript_node.slice, Slice):
        raise UseInferenceDefault

    return iter([inferred])
```

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+if PY310_PLUS:
+    AstroidManager().register_transform(
+        Subscript, inference_tip(infer_parents_subscript), _looks_like_parents_subscript
+    )
```
cdce8p, line 36:
```suggestion
AstroidManager().register_transform(
    Subscript, inference_tip(infer_parents_subscript), _looks_like_parents_subscript
)
```
The subscript isn't limited to `3.10`. Just `[2:]`.

cdce8p (changes requested):

astroid/brain/brain_pathlib.py
deepyaman: Got it, tested it and (of course) it's as you all said!

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,63 @@
...
+    parent_path = current_path.parents[:2]
+    parent_path
+    """
+    )
+    inferred = next(name_node.infer())
+    assert isinstance(inferred, bases.Instance)
+    assert inferred.qname() == "builtins.tuple"
+
+
+def test_inference_parents_subscript_not_path():
```
deepyaman, line 50:
```suggestion
def test_inference_parents_subscript_not_path() -> None:
```

astroid/brain/brain_pathlib.py
deepyaman: Done!

tests/unittest_brain_pathlib.py
deepyaman: Done! Used `.inferred()` instead of `.infer()` and checked the length of that, because that's what astroid/tests/unittest_brain_dataclasses.py does, although I'm not entirely sure what the difference is (`.inferred()` leverages an inference cache?).

astroid/brain/brain_pathlib.py
deepyaman: Comment added!

tests/unittest_brain_pathlib.py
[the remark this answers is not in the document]
deepyaman: Added a note in the implementation, as well as adding a branch for the test case if Python <3.10.

astroid/brain/brain_pathlib.py
deepyaman: As you mentioned above, the result pre-3.10 is to always `Uninferable`, which is also OK; added a comment to explain.

tests/unittest_brain_pathlib.py
deepyaman: Done!

tests/unittest_brain_pathlib.py
deepyaman: I feel like I'm almost there, but it's not currently working for:

```py
from pathlib import Path


def _main() -> None:
    current_path = Path().resolve()
    path_parents = current_path.parents
    parent_path = path_parents[2].resolve()
    print(f"{current_path=}")
    print(f"{parent_path=}")


if __name__ == "__main__":
    _main()
```

It fails with the same `test.py:7:18: E1101: Instance of 'tuple' has no 'resolve' member (no-member)`.

Seems to be why my attempt at adding a nicer-looking test failed:

```py
def test_inference_parents_subscript_index() -> None:
    """Test inference of ``pathlib.Path.parents``, accessed by index."""
    parents, path = astroid.extract_node(
        """
    from pathlib import Path

    current_path = Path().resolve()
    path_parents = current_path.parents
    path_parents  #@
    path_parents[2]  #@
    """
    )
    inferred = parents.inferred()
    assert len(inferred) == 1
    assert isinstance(inferred[0], bases.Instance)
    assert inferred[0].qname() == "pathlib._PathParents"

    inferred = path.inferred()
    assert len(inferred) == 1
    if PY310_PLUS:
        assert isinstance(inferred[0], bases.Instance)
        assert inferred[0].qname() == "pathlib.Path"
    else:
        assert inferred[0] is Uninferable
```

astroid/brain/brain_pathlib.py
DanielNoord: I think the check is somewhat unnecessary because `subscript_node.value` is `nodes.Attribute` due to `_looks_like_parents_subscript`.
However, we don't really "know" this when we are in this function and therefore once we turn on `mypy` it will (rightly) complain that `qname` is not defined on `value`. I think it makes sense to add these `isinstance` checks as well.

tests/unittest_brain_pathlib.py
```diff
@@ -0,0 +1,84 @@
...
+    parent_path
+    """
+    )
+    inferred = name_node.inferred()
+    assert len(inferred) == 1
+    if PY310_PLUS:
+        assert isinstance(inferred[0], bases.Instance)
+        assert inferred[0].qname() == "pathlib.Path"
+    else:
+        assert inferred[0] is Uninferable
```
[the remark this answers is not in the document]
deepyaman: `tuple` should not be correct; when passed an integral index (as opposed to a slice), you get a single path.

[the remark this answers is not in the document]
DanielNoord: No, but since this PR changes the behaviour here shouldn't we try and infer the correct value? Or do you think that is outside of the scope of this PR?

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,42 @@
...
+
+if PY310_PLUS:
+    # Before adding support for slices to the parents sequence in Python
+    # 3.10, the return value for ``pathlib._PathParents.__getitem__`` is
+    # always ``Uninferable``, so we only apply the brain tip since 3.10.
+    AstroidManager().register_transform(
+        nodes.Subscript,
+        inference_tip(infer_parents_subscript),
+        _looks_like_parents_subscript,
+    )
```
deepyaman, line 42:
```suggestion
AstroidManager().register_transform(
    nodes.Subscript,
    inference_tip(infer_parents_subscript),
    _looks_like_parents_subscript,
)
```

tests/unittest_brain_pathlib.py
deepyaman, line 47:
```suggestion
    assert isinstance(inferred[0], bases.Instance)
    assert inferred[0].qname() == "pathlib.Path"
```

ChangeLog
```diff
@@ -22,6 +22,10 @@ Release date: TBA
 
   Closes #1410
 
+* Add ``pathlib`` brain to handle ``pathlib.PurePath.parents`` inference in Python 3.10+.
```
deepyaman, line 25:
```suggestion
* Add ``pathlib`` brain to handle ``pathlib.PurePath.parents`` inference.
```

tests/unittest_brain_pathlib.py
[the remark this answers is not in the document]
deepyaman: @DanielNoord Sorry for the delay in taking a look at this. I've removed the `if PY310_PLUS` condition, and the brain tip seems to be testing fine for older Python versions.

astroid/brain/brain_pathlib.py
deepyaman: @cdce8p @DanielNoord I have updated this to add the `isinstance` checks to my implementation.

deepyaman:
@DanielNoord @cdce8p Sorry for the delay in addressing your comments, but I believe I've done it now.

I'm still not sure how to resolve this scenario:

```python
(astroid-3.10) WASH-178551-C02X31K9JHD4:astroid deepyaman_datta$ cat test.py 
import astroid

code = """from pathlib import Path
current_path = Path().resolve()
path_parents = current_path.parents
path_parents[2]
"""


module = astroid.parse(code)
node = module.body[-1].value
print(node.inferred())
(astroid-3.10) WASH-178551-C02X31K9JHD4:astroid deepyaman_datta$ python test.py 
[<Instance of builtins.tuple at 0x140247194163456>]
```

Even if I can make the inference tip activate on a `Name` node, I can't access the contextual information I need to determine the type. Should I try and make it return `Uninferable` instead?

I'm also not completely sure why I'm getting the Flake8 issues even after updating to latest `main`.

DanielNoord:
> @DanielNoord @cdce8p Sorry for the delay in addressing your comments, but I believe I've done it now.
> 
> I'm still not sure how to resolve this scenario:
> 
> ```python
> (astroid-3.10) WASH-178551-C02X31K9JHD4:astroid deepyaman_datta$ cat test.py 
> import astroid
> 
> code = """from pathlib import Path
> current_path = Path().resolve()
> path_parents = current_path.parents
> path_parents[2]
> """
> 
> 
> module = astroid.parse(code)
> node = module.body[-1].value
> print(node.inferred())
> (astroid-3.10) WASH-178551-C02X31K9JHD4:astroid deepyaman_datta$ python test.py 
> [<Instance of builtins.tuple at 0x140247194163456>]
> ```
> 
> Even if I can make the inference tip activate on a `Name` node, I can't access the contextual information I need to determine the type. Should I try and make it return `Uninferable` instead?

What contextual information are you referring to? I think `uninferable` is probably better than an incorrect inference.

> I'm also not completely sure why I'm getting the Flake8 issues even after updating to latest `main`.

Seems like our CI is broken for flake8. We'll need to change some stuff to fix those issues.

Pierre-Sassoulas:
> Seems like our CI is broken for flake8. We'll need to change some stuff to fix those issues.

I think it's due to ``additional_dependencies: [flake8-bugbear,`` where bugbear version is not set. Interestingly it's the same check than what we just added in pylint 2.14.

deepyaman:
> What contextual information are you referring to? I think `uninferable` is probably better than an incorrect inference.

@DanielNoord If I change `_looks_like_parents_subscript` to also trigger on `isinstance(node.value, nodes.Name)`:

```python
def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
    return isinstance(node.value, nodes.Attribute) and node.value.attrname == "parents" or isinstance(node.value, nodes.Name)
```

Everything looks good inside `infer_parents_subscript`:

```python
(Pdb) subscript_node
<Subscript l.4 at 0x7f8ceb29ffd0>
(Pdb) subscript_node.value
<Name.path_parents l.4 at 0x7f8ceb29fa90>
(Pdb) value = next(subscript_node.value.infer())
(Pdb) value
<Instance of pathlib._PathParents at 0x140243218564960>
(Pdb) isinstance(value, bases.Instance) and isinstance(value._proxied, nodes.ClassDef) and value.qname() == "pathlib._PathParents"
True
(Pdb) subscript_node.slice
<Const.int l.4 at 0x7f8ceb29fbe0>
```

However, I can't figure out any way to infer the type of `current_path` from this point, which is what I'd normally use to set the return type of the `.parents` subscript. There's nothing I can find like `subscript_node.value.expr`.

However, I'm not sure if it's right to handle `Name` nodes explicitly like this, or to make them `Uninferable`, since it seems (in my head, I don't understand `astroid` in depth to actually know) like there would be so many instances where passing a `Name` node/assigning to an intermediate variable would lose a lot of contextual type information. I couldn't find other code handling `Name` nodes so explicitly within `astroid`, at a glance.

deepyaman:
> > What contextual information are you referring to? I think `uninferable` is probably better than an incorrect inference.
> 
> @DanielNoord If I change `_looks_like_parents_subscript` to also trigger on `isinstance(node.value, nodes.Name)`:
> 
> ```python
> def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
>     return isinstance(node.value, nodes.Attribute) and node.value.attrname == "parents" or isinstance(node.value, nodes.Name)
> ```
> 
> Everything looks good inside `infer_parents_subscript`:
> 
> ```python
> (Pdb) subscript_node
> <Subscript l.4 at 0x7f8ceb29ffd0>
> (Pdb) subscript_node.value
> <Name.path_parents l.4 at 0x7f8ceb29fa90>
> (Pdb) value = next(subscript_node.value.infer())
> (Pdb) value
> <Instance of pathlib._PathParents at 0x140243218564960>
> (Pdb) isinstance(value, bases.Instance) and isinstance(value._proxied, nodes.ClassDef) and value.qname() == "pathlib._PathParents"
> True
> (Pdb) subscript_node.slice
> <Const.int l.4 at 0x7f8ceb29fbe0>
> ```
> 
> However, I can't figure out any way to infer the type of `current_path` from this point, which is what I'd normally use to set the return type of the `.parents` subscript. There's nothing I can find like `subscript_node.value.expr`.
> 
> However, I'm not sure if it's right to handle `Name` nodes explicitly like this, or to make them `Uninferable`, since it seems (in my head, I don't understand `astroid` in depth to actually know) like there would be so many instances where passing a `Name` node/assigning to an intermediate variable would lose a lot of contextual type information. I couldn't find other code handling `Name` nodes so explicitly within `astroid`, at a glance.

Hi @DanielNoord @cdce8p! I was wondering if you may have some guidance on how to proceed with this? Would love to close out this issue; just not sure what's the best approach to resolving the above issue.

DanielNoord:
@deepyaman Sorry for letting this rest so long. I have been quite busy and haven't had the time to look into this enough. I hope to find some time for it this weekend. Again, sorry for the long delay!

DanielNoord:
@deepyaman Sorry for letting this rest so long. I have been quite busy and haven't had the time to look into this enough. I hope to find some time for it this weekend. Again, sorry for the long delay!

deepyaman:
> @deepyaman Sorry for letting this rest so long. I have been quite busy and haven't had the time to look into this enough. I hope to find some time for it this weekend. Again, sorry for the long delay!

No problem, and thanks again for taking the time to guide me through this!

DanielNoord (approved):
@deepyaman Sorry this took so long to review. When I finally got to it I was in so deep that it seemed easier to just push the fixes I found.

Sorry for taking over your PR with prior notice like that.

In this new iteration we now also correctly infer an index lookup. I do this by creating an inference of the normal `ClassDef` for `Path` and then instantiating an instance from it.
I am not completely sure `Uninferable` for a `Slice` is correct on <3.9. I'll let another maintainer comment on that: do we infer `Uninferable` for code that would raise warnings?

I'll re-ask @cdce8p for a review since he had some valuable input in the first round.

tests/unittest_brain_pathlib.py
```diff
@@ -72,7 +77,7 @@ class A:
         parents = 42
 
     c = A()
-    error = c.parents[2]
+    error = c.parents[:2]
```
deepyaman, line 80: Did this have to be changed?

astroid/brain/brain_pathlib.py
```diff
@@ -2,36 +2,50 @@
...
 
     if not (
         isinstance(value, bases.Instance)
         and isinstance(value._proxied, nodes.ClassDef)
         and value.qname() == "pathlib._PathParents"
     ):
-        raise UseInferenceDefault
+        return False
+
+    return True
```
deepyaman, line 39: Maybe just:

```suggestion
    return (
        isinstance(value, bases.Instance)
        and isinstance(value._proxied, nodes.ClassDef)
        and value.qname() == "pathlib._PathParents"
    )
```

deepyaman (commented):
@DanielNoord No problem! Thanks for resolving it!

tests/unittest_brain_pathlib.py
DanielNoord: Originally I hadn't fixed the normal index look up, so this was not really testing anything:
The brain returned `Uninferable` and the non-brain would return `Uninferable`. So we weren't actually testing if this doesn't actually enter the brain. By using the slice we can see that this doesn't trigger the brain as it would give a different result for `PY3.10+`.

Now that I fixed the brain to work with index as well the change is perhaps not necessary anymore.

astroid/brain/brain_pathlib.py
DanielNoord: yeah that's probably better! Committed!

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,51 @@
...
+from astroid import bases, context, extract_node, inference_tip, nodes
+from astroid.exceptions import InferenceError, UseInferenceDefault
+from astroid.manager import AstroidManager
+
+PATH = extract_node(
+    """from pathlib import Path
+Path
+"""
+)
+PATH_CLASSDEF: nodes.ClassDef = next(PATH.infer())
```
cdce8p, line 18: Inference calls are non-trivial. I'm don't think it's a good idea to do them at import. Better to only define the code template as global const and call `extract_node` (or even `_extract_single_node`) in `infer_parents_subscript`.

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
+    if not isinstance(node.value, nodes.Name) and not (
+        isinstance(node.value, nodes.Attribute) and node.value.attrname == "parents"
+    ):
```
cdce8p, line 24:
```suggestion
    if not (
        isinstance(node.value, nodes.Name)
        or isinstance(node.value, nodes.Attribute)
        and node.value.attrname == "parents"
    ):
```
`not ... and not (... and ...)` is quite complicated and difficult to follow.

astroid/brain/brain_pathlib.py
```diff
@@ -0,0 +1,51 @@
...
+        and isinstance(value._proxied, nodes.ClassDef)
+        and value.qname() == "pathlib._PathParents"
+    )
+
+
+def infer_parents_subscript(
+    subscript_node: nodes.Subscript, ctx: context.InferenceContext | None = None
+) -> Iterator[bases.Instance]:
+    if isinstance(subscript_node.slice, nodes.Const):
+        return iter((PATH_CLASSDEF.instantiate_class(),))
```
cdce8p, line 42: `_extract_single_node` should be done here.

cdce8p (changes requested):

astroid/brain/brain_pathlib.py
DanielNoord: But it isn't it better to use the actual `ClassDef`? You're more experienced here, but this seems to pass tests and doesn't this give us more information than a constructed mimic of `Path`?

ChangeLog
```diff
@@ -39,6 +43,7 @@ Release date: TBA
 * The Qt brain now correctly treats calling ``.disconnect()`` (with no
   arguments) on a slot as valid.
 
+
```
cdce8p, line 46:
```suggestion
```

astroid/brain/brain_pathlib.py
cdce8p: It's probably fine to use `extract_node` and `infer` to get the `ClassDef`, just don't do it at Module level / import time.

As example, this is how the `typing` brain does something similar
https://github.com/PyCQA/astroid/blob/e57ac42bca80c105bbdad4bae61637b39288a47b/astroid/brain/brain_typing.py#L99-L103
https://github.com/PyCQA/astroid/blob/e57ac42bca80c105bbdad4bae61637b39288a47b/astroid/brain/brain_typing.py#L167-L168

DanielNoord: I'll refactor!

cdce8p:
@DanielNoord I moved `extract_node` inside `infer_parents_subscript` as well. We should try to do as little computation as possible in the global scope. With `iter`, I switched the one-item tuple with a list. Although correct, it's quite easy to accidentally remove the trailing comma or overlook it. A list is a bit safer there.

cdce8p (approved):
Good to merge once the CI checks pass.

DanielNoord:
> @DanielNoord I moved `extract_node` inside `infer_parents_subscript` as well. We should try to do as little computation as possible in the global scope. With `iter`, I switched the one-item tuple with a list. Although correct, it's quite easy to accidentally remove the trailing comma or overlook it. A list is a bit safer there.

Thanks! I'm still finding my way through the `astroid` brains. I'll keep the global computation in mind 😄

cdce8p:
> I'll keep the global computation in mind 😄

Off-topic: There is an open proposal for lazy imports in Python: [PEP 690](https://peps.python.org/pep-0690/). Thankfully opt-in, but nevertheless will be quite challenging as it is to make pylint and astroid compatible.

DanielNoord:
Thanks @deepyaman for the initial PR and thanks to @cdce8p for the reviews!

Happy to finally see this merged!

PR #1443, opened 2022-03-04
Introduce end-of-line normalization

Pierre-Sassoulas:
## Description

This introduce end of line normalization. Some file were formatted for Windows.

## Type of Changes

## Related Issue

Required for #1441


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
DanielNoord, line 1: Shouldn't this have an empty new line? And perhaps a comment to explain what this does. Neat little function by the way.

Pierre-Sassoulas: Regarding the end of line there's one, we have a hook to check that there is exactly one :)

DanielNoord: Huh that's weird. I thought Github would display those in "file-view" or in "raw-view" but I didn't see one so I assumed it was missing.

Thanks for the comment!

tests/testdata/python3/data/recursion.py
```diff
@@ -1,3 +1,3 @@
-""" For issue #25 """
-class Base(object):
+""" For issue #25 """
+class Base(object):
     pass
```
DanielNoord, line 3: There is a missing line missing here though @Pierre-Sassoulas 😄 

But no reason to add it now!

DanielNoord (approved):

PR #1447, opened 2022-03-06
Replace sys.version_info checks with getattr [rebuilder]

cdce8p:
## Description

Simplify maintenance and reduce code duplication in `astroid.rebuilder` by using `getattr` instead of `sys.version_info` to guard against missing `end_lineno` and `end_col_offset` attributes.

This change is a tradeoff between linter support and reduced code duplication. I didn't saw a noticeable performance difference during testing. In the end, I think the simplified maintenance is slightly preferable here.

Once support for Python 3.8 / 3.9 is removed, the `getattr` calls can / should be replaced with normal attribute.

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -1030,16 +1030,14 @@ class TreeRebuilder:
 
     def visit_assert(self, node: "ast.Assert", parent: NodeNG) -> nodes.Assert:
         """visit a Assert node by returning a fresh instance of it"""
-        if sys.version_info >= (3, 8):
-            newnode = nodes.Assert(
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Make sense, maybe something to add in a todo list in a "drop python 3.7" issue ?

DanielNoord (approved):

cdce8p:
> Make sense, maybe something to add in a todo list in a "drop python 3.7" issue ?

The `TreeRebuilder` has so many Python version dependent things, that it's always good advice to look over it from time to time. But those changes don't need to happen together with a release, they can follow afterwards.

PR #1450, opened 2022-03-07
Add missing `shape` parameter to numpy methods

mbyrnepr2:
Add the missing `shape` parameter to the brain for:
- [numpy.zeros_like](https://numpy.org/doc/stable/reference/generated/numpy.zeros_like.html#numpy-zeros-like)
- [numpy.full_like](https://numpy.org/doc/stable/reference/generated/numpy.full_like.html?highlight=full_like#numpy-full-like)
- [numpy.ones_like](https://numpy.org/doc/stable/reference/generated/numpy.ones_like.html#numpy-ones-like)

## Steps

## Type of Changes

## Related Issue

Closes with relevant test PyCQA/pylint#5871

### DIFF
--- a/tests/unittest_brain_numpy_core_numeric.py
+++ b/tests/unittest_brain_numpy_core_numeric.py
@@ -83,7 +83,7 @@ def test_function_parameters(method: str, expected_args: List[str]) -> None:
     numpy.{method} #@
     """
     )
-    actual_args = list(instance.infer())[0].args.args
+    actual_args = instance.inferred()[0].args.args
     assert [arg.name for arg in actual_args] == expected_args
 
 
--- a/tests/unittest_brain_numpy_core_numeric.py
+++ b/tests/unittest_brain_numpy_core_numeric.py
@@ -7,6 +7,7 @@
 
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+import pytest
 import unittest
 
 try:
@@ -60,6 +61,27 @@ class BrainNumpyCoreNumericTest(unittest.TestCase):
                         func_[0], inferred_values[-1].pytype()
                     ),
                 )
+
+
+@pytest.mark.skipif(not HAS_NUMPY, reason="This test requires the numpy library.")
+@pytest.mark.parametrize(
+    "method, expected_args",
+    [
+        ("zeros_like", ["a", "dtype", "order", "subok", "shape"]),
+        ("full_like", ["a", "fill_value", "dtype", "order", "subok", "shape"]),
+        ("ones_like", ["a", "dtype", "order", "subok", "shape"]),
+        ("ones", ["shape", "dtype", "order"]),
+    ],
+)
+def test_parameters(method, expected_args) -> None:
+    instance = builder.extract_node(
+        f"""
+    import numpy
+    numpy.{method} #@
+    """
+    )
+    actual_args = list(instance.infer())[0].args.args
+    assert [arg.name for arg in actual_args] == expected_args
 
 
 if __name__ == "__main__":
--- a/tests/unittest_brain_numpy_core_numeric.py
+++ b/tests/unittest_brain_numpy_core_numeric.py
@@ -4,11 +4,12 @@
 # Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
 # Copyright (c) 2021 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
-
-import unittest
-
 # Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+
+import unittest
+from typing import List
+
 import pytest
 
 try:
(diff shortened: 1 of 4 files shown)


DanielNoord:
@mbyrnepr2 👍 
Could you add a test for this?

mbyrnepr2:
@DanielNoord sure thing. I am less familiar with astroid code so I will take a look at the test structure for this hopefully later today. If you can give me any pointers I would appreciate.

DanielNoord:
> @DanielNoord sure thing. I am less familiar with astroid code so I will take a look at the test structure for this hopefully later today. If you can give me any pointers I would appreciate. 

I'm on mobile so I can't look up code easily right now.
Searching for `numpy` or `np` should probably show the other tests for this brain. Often we simply test that the thing we're inferring is correct.
Note: some tests use `next(node.inferred())`. We're trying to avoid this as this doesn't show whether there is ambiguity in the inference (which we often don't want). So it's often better to cast `inferred()` to a list and check that the length of that list is 1.

mbyrnepr2:
> > @DanielNoord sure thing. I am less familiar with astroid code so I will take a look at the test structure for this hopefully later today. If you can give me any pointers I would appreciate.
> 
> I'm on mobile so I can't look up code easily right now. Searching for `numpy` or `np` should probably show the other tests for this brain. Often we simply test that the thing we're inferring is correct. Note: some tests use `next(node.inferred())`. We're trying to avoid this as this doesn't show whether there is ambiguity in the inference (which we often don't want). So it's often better to cast `inferred()` to a list and check that the length of that list is 1.

Great that's good to know, will try it out later @DanielNoord

Pierre-Sassoulas (commented):
Thank you for your First astroid contribution @mbyrnepr2 ;)

cdce8p:
> I am less familiar with astroid code so I will take a look at the test structure for this hopefully later today. If you can give me any pointers I would appreciate.

Adding another item here should work.
https://github.com/PyCQA/astroid/blob/29dcc35ac14ac21fe97502ac890a1faac936054e/tests/unittest_brain_numpy_core_numeric.py#L28-L29

mbyrnepr2:
> > I am less familiar with astroid code so I will take a look at the test structure for this hopefully later today. If you can give me any pointers I would appreciate.
> 
> Adding another item here should work.
> 
> https://github.com/PyCQA/astroid/blob/29dcc35ac14ac21fe97502ac890a1faac936054e/tests/unittest_brain_numpy_core_numeric.py#L28-L29

Thanks @cdce8p! I was looking at this one earlier. It seems that the tests will pass with or without adding the parameter to this line; but that seems to be the structure for much of the numpy tests that I've seen.
So, for example, I have it set to `("zeros_like", "[1, 2]", "None", "K", "True", "None"),` in my env. but the results are the same if I leave these optional parameters out also.

mbyrnepr2:
> Thank you for your First astroid contribution @mbyrnepr2 ;)

Thanks for the warm welcome, nice to be here!

tests/unittest_brain_numpy_core_numeric.py
```diff
@@ -62,5 +64,26 @@ def test_numpy_function_calls_inferred_as_ndarray(self):
...
+@pytest.mark.parametrize(
+    "method, expected_args",
+    [
+        ("zeros_like", ["a", "dtype", "order", "subok", "shape"]),
+        ("full_like", ["a", "fill_value", "dtype", "order", "subok", "shape"]),
+        ("ones_like", ["a", "dtype", "order", "subok", "shape"]),
+        ("ones", ["shape", "dtype", "order"]),
+    ],
+)
+def test_parameters(method, expected_args) -> None:
```
DanielNoord, line 77:
```suggestion
def test_function_parameters(method: str, expected_args: List[str]) -> None:
```

DanielNoord (changes requested):
Change itself looks good to me!

tests/unittest_brain_numpy_core_numeric.py
mbyrnepr2: Why do I always forget typehints :D
Thanks for the review @DanielNoord

tests/unittest_brain_numpy_core_numeric.py
```diff
@@ -4,10 +4,13 @@
 # Copyright (c) 2021 Pierre Sassoulas <pierre.sassoulas@gmail.com>
 # Copyright (c) 2021 Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
```
DanielNoord, line 6:
```suggestion
# Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>

```

DanielNoord (approved):

cdce8p (approved):
Added some last changes in https://github.com/PyCQA/astroid/pull/1450/commits/aeb1bf8907fb50eff5c38987f65b7d2c5558184a. Looks good now!

Although not strictly necessary as astroid can't infer numpy natively anyway, I added the `>1.17.0` requirement. According to the docs, that's where `shape` was added.
https://numpy.org/doc/stable/reference/generated/numpy.zeros_like.html

mbyrnepr2:
> Added some last changes in [aeb1bf8](https://github.com/PyCQA/astroid/commit/aeb1bf8907fb50eff5c38987f65b7d2c5558184a). Looks good now!
> 
> Although not strictly necessary as astroid can't infer numpy natively anyway, I added the `>1.17.0` requirement. According to the docs, that's where `shape` was added. https://numpy.org/doc/stable/reference/generated/numpy.zeros_like.html

Thank you for this @cdce8p!

cdce8p:
And it's merged. Congrats on your first astroid contribution @mbyrnepr2 🚀

PR #1451, opened 2022-03-07
No longer pass ``doc`` to various node constructors

DanielNoord:
## Steps

## Description

Missed some stuff.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -280,9 +280,7 @@ class AstroidBuilder(raw_building.InspectBuilder):
 
 
 def build_namespace_package_module(name: str, path: List[str]) -> nodes.Module:
-    node = nodes.Module(name, path=path, package=True)
-    node.postinit(body=[], doc_node=nodes.Const(value=""))
-    return node
+    return nodes.Module(name, path=path, package=True)
 
 
 def parse(code, module_name="", path=None, apply_transforms=True):
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -280,7 +280,9 @@ class AstroidBuilder(raw_building.InspectBuilder):
 
 
 def build_namespace_package_module(name: str, path: List[str]) -> nodes.Module:
-    return nodes.Module(name, doc="", path=path, package=True)
+    node = nodes.Module(name, path=path, package=True)
+    node.postinit(body=[], doc_node=nodes.Const(value=""))
+    return node
 
 
 def parse(code, module_name="", path=None, apply_transforms=True):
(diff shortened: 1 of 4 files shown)


astroid/builder.py
```diff
@@ -280,7 +280,9 @@ def delayed_assattr(self, node):
 
 
 def build_namespace_package_module(name: str, path: List[str]) -> nodes.Module:
-    return nodes.Module(name, doc="", path=path, package=True)
+    node = nodes.Module(name, path=path, package=True)
+    node.postinit(body=[], doc_node=nodes.Const(value=""))
```
DanielNoord, line 284: Or do we want to make `doc_node` `None` here?

cdce8p: I think it would be safe to leave it as `None`. Code already needs to handle it for Modules created by the TreeRebuilder, so this wouldn't introduce anything new.

Since we no longer need the `postinit` call then, I suggest to return the Module node directly instead.

astroid/rebuilder.py
```diff
@@ -278,10 +278,9 @@ def visit_module(
 
         Note: Method not called by 'visit'
         """
-        node, doc_ast_node, doc = self._get_doc(node)
+        node, doc_ast_node, _ = self._get_doc(node)
```
cdce8p, line 281: Anything against updating ` self._get_doc`, too? It's an internal method, so no need to keep backwards compatibility there.

cdce8p (changes requested):

astroid/raw_building.py
```diff
@@ -110,7 +110,11 @@ def attach_import_node(node, modname, membername):
...
 def build_module(name: str, doc: Optional[str] = None) -> nodes.Module:
     """create and initialize an astroid Module node"""
-    node = nodes.Module(name, doc, pure_python=False)
+    node = nodes.Module(name, pure_python=False)
+    node.postinit(
+        body=[],
+        doc_node=nodes.Const(value=doc) if doc else None,
+    )
     node.package = False
     node.parent = None
```
cdce8p, line 119: `package` can be assigned with `__init__`. For `parent`, `None` is the default.

astroid/rebuilder.py
```diff
@@ -116,9 +116,7 @@ def __init__(
             self._parser_module = parser_module
         self._module = self._parser_module.module
 
-    def _get_doc(
-        self, node: T_Doc
-    ) -> Tuple[T_Doc, Optional["ast.Constant | ast.Str"], Optional[str]]:
+    def _get_doc(self, node: T_Doc) -> Tuple[T_Doc, Optional["ast.Constant | ast.Str"]]:
         """Return the doc ast node and the actual docstring."""
```
DanielNoord, line 120:
```suggestion
        """Return the doc ast node."""
```

cdce8p (approved):

PR #1454, opened 2022-03-08
Fix end_lineno on PyPy 3.8

cdce8p:
## Description
PyPy 3.8 assigns `end_lineno` and `end_col_offset` inconsistently. Some nodes set it correctly, other only set `-1`, and some only set a value if certain conditions are met (`ast.If` is a really interesting case here). If `end_lineno` and `end_col_offset` are set, astroid depends on them for other methods and properties (e.g. the block_range, tolineno, fromlineno). A situation where they are assigned inconsistently just leads the unexpected errors. Fortunately for us PyPy 3.9 seems the have addressed these issues.

The best way forward IMO is to reset the `end_lineno` and `end_col_offset` attributes for **`all`** nodes. That will make the whole situation predictable again. Basically, we only start using these attributes in Python 3.9 on PyPy.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -336,7 +336,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        python-version: ["pypy-3.6", "pypy-3.7"]
+        python-version: ["pypy-3.6", "pypy-3.7", "pypy-3.8"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
@@ -381,7 +381,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["pypy-3.6", "pypy-3.7"]
+        python-version: ["pypy-3.6", "pypy-3.7", "pypy-3.8"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.0


Pierre-Sassoulas (approved):
LGTM. 

Nitpick: I think we could (re)name ``IMPLEMENTATION_PYPY`` to ``INTERPRETER_IS_PYPY`` or ``IS_PYPY``

cdce8p:
> Nitpick: I think we could (re)name `IMPLEMENTATION_PYPY` to `INTERPRETER_IS_PYPY` or `IS_PYPY`

Changed it to just `PYPY`.

Pierre-Sassoulas (approved):

ChangeLog
```diff
@@ -18,6 +18,11 @@ Release date: TBA
 
   Closes #1410
 
+* Set ``end_lineno`` and ``end_col_offset`` attributes to ``None`` for all nodes
+  with PyPy 3.8. PyPy assigned these attributes inconsistently which could lead
+  to unexpected errors. Overwriting them with ``None`` will cause a fallback
+  to the already support way of PyPy 3.7.
```
DanielNoord, line 24:
```suggestion
* Set ``end_lineno`` and ``end_col_offset`` attributes to ``None`` for all nodes
  with PyPy 3.8. PyPy 3.8 assigns these attributes inconsistently which could lead
  to unexpected errors. Overwriting them with ``None`` will cause a fallback
  to the already supported way of PyPy 3.7.
```

astroid/const.py
```diff
@@ -12,7 +12,7 @@
 
 WIN32 = sys.platform == "win32"
 
-IMPLEMENTATION_PYPY = platform.python_implementation() == "PyPy"
+PYPY = platform.python_implementation() == "PyPy"
```
DanielNoord, line 15: Just a thought, can this be done with `sys` as well? I checked the import time of `platform` recently and it's quite a lot for what we are using it for. Are there any possibilities with already imported packages?

tests/unittest_builder.py
```diff
@@ -78,9 +79,12 @@ def test_callfunc_lineno(self) -> None:
         self.assertEqual(name.tolineno, 4)
         strarg = callfunc.args[0]
         self.assertIsInstance(strarg, nodes.Const)
-        if hasattr(sys, "pypy_version_info"):
+        if PYPY:
             self.assertEqual(strarg.fromlineno, 4)
-            self.assertEqual(strarg.tolineno, 4)
+            if not PY39_PLUS:
+                self.assertEqual(strarg.tolineno, 4)
+            else:
```
DanielNoord, line 86: The `else` is not "covered" right now, right?

tests/unittest_builder.py
```diff
@@ -38,11 +38,12 @@
 import tempfile
 import textwrap
 import unittest
+import unittest.mock
```
DanielNoord, line 41: Probably missing something, but is this being used?

tests/unittest_builder.py
```diff
@@ -183,7 +187,7 @@ class C:
 
         c = ast_module.body[2]
         assert isinstance(c, nodes.ClassDef)
-        if not PY38_PLUS:
+        if not PY38_PLUS or PY38 and PYPY:
             # Not perfect, but best we can do for Python 3.7
```
DanielNoord, line 191:
```suggestion
            # Not perfect, but best we can do for Python 3.7 and PyPy 3.8
```

DanielNoord (changes requested):

astroid/const.py
cdce8p: We should be able to use `sys.implementation.name` instead. Let me do this in a followup. I'll revert the renaming for now to keep the diff smaller. https://docs.python.org/3/library/sys.html#sys.implementation

tests/unittest_builder.py
cdce8p: Correct. I tested PyPy 3.9 locally and this seems to be the only necessary change regarding the lineno and col_offset attributes. PyPy 3.9 is basically equivalent with CPython there.

_Unfortunately there is another issue that prevents us from adding PyPy 3.9 here, too._

tests/unittest_builder.py
cdce8p: It depends, running the whole test suite works. However `pytest tests/unittest_builder.py` currently fails. Adding that line fixes it. _Another indicator is that VS Code intellisense only works with it._
```
AttributeError: module 'unittest' has no attribute 'mock'
```
https://github.com/PyCQA/astroid/blob/29dcc35ac14ac21fe97502ac890a1faac936054e/tests/unittest_builder.py#L1020

DanielNoord: Ah, unrelated to this PR then. I was wondering why I didn't see `mock` anywhere in the diff.
Fine with me to keep this in!

DanielNoord (approved):

DanielNoord:
@cdce8p Do you want to reorder the commits to separate the CI job commit, or just squash merge?

PR #1453, opened 2022-03-08
Deprecate passing the ``doc`` argument

DanielNoord:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description

Based on @cdce8p's work in https://github.com/PyCQA/astroid/issues/1414.

I have not done anything to the tests other than testing if the decorator works. I'll leave changing the tests to a later PR.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |
| ✓   | :hammer: Refactoring   |

PR #1455, opened 2022-03-09
Do no pass ``doc`` attribute to ``nodes.Module`` in tests

DanielNoord:
## Steps

## Description

Follow up to #1453. Can be merged already.

## Type of Changes

## Related Issue

### DIFF
--- a/tests/unittest_scoped_nodes.py
+++ b/tests/unittest_scoped_nodes.py
@@ -215,8 +215,7 @@ class ModuleNodeTest(ModuleLoader, unittest.TestCase):
 
     def test_relative_to_absolute_name(self) -> None:
         # package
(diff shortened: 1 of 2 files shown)


tests/unittest_scoped_nodes.py
```diff
@@ -215,7 +215,7 @@ def test_module_getattr(self) -> None:
 
     def test_relative_to_absolute_name(self) -> None:
         # package
-        mod = nodes.Module("very.multi.package", "doc")
+        mod = nodes.Module("very.multi.package")
         mod.package = True
```
cdce8p, line 219: Can you assign `package` in the `Module` call? Same for the other two below.

cdce8p (changes requested):

PR #1458, opened 2022-03-09
Do no pass ``doc`` attribute to ``nodes.ClassDef`` in tests

DanielNoord:
## Steps

## Description

Follow up to #1453. Can be merged already.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/brain/brain_argparse.py
+++ b/astroid/brain/brain_argparse.py
@@ -10,12 +10,6 @@ def infer_namespace(node, context=None):
         raise UseInferenceDefault()
 
     class_node = nodes.ClassDef("Namespace", parent=node.parent)
-    class_node.postinit(
-        bases=[],
-        body=[],
-        decorators=None,
-        doc_node=nodes.Const(value="docstring"),
-    )
     for attr in set(callsite.keyword_arguments):
         fake_node = nodes.EmptyNode()
         fake_node.parent = class_node
--- a/astroid/brain/brain_argparse.py
+++ b/astroid/brain/brain_argparse.py
@@ -9,8 +9,13 @@ def infer_namespace(node, context=None):
         # Cannot make sense of it.
         raise UseInferenceDefault()
 
-    class_node = nodes.ClassDef("Namespace", "docstring")
-    class_node.parent = node.parent
+    class_node = nodes.ClassDef("Namespace", parent=node.parent)
+    class_node.postinit(
+        bases=[],
+        body=[],
+        decorators=None,
+        doc_node=nodes.Const(value="docstring"),
+    )
     for attr in set(callsite.keyword_arguments):
         fake_node = nodes.EmptyNode()
         fake_node.parent = class_node
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -157,10 +157,14 @@ def infer_func_form(
     # we know it is a namedtuple anyway.
     name = name or "Uninferable"
     # we want to return a Class node instance with proper attributes set
-    class_node = nodes.ClassDef(name, "docstring")
-    class_node.parent = node.parent
-    # set base class=tuple
-    class_node.bases.append(base_type)
+    class_node = nodes.ClassDef(name, parent=node.parent)
+    class_node.postinit(
+        # set base class=tuple
+        bases=[base_type],
+        body=[],
+        decorators=None,
+        doc_node=nodes.Const(value="docstring"),
+    )
     # XXX add __init__(*attributes) method
     for attr in attributes:
         fake_node = nodes.EmptyNode()
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -163,7 +163,6 @@ def infer_func_form(
         bases=[base_type],
         body=[],
         decorators=None,
-        doc_node=nodes.Const(value="docstring"),
     )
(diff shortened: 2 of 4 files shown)


astroid/raw_building.py
```diff
@@ -467,18 +467,28 @@ def _astroid_bootstrapping():
     # Set the builtin module as parent for some builtins.
     nodes.Const._proxied = property(_set_proxied)
 
-    _GeneratorType = nodes.ClassDef(
-        types.GeneratorType.__name__, types.GeneratorType.__doc__
-    )
+    _GeneratorType = nodes.ClassDef(types.GeneratorType.__name__)
     _GeneratorType.parent = astroid_builtin
```
DanielNoord, line 471: Passing `parent` on L478 breaks a test as the `init` of `ClassDef` does something with `set_local` if `parent is not None`. That's why this is a separate line here.

astroid/raw_building.py
```diff
@@ -467,18 +467,28 @@ def _astroid_bootstrapping():
...
+        )
     bases.Generator._proxied = _GeneratorType
     builder.object_build(bases.Generator._proxied, types.GeneratorType)
 
     if hasattr(types, "AsyncGeneratorType"):
-        _AsyncGeneratorType = nodes.ClassDef(
-            types.AsyncGeneratorType.__name__, types.AsyncGeneratorType.__doc__
-        )
+        _AsyncGeneratorType = nodes.ClassDef(types.AsyncGeneratorType.__name__)
         _AsyncGeneratorType.parent = astroid_builtin
```
DanielNoord, line 484: This does not break a test, but because it does above I kept this the same.

astroid/raw_building.py
```diff
@@ -495,10 +505,17 @@ def _astroid_bootstrapping():
     )
     for _type in builtin_types:
         if _type.__name__ not in astroid_builtin:
-            cls = nodes.ClassDef(_type.__name__, _type.__doc__)
-            cls.parent = astroid_builtin
-            builder.object_build(cls, _type)
-            astroid_builtin[_type.__name__] = cls
+            klass = nodes.ClassDef(_type.__name__)
+            klass.parent = astroid_builtin
```
DanielNoord, line 509: This does not break a test, but because it does above I kept this the same (although I changed the name).

astroid/brain/brain_argparse.py
```diff
@@ -9,8 +9,13 @@ def infer_namespace(node, context=None):
...
 
-    class_node = nodes.ClassDef("Namespace", "docstring")
-    class_node.parent = node.parent
+    class_node = nodes.ClassDef("Namespace", parent=node.parent)
+    class_node.postinit(
+        bases=[],
+        body=[],
+        decorators=None,
+        doc_node=nodes.Const(value="docstring"),
+    )
```
cdce8p, line 18: Is this necessary? From what I can tell, it can be removed. The tests also pass without it.

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -157,10 +157,14 @@ def infer_func_form(
...
-    class_node.parent = node.parent
-    # set base class=tuple
-    class_node.bases.append(base_type)
+    class_node = nodes.ClassDef(name, parent=node.parent)
+    class_node.postinit(
+        # set base class=tuple
+        bases=[base_type],
+        body=[],
+        decorators=None,
+        doc_node=nodes.Const(value="docstring"),
```
cdce8p, line 166: Same here.

astroid/raw_building.py
```diff
@@ -495,10 +505,17 @@ def _astroid_bootstrapping():
...
+            if _type.__doc__:
+                klass.postinit(
```
cdce8p, line 511: This feels wrong. It would be better to always call `postinit` and assign a conditional and / or precomputed `doc_node` attribute.

cdce8p (changes requested):

astroid/raw_building.py
```diff
@@ -467,18 +467,28 @@ def _astroid_bootstrapping():
     # Set the builtin module as parent for some builtins.
     nodes.Const._proxied = property(_set_proxied)
 
-    _GeneratorType = nodes.ClassDef(
-        types.GeneratorType.__name__, types.GeneratorType.__doc__
-    )
+    _GeneratorType = nodes.ClassDef(types.GeneratorType.__name__)
     _GeneratorType.parent = astroid_builtin
+    if types.GeneratorType.__doc__:
```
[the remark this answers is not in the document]
DanielNoord: This is more future-proof. If `types` maintainers add a docstring at some point we would miss them.

astroid/raw_building.py
DanielNoord: Fine with me!

astroid/brain/brain_namedtuple_enum.py
DanielNoord: This sets the base class, so I guess this is doing more than only set the docstring?

cdce8p: Only remove the `doc_node` attribute here

astroid/raw_building.py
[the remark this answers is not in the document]
cdce8p: Just wondering, why would they set `__doc__` or better what should it be used for?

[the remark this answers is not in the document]
cdce8p: If we do keep it, the pattern should change similar to the other comment.

cdce8p (approved):

PR #1460, opened 2022-03-09
Move ``builtin_lookup`` out of main ``scoped_nodes`` file

DanielNoord:
## Steps

## Description

I was looking into some of the code in `scoped_nodes` for a potential refactor to remove some circular dependencies. 
Noticed that this function could be optimised by not building the `Module` node every time, but only doing this once.

Tested with `pylint` over `pylint`. Went from 80th to 145th in terms of time spent in function and child-calls.

_This prepares for the move of all mixins from `scoped_nodes` for which a multiple years old TODO is still included._

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/scoped_nodes/utils.py
+++ b/astroid/nodes/scoped_nodes/utils.py
@@ -6,7 +6,7 @@ This module contains utility functions for scoped nodes.
 """
 
 import builtins
-from typing import TYPE_CHECKING, List, Optional, Tuple
+from typing import TYPE_CHECKING, List, Tuple
 
 from astroid.manager import AstroidManager
 
@@ -14,7 +14,7 @@ if TYPE_CHECKING:
     from astroid import nodes
 
 
-BUILTINS_AST: Optional["nodes.Module"] = None
+_builtin_astroid: "nodes.Module | None" = None
 
 
 def builtin_lookup(name: str) -> Tuple["nodes.Module", List["nodes.NodeNG"]]:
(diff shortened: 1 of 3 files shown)


astroid/nodes/scoped_nodes/utils.py
```diff
@@ -0,0 +1,35 @@
...
+import builtins
+from typing import TYPE_CHECKING, List, Tuple
+
+from astroid.manager import AstroidManager
+
+if TYPE_CHECKING:
+    from astroid import nodes
+
+
+_builtin_astroid: "nodes.Module | None" = None
```
cdce8p, line 17: I've renamed it to lowercase, only constants should be uppercase, and added an underscore since it's only used in this module.

DanielNoord: Thanks! That was quite obvious.

astroid/nodes/scoped_nodes/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    Return the list of matching statements and the ast for the builtin module
+    """
+    # pylint: disable-next=global-statement
+    global _builtin_astroid
+    if _builtin_astroid is None:
+        _builtin_astroid = AstroidManager().ast_from_module(builtins)
+    if name == "__dict__":
+        return _builtin_astroid, ()
+    try:
+        stmts = _builtin_astroid.locals[name]
```
cdce8p, line 32:
```suggestion
        stmts: Sequence["nodes.NodeNG"] = _builtin_astroid.locals[name]
```
Necessary for mypy. Otherwise `stmts` would be inferred as `List` and `stmts = ()` would be an error.

cdce8p (changes requested):

Pierre-Sassoulas (commented):
Removing a circular import is nice...

> Tested with pylint over pylint. Went from 80th to 145th in terms of time spent in function and child-calls.

... but this is s very nice 😄

astroid/nodes/scoped_nodes/utils.py
```diff
@@ -0,0 +1,35 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+
+"""
+This module contains utility functions for scoped nodes.
+"""
+
+import builtins
+from typing import TYPE_CHECKING, Tuple
```
DanielNoord, line 9:
```suggestion
from typing import TYPE_CHECKING, Sequence, Tuple
```

Pierre-Sassoulas (approved):

PR #1463, opened 2022-03-09
Move Mixin classes out of main ``scoped_nodes`` file

DanielNoord:
## Steps

## Description

Original TODO was added in: https://github.com/PyCQA/astroid/commit/bec4ee12322d3f10df273e0764290e41cc6c701b.

We could add these `mixins` to the `mixin.py` file, but I think it is better to make `scoped_modules` a self contained module (as much as possible). In the end, self-contained modules make the risk of circular dependencies much less likely.

## Type of Changes

## Related Issue


Pierre-Sassoulas (approved):
LGTM, nice cleanup of a 12 year old todo :tada: !

cdce8p (approved):
Looks good!

PR #1464, opened 2022-03-09
Minor changes

cdce8p:
## Description
Spelling + unused type ignore

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -261,7 +261,7 @@ class NodeNG:
         """An optimized version of list(get_children())[-1]"""
         for field in self._astroid_fields[::-1]:
             attr = getattr(self, field)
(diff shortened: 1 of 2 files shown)


DanielNoord:
Do we want to turn on `warn_unused_ignores`?

DanielNoord (approved):

cdce8p:
> Do we want to turn on `warn_unused_ignores`?

Someday 😅
These were just some things I had gathered over the last weeks.

cdce8p:
> Do we want to turn on `warn_unused_ignores`?

There is also the issue of `python_version` which should really be added add some point. Just to make runs with different versions reproducible. There just isn't any point in doing so before we've fixed all existing errors.

_Something for another time. I need to take a step back from astroid / pylint soon to focus on other things, so please don't start with that now. Will have more time in a month or two._

PR #1466, opened 2022-03-11
Upgrade mypy to 0.940

DanielNoord:
## Steps

## Type of Changes

## Related Issue

Can't merge yet as the ``pre-commit`` mypy mirror hasn't been updated yet, see: https://github.com/pre-commit/mirrors-mypy

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,14 +63,14 @@ repos:
           ]
         exclude: tests/testdata|conf.py
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.931
+    rev: v0.940
     hooks:
       - id: mypy
         name: mypy
         entry: mypy
         language: python
         types: [python]
-        args: []
+        args: [--enable-error-code ignore-without-code]
         require_serial: true
         additional_dependencies:
           [
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,7 +70,7 @@ repos:
         entry: mypy
         language: python
         types: [python]
-        args: [--enable-error-code ignore-without-code]
+        args: []
         require_serial: true
         additional_dependencies:
           [
--- a/tests/unittest_nodes.py
+++ b/tests/unittest_nodes.py
@@ -1222,7 +1222,7 @@ def test_type_comments_with() -> None:
         """
     with a as b: # type: int
         pass
-    with a as b: # type: ignore[misc]
+    with a as b: # type: ignore[name-defined]
         pass
     """
     )
@@ -1239,7 +1239,7 @@ def test_type_comments_for() -> None:
         """
     for a, b in [1, 2, 3]: # type: List[int]
         pass
-    for a, b in [1, 2, 3]: # type: ignore[misc]
+    for a, b in [1, 2, 3]: # type: ignore[name-defined]
         pass
     """
     )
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -63,14 +63,14 @@ repos:
...
-    rev: v0.931
+    rev: v0.940
     hooks:
       - id: mypy
         name: mypy
         entry: mypy
         language: python
         types: [python]
-        args: []
+        args: [--enable-error-code ignore-without-code]
```
cdce8p, line 73: You can add it directly to `setup.cfg`. In the `mypy` section add
```
enable_error_code = ignore-without-code
```

cdce8p (changes requested):

cdce8p (commented):
Could you update existing `type: ignore` without error codes?
From what I've seen there are only a few in `unittest_nodes` and `unittest_nodes_lineno`.

tests/unittest_nodes.py
```diff
@@ -1222,7 +1222,7 @@ def test_type_comments_with() -> None:
         """
     with a as b: # type: int
         pass
-    with a as b: # type: ignore
+    with a as b: # type: ignore[misc]
```
DanielNoord, line 1225: Not sure what to put here. `mypy` complains about them...

DanielNoord: Never mind.

tests/unittest_nodes_lineno.py
```diff
@@ -1153,7 +1153,7 @@ def test_end_lineno_comprehension() -> None:
         c1 = ast_nodes[0]
         assert isinstance(c1, nodes.ListComp)
         assert isinstance(c1.elt, nodes.Name)
-        assert isinstance(c1.generators[0], nodes.Comprehension)  # type: ignore
+        assert isinstance(c1.generators[0], nodes.Comprehension) # type: ignore[index]
```
cdce8p, line 1156:
```suggestion
        assert isinstance(c1.generators[0], nodes.Comprehension)  # type: ignore[index]
```

tests/unittest_nodes.py
```diff
@@ -1222,7 +1222,7 @@ def test_type_comments_with() -> None:
         """
     with a as b: # type: int
         pass
-    with a as b: # type: ignore
+    with a as b: # type: ignore[name-defined]
```
cdce8p, line 1225: Didn't saw that those these here are inside code blocks 😅
Doesn't matter now I guess.

cdce8p (approved):

tests/unittest_nodes.py
DanielNoord: It's interesting that `mypy` picks up on those. Do you know if code in code blocks is checked? If not, this option also probably shouldn't?

cdce8p: > It's interesting that `mypy` picks up on those. Do you know if code in code blocks is checked? If not, this option also probably shouldn't?

Mypy didn't notice those. I just searched for `type: ignore` and looked which ones didn't have error codes 😄

DanielNoord: If I'm not mistaken they did. I think I saw an error message about not having the error specified for these.

cdce8p: Hmm 🤔 I just removed the comment and ran mypy. Didn't saw it.
Under the hood mypy uses ast parsing  and this would be just one string node. So hard to imagine that it would mark it.

DanielNoord: Probably my memory messing up 😅

PR #1470, opened 2022-03-12
Add conditional overloads [TreeRebuilder]

cdce8p:
## Description
Mypy 0.940 finally includes conditional overload which allow us to clean up the `TreeRebuilder` overloads.

I've waited since last June for this when I first opened the mypy PR. https://github.com/python/mypy/pull/10712
https://mypy.readthedocs.io/en/stable/more_types.html#conditional-overloads

/CC: @Pierre-Sassoulas @DanielNoord

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -377,11 +377,6 @@ class TreeRebuilder:
     def visit(self, node: "ast.Expr", parent: NodeNG) -> nodes.Expr:
         ...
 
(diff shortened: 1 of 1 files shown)


DanielNoord (approved):

Pierre-Sassoulas (approved):
That's a very nice cleanup ! I'm going to trust you blindly regarding the minute detail though 😄 

Should we rebase or squash ?

PR #1471, opened 2022-03-12
Fix typing in ``context.py``

DanielNoord:
## Steps

## Description

Had this laying around for quite some time. Mostly moving around some stuff and making sure `keywords` is not reassigned to another type.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -163,10 +163,10 @@ class CallContext:
     ):
         self.args = args  # Call positional arguments
         if keywords:
-            keywords_list = [(arg.arg, arg.value) for arg in keywords]
+            keywords_tuples = [(arg.arg, arg.value) for arg in keywords]
         else:
-            keywords_list = []
-        self.keywords = keywords_list  # Call keyword arguments
+            keywords_tuples = []
+        self.keywords = keywords_tuples  # Call keyword arguments
         self.callee = callee  # Function being called
 
 
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -163,10 +163,10 @@ class CallContext:
     ):
         self.args = args  # Call positional arguments
         if keywords:
-            keywords_tuples = [(arg.arg, arg.value) for arg in keywords]
+            arg_value_pairs = [(arg.arg, arg.value) for arg in keywords]
         else:
-            keywords_tuples = []
-        self.keywords = keywords_tuples  # Call keyword arguments
+            arg_value_pairs = []
+        self.keywords = arg_value_pairs  # Call keyword arguments
         self.callee = callee  # Function being called
 
 
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -5,12 +5,12 @@
 """Various context related utilities, including inference and call contexts."""
 import contextlib
 import pprint
-from typing import TYPE_CHECKING, List, MutableMapping, Optional, Sequence, Tuple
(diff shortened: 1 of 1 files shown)


astroid/context.py
```diff
@@ -10,11 +10,14 @@
 if TYPE_CHECKING:
     from astroid.nodes.node_classes import Keyword, NodeNG
 
+_InferenceCache = MutableMapping[
```
cdce8p, line 13: Any idea why it was `MutableMapping` instead of `Dict`?

Do we want to start adding a `Type` suffix to type aliases?

astroid/context.py
```diff
@@ -164,10 +163,10 @@ def __init__(
     ):
         self.args = args  # Call positional arguments
         if keywords:
-            keywords = [(arg.arg, arg.value) for arg in keywords]
+            keywords_list = [(arg.arg, arg.value) for arg in keywords]
```
Pierre-Sassoulas, line 166: I'm not a fan of putting the type in the variable name: it increase the mental load, you can check the type if you need it and the plural form indicate it's an iterable.

DanielNoord: I changed it to `keywords_tuples`. Open for any better suggestions.

astroid/context.py
DanielNoord: PR that added it is https://github.com/PyCQA/astroid/pull/1009.

Perhaps adding `Type` is something to do after we can run `mypy` successfully? I'm still running into issues with types not actually being correct or missing information. I think we can do a "cleanup" `mypy` PR after we can get it to pass.

astroid/context.py
cdce8p: > I changed it to `keywords_tuples`. Open for any better suggestions.

I think that's worse, let's just leave it at `keywords`. It's assigned to `self.keywords` after all.

Pierre-Sassoulas: I meant that ``keywords`` was fine :)

DanielNoord: `Mypy` will complain then. `keywords` is defined as `keywords: Optional[List["Keyword"]] = None,` on L161. That's why I introduced the name change.

DanielNoord:
```python
keywords: List[Tuple[str, str]] = [(arg.arg, arg.value) for arg in keywords]
```

Will raise a `no-redef` warning..

Pierre-Sassoulas: Hmm, maybe ``keywords_as_list`` then ?

cdce8p: > Hmm, maybe `keywords_as_list` then ?

They are a list already. Maybe `keywords_list` wasn't that bad after all.

DanielNoord: Went with `arg_value_pairs` 😅 Is that okay?

astroid/context.py
cdce8p: > PR that added it is #1009.

Let's use `Dict`. Usually the return type should be precise whereas the parameter type should be open. `Sequence["NodeNG"]` is ok though I think. We assign a list converted to a tuple to it.

> Perhaps adding `Type` is something to do after we can run `mypy` successfully? I'm still running into issues with types not actually being correct or missing information. I think we can do a "cleanup" `mypy` PR after we can get it to pass.

👍🏻

cdce8p (approved):

PR #1472, opened 2022-03-12
Promoted ``getattr()`` from FunctionDef to Lambda

jacobtylerwalls:
## Steps

## Type of Changes

## Related Issue

Lambdas can have special attributes, such as `__code__`, or even arbitrary attributes. Attempting to use `getattr` on the `scoped_nodes.Lambda` instance, though, raised `AttributeError`.

Now, it won't: the PR promotes `getattr()` from `FunctionDef` to its parent, `Lambda`.

Discovered by pylint's primer tests during an attempt to remove an `except: AttributeError` from pylint's `TypeChecker`, discussed [here](https://github.com/PyCQA/pylint/pull/5544#discussion_r825305650).

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -39,7 +39,7 @@ from astroid.exceptions import (
 from astroid.interpreter.dunder_lookup import lookup
 from astroid.interpreter.objectmodel import ClassModel, FunctionModel, ModuleModel
 from astroid.manager import AstroidManager
-from astroid.nodes import Arguments, Const, node_classes
+from astroid.nodes import Arguments, Const, NodeNG, node_classes
 from astroid.nodes.scoped_nodes.mixin import ComprehensionScope, LocalsDictNodeNG
 from astroid.nodes.scoped_nodes.utils import builtin_lookup
 from astroid.nodes.utils import Position
@@ -1135,7 +1135,7 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
         :type: list(NodeNG)
         """
 
-        self.instance_attrs: Dict[str, List[nodes.NodeNG]] = {}
+        self.instance_attrs: Dict[str, List[NodeNG]] = {}
 
         super().__init__(
             lineno=lineno,
@@ -1269,7 +1269,7 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
 
     def getattr(
         self, name: str, context: Optional[InferenceContext] = None
-    ) -> List[nodes.NodeNG]:
+    ) -> List[NodeNG]:
         if not name:
             raise AttributeInferenceError(target=self, attribute=name, context=context)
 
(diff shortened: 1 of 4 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1074,6 +1074,11 @@ class Lambda(mixins.FilterStmtsMixin, LocalsDictNodeNG):
     _other_other_fields = ("locals",)
     name = "<lambda>"
     is_lambda = True
+    special_attributes = FunctionModel()
+    """The names of special attributes that this function has.
+
+    :type: objectmodel.FunctionModel
+    """
```
DanielNoord, line 1081:
```suggestion
    special_attributes = FunctionModel()
    """The names of special attributes that this function has."""
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1133,6 +1138,8 @@ def __init__(
         :type: list(NodeNG)
         """
 
+        self.instance_attrs = {}
```
DanielNoord, line 1141:
```suggestion
        self.instance_attrs: Dict[str, List[nodes.NodeNG]] = {}
```

Perhaps we can narrow it down even more, but for now this is enough I think?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1263,6 +1270,19 @@ def frame(self: T, *, future: Literal[None, True] = None) -> T:
         """
         return self
 
+    def getattr(self, name, context=None):
```
DanielNoord, line 1273:
```suggestion
    def getattr(self, name: str, context: Optional[InferenceContext]=None) -> List[nodes.NodeNG]:
```

tests/unittest_inference.py
```diff
@@ -4388,7 +4388,7 @@ def test_lambda(self) -> None:
         """
         )
         inferred = next(node.infer())
-        self.assertEqual(inferred, util.Uninferable)
+        self.assertIsInstance(inferred, nodes.Const)
```
DanielNoord, line 4391: Do we want to check if this is indeed `False` and not `None` or `True`?

DanielNoord (changes requested):
Just adding some typing while we're at it 😄 

Nice that that test suddenly works 👍

DanielNoord (approved):
We could consider this a bug fix and test @Pierre-Sassoulas's new release process 😄  Considering the fact that this is breaking in the `pylint` primer and we're handling the bug, there is an argument to be made to do so. I'll leave that to the others though.

_We would only need to change the position of the changelog entry._

Thanks @jacobtylerwalls 👍

jacobtylerwalls:
Makes sense. But to be clear, it's not urgent -- nothing is blocking pylint -- we were just exploring whether we could remove an except case in pylint while working on PyCAQ/pylint#5544. However, at this point, any changes would be better off in a new PR with pylint regression tests.

Pierre-Sassoulas (changes requested):
LGTM. Let's put this in 2.11.1 changelog :)

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #1475, opened 2022-03-15
Handle inference of ``ComprehensionScope`` better

DanielNoord:
## Steps

## Description

This makes it so that `ComprehensionScopes` are actually inferred. This requires some changes in `pylint` as we rely on the fact that these don't get inferred currently.
I have made a related PR to `pylint` which shows what these changes would be. I think it will be useful to review this simultaneously, as they show how this impacts `astroid's` output.

One problem I couldn't overcome: this will break `pre-commit` for the project that merges this first. The change in behaviour changes various checks and we need to resolve an order in which these get merged.

Let me know what you guys think!
_We might want to add some additional tests._

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/mixin.py
+++ b/astroid/nodes/scoped_nodes/mixin.py
@@ -170,3 +170,6 @@ class ComprehensionScope(LocalsDictNodeNG):
     """Scoping for different types of comprehensions."""
 
     scope_lookup = LocalsDictNodeNG._scope_lookup
+
+    generators: List["nodes.Comprehension"]
+    """The generators that are looped through."""
--- a/astroid/nodes/scoped_nodes/mixin.py
+++ b/astroid/nodes/scoped_nodes/mixin.py
@@ -4,6 +4,7 @@
 
 """This module contains mixin classes for scoped nodes."""
 
+import abc
 from typing import TYPE_CHECKING, Dict, List, TypeVar
 
 from astroid.filter_statements import _filter_stmts
@@ -173,3 +174,7 @@ class ComprehensionScope(LocalsDictNodeNG):
 
     generators: List["nodes.Comprehension"]
     """The generators that are looped through."""
+
+    @abc.abstractmethod
+    def pytype(self) -> str:
+        pass
--- a/astroid/nodes/scoped_nodes/mixin.py
+++ b/astroid/nodes/scoped_nodes/mixin.py
@@ -5,7 +5,7 @@
 """This module contains mixin classes for scoped nodes."""
 
 import abc
-from typing import TYPE_CHECKING, Dict, List, TypeVar
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, TypeVar
 
 from astroid.filter_statements import _filter_stmts
 from astroid.nodes import node_classes, scoped_nodes
@@ -13,6 +13,8 @@ from astroid.nodes.scoped_nodes.utils import builtin_lookup
 
 if TYPE_CHECKING:
     from astroid import nodes
+    from astroid.context import InferenceContext
+
 
 _T = TypeVar("_T")
 
@@ -174,6 +176,15 @@ class ComprehensionScope(LocalsDictNodeNG):
 
     generators: List["nodes.Comprehension"]
     """The generators that are looped through."""
+
+    def qname(self):
+        """Get the 'qualified' name of the node."""
+        return self.pytype()
+
+    def infer(
+        self: _T, context: Optional["InferenceContext"] = None, **kwargs: Any
+    ) -> Iterator[_T]:
+        yield self
 
     @abc.abstractmethod
     def pytype(self) -> str:
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -750,6 +750,9 @@ class GeneratorExp(ComprehensionScope):
 
         yield from self.generators
 
+    def pytype(self) -> str:
+        return "builtins.generator"
+
 
 class DictComp(ComprehensionScope):
     """Class representing an :class:`ast.DictComp` node.
@@ -849,6 +852,9 @@ class DictComp(ComprehensionScope):
 
         yield from self.generators
 
+    def pytype(self) -> str:
+        return "builtins.dict"
+
 
 class SetComp(ComprehensionScope):
     """Class representing an :class:`ast.SetComp` node.
@@ -936,6 +942,9 @@ class SetComp(ComprehensionScope):
 
         yield from self.generators
 
+    def pytype(self) -> str:
+        return "builtins.set"
+
 
 class ListComp(ComprehensionScope):
     """Class representing an :class:`ast.ListComp` node.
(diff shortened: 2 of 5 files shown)


astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -170,3 +173,19 @@ class ComprehensionScope(LocalsDictNodeNG):
     """Scoping for different types of comprehensions."""
 
     scope_lookup = LocalsDictNodeNG._scope_lookup
+
+    generators: List["nodes.Comprehension"]
+    """The generators that are looped through."""
+
+    def qname(self):
```
DanielNoord, line 180: This is another instance of https://github.com/pycqa/astroid/issues/278. We could also change `qname()` on `LocalsDictNodeNG` not to use `name()` if it doesn't exist.

I don't know what is better.

DanielNoord:
```suggestion
    def qname(self) -> str:
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -721,14 +719,13 @@ def __init__(
             parent=parent,
         )
 
-    def postinit(self, elt=None, generators=None):
+    def postinit(self, elt=None, generators: Optional["nodes.Comprehension"] = None):
```
Pierre-Sassoulas, line 722: So it wasn't a List of Comprehension but a Comprehension ? (I.e. old value was wrong ?)

Pierre-Sassoulas:
Forgot to point out that the coverage decreased by more than 0.05% and coverall did not fail. I must have set it up wrong.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: No, this is a mistake.. 😅

Pierre-Sassoulas (commented):
The coverage decreased a lot, is this coverall acting up again ?

DanielNoord:
Hmm I'll have a good look.

I think this might actually fix some `StopIteration` except handlers as we now "can" infer these Scope nodes. See https://coveralls.io/builds/47720462/source?filename=astroid%2Fbrain%2Fbrain_builtin_inference.py#L207.

Part of it is also the merge of `main` which @jacobtylerwalls mentioned in another PR today. That seems to give coveralls issues.

DanielNoord:
@Pierre-Sassoulas Can we merge this and backport this to the upcoming patch release? Would be good to fix the coverage in `pylint` before `2.14`.

Pierre-Sassoulas:
> Can we merge this and backport this to the upcoming patch release?

Sure, but do you mean astroid patch release or pylint's patch release, or both ?

DanielNoord:
> > Can we merge this and backport this to the upcoming patch release?
> 
> Sure, but do you mean astroid patch release or pylint's patch release, or both ?

`astroid` `patch`!

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -170,3 +173,19 @@ class ComprehensionScope(LocalsDictNodeNG):
     """Scoping for different types of comprehensions."""
 
     scope_lookup = LocalsDictNodeNG._scope_lookup
+
+    generators: List["nodes.Comprehension"]
+    """The generators that are looped through."""
+
+    def qname(self) -> str:
+        """Get the 'qualified' name of the node."""
+        return self.pytype()
```
jacobtylerwalls, line 182: @cdce8p do you have thoughts about this? In the past I believe you were a vote against, see https://github.com/PyCQA/astroid/pull/1284#pullrequestreview-823449216.

DanielNoord:
Moving this to `2.12` as we can realise the patch version without this.

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -172,3 +175,16 @@ class ComprehensionScope(LocalsDictNodeNG):
...
     """The generators that are looped through."""
+
+    def qname(self) -> str:
+        """Get the 'qualified' name of the node."""
+        return self.pytype()
+
+    def _infer(
+        self: _T, context: Optional["InferenceContext"] = None, **kwargs: Any
+    ) -> Iterator[_T]:
+        yield self
```
cdce8p, line 186: What do we get by returning `self`, i.e. the AST node, here?

tests/unittest_inference.py
```diff
@@ -2642,11 +2642,11 @@ def true_value():
...
         self.assertTrue(klass.bool_value())
         dict_comp = next(module["dict_comp"].infer())
-        self.assertEqual(dict_comp, util.Uninferable)
+        assert isinstance(dict_comp, nodes.DictComp)
         set_comp = next(module["set_comp"].infer())
-        self.assertEqual(set_comp, util.Uninferable)
+        assert isinstance(set_comp, nodes.SetComp)
         list_comp = next(module["list_comp"].infer())
-        self.assertEqual(list_comp, util.Uninferable)
+        assert isinstance(list_comp, nodes.ListComp)
```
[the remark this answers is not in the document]
DanielNoord: Hm, yeah I didn't think of it like that. That would decouple the value of `lst` from the actual `ast` tree though.

For example, we would then have a `ListComp` node in `nodes.Module.body` but a `List` node for `lst`? If we create that `List` when calling `_infer` on `ListComp` what would then be the `lineno` etc.?

Wouldn't it be better to say that `lst` is a `ListComp` and let downstream users decide if they want to handle `ListComp` and `List` similarly? Like we often do with `FunctionDef` and `AsyncFunctionDef`.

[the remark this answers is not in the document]
cdce8p: > That would decouple the value of `lst` from the actual `ast` tree though.

Yeah, it's an inference **result** after all.

> If we create that `List` when calling `_infer` on `ListComp` what would then be the `lineno` etc.?

`None`? Both `lineno` and `col_offset` attributes are typed as optional.
Or the same one as the original `ListComp`.

> Wouldn't it be better to say that `lst` is a `ListComp` and let downstream users decide if they want to handle `ListComp` and `List` similarly? Like we often do with `FunctionDef` and `AsyncFunctionDef`.

The downstream user still has full control over it. If he once the "raw" data, he can just use the node itself. No need to call `infer` for that. Regarding `FunctionDef`, the infer method can actually return a `Property` if the function is decorated. Similarly, inferring a `DictUnpack` will return a single dict or a comparison just a bool constant
```py
{1: 'A', **{2: 'B'}}
True == True
```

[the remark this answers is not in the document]
DanielNoord: I guess we should probably handle this comprehension per comprehension then. Instead of one larger PR.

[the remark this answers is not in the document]
cdce8p: > I guess we should probably handle this comprehension per comprehension then. Instead of one larger PR.

Yeah. There are multiple parts to be aware of: evaluating the iterator, matching the loop var(s) with optional sequence unpacking, checking any if statements and doing late assignments via `:=`, evaluating the loop expression. Likely I'm forgetting something.

```py
[a + c for a, b in func() if (c := b * 2)]
```

If you, or someone else wants to go down that route, it's probably best to start with a fairly limited set constraints and only one comprehension type. Once that is down extending it should be fairly easy in comparison.

cdce8p:
Going to close this PR since we would need a different approach for inference.
See https://github.com/PyCQA/astroid/pull/1475#discussion_r868142345

PR #1476, opened 2022-03-15
Define ``generators`` on ``ComprehensionScope``

DanielNoord:
## Steps

## Description

Ref https://github.com/PyCQA/astroid/pull/1475.

This is necessary for `ComprehensionScope` to be able to access `generators` without `mypy` complaining. It doesn't really change behaviour either way and actually fixes an unintentional bug where `ListComp.generators == None` even though `postinit` is called.
_I see no reason why that behaviour shouldn't be the same as for all other `generators` attributes._

## Type of Changes

## Related Issue


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -724,14 +722,13 @@ def __init__(
             parent=parent,
         )
 
-    def postinit(self, elt=None, generators=None):
+    def postinit(self, elt=None, generators: Optional["nodes.Comprehension"] = None):
         """Do some setup after initialisation.
 
         :param elt: The element that forms the output of the expression.
         :type elt: NodeNG or None
```
Pierre-Sassoulas, line 729:
```suggestion
    def postinit(self, elt: "nodes.NodeNG" =None, generators: Optional["nodes.Comprehension"] = None):
        """Do some setup after initialisation.

        :param elt: The element that forms the output of the expression.
```
Wondering also if it should be ``Optional[List["nodes.Comprehension"]]``

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -822,7 +814,9 @@ def __init__(
             parent=parent,
         )
 
-    def postinit(self, key=None, value=None, generators=None):
+    def postinit(
+        self, key=None, value=None, generators: Optional["nodes.Comprehension"] = None
```
Pierre-Sassoulas, line 818:
```suggestion
        self, key=None, value=None, generators: Optional[List["nodes.Comprehension"]] = None
```
?

Pierre-Sassoulas (commented):
I did not comment on all the Optional[nodes.Comprehension], I suppose I'm missing something.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Well, that was quite stupid... Fixed!

Pierre-Sassoulas (approved):

PR #1477, opened 2022-03-15
Fix crash on direct inference via nodes.FunctionDef._infer

### LINKED ISSUE #817: IndexError in FunctionDef._infer()
### Steps to reproduce
Run following code:
```
import astroid
code = "def func(): pass"
funcdef = astroid.extract_node(code)
print(next(funcdef._infer()))
```

### Current behavior
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-4-0a19b019c67b> in <module>
----> 1 print(next(funcdef._infer()))

/dat/uber/prod/python/venv/py_3.7.4_prod_20200604_0/lib/python3.7/site-packages/astroid/inference.py in _cached_generator(func, instance, args, kwargs, _cache)
    957 @wrapt.decorator
    958 def _cached_generator(func, instance, args, kwargs, _cache={}):
--> 959     node = args[0]
    960     try:
    961         return iter(_cache[func, id(node)])

IndexError: tuple index out of range
```

This is actually a combination of two bugs:
1. `astroid.inference._cached_generator` is decorated by `wrapt.decorator`. And in the mean time you try to access node instance via `node = args[0]`. This is incorrect, you should use `node = instance` instead.
2. `NodeNG.infer` calls `_infer` with `context` as positional argument instead of keyword one. See https://github.com/PyCQA/astroid/blob/master/astroid/node_classes.py#L363 and https://github.com/PyCQA/astroid/blob/master/astroid/node_classes.py#L371

Since `_cached_generator` works not as expected, this may also have some impact on inference caching, but I did not encounter any other issues here.

### Expected behavior
No error

2.4.1

DanielNoord:
## Steps

## Description

Fix as suggested in the issue.

## Type of Changes

## Related Issue

Closes #817

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,9 @@ What's New in astroid 2.12.0?
 =============================
 Release date: TBA
 
+* Fixed crash on direct inference via ``nodes.FunctionDef._infer``.
+
+  Closes #817
 
 
 What's New in astroid 2.11.1?
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,9 +6,6 @@ What's New in astroid 2.12.0?
 =============================
 Release date: TBA
 
-* Fixed crash on direct inference via ``nodes.FunctionDef._infer``.
-
-  Closes #817
 
 
 What's New in astroid 2.11.1?
@@ -18,6 +15,9 @@ Release date: TBA
 * Promoted ``getattr()`` from ``astroid.scoped_nodes.FunctionDef`` to its parent
   ``astroid.scoped_nodes.Lambda``.
 
+* Fixed crash on direct inference via ``nodes.FunctionDef._infer``.
+
+  Closes #817
 
 
 What's New in astroid 2.11.0?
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas:
Shouldn't this be in 2.11.1 ?

DanielNoord:
> I was thinking crashes and bugs should be backported. But it's true that it increase the maintenance burden. Maybe only crashes ?

Then we would need to back port this as this prevents an `IndexError` 😅 

I think not releasing a hotfix if nobody really depends on it asap is fine. We have normal releases for "cold fixes" 😄

Pierre-Sassoulas (changes requested):
I can't suggest the change in interface for the changelog 2.11.1 is outside the diff. The rest LGTM.

DanielNoord:
@Pierre-Sassoulas Do you know you can use `.` on the Github website to launch a web-version of VS Code to make simple changes like those? Since you're maintainer of the repo you have write access to my branches.

Pierre-Sassoulas:
Wow, this is actually impressive (and a little slow too, but less than doing adding a remote and modifying locally). Thanks for the tip !

Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas Yeah I never used it much, but found out recently how useful it is for small changes to a PR

PR #1480, opened 2022-03-19
Make `else` its own scope

DanielNoord:
## Steps

## Description

This helps to identify where the `else` keyword is. It can be very useful, for example, for this issue: https://github.com/PyCQA/pylint/issues/872.
With a line number for the `else` keyword we can separate the different blocks in the `if...else` block and handle the disables accordingly.

Let me know what you guys think of this approach.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,11 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Changed `block_range` to consider `else` its own block, allowing `pylint` to apply
+  disables to just the block.
+
+  References pylint-dev/pylint#872
+
 
 What's New in astroid 4.1.2?
 ============================
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -244,15 +244,22 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
     def _elsed_block_range(
         self, lineno: int, orelse: list[nodes.NodeNG], last: int | None = None
     ) -> tuple[int, int]:
-        """Handle block line numbers range for try/finally, for, if and while
-        statements.
+        """handle block line numbers range for try/finally, for, if and while
+        statements
         """
-        if lineno == self.fromlineno:
+        # If at the end of the node, return same line
+        if lineno == self.tolineno:
             return lineno, lineno
         if orelse:
-            if lineno >= orelse[0].fromlineno:
+            # If the lineno is beyond the body of the node we check the orelse
+            if lineno >= self.body[-1].tolineno + 1:
+                # If the orelse has a scope of its own we determine the block range there
+                if isinstance(orelse[0], BlockRangeMixIn):
+                    return orelse[0]._elsed_block_range(lineno, orelse[0].orelse)
+                # Return last line of orelse
                 return lineno, orelse[-1].tolineno
-            return lineno, orelse[0].fromlineno - 1
+            # If the lineno is within the body we take the last line of the body
+            return lineno, self.body[-1].tolineno
         return lineno, last or self.tolineno
 
 
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -255,6 +255,8 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
         :returns: The range of line numbers that this node belongs to,
             starting at the given line number.
         """
+        if lineno < self.fromlineno:
+            return lineno, self.tolineno
         if lineno == self.body[0].fromlineno:
             return lineno, lineno
         if lineno <= self.body[-1].tolineno:
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -240,9 +240,26 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
     body: list[NodeNG]
     """The contents of the block."""
 
+    orelse: list[NodeNG]
+    """The contents of the ``else`` block."""
+
     @cached_property
     def blockstart_tolineno(self):
         return self.lineno
+
+    def block_range(self, lineno: int) -> tuple[int, int]:
+        """Get a range from the given line number to where this node ends.
+
+        :param lineno: The line number to start the range at.
+
+        :returns: The range of line numbers that this node belongs to,
+            starting at the given line number.
+        """
+        if lineno == self.body[0].fromlineno:
+            return lineno, lineno
+        if lineno <= self.body[-1].tolineno:
+            return lineno, self.body[-1].tolineno
+        return self._elsed_block_range(lineno, self.orelse, self.body[0].fromlineno - 1)
 
     def _elsed_block_range(
         self, lineno: int, orelse: list[nodes.NodeNG], last: int | None = None
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -237,6 +237,9 @@ class MultiLineBlockNode(NodeNG):
 class MultiLineWithElseBlockNode(MultiLineBlockNode):
     """Base node for multi-line blocks that can have else statements."""
 
+    body: list[NodeNG]
+    """The contents of the block."""
+
     @cached_property
     def blockstart_tolineno(self):
         return self.lineno
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -244,8 +244,8 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
     def _elsed_block_range(
         self, lineno: int, orelse: list[nodes.NodeNG], last: int | None = None
     ) -> tuple[int, int]:
-        """handle block line numbers range for try/finally, for, if and while
-        statements
+        """Handle block line numbers range for try/finally, for, if and while
+        statements.
         """
         # If at the end of the node, return same line
         if lineno == self.tolineno:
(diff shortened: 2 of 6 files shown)


astroid/rebuilder.py
```diff
@@ -1382,6 +1382,20 @@ def visit_global(self, node: "ast.Global", parent: NodeNG) -> nodes.Global:
...
         return newnode
 
+    def _find_else_keyword(self, node: "ast.If") -> Tuple[Optional[int], Optional[int]]:
+        """Get the line number and column offset of the `else` keyword."""
+        if not self._data or not node.orelse:
+            return None, None
+
+        end_lineno = node.orelse[0].lineno - 1
+
+        # pylint: disable-next=unsubscriptable-object
```
Pierre-Sassoulas, line 1392: Is there an issue in pylint for that false positive ?

Pierre-Sassoulas (approved):
LGTM, that would be very useful indeed.

astroid/rebuilder.py
DanielNoord: Issue in `pylint` I think. Marc added disables as well in his PRs.

DanielNoord:
Drafting this. Thought about this some more over the weekend and I think it makes sense to also pick up the `elif` when available.
For:
```python
if:
    ...
elif:
    ...
else:
    ...
```

The `elif` is the `orelse` of the first `if`. I think it makes sense to pick up that keyword as well.

Pierre-Sassoulas (dismissed):
Looks good, I think a review by Marc would be nice but if he does not have the time this can go :)

cdce8p (changes requested):
The current solution is quite inefficient. In worse case, the whole token list is generated multiple times for a single node. It would be better to calculate the `end_line` beforehand and using the `generate_tokens` iterator directly. For `_get_position_info` I've used `node.end_lineno` (were available) or iterated until the last line. Since it's a generator, we can exit early easily.

Besides the performance issue, I'm not sure a tokenize solution is the right approach. We need to search a whole range of lines which can contain basically any other nodes. For example, what about a nested function with an `IfExp` node in its body? There is also the issue of `else` which isn't exclusive to `if - else`. It can also be used with `try` and `for`. All in all, you would need to keep track of a lot of different things which I imaging will be pretty complicated.

Thinking about a different solution, is something preventing us to look at the first node in `orelse` and using its `lineno` and `col_offset`?

Some other questions:
* If that works, do we really need to add new node attributes in astroid? (Or can pylint deal with it itself.)
* What about other block nodes `for` with `else`, `try` with `except`, `else`, and `finally`? How do we deal with them?

In the end, is a change worth the effort?

DanielNoord:
> Thinking about a different solution, is something preventing us to look at the first node in `orelse` and using its `lineno` and `col_offset`?

```python
if condition:
    do_something()
elif other_condition:
    do_something_different()
else:
    # We don't want to do anything here
    # But we don't to disable a message: # pylint: disable=invalid-name
    dont_do_anything()
```

The issue with the approach and the code above is that we don't think comments into account. I thought this might become problematic, especially since the use-case for this is the handle disable comments in these blocks better. Doing `orelse[0].lineno - 1` doesn't guarantee we actually hit the `else` line.

The issue then became that I had to take quite large range of lines in order not to pass `SyntaxError` raising code to `generate_tokens`...

DanielNoord:
@cdce8p I changed the code to use a pretty naive but working solution. I'm not sure what the impact of `rstrip` is but it should be much better than `generate_tokens` right?

Let me know what you think!

cdce8p (commented):
Not particularly pretty, but I guess it should work.

Could you add some more test cases with `for - else` and `try - except - else` in addition to the nested one? This shouldn't be an issue with the current implementation, but those cases are never the less good as regression tests if we ever need to touch this part again.

DanielNoord:
@cdce8p After your comments I found out that we were actually missing some nodes which could have access to the same attributes.
`IfExp` is now the only node which has a `orelse` attribute but not a `orelse_lineno` attribute as it doesn't really seem relevant there. If you want I can add it there as well, but it is probably best to do so in a follow-up when we actually need it.

astroid/rebuilder.py
```diff
@@ -248,6 +248,26 @@ def _reset_end_lineno(self, newnode: nodes.NodeNG) -> None:
...
+        if not self._data or not node.orelse:
+            return None, None
+
+        start = node.orelse[0].lineno
+
+        # pylint: disable-next=unsubscriptable-object
+        for index, line in enumerate(self._data[start - 1 : node.lineno : -1]):
+            if line.lstrip().startswith("else"):
+                # if start - index == 37:
+                # breakpoint()
```
DanielNoord, line 264:
```suggestion
```

cdce8p:
> @cdce8p After your comments I found out that we were actually missing some nodes which could have access to the same attributes.
> `IfExp` is now the only node which has a `orelse` attribute but not a `orelse_lineno` attribute as it doesn't really seem relevant there. If you want I can add it there as well, but it is probably best to do so in a follow-up when we actually need it.

Let's step back for a moment. Just for me to understand, once the PR is merged how would you integrate the new information so that the issue is fixed? If I remember correctly, wouldn't you need to modify the `block_range` methods?

Something else, although I'm not sure it's an issue, with this PR we're only addressing `else` blocks. What about `finally` and `except`? Furthermore, is `col_offset` necessary or `lineno` enough?

DanielNoord:
@cdce8p Current iteration is with changing `block_range`. As you can see this will fail some tests as we would scopes differently. That's why I originally opted for a completely separate system.

The issue is basically the following example:
```python
if x:
    1
elif y:  # pylint: disable=pointless-statement
    1
else:
    1
```

Which lines should be covered by the disable? With this proposal it would be 3 and 4. Similar examples can be seen in the tests. The PR would make it so that every keyword creates its own scope "disable-wise". That's not completely backwards compatible I think, but it does seem a little bit more logical and also more in line with other tools such as `coverage.py` I think.

tests/test_nodes.py
```diff
@@ -440,13 +451,50 @@ def test_if_elif_else_node(self) -> None:
...
 
     def test_block_range(self) -> None:
-        # XXX ensure expected values
-        self.assertEqual(self.astroid.block_range(1), (0, 22))
-        self.assertEqual(self.astroid.block_range(10), (0, 22))  # XXX (10, 22) ?
+        """Test block_range of various scope constructs"""
+        # Module
+        self.assertEqual(self.astroid.block_range(1), (0, 33))
+        # NOTE: Module does not consider the lineno argument. It would be more consistent to make
+        # thi return (10, 33) but without a use case it seems better to not change behaviour.
```
Pierre-Sassoulas:
```suggestion
        # this return (10, 33) but without a use case it seems better to not change behaviour.
```

Pierre-Sassoulas (changes requested):
Maybe we could add the try-except-else tests that Marc talked about ?

tests/test_nodes.py
```diff
@@ -465,82 +513,18 @@ class TryNodeTest(_NodeTest):
...
         assert try_node.block_range(7) == (7, 7)
-        assert try_node.block_range(8) == (8, 8)
+        assert try_node.block_range(8) == (8, 9)
         assert try_node.block_range(9) == (9, 9)
-        assert try_node.block_range(10) == (10, 10)
+        assert try_node.block_range(10) == (10, 11)
         assert try_node.block_range(11) == (11, 11)
 
 
-class TryExceptNodeTest(_NodeTest):
```
DanielNoord, line 481: No longer needed, this is not a separate node type anymore and falls under `TryNodeTest` now. Same for the other ones.

tests/test_nodes_lineno.py
```diff
@@ -878,6 +878,10 @@ def test_end_lineno_while() -> None:
         assert (w1.body[0].end_lineno, w1.body[0].end_col_offset) == (2, 8)
         assert (w1.orelse[0].lineno, w1.orelse[0].col_offset) == (4, 4)
         assert (w1.orelse[0].end_lineno, w1.orelse[0].end_col_offset) == (4, 8)
+        assert w1.block_range(1) == (1, 2)
```
DanielNoord, line 881: Add tests for `While.block_range`

ChangeLog
```diff
@@ -7,6 +7,11 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Changed `block_range` to consider `else` its own block, allowing `pylint` to apply
```
DanielNoord, line 10: This remains the main change, as well as deduplicating a lot of code

tests/test_nodes.py
```diff
@@ -465,82 +513,18 @@ class TryNodeTest(_NodeTest):
 
     def test_block_range(self) -> None:
         try_node = self.astroid.body[0]
-        assert try_node.block_range(1) == (1, 11)
-        assert try_node.block_range(2) == (2, 2)
+        assert try_node.block_range(2) == (2, 3)
         assert try_node.block_range(3) == (3, 3)
-        assert try_node.block_range(4) == (4, 4)
+        assert try_node.block_range(4) == (4, 5)
```
DanielNoord, line 519: Brings this in line with the other nodes: a block starts with a keyword and contains the full body after said keyword

astroid/nodes/_base_nodes.py
```diff
@@ -237,22 +237,49 @@ def _assign_nodes_in_scope(self) -> list[nodes.Assign]:
...
+            # If the lineno is beyond the body of the node we check the orelse
+            if lineno >= self.body[-1].tolineno + 1:
+                # If the orelse has a scope of its own we determine the block range there
+                if isinstance(orelse[0], MultiLineWithElseBlockNode):
+                    return orelse[0]._elsed_block_range(lineno, orelse[0].orelse)
+                # Return last line of orelse
                 return lineno, orelse[-1].tolineno
-            return lineno, orelse[0].fromlineno - 1
+            # If the lineno is within the body we take the last line of the body
+            return lineno, self.body[-1].tolineno
```
DanielNoord, line 284: This remains the main change, as well as the simplification and standardization of the implementation across all nodes

Pierre-Sassoulas (approved):
Great ! Now should it be in astroid 5 or astroid 4.2 ? 😄

DanielNoord:
I have started the release process for `4.2.0b1`. We'll first need a patch release though...

cdce8p:
FWIW I think minor versions for astroid and pylint are enough. The way I understand it, it only changes where `pylint: disable` comments need to be placed. That's not all that different from us adding a new check.

_Just one data point but for Home Assistant, I didn't need to change anything._

PR #1485, opened 2022-03-23
Fix incorrect xfailed test

DanielNoord:
## Steps

## Description

I might be missing something, but:
```console
❯ cat test.py
class X(object):
    M = lambda self, arg: arg + 1
    x = 24

    def __init__(self):
        y = self
        self.m = y.M(1) + y.z

class C(object):
    z = 24

class Y(X, C):
    M = lambda self, arg: arg + 1

    def blurb(self):
        print(self.m)

Y().blurb()%                                                                                                                                         
❯ python test.py
26
```

I don't think this test is correct and the tested value should be 26.
The underlying issue which originally prompted the xfail has probably been solved.

While I was at it I also changed the test to use `pytest` asserts instead of unittest.

## Type of Changes

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -4056,10 +4056,7 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         self.assertIsInstance(inferred, nodes.Const)
         self.assertEqual(inferred.value, 25)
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Nice catch !

PR #1486, opened 2022-03-23
Astroid manager

DanielNoord:
## Steps

## Description

I could split out the comment commit, but I had that stored so wanted to get it in somewhere. Comment is outdated, see commit message.

Refactor is because of `mypy` and `pylint` complaining about the attributes. Not sure if this is the most elegant way to do this, but both `astroid` and `pylint` tests passed locally with this.

## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -319,7 +319,6 @@ class Instance(BaseInstance):
         return result
 
     def getitem(self, index, context=None):
-        # TODO: Rewrap index to Const for this case
         new_context = bind_context_to_node(context, self)
         if not context:
             context = new_context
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
LGTM. It's ok, for the other commit, let's just rebase.

knedlsepp:
@Pierre-Sassoulas This PR turns out to introduce a regression with pylint and CPython extensions: https://github.com/pylint-dev/pylint/issues/7433
I have not yet understood what exactly is wrong here, I will need to dig further. Given that this is only a refactoring would you also accept a PR that just reverts this?

PR #1488, opened 2022-03-25
Update ``pylint`` to ``2.13``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,5 +1,6 @@
 -r requirements_test_min.txt
 -r requirements_test_pre_commit.txt
+contributors-txt>=0.7.3
 coveralls~=3.3
 coverage~=5.5
 pre-commit~=2.17
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,6 +1,6 @@
 -r requirements_test_min.txt
 -r requirements_test_pre_commit.txt
-contributors-txt>=0.7.3
+contributors-txt>=0.7.4
 coveralls~=3.3
 coverage~=5.5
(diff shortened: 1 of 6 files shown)


requirements_test_pre_commit.txt
```diff
@@ -1,5 +1,5 @@
 black==22.1.0
-pylint==2.12.2
+pylint==2.13
```
Pierre-Sassoulas, line 2:
```suggestion
pylint~=2.13.0
```
I'm surprised 2.13 without the 0 is working ? TIL !

requirements_test.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements_test_min.txt
 -r requirements_test_pre_commit.txt
+contributors-txt>=0.7.3
```
Pierre-Sassoulas, line 3: It wasn't there because it was not compatible with python 3.7, there's the runtime pip install to remove in the tbump configuration now that it is :)

Pierre-Sassoulas (approved):

requirements_test.txt
Pierre-Sassoulas:
```suggestion
contributors-txt>=0.7.4
```

DanielNoord: Ah that's another reason not to wait too long with moving `pylint` to `3.7`: we can't do `astroid` until `pylint` has deprecated it.
Seeing the time it takes to merge a PR in `astroid` if we don't start the process soon we'll have `3.6` unsupported for `astroid` somewhere in 2025 😅

PR #1489, opened 2022-03-26
Avoid adding the name of a parent namedtuple to child's locals

jacobtylerwalls:
## Description

When subclassing a `namedtuple`, the behavior before d30592a5d27ca801aae36248ed169ae3dee5f4c9 was for astroid to avoid adding the name of the namedtuple to the child class's locals. Apparently doing so caused the crash in PyCQA/pylint#5982.

## Type of Changes

## Related Issue

Refs PyCQA/pylint#5982

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -133,7 +133,11 @@ def infer_func_form(
     # we know it is a namedtuple anyway.
     name = name or "Uninferable"
     # we want to return a Class node instance with proper attributes set
-    class_node = nodes.ClassDef(name, parent=node.parent)
+    class_node = nodes.ClassDef(name)
+    # A typical ClassDef automatically adds its name to the parent scope,
+    # but doing so causes problems, so defer setting parent until after init
+    # see: https://github.com/PyCQA/pylint/issues/5982
+    class_node.parent = node.parent
     class_node.postinit(
         # set base class=tuple
         bases=[base_type],
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,9 @@ What's New in astroid 2.11.2?
 =============================
 Release date: TBA
 
+* Avoided adding the name of a parent namedtuple to its child's locals.
+
+  Refs PyCQA/pylint#5982
 
 
 What's New in astroid 2.11.1?
--- a/tests/unittest_brain.py
+++ b/tests/unittest_brain.py
@@ -159,6 +159,8 @@ class NamedTupleTest(unittest.TestCase):
         self.assertEqual(
             [anc.name for anc in klass.ancestors()], ["X", "tuple", "object"]
         )
+        # See: https://github.com/PyCQA/pylint/issues/5982
+        self.assertNotIn("X", klass.locals)
         for anc in klass.ancestors():
             self.assertFalse(anc.parent is None)
 


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -133,7 +133,11 @@ def infer_func_form(
     # we know it is a namedtuple anyway.
     name = name or "Uninferable"
     # we want to return a Class node instance with proper attributes set
-    class_node = nodes.ClassDef(name, parent=node.parent)
+    class_node = nodes.ClassDef(name)
+    # A typical ClassDef automatically adds its name to the parent scope,
+    # but doing so causes problems, so defer setting parent until after init
+    # see: https://github.com/PyCQA/pylint/issues/5982
+    class_node.parent = node.parent
```
jacobtylerwalls, line 140: I borrowed the wording from: 
https://github.com/PyCQA/astroid/blob/16322e84801fc143906db33868abf5813875ae30/astroid/objects.py#L265-L267

Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
I'm going to release the hot fix in 2.11.2 / 2.13.2, when @DanielNoord is available for review we can integrate his review in 2.12.

DanielNoord:
I'm not sure this is correct. Doesn't this code check something tricky?
```python
from collections import namedtuple

class X(namedtuple("X", ["a", "b", "c"])):
    pass
```

I believe `X` should be in the locals of the ClassDef as that `X` is referring to `NamedTuple` `X`.

For:
```python
from collections import namedtuple

class Y(namedtuple("X", ["a", "b", "c"])):
    pass
```

`X` should also be in the locals of `Y`.

```console
(Pdb) klass.locals
{'__module__': [<Const.str l.4 at 0x10ebbfa30>], '__qualname__': [<Const.str l.4 at 0x10ebbf940>], 'X': [<ClassDef.X l.4 at 0x10ebbfb80>]}
(Pdb) klass
<ClassDef.X l.4 at 0x10ebbfd60>
(Pdb) klass.locals["X"]
[<ClassDef.X l.4 at 0x10ebbfb80>]
(Pdb) klass.locals["X"] == klass
False
```

Note that these are two different `ClassDefs`. Their address doesn't match.

jacobtylerwalls:
Yes, they are different `ClassDef`s. But why should the name of the parent class be in the child class's locals? It's not even in the parent class's locals, is it?
```
>>> X = namedtuple("X", [])
>>> 'X' in dir(X)
False
```
So why would we want 'X' to be in the locals of Y when it's not even in X?

DanielNoord:
> Yes, they are different `ClassDef`s. But why should the name of the parent class be in the child class's locals? It's not even in the parent class's locals, is it?
> 
> ```
> >>> X = namedtuple("X", [])
> >>> 'X' in dir(X)
> False
> ```
> 
> So why would we want 'X' to be in the locals of Y when it's not even in X?

Ah yeah, didn't think about that.

It's getting late here, but are there any other examples where something gets defined in the "ancestors space" of a class definition? Do we add them as well?
I guess the fix is actually correct then, but it does feel a bit weird. Like, wouldn't it be better to handle this in the `__init__` instead of arbitrary not passing `parent`?

jacobtylerwalls:
> Like, wouldn't it be better to handle this in the __init__ instead of arbitrary not passing parent?

Oh it definitely feels janky. I was just trying to restore status quo before the commit that caused the failure. You're right this could be audited and fixed in the `__init__`.

DanielNoord:
Would you mind opening an issue for it? Then at least we don't forget about it :)

jacobtylerwalls:
done!

PR #1504, opened 2022-04-07
Fix flake8 bugbear version and fix some flake8 issues

Pierre-Sassoulas:
## Type of Changes

### DIFF
--- a/tests/unittest_inference_calls.py
+++ b/tests/unittest_inference_calls.py
@@ -522,8 +522,8 @@ def test_instance_method_inherited() -> None:
     A.method(B())  #@
     """
     )
-    expected = ["A", "A", "B", "B", "B"]
-    for node, expected in zip(nodes_, expected):
+    expected_names = ["A", "A", "B", "B", "B"]
+    for node, expected in zip(nodes_, expected_names):
         assert isinstance(node, nodes.NodeNG)
         inferred = node.inferred()
         assert len(inferred) == 1
@@ -553,8 +553,8 @@ def test_class_method_inherited() -> None:
     B.method()  #@
     """
     )
-    expected = ["A", "A", "B", "B"]
-    for node, expected in zip(nodes_, expected):
+    expected_names = ["A", "A", "B", "B"]
+    for node, expected in zip(nodes_, expected_names):
         assert isinstance(node, nodes.NodeNG)
         inferred = node.inferred()
         assert len(inferred) == 1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 5
+  CACHE_VERSION: 6
   DEFAULT_PYTHON: 3.8
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,8 @@ repos:
     rev: 4.0.1
     hooks:
       - id: flake8
-        additional_dependencies: [flake8-bugbear, flake8-typing-imports==1.12.0]
+        additional_dependencies:
+          [flake8-bugbear==22.3.23, flake8-typing-imports==1.12.0]
         exclude: tests/testdata|doc/conf.py
   - repo: local
     hooks:


astroid/interpreter/objectmodel.py
```diff
@@ -60,6 +61,26 @@ def _dunder_dict(instance, attributes):
     return obj
 
 
+@lru_cache(maxsize=None)
+def str_startswith_attr_(obj: str) -> bool:
```
DanielNoord, line 65: Should we really cache this? Seems a bit a overkill?

astroid/interpreter/objectmodel.py
```diff
@@ -60,6 +61,26 @@ def _dunder_dict(instance, attributes):
...
+    """Done this way because ObjectModel.attributes was cached
+
+    See B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead
+    to memory leaks. The cache may retain instance references, preventing garbage
+    collection."""
+    return obj.startswith(IMPL_PREFIX)
+
+
+@lru_cache(maxsize=None)
+def after_attr__in_str(obj: str) -> str:
```
DanielNoord, line 75: Same here..

astroid/interpreter/objectmodel.py
```diff
@@ -100,20 +121,16 @@ def __get__(self, instance, cls=None):
     def __contains__(self, name):
         return name in self.attributes()
 
-    @lru_cache(maxsize=None)
-    def attributes(self):
+    def attributes(self) -> List[str]:
```
DanielNoord, line 124: The expensive part of this function is (I assume) the iteration over the `dir`. The other two function seem fairly cost efficient? Perhaps just remove the caching here altogether?

astroid/nodes/node_classes.py
```diff
@@ -366,21 +366,19 @@ def get_children(self):
 class LookupMixIn:
     """Mixin to look up a name in the right scope."""
 
-    @lru_cache(maxsize=None)
-    def lookup(self, name):
+    @lru_cache(maxsize=None)  # pylint: disable=cache-max-size-none # noqa
```
DanielNoord, line 369: Don't we want a max here? `128` on the same `LookupMixIn` seems like it should fit most of our uses in `pylint`?

astroid/transforms.py
```diff
@@ -3,10 +3,13 @@
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
 import collections
-from functools import lru_cache
+from typing import TYPE_CHECKING
 
 from astroid.context import _invalidate_cache
 
+if TYPE_CHECKING:
+    from astroid.nodes.node_ng import NodeNG
```
DanielNoord, line 11: I know we have different opinions about this, but I like importing `nodes` here as it also keeps the line much shorter and gives some consistency with the `pylint` codebase. I'll leave this to you!

astroid/transforms.py
```diff
@@ -22,18 +25,21 @@ class TransformVisitor:
     def __init__(self):
         self.transforms = collections.defaultdict(list)
 
-    @lru_cache(maxsize=TRANSFORM_MAX_CACHE_SIZE)
-    def _transform(self, node):
+    def _transform(self, node: "NodeNG") -> "NodeNG":
```
DanielNoord, line 28: Wasn't #1242 looking at how to cache this without running into memory issues?

tests/unittest_inference_calls.py
```diff
@@ -522,8 +522,8 @@ class B(A):
     A.method(B())  #@
     """
     )
-    expected = ["A", "A", "B", "B", "B"]
-    for node, expected in zip(nodes_, expected):
+    expecteds = ["A", "A", "B", "B", "B"]
+    for node, expected in zip(nodes_, expecteds):
```
DanielNoord, line 526:
```suggestion
    expected_names = ["A", "A", "B", "B", "B"]
    for node, expected in zip(nodes_, expected_names):
```

tests/unittest_inference_calls.py
```diff
@@ -553,8 +553,8 @@ class B(A):
     B.method()  #@
     """
     )
-    expected = ["A", "A", "B", "B"]
-    for node, expected in zip(nodes_, expected):
+    expecteds = ["A", "A", "B", "B"]
+    for node, expected in zip(nodes_, expecteds):
```
DanielNoord, line 557:
```suggestion
    expected_names = ["A", "A", "B", "B"]
    for node, expected in zip(nodes_, expected_names):
```

Pierre-Sassoulas:
Maybe we can remove all the caching temporarily and review/merge #1242 ?

DanielNoord:
> Maybe we can remove all the caching temporarily and review/merge #1242 ?

I think that PR was almost good to go?

Pierre-Sassoulas:
I fixed the merge conflict when merging main on 1242, it has gone stale for a long time. What about merging the non-cache fix here first even if it fails, then #1242 ?

DanielNoord:
> I fixed the merge conflict when merging main on 1242, it has gone stale for a long time. What about merging the non-cache fix here first even if it fails, then #1242 ?

Let's add `TODOs` then to make sure we don't forget about this. And perhaps try to contact that PR owner to see if we can get it merged.
I know last time I looked at it I hadn't looked at the implications of my new knowledge about `lru_caches` and that PR. That's the only thing we need to fix I think: making sure the new cache isn't used on methods or handles `self` better.

Pierre-Sassoulas:
> Let's add TODOs then to make sure we don't forget about this. And perhaps try to contact that PR owner to see if we can get it merged.

Well we won't forget about this if we merge with pre-commit failing 😄

PR #1505, opened 2022-04-08
Fix instance_attrs building error in Qt brain

jacobtylerwalls:
## Description
Fix root cause of PyCQA/pylint#6221. `instance_attrs` should have a list, not a bare Node:
https://github.com/PyCQA/astroid/blob/6578a7d04dc28a539402dc27ef50ac756bc9a6d7/astroid/nodes/scoped_nodes/scoped_nodes.py#L1135

## Type of Changes

## Related Issue

Closes PyCQA/pylint#6221

Supersedes PyCQA/pylint#6235

### DIFF
--- a/.github/workflows/release-tests.yml
+++ b/.github/workflows/release-tests.yml
@@ -31,3 +31,31 @@ jobs:
           . venv2\scripts\activate
           echo "import distutils.util  # pylint: disable=unused-import" > test.py
           pylint test.py
+
+  qt6-test:
+    # Regression test added in https://github.com/PyCQA/astroid/pull/1505
+    name: Regression test for Qt6 brain
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.0.0
+      - name: Set up Python
+        id: python
+        uses: actions/setup-python@v3.1.0
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+      - name: Install Qt
+        run: |
+          sudo apt-get install build-essential libgl1-mesa-dev
+      - name: Create Python virtual environment
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt -r requirements_test_brain.txt
+          pip install -e .
+      - name: Run brain_qt tests
+        run: |
+          . venv/bin/activate
+          pytest tests/unittest_brain_qt.py
--- a/.github/workflows/release-tests.yml
+++ b/.github/workflows/release-tests.yml
@@ -32,9 +32,8 @@ jobs:
           echo "import distutils.util  # pylint: disable=unused-import" > test.py
           pylint test.py
 
-  qt6-test:
-    # Regression test added in https://github.com/PyCQA/astroid/pull/1505
-    name: Regression test for Qt6 brain
+  additional-dependencies-linux-tests:
+    name: Regression tests w/ additional dependencies (Linux)
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
@@ -56,6 +55,7 @@ jobs:
           pip install -U -r requirements_test.txt -r requirements_test_brain.txt
           pip install -e .
       - name: Run brain_qt tests
+        # Regression test added in https://github.com/PyCQA/astroid/pull/1505
         run: |
           . venv/bin/activate
           pytest tests/unittest_brain_qt.py
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,9 @@ What's New in astroid 2.11.3?
 =============================
 Release date: TBA
 
+* Fixed an error in the Qt brain when building ``instance_attrs``.
+
+  Closes PyCQA/pylint#6221
 
 
 What's New in astroid 2.11.2?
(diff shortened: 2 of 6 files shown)


tests/unittest_brain_qt.py
```diff
@@ -0,0 +1,36 @@
...
+
+from astroid import extract_node
+from astroid.bases import UnboundMethod
+from astroid.manager import AstroidManager
+from astroid.nodes import FunctionDef
+
+HAS_PYQT6 = find_spec("PyQt6")
+
+
+@pytest.mark.skipif(HAS_PYQT6 is None, reason="This test requires the PyQt6 library.")
```
jacobtylerwalls, line 17: Maybe this test is platform specific? It passes on my mac.

jacobtylerwalls:
Adding the typing changes but not the functional change yields:
```
astroid/brain/brain_qt.py:37: error: Incompatible types in assignment (expression has type "NodeNG", target has type "List[NodeNG]")  [assignment]
```

tests/unittest_brain_qt.py
jacobtylerwalls: (I don't know how to verify the C extensions were installed on linux.)

Pierre-Sassoulas: Do you have ``PyQt6`` installed ? Maybe it was skipped because of line 34 ?

jacobtylerwalls: Yes the test passes for me on a Mac with PyQt6 installed and skips properly when it's not installed. I added a `pip list` command to another branch and opened a PR against my own fork, and verified that `PyQt6` is still installed on CI. Somehow the inference is still `Uninferable`. I'm wondering if that's because on linux `pip install` doesn't necessary install the c bindings. I looked at the docs. Something about `qmake must be on your PATH`?

Pierre-Sassoulas: >  I'm wondering if that's because on linux pip install doesn't necessary install the c bindings.

Yes, we need the c-binding installed for the system before launching pip. But we should probably separate tests with pyqt installed vs our standard tests. We don't want astroid to unexpectedly break if pyqt6 is not installed. Right now we're skipping some tests if numpy/pandas/pyqt is not installed, but we might want to create a mark and specific runner with the required dependencies to cover all the tests that requires a dependency or another. I

ChangeLog
```diff
@@ -12,6 +12,9 @@ What's New in astroid 2.11.3?
 =============================
 Release date: TBA
 
+* Fixed an error in the Qt brain when building ``instance_attrs``.
+
+  Closes PyCQA/pylint#6221
```
jacobtylerwalls, line 17: I can move this to 2.12, but I was wondering if we will be releasing it before pylint 2.14 to prevent user crashes.

Pierre-Sassoulas: Sorry oversight on my part, it should be the next patch release indeed.

tests/unittest_brain_qt.py
jacobtylerwalls: Added a test to the "release tests" workflow that when manually triggered on my fork, passed!

DanielNoord (changes requested):
Change itself looks good (and a pretty big oversight when this created). Some minor other comments.

.github/workflows/release-tests.yml
```diff
@@ -31,3 +31,31 @@ jobs:
           . venv2\scripts\activate
           echo "import distutils.util  # pylint: disable=unused-import" > test.py
           pylint test.py
+
+  qt6-test:
```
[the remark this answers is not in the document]
jacobtylerwalls: I don't think it's worth it. I'd have to look up and test the commands for installing Qt on Windows. If we have more regression tests for Linux, I think we should add them to this one for Qt, but I think we'll make it harder to debug future failures in the `virtualenv==15` test if we start installing other dependencies into it. For tests run a few times a year, I think the setup time is okay. WDYT?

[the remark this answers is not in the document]
jacobtylerwalls: I'd be open to renaming the job something like "Regression tests - linux - additional dependencies" or sth to encourage future tests to be consolidated there.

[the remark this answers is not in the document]
DanielNoord: Ah I didn't see the name of this file. Makes sense to keep it like this, but I do like the name change suggestion!

DanielNoord (approved):

DanielNoord:
@jacobtylerwalls Shall we merge this? CI failures are unrelated.

.github/workflows/release-tests.yml
```diff
@@ -31,3 +31,31 @@ jobs:
...
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.0.0
+      - name: Set up Python
+        id: python
+        uses: actions/setup-python@v3.1.0
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+      - name: Install Qt
+        run: |
+          sudo apt-get install build-essential libgl1-mesa-dev
```
The-Compiler, line 49: As mentioned over in https://github.com/PyCQA/astroid/pull/1531#discussion_r860683187:

> I'm confused - why does "Install Qt" actually only do `sudo apt-get install build-essential libgl1-mesa-dev`? You are installing PyQt from binary wheels, so that 40s step can probably just be dropped entirely.

Issue #562, opened 2018-06-18
brain_pytest and brain_gi should not import their underlying module (gi and pytest)


PCManticore:
This breaks the static analysis contract since we're now importing third party libraries to build AST trees from them. While this makes sense for plugins, we shouldn't do this for core astroid, as it means we're now breaking the contract by importing live modules. Note that we currently do that with `raw_building`, but we only do that by default for builtin modules, while for third party modules we rely on `--extension-pkg-whitelist` option.

lovetox:
I looked into this and i have some questions

how does pylint discover all members of a object that is imported from a third party module?
does it search that module on harddisk and uses astroid to parse the python source code of said module?

If yes this is impossible for gi. gi is a wrapper around C libs that use the gobject framework. No parsing of python source code will yield the necessary class definitions and propertys to infer anything, because there is no python source code.

So what would be a good way forward?

Maybe a plugin that simply provides all the class definitions of all the libs gi wrapps? like a big stub file? And people could simply install this plugin if they need it?

lovetox:
@PCManticore a comment would be appreciated, i would like to work on this, but i have no idea on the direction how to solve the whole gi dilema, thanks

ernestask:
Yes, a comment would be nice if there is someone willing to solve the whole gi dilemma.

gnacsa:
I also would be grateful for a solution.

PR #1514, opened 2022-04-17
Capture and log messages emitted by C modules when importing them

jacobtylerwalls:
## Steps

## Description

~Pipe stdout to stderr~ Capture and log messages emitted while importing C modules. This prevents contaminating programmatic output, e.g. pylint's JSON reporter.

An alternative would be to silence stdout altogether, but there may be users who appreciate the notice that code is being executed. This way no information is lost, while still allowing clean programmatic output.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#3518

/cc @piotr-machura, what do you think of this approach?

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -19,16 +19,22 @@ from __future__ import annotations
 import importlib
 import importlib.machinery
 import importlib.util
+import io
 import itertools
+import logging
 import os
 import sys
 import sysconfig
 import types
+from contextlib import redirect_stderr, redirect_stdout
 from functools import lru_cache
 from pathlib import Path
 
 from astroid.const import IS_JYTHON, IS_PYPY
 from astroid.interpreter._import import spec, util
+
+logger = logging.getLogger(__name__)
+
 
 if sys.platform.startswith("win"):
     PY_SOURCE_EXTS = ("py", "pyw")
@@ -171,7 +177,25 @@ def load_module_from_name(dotted_name: str) -> types.ModuleType:
     except KeyError:
         pass
 
-    return importlib.import_module(dotted_name)
+    # Capture and log anything emitted during import to avoid
+    # contaminating JSON reports in pylint
+    with redirect_stderr(io.StringIO()) as stderr, redirect_stdout(
+        io.StringIO()
+    ) as stdout:
+        module = importlib.import_module(dotted_name)
+
+    stderr_value = stderr.getvalue()
+    if stderr_value:
+        logger.error(
+            "Captured stderr while importing %s:\n%s", dotted_name, stderr_value
+        )
+    stdout_value = stdout.getvalue()
+    if stdout_value:
+        logger.info(
+            "Captured stdout while importing %s:\n%s", dotted_name, stdout_value
+        )
+
+    return module
 
 
 def load_module_from_modpath(parts):
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -5,7 +5,6 @@ coveralls~=3.3
 coverage~=6.3.3
 pre-commit~=2.19
 pytest-cov~=3.0
-contributors-txt>=0.7.3
 tbump~=6.9.0
 types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.3
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -5,6 +5,7 @@ coveralls~=3.3
 coverage~=6.3.3
 pre-commit~=2.19
 pytest-cov~=3.0
+contributors-txt>=0.7.3
 tbump~=6.9.0
 types-typed-ast; implementation_name=="cpython" and python_version<"3.8"
 types-pkg_resources==0.1.3
--- a/tests/unittest_modutils.py
+++ b/tests/unittest_modutils.py
@@ -57,7 +57,7 @@ class ModuleFileTest(unittest.TestCase):
 
 
 class LoadModuleFromNameTest(unittest.TestCase):
-    """load a python module from it's name"""
+    """load a python module from its name"""
 
     def test_known_values_load_module_from_name_1(self) -> None:
         self.assertEqual(modutils.load_module_from_name("sys"), sys)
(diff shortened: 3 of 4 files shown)


Pierre-Sassoulas (commented):
I'm wondering if the solution would be to add debug logging to astroid. It seems to me that this output is not error so technically should not be printed to stderr (although it's convenient to do so because we would not have to setup logging). I don't think we have logging in astroid yet, but we don't have to do the configuration, probably just clarifying in the documentation that you need to activate logging with debug level if you want to see the stdout from loading c module.

jacobtylerwalls:
> It seems to me that this output is not error so technically should not be printed to stderr (although it's convenient to do so because we would not have to setup logging).

Thanks, I thought about this for a bit. I think the fact that astroid imports C modules (that are allowlisted, or even just that have a brain, see #562), is worth warning the user, and that's an okay use for stderr.  "Welcome to pygame!" is not an error in `pygame`, but IMO when it happens in `astroid` it's worth a warning because it could be unexpected.

I found a couple interesting things at the Python [logging docs for library developers](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library): 

> If the using application does not use logging, and library code makes logging calls, then (as described in the previous section) events of severity WARNING and greater will be printed to sys.stderr. This is regarded as the best default behaviour.

So I think even if we go the logging route we'll end up generating messages in `stderr`, is that right?

jacobtylerwalls:
Ah, maybe we could wrap it in a UserWarning or something that can be filtered idiomatically.

Pierre-Sassoulas:
> So I think even if we go the logging route we'll end up generating messages in stderr,

I would use the "info" log level for what comes from stdout, debug is a little low. If the output comes from stderr originally we should use the error level though. Reading your documentation it seems we should add a ``NullHandler`` so nothing is printed unless lib users add a configuration. 

> If for some reason you don’t want these messages printed in the absence of any logging configuration, you can attach a do-nothing handler to the top-level logger for your library. This avoids the message being printed, since a handler will always be found for the library’s events: it just doesn’t produce any output. If the library user configures logging for application use, presumably that configuration will add some handlers, and if levels are suitably configured then logging calls made in library code will send output to those handlers, as normal.

> A do-nothing handler is included in the logging package: [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) (since Python 3.1). An instance of this handler could be added to the top-level logger of the logging namespace used by the library (if you want to prevent your library’s logged events being output to sys.stderr in the absence of logging configuration). If all logging by a library foo is done using loggers with names matching ‘foo.x’, ‘foo.x.y’, etc. then the code:

```python
import logging
logging.getLogger(astroid).addHandler(logging.NullHandler())

logger = logging.getLogger(__file__)
...
if stderr:
    logger.error(...)
else:
    logger.info(...)
```

astroid/modutils.py
```diff
@@ -171,7 +177,23 @@ def load_module_from_name(dotted_name: str) -> types.ModuleType:
...
+    # contaminating JSON reports in pylint
+    with redirect_stderr(io.StringIO()) as stderr, redirect_stdout(
+        io.StringIO()
+    ) as stdout:
+        module = importlib.import_module(dotted_name)
+
+    stderr_value = stderr.getvalue()
+    if stderr_value:
+        logger.error(
+            "Captured stderr from importing %s:\n%s", dotted_name, stderr_value
```
Pierre-Sassoulas, line 190:
```suggestion
            "Captured stderr while importing %s:\n%s", dotted_name, stderr_value
```

astroid/modutils.py
```diff
@@ -171,7 +177,23 @@ def load_module_from_name(dotted_name: str) -> types.ModuleType:
...
+        )
+    stdout_value = stdout.getvalue()
+    if stdout_value:
+        logger.info("Captured stdout from importing %s:\n%s", dotted_name, stdout_value)
```
Pierre-Sassoulas, line 194:
```suggestion
        logger.info("Captured stdout while importing %s:\n%s", dotted_name, stdout_value)
```

tests/unittest_modutils.py
```diff
@@ -57,7 +57,7 @@ def test_find_egg_module(self) -> None:
 
 
 class LoadModuleFromNameTest(unittest.TestCase):
-    """load a python module from it's name"""
+    """load a python module from its name"""
```
Pierre-Sassoulas, line 61: I think it would be nice to have a test mocking ``importlib.import_module`` with a check of capsys stdout and stderr being falsey and caplog outputting what is expected when set in info level.

```python
def test_import_dotted_library(
    capsys: CaptureFixture,
    caplog: LogCaptureFixture,
):
    caplog.set_level(logging.INFO)
    with patch(`"importlib.import_module", function_with_stdout_and_stderr(expected_out, expected_err)):
        ...
    out, err = capsys.readouterr()
    assert not out
    assert not err
    assert expected_out in caplog.text
    assert expected_err in caplog.text
```

Pierre-Sassoulas (changes requested):

tests/unittest_modutils.py
Pierre-Sassoulas: I still have this on my todo list, but unstacking the latest github mail (last in first out 😄) got me too late in the night yet again :)

Pierre-Sassoulas (approved):
LGTM, that's a nice addition to 2.12 !

tests/unittest_modutils.py
jacobtylerwalls: Thank you for figuring it out!

requirements_test.txt
```diff
@@ -5,7 +5,6 @@ coveralls~=3.3
 coverage~=6.3.3
 pre-commit~=2.19
 pytest-cov~=3.0
-contributors-txt>=0.7.3
```
jacobtylerwalls, line 8: @Pierre-Sassoulas did you want to remove this in another MR?

Pierre-Sassoulas: It happened to prevent installing with pip locally I guess it's not that important. (Duplicated entry)

Issue #1516, opened 2022-04-17
TODO list for Python 3.11 support


DanielNoord:
- [x] Re-enable `numpy` as a test dependency once they provide wheels for `3.11` > See #1517
- [x] Remove the `if: ${{ always() }}` line from the `coverage` CI jobs
- [ ] Handle PEP654 (https://github.com/PyCQA/pylint/issues/7703) #2028 
- [x] `FormattedValue.conversion` is a required field (instead of optional). This was an error in the Python documentation which was addressed recently. https://docs.python.org/3.11/library/ast.html#ast.FormattedValue https://github.com/python/cpython/pull/30467 > https://github.com/PyCQA/astroid/pull/2029
- [x] [Optional] We could add an enum for `FormattedValue.conversion`. The int values can be confusion on its own. > Decided not to.

cdce8p:
Some recent AST changes:
* Add support for new `TryStar` node. Almost identical to the `ast.Try` node. We shouldn't use the old `TryExcept` / `TryFinally` structure though! It's probably better to look at #1389
https://docs.python.org/3.11/library/ast.html#ast.TryStar
https://github.com/python/cpython/pull/29581
* `FormattedValue.conversion` is a required field (instead of optional). This was an error in the Python documentation which was addressed recently.
https://docs.python.org/3.11/library/ast.html#ast.FormattedValue
https://github.com/python/cpython/pull/30467
* [Optional] We could add an enum for `FormattedValue.conversion`. The int values can be confusion on its own.

rkhwaja:
The wrapt dependency needs to be ">=1.14" for python 3.11. See [changelog](https://wrapt.readthedocs.io/en/latest/changes.html#version-1-14-0)

Pierre-Sassoulas:
@rkhwaja I think pip handle that, if pylint require wrapt >= 1.12 but python 3.11 require wrapt >= 1.14 then pip download >= 1.14. We should not change our lower bound, as we could imagine that wrapt >=1.14 is incompatible with python 3.7 and changing this would break pylint for python 3.7.

rkhwaja:
You could add a dependency like "wrapt>=1.14,<2;python_version>='3.11'" and then both the old and the new python will work.

I don't know if you have other users than pylint but they would all just work too if you did it this way.

Pierre-Sassoulas:
But is this required at all ? ``wrapt`` is the package that should handle the metadata about the pytho interpreters it supports and pip takes it into account.

[...]

PR #1513, opened 2022-04-17
Find frozen stdlib modules

DanielNoord:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description

Closes #1512.

Related is https://github.com/python/cpython/issues/65935. I'm not sure how to test this on `3.10` though as this is a new feature on `3.11`. That would be the next step.
 
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

PR #1517, opened 2022-04-17
Enable ``3.11`` in the CI

DanielNoord:
]
## Steps

## Type of Changes

## Related Issue

Ref. #1516

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -179,6 +179,7 @@ jobs:
     name: tests / run / ${{ matrix.python-version }} / Windows
     runs-on: windows-latest
     timeout-minutes: 20
+    if: ${{ always() }}
     needs: tests-linux
     strategy:
       fail-fast: true
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,13 +137,9 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.11-dev"]
-    outputs:
-      python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.1
-        with:
-          fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
         id: python
         uses: actions/setup-python@v3.1.2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -81,9 +81,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     strategy:
-      fail-fast: true
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11.0-alpha.7 - 3.11.0"]
```
cdce8p, line 85: Why use `"3.11.0-alpha.7 - 3.11.0"`? Maybe `3.11-dev` would work?

.github/workflows/ci.yaml
```diff
@@ -179,11 +179,12 @@ jobs:
     name: tests / run / ${{ matrix.python-version }} / Windows
     runs-on: windows-latest
     timeout-minutes: 20
+    if: ${{ always() }}
```
cdce8p, line 182: That doesn't look great.

cdce8p (commented):
Python 3.11 is still in alpha at this time. Maybe it's better to add it as a separate job for now, instead of using `always()`. We can merge that back once the beta has gotten stable enough.

.github/workflows/ci.yaml
DanielNoord: This was just to make sure we would get the latest version (which was necessary for `tomllib` for the `pylint` counter-PR). I'll try with `3.11-dev`.

.github/workflows/ci.yaml
DanielNoord: I'll create a new job.

.github/workflows/ci.yaml
```diff
@@ -130,11 +130,65 @@ jobs:
...
+          name: coverage-${{ matrix.python-version }}
+          path: .coverage
+
   coverage:
     name: tests / process / coverage
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    needs: ["tests-linux"]
+    needs: ["tests-linux", "tests-linux-alpha"]
+    if: ${{ always() }}
```
DanielNoord, line 191: @cdce8p This remains necessary for the coverage to include `3.11` which would be good for any changes we introduce to fix regression with that version. I have removed it from the windows job though.

.github/workflows/ci.yaml
```diff
@@ -179,6 +233,7 @@ jobs:
     name: tests / run / ${{ matrix.python-version }} / Windows
     runs-on: windows-latest
     timeout-minutes: 20
+
```
DanielNoord, line 236:
```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -130,11 +130,65 @@ jobs:
...
 
+  tests-linux-alpha:
+    name: tests / run / ${{ matrix.python-version }} / Linux
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        python-version: ["3.11-dev"]
+    outputs:
+      python-key: ${{ steps.generate-python-key.outputs.key }}
```
cdce8p, line 141: The job `output` isn't actually used anywhere. So this is safe to remove.
`${{ steps.python.outputs.python-version }}` does work without it.

.github/workflows/ci.yaml
```diff
@@ -130,11 +130,65 @@ jobs:
...
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.0.1
+        with:
+          fetch-depth: 0
```
cdce8p, line 146: Why set it to `0`? That will fetch the complete history which is unnecessary. A shallow copy (with the default `1`) should be enough. https://github.com/actions/checkout#usage

.github/workflows/ci.yaml
cdce8p: Not sure that is all that useful. The coverage is merged anyway, so the only thing we would get would be the lines with `PY311_PLUS`.

.github/workflows/ci.yaml
```diff
@@ -227,6 +281,59 @@ jobs:
           . venv\\Scripts\\activate
           pytest tests/
 
+  tests-windows-alpha:
+    name: tests / run / ${{ matrix.python-version }} / Windows
+    runs-on: windows-latest
```
cdce8p, line 286: Do we want / need to test Windows? For almost everything the `Linux` test should be enough. Once merged we'll test Windows anyway.

.github/workflows/ci.yaml
DanielNoord: I think that would still be helpful. For example, more "risky" PRs such as https://github.com/PyCQA/astroid/pull/1513 would probably be less risky if we know coverage is okay.

cdce8p (approved):
Pushed some minor changes. Looks good now!

PR #1518, opened 2022-04-17
Fix a crash in the `gi` brain

jacobtylerwalls:
## Description

Fix a crash in the `gi` brain which calls `gi.module.IntrospectionModule.__getattr__()`. It can raise [various](https://gitlab.gnome.org/GNOME/pygobject/-/blob/master/gi/module.py) exceptions, so catching `AttributeError` is too narrow.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#6371.

@senier, would you be able to let us know if this resolves your issue? I haven't reproduced the crash locally because I'd like to avoid installing the C headers for `cairo` if possible.

### DIFF
--- a/astroid/brain/brain_gi.py
+++ b/astroid/brain/brain_gi.py
@@ -74,7 +74,9 @@ def _gi_build_stub(parent):
 
         try:
             obj = getattr(parent, name)
-        except AttributeError:
+        except Exception:  # pylint: disable=broad-except
+            # gi.module.IntrospectionModule.__getattr__() can raise all kinds of things
+            # like ValueError, TypeError, NotImplementedError, RepositoryError, etc
             continue
 
         if inspect.isclass(obj):
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,10 @@ Release date: TBA
 
   Closes PyCQA/pylint#6221
 
+* Fixed a crash in the ``gi`` brain.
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (approved):

astroid/brain/brain_gi.py
```diff
@@ -74,7 +74,9 @@ def _gi_build_stub(parent):
 
         try:
             obj = getattr(parent, name)
-        except AttributeError:
+        except Exception:  # pylint: disable=broad-except
```
DanielNoord, line 77: I assume this can't be tested easily?

jacobtylerwalls: The entire brain is untested. I can't even install `PyGObject` on any of my machines with `pip`.

DanielNoord: 😅 

Seems like this is a likely candidate for a recent discussion we had about moving non-standard `brains` out of the main `astroid` repository into plugins..
But as long as it works it works I guess.

jacobtylerwalls: I agree.

senier:
Thanks for the quick response! I can confirm that this fixes my issue (tried with `pylint==2.13.5` and `astroid==2.12.0.dev0`, i.e. the above branch).

PR #1520, opened 2022-04-20
Fix ``col_offset`` attribute for nodes involving ``with`` on ``PyPy``

DanielNoord:
## Steps

## Description

For some reason these are incorrect on ``PyPy``...

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,7 +13,7 @@ What's New in astroid 2.11.4?
 =============================
 Release date: TBA
 
-
+* Fix ``col_offset`` attribute for nodes involving ``with`` on ``PyPy``.
 
 What's New in astroid 2.11.3?
 =============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,7 @@ Release date: TBA
 
 * Fix ``col_offset`` attribute for nodes involving ``with`` on ``PyPy``.
 
+
 What's New in astroid 2.11.3?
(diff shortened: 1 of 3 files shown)


DanielNoord:
@Pierre-Sassoulas What do we think of postponing running `PyPy` `3.8` on `pylint` CI until after `2.14`?

There seems to be an issue with `PyPy` `3.6` and I don't really want to spend another couple of hours investigating it since we drop support in a couple of weeks.

I'd propose:
1. Release ``pylint`` ``2.14``
2. Drop support for `3.6` in `astroid`
3. Release `astroid` `2.x` with this fix in it.
4. Merge https://github.com/PyCQA/pylint/pull/6406

Pierre-Sassoulas:
Yes, let's drop python 3.6 first !

cdce8p:
@DanielNoord I just pushed some small changes which did fix PyPy on Python 3.6. I believe the issue was with function decorators and resulting different lineno. Turns out though, we can just reuse `position.col_offset` for that.

DanielNoord:
> @DanielNoord I just pushed some small changes which did fix PyPy on Python 3.6. I believe the issue was with function decorators and resulting different lineno. Turns out though, we can just reuse `position.col_offset` for that.

Thanks for taking the time to investigate this a little further. I gave up once I saw it was `3.6`, but this is indeed quite easy and elegant.

PR #1521, opened 2022-04-20
Let `AstroidManager.clear_cache` act on other caches

### LINKED ISSUE #792: Continuously increasing memory usage when pylint is run via its API

jacobtylerwalls:
## Description
Alternative to #1242. See https://github.com/PyCQA/astroid/pull/1242#pullrequestreview-946486121

## Type of Changes

## Related Issue
Resolves #792 (but let's get a benchmark and find out)


tests/unittest_scoped_nodes.py
```diff
@@ -1678,10 +1678,8 @@ def __init__(self):
                 "FinalClass",
                 "ClassB",
                 "MixinB",
-                "",
                 "ClassA",
                 "MixinA",
-                "",
```
jacobtylerwalls, line 1684: This was added in 629c92db2dc1b016f4bf47645c95c42e65fd3bd6

jacobtylerwalls: I'll try to dig in here. It's the call to `clear_inference_tip_cache` in the new test that is causing test isolation problems.

astroid/interpreter/objectmodel.py
```diff
@@ -100,6 +101,7 @@ def __get__(self, instance, cls=None):
     def __contains__(self, name):
         return name in self.attributes()
 
+    @lru_cache()  # noqa
```
DanielNoord, line 104: Are we sure we want to re-add caching here? @jacobtylerwalls Do you have any data that this saves much time?

astroid/manager.py
```diff
@@ -59,6 +61,16 @@ class AstroidManager:
     max_inferable_values: ClassVar[int] = 100
 
     def __init__(self):
+        # pylint: disable=import-outside-toplevel
```
DanielNoord, line 64: Are we sure this is unavoidable? This is likely to generate many cyclic imports headaches at some point..

astroid/modutils.py
```diff
@@ -138,21 +139,19 @@ def _handle_blacklist(blacklist, dirnames, filenames):
             filenames.remove(norecurs)
 
 
-_NORM_PATH_CACHE: Dict[str, str] = {}
+@lru_cache()
+def _cache_normalize_path_(path: str) -> str:
+    return _normalize_path(path)
 
 
 def _cache_normalize_path(path: str) -> str:
```
DanielNoord, line 147: Can't we cache this functions?
What does `_normalize_path(path)` do when `path` is `""`?

tests/unittest_scoped_nodes.py
DanielNoord: 👍

astroid/manager.py
jacobtylerwalls: I moved the imports even later, inside `clear_cache()`. But yes, this is because a cyclic import is unavoidable--those modules import manager.

astroid/interpreter/objectmodel.py
jacobtylerwalls: Yes. :D

When I add a print statement before the clearing of the three LRU caches, and I run:
```python
lint.Run(["pylint"], do_exit=False)
```
I get:
```python
CacheInfo(hits=149138, misses=43009, maxsize=128, currsize=128)  # LookupMixIn.lookup()
CacheInfo(hits=4082, misses=3, maxsize=128, currsize=3)  # _cache_normalize_path_
CacheInfo(hits=75230, misses=13, maxsize=128, currsize=13)  # ObjectModel.attributes()
```

So a realistic benchmark for ObjectModel.attributes would seems to involve ... 7500 hits? Let's try it.

```python
from timeit import timeit
from astroid.interpreter.objectmodel import FunctionModel
fm = FunctionModel()
timeit(fm.attributes, globals=globals(), number=7500)
```
with caching
0.000958
without
0.210987

astroid/modutils.py
jacobtylerwalls: It returns current working directory. (That doesn't seem appropriate to cache, right?)

DanielNoord: It's `os.path.normcase(os.path.realpath(path))` right? Would that ever change during a `astroid` or `pylint` run?

astroid/interpreter/objectmodel.py
DanielNoord: I tested this with `pylint astroid`.

With `lru_cache` I get:
`CacheInfo(hits=77478, misses=26368, maxsize=128, currsize=128)`
With `maxsize=None` I get:
`CacheInfo(hits=83447, misses=20398, maxsize=None, currsize=20398)`

That's not a lot of hits... But it might be worth increasing it a little?

Pierre-Sassoulas: How does a 256 cache size fair ? Otherwise, we're already covering 92% of the possible hits with a cache size of 128, it seems we've reached a nice compromise between speed and low memory footprint.

astroid/manager.py
```diff
@@ -357,6 +358,21 @@ def bootstrap(self):
         raw_building._astroid_bootstrapping()
 
     def clear_cache(self):
-        """Clear the underlying cache. Also bootstraps the builtins module."""
+        """Clear the underlying caches. Also bootstraps the builtins module."""
+        # pylint: disable=import-outside-toplevel
```
Pierre-Sassoulas, line 362: Is it because of cyclic import ? We could state so 
```suggestion
        # pylint: disable=import-outside-toplevel
        # import here because of cyclic imports
```

astroid/modutils.py
```diff
@@ -138,21 +139,19 @@ def _handle_blacklist(blacklist, dirnames, filenames):
             filenames.remove(norecurs)
 
 
-_NORM_PATH_CACHE: Dict[str, str] = {}
+@lru_cache()
+def _cache_normalize_path_(path: str) -> str:
```
Pierre-Sassoulas, line 143: 👌

astroid/modutils.py
jacobtylerwalls: I think we need to imagine long-running processes and IDEs. I'd say let's not introduce a behavior change on a maintenance PR.

astroid/interpreter/objectmodel.py
jacobtylerwalls: I think we can leave it alone, especially because it's an instance method.

tests/unittest_scoped_nodes.py
jacobtylerwalls: Fix is in #1528 (blocked by another issue)

astroid/manager.py
```diff
@@ -365,8 +366,24 @@ def bootstrap(self):
     def clear_cache(self) -> None:
         """Clear the underlying cache, bootstrap the builtins module and
         re-register transforms."""
+        # import here because of cyclic imports
+        # pylint: disable=import-outside-toplevel
+        from astroid.inference_tip import clear_inference_tip_cache
```
DanielNoord, line 371: I'm still hesitant about this because of all the headaches the cyclic imports are giving us here, but I guess it is better than the status quo?

Pierre-Sassoulas: I'm under the impression that if we fix the other cyclic imports issue, then this one will also be resolved. But then I might be wrong and we could be imbricating another ouroboro in our big ball of ouroboros. It's hard to know until we get to the bottom of the nature of the existing cyclic imports.

jacobtylerwalls: I think this would be an alternative. Is this better or worse?
```python
diff --git a/astroid/__init__.py b/astroid/__init__.py
index 14a61c2a..cec22085 100644
--- a/astroid/__init__.py
+++ b/astroid/__init__.py
@@ -80,7 +80,7 @@ from astroid.exceptions import (
     UnresolvableName,
     UseInferenceDefault,
 )
-from astroid.inference_tip import _inference_tip_cached, inference_tip
+from astroid.inference_tip import _inference_tip_cached, inference_tip, clear_inference_tip_cache
 from astroid.objects import ExceptionInstance
 
 # isort: off
diff --git a/astroid/manager.py b/astroid/manager.py
index 30718e24..1ad5c48b 100644
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -12,6 +12,7 @@ import types
 import zipimport
 from typing import TYPE_CHECKING, ClassVar, List, Optional
 
+import astroid
 from astroid.exceptions import AstroidBuildingError, AstroidImportError
 from astroid.interpreter._import import spec
 from astroid.modutils import (
@@ -359,20 +360,14 @@ class AstroidManager:
 
     def clear_cache(self):
         """Clear the underlying caches. Also bootstraps the builtins module."""
-        # import here because of cyclic imports
-        # pylint: disable=import-outside-toplevel
-        from astroid.inference_tip import clear_inference_tip_cache
-        from astroid.interpreter.objectmodel import ObjectModel
-        from astroid.nodes.node_classes import LookupMixIn
-
-        clear_inference_tip_cache()
+        astroid.clear_inference_tip_cache()
 
         self.astroid_cache.clear()
 
         for lru_cache in (
-            LookupMixIn.lookup,
+            astroid.node_classes.LookupMixIn.lookup,
             _cache_normalize_path_,
-            ObjectModel.attributes,
+            astroid.interpreter.objectmodel.ObjectModel.attributes,
         ):
             lru_cache.cache_clear()
```

DanielNoord: Hm, probably better as is. This is just hiding the cyclic import.

jacobtylerwalls: I guess the other alternative is to move the barebones of the cached routines to a new file. But not sure there's much upside. I think there's greater ROI elsewhere.

DanielNoord: Agreed!

DanielNoord (approved):

astroid/modutils.py
```diff
@@ -138,21 +139,19 @@ def _handle_blacklist(blacklist, dirnames, filenames):
...
     # assembling path components. This cache alleviates that.
-    try:
-        return _NORM_PATH_CACHE[path]
-    except KeyError:
-        if not path:  # don't cache result for ''
-            return _normalize_path(path)
-        result = _NORM_PATH_CACHE[path] = _normalize_path(path)
-        return result
+    if not path:  # don't cache result for ''
+        return _normalize_path(path)
```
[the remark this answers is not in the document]
jacobtylerwalls: I added coverage. We talked a bit about it [here](https://github.com/PyCQA/astroid/pull/1521/files/4427b71cba2539916f0e637fda4564978338055c#r855171105).

jacobtylerwalls:
Ah, another test isolation issue to look into.

tests/unittest_modutils.py
```diff
@@ -179,7 +179,7 @@ def test_load_packages_without_init(self) -> None:
         https://github.com/PyCQA/astroid/issues/1327
         """
         tmp_dir = Path(tempfile.gettempdir())
-        self.addCleanup(os.chdir, os.curdir)
+        self.addCleanup(os.chdir, os.getcwd())
```
jacobtylerwalls, line 182: We need to actually find the cwd before changing dir, otherwise this cleanup won't do anything.

Pierre-Sassoulas (approved):
👍

PR #1526, opened 2022-04-22
Add annotations for `ParserModule`

cdce8p:
## Description
Replace `collections.namedtuple` with `typing.NamedTuple`.


Pierre-Sassoulas (approved):
I like the perf improvement hidden in the typing. I hope we could activate mypy in CI at some point because having a CI checks the changes would be reassuring.

PR #1528, opened 2022-04-23
Let `AstroidManager.clear_cache` reload brain plugins

jacobtylerwalls:
## Description
Fixes the incompatible tests in #1521, ~but unfortunately exposes a new problem.~ Fixed.
## Type of Changes

### DIFF
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -362,7 +362,8 @@ class AstroidManager:
         raw_building._astroid_bootstrapping()
 
     def clear_cache(self):
-        """Clear the underlying cache. Also bootstraps the builtins module."""
+        """Clear the underlying cache. Also bootstraps the builtins module and
+        re-registers transforms."""
         # import here because of cyclic imports
         # pylint: disable=import-outside-toplevel
 
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -10,6 +10,7 @@ from various source and using a cache of built modules)
 import os
 import types
 import zipimport
+from importlib.util import find_spec, module_from_spec
 from typing import TYPE_CHECKING, ClassVar, List, Optional
 
 from astroid.exceptions import AstroidBuildingError, AstroidImportError
@@ -362,5 +363,17 @@ class AstroidManager:
 
     def clear_cache(self):
         """Clear the underlying cache. Also bootstraps the builtins module."""
+        # import here because of cyclic imports
+        # pylint: disable=import-outside-toplevel
+
+        from astroid import BRAIN_MODULES_DIRECTORY
+
         self.astroid_cache.clear()
         self.bootstrap()
+
+        # Load brain plugins: currently done in astroid.__init__.py
+        for module in BRAIN_MODULES_DIRECTORY.iterdir():
+            if module.suffix == ".py":
+                module_spec = find_spec(f"astroid.brain.{module.stem}")
+                module_object = module_from_spec(module_spec)
+                module_spec.loader.exec_module(module_object)
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -369,6 +369,7 @@ class AstroidManager:
         from astroid import BRAIN_MODULES_DIRECTORY
 
         self.astroid_cache.clear()
+        AstroidManager.brain["_transform"] = TransformVisitor()
         self.bootstrap()
 
         # Load brain plugins: currently done in astroid.__init__.py
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -13,6 +13,7 @@ import zipimport
 from importlib.util import find_spec, module_from_spec
 from typing import TYPE_CHECKING, ClassVar, List, Optional
 
+from astroid.const import BRAIN_MODULES_DIRECTORY
 from astroid.exceptions import AstroidBuildingError, AstroidImportError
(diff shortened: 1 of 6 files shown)


astroid/manager.py
```diff
@@ -361,6 +362,20 @@ def bootstrap(self):
         raw_building._astroid_bootstrapping()
 
     def clear_cache(self):
-        """Clear the underlying cache. Also bootstraps the builtins module."""
+        """Clear the underlying cache. Also bootstraps the builtins module and
+        re-registers transforms."""
+        # import here because of cyclic imports
+        # pylint: disable=import-outside-toplevel
+
+        from astroid import BRAIN_MODULES_DIRECTORY
```
DanielNoord, line 370: This should probably go in a `const.py` that would fix the circular dependency.

astroid/manager.py
```diff
@@ -361,6 +363,15 @@ def bootstrap(self):
...
     def clear_cache(self):
-        """Clear the underlying cache. Also bootstraps the builtins module."""
+        """Clear the underlying cache. Also bootstraps the builtins module and
+        re-registers transforms."""
         self.astroid_cache.clear()
+        AstroidManager.brain["_transform"] = TransformVisitor()
         self.bootstrap()
+
+        # Load brain plugins: currently done in astroid.__init__.py
+        for module in BRAIN_MODULES_DIRECTORY.iterdir():
```
DanielNoord, line 373: Should we make this a function?

jacobtylerwalls: I wondered that also. I don't know that we need to expose it publicly. I don't know that we would be able to reuse it in `__init__.py` because I suspect that needs to stay as `import_module` rather than `exec_module`.

astroid/manager.py
```diff
@@ -361,6 +363,15 @@ def bootstrap(self):
         raw_building._astroid_bootstrapping()
 
     def clear_cache(self):
```
DanielNoord, line 365:
```suggestion
    def clear_cache(self) -> None:
```

astroid/manager.py
```diff
@@ -361,6 +363,15 @@ def bootstrap(self):
...
-        """Clear the underlying cache. Also bootstraps the builtins module."""
+        """Clear the underlying cache. Also bootstraps the builtins module and
+        re-registers transforms."""
```
DanielNoord, line 367:
```suggestion
        """Clear the underlying cache, bootstrap the builtins module and
        re-register transforms."""
```

astroid/manager.py
```diff
@@ -361,6 +363,15 @@ def bootstrap(self):
...
         self.astroid_cache.clear()
+        AstroidManager.brain["_transform"] = TransformVisitor()
         self.bootstrap()
+
+        # Load brain plugins: currently done in astroid.__init__.py
```
DanielNoord, line 372:
```suggestion
        # Reload brain plugins. During initialisation this is done in astroid.__init__.py
```

tests/resources.py
```diff
@@ -52,8 +52,11 @@ class AstroidCacheSetupMixin:
     @classmethod
     def setup_class(cls):
         cls._builtins = AstroidManager().astroid_cache.get("builtins")
+        cls._transforms = AstroidManager.brain["_transform"]
 
     @classmethod
     def teardown_class(cls):
         if cls._builtins:
             AstroidManager().astroid_cache["builtins"] = cls._builtins
+        if cls._transforms:
```
DanielNoord, line 61: Is that `if` necessary? `setup_class` is always called right?

tests/unittest_manager.py
```diff
@@ -315,5 +316,13 @@ def test_borg(self) -> None:
         self.assertIs(built, second_built)
 
 
+class ClearCacheTest(unittest.TestCase, resources.AstroidCacheSetupMixin):
```
DanielNoord, line 319: Can we make this `pytest`-based? If it is too much of a hassle never mind!

DanielNoord (changes requested):
Just some small nitpicks 👍

tests/unittest_manager.py
jacobtylerwalls: Yeah, I guess I'd kind of prefer not to.

tests/resources.py
```diff
@@ -52,8 +52,9 @@ class AstroidCacheSetupMixin:
     @classmethod
     def setup_class(cls):
         cls._builtins = AstroidManager().astroid_cache.get("builtins")
+        cls._transforms = AstroidManager.brain["_transform"]
```
DanielNoord, line 55: Why is the previous line instantiating an `AstroidManager`? Should that be done here as well?

jacobtylerwalls: Because it's accessing the `astroid_cache`, not the `brain`. (I know...)

DanielNoord: Ah, never mind. I guess things are never simple with `astroid`...

DanielNoord (approved):

tests/resources.py
jacobtylerwalls: I guess I could be rewritten to use `AstroidManager()._transform`. Doesn't feel any clearer, but would be more consistent? 🤔

DanielNoord: I think this is fine!

PR #1529, opened 2022-04-25
Improve recognition of ``_io`` module during bootstrapping on ``PyPy``

DanielNoord:
## Steps

## Description

This was added in https://github.com/PyCQA/astroid/commit/9d090ce96e3f3177708d792613fd66f343309bc0.
However, with this change the test still passes. We have likely mitigated the circular dependency somewhere else. This line did make it so that we did not infer `builtins` defined in `_io` and `io`.
Test show that we now infer this correctly 😄 

I believe this is also the last remaining issue for https://github.com/PyCQA/pylint/pull/6406.
It will also make it possible to run some additional tests on `PyPy` in `pylint.

## Type of Changes

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -26,17 +26,6 @@ TYPE_NOTIMPLEMENTED = type(NotImplemented)
 TYPE_ELLIPSIS = type(...)
 
 
-def _io_discrepancy(member):
-    # _io module names itself `io`: http://bugs.python.org/issue18602
-    member_self = getattr(member, "__self__", None)
-    return (
-        member_self
-        and inspect.ismodule(member_self)
-        and member_self.__name__ == "_io"
-        and member.__module__ == "io"
-    )
-
-
 def _attach_local_node(parent, node, name):
     node.name = name  # needed by add_local_node
     parent.add_local_node(node)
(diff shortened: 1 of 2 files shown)


astroid/raw_building.py
```diff
@@ -402,7 +389,13 @@ def imported_member(self, node, member, name):
...
                 return True
 
-        real_name = {"gtk": "gtk_gtk", "_io": "io"}.get(modname, modname)
+        # On PyPy during bootstrapping we infer _io while _module is
+        # builtins. In CPython _io names itself io, see http://bugs.python.org/issue18602
+        # Therefore, this basically checks whether we are not in PyPy.
+        if modname == "_io" and not self._module.__name__ == "builtins":
+            return False
+
+        real_name = {"gtk": "gtk_gtk"}.get(modname, modname)
```
Pierre-Sassoulas, line 398: I don't understand what this does. Is it it equivalent to this ?
```suggestion
        real_name = "gtk_gtk" if modname == "gtk" else modname
```

DanielNoord: Yes, both work. There doesn't actually seem to be a test for this so I just tried to keep the code as similar to the previous iteration as possible.

Pierre-Sassoulas:
```suggestion
        MODNAME_ALIASES = {"gtk": "gtk_gtk"} # Actually defined outside of the function scope ?
        real_name = MODNAME_ALIASES .get(modname, modname)
```
Would help with comprehension I guess.

DanielNoord: We could then also use your earlier suggestion? If we're going to change stuff anyway. Let me know what has you preference (or just commit directly)

Pierre-Sassoulas: It depends on how likely we are to have another modname aliases, it's probably a good thing if we can extend to two aliases again without too much refactor. Let's keep the current code.

Pierre-Sassoulas (approved):

PR #1531, opened 2022-04-26
Make `FunctionDef.implicit_parameters` return 1 for methods

jacobtylerwalls:
## Description

When a node is inferred as `FunctionDef` instead of `BoundMethod`, `implicit_parameters()` always returns 0, which can cause problems in `pylint` with `no-value-for-parameter` and such. We can at least fall back to the knowledge that it is bound and return 1.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#6464 (because we will not be writing Qt tests there)

- [x] Todo: see if this closes any other pylint issues for `no-value-for-parameter`

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1645,7 +1645,7 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
             False otherwise.
         :rtype: bool
         """
-        return self.type in ("method", "classmethod")
+        return self.type in {"method", "classmethod"}
 
     def is_abstract(self, pass_is_abstract=True, any_raise_is_abstract=False):
         """Check if the method is abstract.
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1581,6 +1581,9 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
         """
         return self.args.tolineno
 
+    def implicit_parameters(self):
+        return 1 if self.is_bound() else 0
+
     def block_range(self, lineno):
         """Get a range from the given line number to where this node ends.
 
@@ -1642,7 +1645,7 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
             False otherwise.
         :rtype: bool
         """
-        return self.type == "classmethod"
+        return self.type in ("method", "classmethod")
 
     def is_abstract(self, pass_is_abstract=True, any_raise_is_abstract=False):
         """Check if the method is abstract.
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1581,7 +1581,7 @@ class FunctionDef(mixins.MultiLineBlockMixin, node_classes.Statement, Lambda):
         """
         return self.args.tolineno
 
-    def implicit_parameters(self):
+    def implicit_parameters(self) -> Literal[0, 1]:
         return 1 if self.is_bound() else 0
 
     def block_range(self, lineno):
--- a/tests/unittest_brain_qt.py
+++ b/tests/unittest_brain_qt.py
@@ -16,6 +16,8 @@ HAS_PYQT6 = find_spec("PyQt6")
 
 @pytest.mark.skipif(HAS_PYQT6 is None, reason="This test requires the PyQt6 library.")
 class TestBrainQt:
+    AstroidManager.brain["extension_package_whitelist"] = {"PyQt6.QtPrintSupport"}
+
     @staticmethod
     def test_value_of_lambda_instance_attrs_is_list():
         """Regression test for https://github.com/PyCQA/pylint/issues/6221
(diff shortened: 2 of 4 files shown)


jacobtylerwalls:
@The-Compiler does this resolve the false positive for you?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1642,7 +1645,7 @@ def is_bound(self):
             False otherwise.
         :rtype: bool
         """
-        return self.type == "classmethod"
+        return self.type in {"method", "classmethod"}
```
jacobtylerwalls, line 1648: This method was untested according to coveralls.

Pierre-Sassoulas (dismissed):
LGTM !

tests/unittest_brain_qt.py
```diff
@@ -28,11 +30,25 @@ def test_value_of_lambda_instance_attrs_is_list():
...
+        """Regression test for https://github.com/PyCQA/pylint/issues/6464"""
+        src = """
+        from PyQt6.QtCore import QTimer
+        timer = QTimer()
+        timer.timeout.connect  #@
+        """
+        node = extract_node(src)
+        attribute_node = node.inferred()[0]
+        if attribute_node is Uninferable:
+            pytest.skip("PyQt6 C bindings may not be installed?")
```
Pierre-Sassoulas, line 52: Don't we want to force the c binding to work correctly in at least one of the CI pipeline ? Otherwise this test can work fine online and fail for someone with QT locally. I think we could create a pytest mark for "qt" and a CI job with a proper installation.

Pierre-Sassoulas: Hmm, it means we're going to catch issues at release time right ? When/how are we launching this workflow ?

jacobtylerwalls: Just manually with a reminder in the release.md.

Installing `qt` [takes 39s](https://github.com/PyCQA/astroid/runs/6024015453?check_suite_focus=true). Unless there's a way to cache it?

Pierre-Sassoulas: For 40s we might as well add it for each commit (one interpreter only ?). There's a lot of available runner for astroid and they run in a reasonable time (compared to pylint anyway).

jacobtylerwalls: Actually, it takes 8s, duh, because most of that linked run was setting up the env!

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1581,6 +1581,9 @@ def blockstart_tolineno(self):
         """
         return self.args.tolineno
 
+    def implicit_parameters(self):
```
DanielNoord, line 1584:
```suggestion
    def implicit_parameters(self):
```

Could you add typing with `Literal` here?

tests/unittest_brain_qt.py
```diff
@@ -28,11 +30,25 @@ def test_value_of_lambda_instance_attrs_is_list():
...
         node = extract_node(src)
         attribute_node = node.inferred()[0]
         if attribute_node is Uninferable:
             pytest.skip("PyQt6 C bindings may not be installed?")
         assert isinstance(attribute_node, UnboundMethod)
         # scoped_nodes.Lambda.instance_attrs is typed as Dict[str, List[NodeNG]]
         assert isinstance(attribute_node.instance_attrs["connect"][0], FunctionDef)
+
+    @staticmethod
+    def test_implicit_parameters():
```
DanielNoord, line 42:
```suggestion
    def test_implicit_parameters() -> None:
```

The-Compiler:
> @The-Compiler does this resolve the false positive for you?

Almost! It still complains on two lines compared to 2.13.7:

https://github.com/qutebrowser/qutebrowser/blob/6fc30a1e851dbc07461ba50d4266bbcae40e1fb1/qutebrowser/misc/throttle.py#L88

and

https://github.com/qutebrowser/qutebrowser/blob/6fc30a1e851dbc07461ba50d4266bbcae40e1fb1/tests/unit/config/test_configtypes.py#L226

The former sounds like pylint/astroid thinks a slot is needed for `.disconnect()`, but that's not the case, it can be called without one.

The latter gives me `too-many-function-args` and is probably an unrelated false-positive regression.

tests/unittest_brain_qt.py
The-Compiler: I'm confused - why does "Install Qt" actually only do `sudo apt-get install build-essential libgl1-mesa-dev`? You are installing PyQt from binary wheels, so that 40s step can probably just be dropped entirely.

jacobtylerwalls: We discovered that the `PyQt` wheel doesn't necessarily install the C bindings on `ubuntu-latest`, see [failing run ](https://github.com/PyCQA/astroid/runs/5893185699?check_suite_focus=true) and [comment](https://github.com/PyCQA/astroid/pull/1505#discussion_r846644697).

jacobtylerwalls:
Thanks for testing. Feel free to open new issues or PRs. Apparently `slot` is optional, but you cannot pass an explicit `None`, see https://github.com/python-qt-tools/PyQt5-stubs/pull/103. I'm a little hesitant to show that `slot=None` is the default value in that case. See also PyCQA/pylint#5264 for general weakness in this area.

tests/unittest_brain_qt.py
jacobtylerwalls: Happy to add it, but when I've seen this in the test suite, it's struck me as oddly specific. Why would a test method ever return anything? 😄

DanielNoord: I just like to add it as I think there can be a benefit to eventually turning on `mypy` even over the test suite. I think `mypy` actually caught some bugs for me in the `pylint` test suite already. Having the signature typed makes `mypy` actually check the function.

jacobtylerwalls: > Having the signature typed makes mypy actually check the function.

Ah, thank you, this was the part that didn't come to mind!

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #1534, opened 2022-04-28
Fix a crash when `_filter_stmts` encounters an `EmptyNode`

jacobtylerwalls:
## Description
An assertion was added in 864d2f31b0d30a25d5986cb8682c5500454da6c8 "to be removed later". `EmptyNode` was failing the condition, see also related issue PyCQA/pylint#2535. This PR removes the assertion and short-circuits.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#6438. Added a regression test without Qt.

### DIFF
--- a/tests/unittest_filter_statements.py
+++ b/tests/unittest_filter_statements.py
@@ -1,0 +1,20 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
+
+import unittest
+
+from astroid.builder import extract_node
+from astroid.filter_statements import _filter_stmts
+from astroid.nodes import EmptyNode
+
+
+class FilterStatementsTest(unittest.TestCase):
+    def test_empty_node(self) -> None:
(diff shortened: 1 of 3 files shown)


tests/unittest_filter_statements.py
```diff
@@ -0,0 +1,20 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
+
+import unittest
```
DanielNoord, line 5: I'd prefer moving `astroid` to `pytest` as well so I tend to change test to `pytest` when I encounter them. However, not sure what the "official policy" is?

jacobtylerwalls: Makes sense if `pytest` is being used for test discovery anyway. unittest test discovery currently doesn't work.

DanielNoord (approved):

tests/unittest_filter_statements.py
jacobtylerwalls: Perhaps. I do kind of worry about the contribution barrier if we no longer accept tests written in unittest. There are a number of things I know how to do in unittest that I don't in pytest because I don't work with it ever.

Pierre-Sassoulas (approved):

PR #1535, opened 2022-04-30
Fix a crash involving two starred expressions inside a call

jacobtylerwalls:
## Steps

## Description
Before, there was an assumption that given an assign statement with a starred node on the right-hand side, there will be a container on the left-hand side. When the starred node is inside a call, that assumption is violated.

## Type of Changes

## Related Issue

Refs PyCQA/pylint#6372

### DIFF
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -690,6 +690,9 @@ def starred_assigned_stmts(
     if isinstance(stmt, nodes.Assign):
         value = stmt.value
         lhs = stmt.targets[0]
+        if not hasattr(lhs, "elts"):
+            yield util.Uninferable
+            return
 
         if sum(1 for _ in lhs.nodes_of_class(nodes.Starred)) > 1:
             raise InferenceError(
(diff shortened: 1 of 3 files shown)


astroid/protocols.py
```diff
@@ -690,6 +690,9 @@ def _determine_starred_iteration_lookups(starred, target, lookups):
     if isinstance(stmt, nodes.Assign):
         value = stmt.value
         lhs = stmt.targets[0]
+        if not hasattr(lhs, "elts"):
+            yield util.Uninferable
+            return
```
DanielNoord, line 695: The last line is uncovered.

astroid/protocols.py
```diff
@@ -690,6 +690,9 @@ def _determine_starred_iteration_lookups(starred, target, lookups):
     if isinstance(stmt, nodes.Assign):
         value = stmt.value
         lhs = stmt.targets[0]
+        if not hasattr(lhs, "elts"):
```
DanielNoord, line 693: I'd prefer using an `isinstance` check here as I think it is more explicit, but I'm not sure how easy it is to come up with a list of nodes that have `elts` defined.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #1536, opened 2022-05-01
Remove dependency on `pkg_resources` from `setuptools`

### LINKED ISSUE #1103: Remove dependency to ``setuptools`` and ``distutil``

### LINKED ISSUE #1320: Decrease the time required to import astroid

jacobtylerwalls:
## Steps

## Description
Remove dependence on `pkg_resources` from `setuptools`, which is discouraged, and which increases import time.

## Type of Changes

## Related Issue

Closes #1103
Closes #1320 
 - [unscientific benchmark]: 50% reduction of import time on an M1 mac, 67% reduction on a cloud console

### DIFF
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -2,8 +2,6 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
-from __future__ import annotations
-
 import sys
 from functools import lru_cache
 from importlib.util import _find_spec_from_path
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -5,8 +5,8 @@
 from __future__ import annotations
 
 import pathlib
+import sys
 from functools import lru_cache
-from importlib.machinery import BuiltinImporter
 from importlib.util import _find_spec_from_path
 
 
@@ -26,6 +26,9 @@ def _is_setuptools_namespace(location: pathlib.Path) -> bool:
 
 @lru_cache(maxsize=4096)
 def is_namespace(modname: str) -> bool:
+    if modname in sys.builtin_module_names:
+        return False
+
     found_spec = None
 
     # find_spec() attempts to import parent packages when given dotted paths.
@@ -45,10 +48,6 @@ def is_namespace(modname: str) -> bool:
         last_parent = working_modname
 
     if found_spec is None:
-        return False
-
-    if found_spec.loader is BuiltinImporter:
-        # TODO(Py39): remove, since found_spec.origin will be "built-in"
         return False
 
     return found_spec.origin is None
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -2,7 +2,23 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
+import os
+import pathlib
 from importlib.util import _find_spec_from_path
+
+
+def _is_setuptools_namespace(location):
+    try:
+        with open(os.path.join(location, "__init__.py"), "rb") as stream:
+            data = stream.read(4096)
+    except OSError:
+        return None
+    else:
+        extend_path = b"pkgutil" in data and b"extend_path" in data
+        declare_namespace = (
+            b"pkg_resources" in data and b"declare_namespace(__name__)" in data
+        )
+        return extend_path or declare_namespace
 
 
 def is_namespace(modname: str) -> bool:
@@ -29,5 +45,14 @@ def is_namespace(modname: str) -> bool:
 
     if found_spec is None:
         return False
+
+    if found_spec.submodule_search_locations and any(
+        _is_setuptools_namespace(directory)
+        for directory in pathlib.Path(
+            found_spec.submodule_search_locations._path[0]
+        ).iterdir()
+        if directory.is_dir()
+    ):
+        return True
 
     return found_spec.origin == "namespace"
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -31,14 +31,11 @@ def is_namespace(modname: str) -> bool:
     # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
     # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
     # be specced from left to right.
-    working_modname = ""
+    processed_components = []
     last_parent = None
     for component in modname.split("."):
-        if working_modname:
-            working_modname += "." + component
-        else:
-            # First component
-            working_modname = component
+        processed_components.append(component)
+        working_modname = ".".join(processed_components)
         try:
             found_spec = _find_spec_from_path(working_modname, last_parent)
         except ValueError:
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -4,15 +4,14 @@
 
 from __future__ import annotations
 
-import os
 import pathlib
 from functools import lru_cache
 from importlib.util import _find_spec_from_path
 
 
-def _is_setuptools_namespace(location: str | pathlib.Path) -> bool:
+def _is_setuptools_namespace(location: pathlib.Path) -> bool:
     try:
-        with open(os.path.join(location, "__init__.py"), "rb") as stream:
+        with open(location / "__init__.py", "rb") as stream:
             data = stream.read(4096)
     except OSError:
         return False
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -2,15 +2,34 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
-from importlib.util import find_spec
+from importlib.util import _find_spec_from_path
 
 
-def is_namespace(modname):
-    try:
-        found_spec = find_spec(modname)
-    except ValueError:
-        # execution of .pth files may not have __spec__
-        return True
+def is_namespace(modname: str) -> bool:
+    found_spec = None
+
+    # find_spec() attempts to import parent packages when given dotted paths.
+    # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
+    # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
+    # be specced from left to right.
+    working_modname = ""
+    last_parent = None
+    for component in modname.split("."):
+        if working_modname:
+            working_modname += "." + component
+        else:
+            # First component
+            working_modname = component
+        try:
+            found_spec = _find_spec_from_path(working_modname, last_parent)
+        except (
+            AttributeError,  # TODO: remove AttributeError when 3.7+ is min
+            ValueError,
+        ):
+            # executed .pth files may not have __spec__
+            return True
+        last_parent = working_modname
+
     if found_spec is None:
         return
     # origin can be either a string on older Python versions
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -2,13 +2,15 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
+from __future__ import annotations
+
 import os
 import pathlib
 from functools import lru_cache
 from importlib.util import _find_spec_from_path
 
 
-def _is_setuptools_namespace(location) -> bool:
+def _is_setuptools_namespace(location: str | pathlib.Path) -> bool:
     try:
         with open(os.path.join(location, "__init__.py"), "rb") as stream:
             data = stream.read(4096)
@@ -48,6 +50,9 @@ def is_namespace(modname: str) -> bool:
     if found_spec is None:
         return False
 
+    if found_spec.origin == "namespace":
+        return True
+
     if found_spec.submodule_search_locations is not None:
         for search_location in found_spec.submodule_search_locations:
             if any(
@@ -57,4 +62,4 @@ def is_namespace(modname: str) -> bool:
             ):
                 return True
 
-    return found_spec.origin == "namespace"
+    return False
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -4,6 +4,7 @@
 
 import os
 import pathlib
+from functools import lru_cache
 from importlib.util import _find_spec_from_path
 
 
@@ -21,6 +22,7 @@ def _is_setuptools_namespace(location) -> bool:
         return extend_path or declare_namespace
 
 
+@lru_cache(maxsize=4096)
 def is_namespace(modname: str) -> bool:
     found_spec = None
 
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -2,15 +2,18 @@
 # For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
 
-try:
-    import pkg_resources
-except ImportError:
-    pkg_resources = None  # type: ignore[assignment]
+from importlib.util import find_spec
 
 
 def is_namespace(modname):
-    return (
-        pkg_resources is not None
-        and hasattr(pkg_resources, "_namespace_packages")
-        and modname in pkg_resources._namespace_packages
-    )
+    try:
+        found_spec = find_spec(modname)
+    except ValueError:
+        # execution of .pth files may not have __spec__
+        return True
+    if found_spec is None:
+        return
+    # origin can be either a string on older Python versions
+    # or None in case it is a namespace package:
+    # https://github.com/python/cpython/pull/5481
+    return found_spec.origin in {None, "namespace"}
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -4,24 +4,9 @@
 
 from __future__ import annotations
 
-import pathlib
 import sys
 from functools import lru_cache
 from importlib.util import _find_spec_from_path
-
-
-def _is_setuptools_namespace(location: pathlib.Path) -> bool:
-    try:
-        with open(location / "__init__.py", "rb") as stream:
-            data = stream.read(4096)
-    except OSError:
-        return False
-    else:
-        extend_path = b"pkgutil" in data and b"extend_path" in data
-        declare_namespace = (
-            b"pkg_resources" in data and b"declare_namespace(__name__)" in data
-        )
-        return extend_path or declare_namespace
 
 
 @lru_cache(maxsize=4096)
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import pathlib
 from functools import lru_cache
+from importlib.machinery import BuiltinImporter
 from importlib.util import _find_spec_from_path
 
 
@@ -44,6 +45,10 @@ def is_namespace(modname: str) -> bool:
         last_parent = working_modname
 
     if found_spec is None:
+        return False
+
+    if found_spec.loader is BuiltinImporter:
+        # TODO(Py39): remove, since found_spec.origin will be "built-in"
         return False
 
     if found_spec.origin is None:
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -49,7 +49,7 @@ def is_namespace(modname: str) -> bool:
     if found_spec.submodule_search_locations and any(
         _is_setuptools_namespace(directory)
         for directory in pathlib.Path(
-            found_spec.submodule_search_locations._path[0]
+            found_spec.submodule_search_locations[0]
         ).iterdir()
         if directory.is_dir()
     ):
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -7,12 +7,12 @@ import pathlib
 from importlib.util import _find_spec_from_path
 
 
-def _is_setuptools_namespace(location):
+def _is_setuptools_namespace(location) -> bool:
     try:
         with open(os.path.join(location, "__init__.py"), "rb") as stream:
             data = stream.read(4096)
     except OSError:
-        return None
+        return False
     else:
         extend_path = b"pkgutil" in data and b"extend_path" in data
         declare_namespace = (
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -51,16 +51,4 @@ def is_namespace(modname: str) -> bool:
         # TODO(Py39): remove, since found_spec.origin will be "built-in"
         return False
 
-    if found_spec.origin is None:
-        return True
-
-    if found_spec.submodule_search_locations is not None:
-        for search_location in found_spec.submodule_search_locations:
-            if any(
-                _is_setuptools_namespace(directory)
-                for directory in pathlib.Path(search_location).iterdir()
-                if directory.is_dir()
-            ):
-                return True
-
-    return False
+    return found_spec.origin is None
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -31,7 +31,7 @@ def is_namespace(modname: str) -> bool:
         last_parent = working_modname
 
     if found_spec is None:
-        return
+        return False
     # origin can be either a string on older Python versions
     # or None in case it is a namespace package:
     # https://github.com/python/cpython/pull/5481
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -46,13 +46,13 @@ def is_namespace(modname: str) -> bool:
     if found_spec is None:
         return False
 
-    if found_spec.submodule_search_locations and any(
-        _is_setuptools_namespace(directory)
-        for directory in pathlib.Path(
-            found_spec.submodule_search_locations[0]
-        ).iterdir()
-        if directory.is_dir()
-    ):
-        return True
+    if found_spec.submodule_search_locations is not None:
+        for search_location in found_spec.submodule_search_locations:
+            if any(
+                _is_setuptools_namespace(directory)
+                for directory in pathlib.Path(search_location).iterdir()
+                if directory.is_dir()
+            ):
+                return True
 
     return found_spec.origin == "namespace"
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -46,7 +46,7 @@ def is_namespace(modname: str) -> bool:
     if found_spec is None:
         return False
 
-    if found_spec.origin == "namespace":
+    if found_spec.origin is None:
         return True
 
     if found_spec.submodule_search_locations is not None:
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -22,17 +22,12 @@ def is_namespace(modname: str) -> bool:
             working_modname = component
         try:
             found_spec = _find_spec_from_path(working_modname, last_parent)
-        except (
-            AttributeError,  # TODO: remove AttributeError when 3.7+ is min
-            ValueError,
-        ):
+        except ValueError:
             # executed .pth files may not have __spec__
             return True
         last_parent = working_modname
 
     if found_spec is None:
         return False
-    # origin can be either a string on older Python versions
-    # or None in case it is a namespace package:
-    # https://github.com/python/cpython/pull/5481
-    return found_spec.origin in {None, "namespace"}
+
+    return found_spec.origin == "namespace"
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -10,8 +10,6 @@ import unittest
 from contextlib import contextmanager
 from typing import Iterator
 
-import pkg_resources
-
 import astroid
 from astroid import manager, test_utils
 from astroid.const import IS_JYTHON
@@ -128,7 +126,6 @@ class AstroidManagerTest(
     def test_namespace_package_pth_support(self) -> None:
         pth = "foogle_fax-0.12.5-py2.7-nspkg.pth"
         site.addpackage(resources.RESOURCE_PATH, pth, [])
-        pkg_resources._namespace_packages["foogle"] = []
 
         try:
             module = self.manager.ast_from_module_name("foogle.fax")
@@ -138,18 +135,14 @@ class AstroidManagerTest(
             with self.assertRaises(AstroidImportError):
                 self.manager.ast_from_module_name("foogle.moogle")
         finally:
-            del pkg_resources._namespace_packages["foogle"]
             sys.modules.pop("foogle")
 
     def test_nested_namespace_import(self) -> None:
         pth = "foogle_fax-0.12.5-py2.7-nspkg.pth"
         site.addpackage(resources.RESOURCE_PATH, pth, [])
-        pkg_resources._namespace_packages["foogle"] = ["foogle.crank"]
-        pkg_resources._namespace_packages["foogle.crank"] = []
         try:
             self.manager.ast_from_module_name("foogle.crank")
         finally:
-            del pkg_resources._namespace_packages["foogle"]
             sys.modules.pop("foogle")
 
     def test_namespace_and_file_mismatch(self) -> None:
@@ -158,12 +151,10 @@ class AstroidManagerTest(
         self.assertEqual(ast.name, "unittest")
         pth = "foogle_fax-0.12.5-py2.7-nspkg.pth"
         site.addpackage(resources.RESOURCE_PATH, pth, [])
-        pkg_resources._namespace_packages["foogle"] = []
         try:
             with self.assertRaises(AstroidImportError):
                 self.manager.ast_from_module_name("unittest.foogle.fax")
         finally:
-            del pkg_resources._namespace_packages["foogle"]
             sys.modules.pop("foogle")
 
     def _test_ast_from_zip(self, archive: str) -> None:
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -109,6 +109,10 @@ class AstroidManagerTest(
         self._test_ast_from_old_namespace_package_protocol("pkg_resources")
 
     def test_identify_old_namespace_package_protocol(self) -> None:
+        # Like the above cases, this package follows the old namespace package protocol
+        # astroid currently assumes such packages are in sys.modules, so import it
+        import tests.testdata.python3.data.path_pkg_resources_1.package.foo  # noqa
+
         self.assertTrue(
             util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
         )
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -325,6 +325,7 @@ class ClearCacheTest(unittest.TestCase, resources.AstroidCacheSetupMixin):
         lrus = (
             astroid.nodes.node_classes.LookupMixIn.lookup,
             astroid.modutils._cache_normalize_path_,
+            util.is_namespace,
             astroid.interpreter.objectmodel.ObjectModel.attributes,
         )
 
@@ -334,6 +335,7 @@ class ClearCacheTest(unittest.TestCase, resources.AstroidCacheSetupMixin):
         # Generate some hits and misses
         ClassDef().lookup("garbage")
         is_standard_module("unittest", std_path=["garbage_path"])
+        util.is_namespace("unittest")
         astroid.interpreter.objectmodel.ObjectModel().attributes()
 
         # Did the hits or misses actually happen?
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -113,7 +113,8 @@ class AstroidManagerTest(
     def test_identify_old_namespace_package_protocol(self) -> None:
         # Like the above cases, this package follows the old namespace package protocol
         # astroid currently assumes such packages are in sys.modules, so import it
-        import tests.testdata.python3.data.path_pkg_resources_1.package.foo  # noqa
+        # pylint: disable-next=import-outside-toplevel
+        import tests.testdata.python3.data.path_pkg_resources_1.package.foo as _  # noqa
 
         self.assertTrue(
             util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -14,6 +14,7 @@ import astroid
 from astroid import manager, test_utils
 from astroid.const import IS_JYTHON
 from astroid.exceptions import AstroidBuildingError, AstroidImportError
+from astroid.interpreter._import import util
 from astroid.nodes import Const
 
 from . import resources
@@ -106,6 +107,11 @@ class AstroidManagerTest(
 
     def test_ast_from_namespace_pkg_resources(self) -> None:
         self._test_ast_from_old_namespace_package_protocol("pkg_resources")
+
+    def test_identify_old_namespace_package_protocol(self) -> None:
+        self.assertTrue(
+            util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
+        )
 
     def test_implicit_namespace_package(self) -> None:
         data_dir = os.path.dirname(resources.find("data/namespace_pep_420"))
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,10 @@ Release date: TBA
 
   Closes #1512
 
+* Remove dependency on ``pkg_resources`` from ``setuptools``.
+
+  Closes #1103
+
 * Rename ``ModuleSpec`` -> ``module_type`` constructor parameter to match attribute
   name and improve typing. Use ``type`` instead.
 
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -147,7 +147,7 @@ class ImportlibFinder(Finder):
             # Builtin.
             return None
 
-        if util._is_setuptools_namespace(spec.location):
+        if util._is_setuptools_namespace(Path(spec.location)):
             # extend_path is called, search sys.path for module/packages
             # of this name see pkgutil.extend_path documentation
             path = [
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -10,6 +10,7 @@ import importlib
 import importlib.machinery
 import importlib.util
 import os
+import pathlib
 import sys
 import zipimport
 from collections.abc import Sequence
@@ -147,7 +148,7 @@ class ImportlibFinder(Finder):
             # Builtin.
             return None
 
-        if util._is_setuptools_namespace(Path(spec.location)):
+        if _is_setuptools_namespace(Path(spec.location)):
             # extend_path is called, search sys.path for module/packages
             # of this name see pkgutil.extend_path documentation
             path = [
@@ -254,6 +255,20 @@ _SPEC_FINDERS = (
     PathSpecFinder,
     ExplicitNamespacePackageFinder,
 )
+
+
+def _is_setuptools_namespace(location: pathlib.Path) -> bool:
+    try:
+        with open(location / "__init__.py", "rb") as stream:
+            data = stream.read(4096)
+    except OSError:
+        return False
+    else:
+        extend_path = b"pkgutil" in data and b"extend_path" in data
+        declare_namespace = (
+            b"pkg_resources" in data and b"declare_namespace(__name__)" in data
+        )
+        return extend_path or declare_namespace
 
 
 @lru_cache()
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -171,7 +171,7 @@ class ImportlibFinder(Finder):
 
 
 class ExplicitNamespacePackageFinder(ImportlibFinder):
-    """A finder for the explicit namespace packages, generated through pkg_resources."""
+    """A finder for the explicit namespace packages."""
 
     def find_module(self, modname, module_parts, processed, submodule_path):
         if processed:
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -144,7 +144,7 @@ class ImportlibFinder(Finder):
             # Builtin.
             return None
 
-        if _is_setuptools_namespace(spec.location):
+        if util._is_setuptools_namespace(spec.location):
             # extend_path is called, search sys.path for module/packages
             # of this name see pkgutil.extend_path documentation
             path = [
@@ -251,20 +251,6 @@ _SPEC_FINDERS = (
     PathSpecFinder,
     ExplicitNamespacePackageFinder,
 )
-
-
-def _is_setuptools_namespace(location):
-    try:
-        with open(os.path.join(location, "__init__.py"), "rb") as stream:
-            data = stream.read(4096)
-    except OSError:
-        return None
-    else:
-        extend_path = b"pkgutil" in data and b"extend_path" in data
-        declare_namespace = (
-            b"pkg_resources" in data and b"declare_namespace(__name__)" in data
-        )
-        return extend_path or declare_namespace
 
 
 @lru_cache()
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -17,7 +17,7 @@ from typing import TYPE_CHECKING, ClassVar
 
 from astroid.const import BRAIN_MODULES_DIRECTORY
 from astroid.exceptions import AstroidBuildingError, AstroidImportError
-from astroid.interpreter._import import spec
+from astroid.interpreter._import import spec, util
 from astroid.modutils import (
     NoSourceFile,
     _cache_normalize_path_,
@@ -382,6 +382,7 @@ class AstroidManager:
         for lru_cache in (
             LookupMixIn.lookup,
             _cache_normalize_path_,
+            util.is_namespace,
             ObjectModel.attributes,
         ):
             lru_cache.cache_clear()
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,6 @@ packages = find:
 install_requires =
     lazy_object_proxy>=1.4.0
     wrapt>=1.11,<2
-    setuptools>=20.0
     typed-ast>=1.4.0,<2.0;implementation_name=="cpython" and python_version<"3.8"
     typing-extensions>=3.10;python_version<"3.10"
 python_requires = >=3.6.2


DanielNoord:
Haven't looked at the code, but please see https://github.com/PyCQA/astroid/pull/1326 and it's reversal in https://github.com/PyCQA/astroid/pull/1409.

I don't have the exact reasons on the top of my head, but we should be careful here as this already regressed once.

jacobtylerwalls:
> the exact reasons

Thanks for the heads up! Looks like it was because parent packages were actually being imported. I believe I worked around that (I see one fewer failure in the pylint test suite that leads me to that belief).

There is still this set of failures in `pylint`, though:

<details>
________________________________________________________________________________________________________ TestRunTC.test_regression_parallel_mode_without_filepath _________________________________________________________________________________________________________

self = <test_self.TestRunTC object at 0x104894c10>

    def test_regression_parallel_mode_without_filepath(self) -> None:
        # Test that parallel mode properly passes filepath
        # https://github.com/PyCQA/pylint/issues/3564
        path = join(
            HERE, "regrtest_data", "regression_missing_init_3564", "subdirectory/"
        )
>       self._test_output([path, "-j2"], expected_output="No such file or directory")

tests/test_self.py:1063: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_self.TestRunTC object at 0x104894c10>, args = ['--rcfile=/Users/jacobwalls/pylint/pylint/testutils/testing_pylintrc', '/Users/jacobwalls/pylint/tests/regrtest_data/regression_missing_init_3564/subdirectory/', '-j2']
expected_output = 'No such file or directory'

    def _test_output(self, args: list[str], expected_output: str) -> None:
        out = StringIO()
        args = _add_rcfile_default_pylintrc(args)
        self._run_pylint(args, out=out)
        actual_output = self._clean_paths(out.getvalue())
        expected_output = self._clean_paths(expected_output)
>       assert expected_output.strip() in actual_output.strip()
E       AssertionError: assert 'No such file or directory' in ''
E        +  where 'No such file or directory' = <built-in method strip of str object at 0x1048569c0>()
E        +    where <built-in method strip of str object at 0x1048569c0> = 'No such file or directory'.strip
E        +  and   '' = <built-in method strip of str object at 0x101478030>()
E        +    where <built-in method strip of str object at 0x101478030> = ''.strip

tests/test_self.py:169: AssertionError
___________________________________________________________________________________________________________________ TestRunTC.test_regression_recursive ___________________________________________________________________________________________________________________

self = <test_self.TestRunTC object at 0x1048963e0>

    def test_regression_recursive(self):
>       self._test_output(
            [join(HERE, "regrtest_data", "directory", "subdirectory"), "--recursive=n"],
            expected_output="No such file or directory",
        )

tests/test_self.py:1232: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_self.TestRunTC object at 0x1048963e0>, args = ['--rcfile=/Users/jacobwalls/pylint/pylint/testutils/testing_pylintrc', '/Users/jacobwalls/pylint/tests/regrtest_data/directory/subdirectory', '--recursive=n'], expected_output = 'No such file or directory'

    def _test_output(self, args: list[str], expected_output: str) -> None:
        out = StringIO()
        args = _add_rcfile_default_pylintrc(args)
        self._run_pylint(args, out=out)
        actual_output = self._clean_paths(out.getvalue())
        expected_output = self._clean_paths(expected_output)
>       assert expected_output.strip() in actual_output.strip()
E       AssertionError: assert 'No such file or directory' in ''
E        +  where 'No such file or directory' = <built-in method strip of str object at 0x1048569c0>()
E        +    where <built-in method strip of str object at 0x1048569c0> = 'No such file or directory'.strip
E        +  and   '' = <built-in method strip of str object at 0x101478030>()
E        +    where <built-in method strip of str object at 0x101478030> = ''.strip

tests/test_self.py:169: AssertionError
_____________________________________________________________________________________________________________ TestRunTC.test_regression_recursive_current_dir _____________________________________________________________________________________________________________

self = <test_self.TestRunTC object at 0x104896740>

    def test_regression_recursive_current_dir(self):
        with _test_sys_path():
            # pytest is including directory HERE/regrtest_data to sys.path which causes
            # astroid to believe that directory is a package.
            sys.path = [
                path
                for path in sys.path
                if not os.path.basename(path) == "regrtest_data"
            ]
            with _test_cwd():
                os.chdir(join(HERE, "regrtest_data", "directory"))
>               self._test_output(
                    ["."],
                    expected_output="No such file or directory",
                )

tests/test_self.py:1270: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_self.TestRunTC object at 0x104896740>, args = ['--rcfile=/Users/jacobwalls/pylint/pylint/testutils/testing_pylintrc', '.'], expected_output = 'No such file or directory'

    def _test_output(self, args: list[str], expected_output: str) -> None:
        out = StringIO()
        args = _add_rcfile_default_pylintrc(args)
        self._run_pylint(args, out=out)
        actual_output = self._clean_paths(out.getvalue())
        expected_output = self._clean_paths(expected_output)
>       assert expected_output.strip() in actual_output.strip()
E       AssertionError: assert 'No such file or directory' in ''
E        +  where 'No such file or directory' = <built-in method strip of str object at 0x1048569c0>()
E        +    where <built-in method strip of str object at 0x1048569c0> = 'No such file or directory'.strip
E        +  and   '' = <built-in method strip of str object at 0x101478030>()
E        +    where <built-in method strip of str object at 0x101478030> = ''.strip

tests/test_self.py:169: AssertionError
</details>

jacobtylerwalls:
No more failures in pylint. I'd be more comfortable merging after dropping python3.6, but apparently this isn't blocked any more.

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,32 @@
...
+    working_modname = ""
+    last_parent = None
+    for component in modname.split("."):
+        if working_modname:
+            working_modname += "." + component
+        else:
+            # First component
+            working_modname = component
+        try:
+            found_spec = _find_spec_from_path(working_modname, last_parent)
```
DanielNoord, line 24: Is there any benefit to caching this?
Importlib usually is quite expensive, but perhaps this private function isn't.

tests/unittest_manager.py
```diff
@@ -128,7 +126,6 @@ def test_implicit_namespace_package(self) -> None:
     def test_namespace_package_pth_support(self) -> None:
```
DanielNoord, line 128: Do these tests still fail without certain code? I thought these lines were needed to make these packages namespace packages in a hacky way.

jacobtylerwalls: Yeah, if you just make `is_namespace()` unconditionally return False, these tests fail. My understanding is that the hacks were just for the reliance on the private variables of `pkg_resources`. (Part of the reason `pkg_resources` is discouraged now, I take it!)

DanielNoord: I haven't gotten around to testing it, but are we sure this still catches namespace packages defined in the `pkg_resources` way? I know I had made a test package when I initially worked on this and was then able to keep the same tests. See: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages

jacobtylerwalls:
As I can't hit the now uncovered lines on my CPython install even on main, I'm assuming the coverage reduction is due to PyPy objects being built more like CPython now?

tests/unittest_manager.py
DanielNoord: [TestPackage.zip](https://github.com/PyCQA/astroid/files/8620555/TestPackage.zip)

The above zip contains a package as defined by that guideline. With the following diff:
```diff
diff --git a/astroid/interpreter/_import/spec.py b/astroid/interpreter/_import/spec.py
index 7a1de6cdb..cc963d722 100644
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -328,6 +328,9 @@ def _find_spec_with_path(search_path, modname, module_parts, processed, submodul
         spec = finder.find_module(modname, module_parts, processed, submodule_path)
         if spec is None:
             continue
+        if spec.type == ModuleType.PY_NAMESPACE:
+            print(spec)
+            print(finder)
         return finder, spec
 
     raise ImportError(f"No module named {'.'.join(module_parts)}")
```
And the following command: `pylint ../namespace_package/myotherpackage`

`main` returns:
```console
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x1062e12d0>
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x1062e17b0>
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x1062e18d0>
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x1062e1000>
```

Whereas this PR returns:
```console
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x1053054b0>
```

Without the `if`:
```console
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x10df415a0>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df41570>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df412a0>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df40340>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df41c60>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df41e40>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df40340>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df417b0>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df41e40>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/Users/daniel/DocumentenLaptop/Programming/Github/namespace_package/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x10df41e70>
```

`mynamespace` is thus interpreted as a different type of package. I don't know if this is actually a bug or if we are actually fixing a bug here, but I would need to do some more investigation before signing off on this. Perhaps you already know why this is different and whether this is a good thing from your own investigations?

jacobtylerwalls: Thanks for looking into this. I see the same output on `main` vs. the PR. Would you mind checking one more time? I also tried varying my `cwd` and invoking `pylint` differently.

I see `PKG_DIRECTORY: 3` for `mynamespace` and `PY_NAMESPACE: 10` for `myotherpackage` on both:

```shell
% python3 -m pylint mynamespace --disable=all 
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='./mynamespace', origin=None, submodule_search_locations=None)

Your code has been rated at 10.00/10 (previous run: 0.00/10, +10.00)

% python3 -m pylint myotherpackage --disable=all
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/Users/myuser/TestPackage/./myotherpackage'])

Your code has been rated at 10.00/10 (previous run: 0.00/10, +10.00)
```

DanielNoord: Setup:

```console
❯ cp -R /Users/daniel/Documents/Downloads/TestPackage /tmp/TestPackage
❯ cd /tmp/TestPackage
❯ virtualenv venv
created virtual environment CPython3.10.0.final.0-64 in 614ms
...
❯ source venv/bin/activate
❯ python -m pip install pylint
...
❯ python -m pip install -e /Users/daniel/DocumentenLaptop/Programming/Github/astroid
...
❯ python -m pip install -e .
...
```

On `main`:
```console
❯ python -m pylint myotherpackage
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/private/tmp/TestPackage/./myotherpackage', '/private/tmp/TestPackage/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x10e9f2440>
************* Module mynamespace.subpackage_b.module
myotherpackage/mynamespace/subpackage_b/module.py:1:0: C0114: Missing module docstring (missing-module-docstring)
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/private/tmp/TestPackage/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x10e9f27d0>
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/private/tmp/TestPackage/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x10e9f2ce0>
ModuleSpec(name='mynamespace', type=<ModuleType.PY_NAMESPACE: 10>, location='', origin='namespace', submodule_search_locations=['/private/tmp/TestPackage/mynamespace'])
<astroid.interpreter._import.spec.ExplicitNamespacePackageFinder object at 0x10e9f25c0>
myotherpackage/mynamespace/subpackage_b/module.py:4:0: W0104: Statement seems to have no effect (pointless-statement)

Your code has been rated at 3.33/10 (previous run: 6.67/10, -3.33)
```

On `PR`:
```console
❯ python -m pylint myotherpackage
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/private/tmp/TestPackage/./myotherpackage', '/private/tmp/TestPackage/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x10be7dc30>
************* Module mynamespace.subpackage_b.module
myotherpackage/mynamespace/subpackage_b/module.py:1:0: C0114: Missing module docstring (missing-module-docstring)
myotherpackage/mynamespace/subpackage_b/module.py:4:0: W0104: Statement seems to have no effect (pointless-statement)

Your code has been rated at 3.33/10 (previous run: 3.33/10, +0.00)
```

So when linting `myotherpackage` on this PR `mynamespace` is no longer inferred as `ModuleType.PY_NAMESPACE`.
By removing the `if` from the `diff` I can see that it is now inferred as:

```console
❯ python -m pylint myotherpackage
ModuleSpec(name='myotherpackage', type=<ModuleType.PY_NAMESPACE: 10>, location=None, origin=None, submodule_search_locations=['/private/tmp/TestPackage/./myotherpackage', '/private/tmp/TestPackage/myotherpackage'])
<astroid.interpreter._import.spec.PathSpecFinder object at 0x107dc59c0>
************* Module mynamespace.subpackage_b.module
myotherpackage/mynamespace/subpackage_b/module.py:1:0: C0114: Missing module docstring (missing-module-docstring)
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc5e70>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc5e10>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc5d50>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc5900>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc6530>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc5fc0>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc6590>
ModuleSpec(name='subpackage_a', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace/subpackage_a', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc6260>
ModuleSpec(name='mynamespace', type=<ModuleType.PKG_DIRECTORY: 3>, location='/private/tmp/TestPackage/mynamespace', origin=None, submodule_search_locations=None)
<astroid.interpreter._import.spec.ImportlibFinder object at 0x107dc65c0>
myotherpackage/mynamespace/subpackage_b/module.py:4:0: W0104: Statement seems to have no effect (pointless-statement)

Your code has been rated at 3.33/10 (previous run: 3.33/10, +0.00)
```

So `ModuleType.PKG_DIRECTORY`.

Note that I'm not linting `mynamespace` itself, but looking at how `mynamespace` is inferred when linting `myotherpackage`. Hope this helps make it reproducible!

jacobtylerwalls:
@DanielNoord I played around with the old-style setuptools packages here:
https://github.com/PyCQA/astroid/blob/826f6efc14bb63531fafe67da26802e6e77005b4/tests/testdata/python3/data/path_pkg_resources_1/package/__init__.py#L1

I did manage to find a behavior difference with `main`, so I'll push changes in a second, but it doesn't seem to be the one you observed. If this package already in the test suite is not correct somehow, let's try to identify how. 🤔

DanielNoord:
> @DanielNoord I played around with the old-style setuptools packages here:
> 
> https://github.com/PyCQA/astroid/blob/826f6efc14bb63531fafe67da26802e6e77005b4/tests/testdata/python3/data/path_pkg_resources_1/package/__init__.py#L1
> 
> I did manage to find a behavior difference with `main`, so I'll push changes in a second, but it doesn't seem to be the behavior change you observed. If this package already in the test suite is not correct somehow, let's try to identify how. 🤔
> 
> This is what I observed:
> 
> ```python
> >>> import astroid
> >>> import tests.testdata.python3.data.path_pkg_resources_1.package.foo
> >>> astroid.interpreter._import.util.is_namespace('tests.testdata.python3.data.path_pkg_resources_1')  # was False with my original PR, True on main
> True
> ```

This was the original test case that my attempt bonked on as well. I'm excited to see what you came up with!

tests/unittest_manager.py
```diff
@@ -107,6 +108,11 @@ def test_ast_from_namespace_pkgutil(self) -> None:
     def test_ast_from_namespace_pkg_resources(self) -> None:
         self._test_ast_from_old_namespace_package_protocol("pkg_resources")
 
+    def test_identify_old_namespace_package_protocol(self) -> None:
+        self.assertTrue(
+            util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
```
jacobtylerwalls, line 113: hah, pre-commit just straight up deleted one of my imports to set up this test case, hang on

tests/unittest_manager.py
jacobtylerwalls: Thanks. Still can't do it. What is your current working directory when invoking `pylint`? I find that leads to different results.

jacobtylerwalls: Ah, you did say `cd /tmp/TestPackage`. Weird.

DanielNoord:
```ini
[MASTER]

# A comma-separated list of package or module names from where C extensions may
# be loaded. Extensions are loading into the active Python interpreter and may
# run arbitrary code.
extension-pkg-allow-list=

# A comma-separated list of package or module names from where C extensions may
# be loaded. Extensions are loading into the active Python interpreter and may
# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
# for backward compatibility.)
extension-pkg-whitelist=

# Return non-zero exit code if any of these messages/categories are detected,
# even if score is above --fail-under value. Syntax same as enable. Messages
# specified are enabled, while categories only check already-enabled messages.
fail-on=

# Specify a score threshold to be exceeded before program exits with error.
fail-under=10.0

# Files or directories to be skipped. They should be base names, not paths.
ignore=CVS

# Add files or directories matching the regex patterns to the ignore-list. The
# regex matches against paths and can be in Posix or Windows format.
ignore-paths=

# Files or directories matching the regex patterns are skipped. The regex
# matches against base names, not paths. The default value ignores emacs file
# locks
ignore-patterns=^\.#

# Python code to execute, usually for sys.path manipulation such as
# pygtk.require().
#init-hook=

# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
# number of processors available to use.
jobs=1

# Control the amount of potential inferred values when inferring a single
# object. This can help the performance when dealing with large functions or
# complex, nested conditions.
limit-inference-results=100

# List of plugins (as comma separated values of python module names) to load,
# usually to register additional checkers.
load-plugins=

# Pickle collected data for later comparisons.
persistent=yes

# Minimum Python version to use for version dependent checks. Will default to
# the version used to run pylint.
py-version=3.10

# Discover python modules and packages in the file system subtree.
recursive=no

# When enabled, pylint would attempt to guess common misconfiguration and emit
# user-friendly hints instead of false-positive error messages.
suggestion-mode=yes

# Allow loading of arbitrary C extensions. Extensions are imported into the
# active Python interpreter and may run arbitrary code.
unsafe-load-any-extension=no


[MESSAGES CONTROL]

# Only show warnings with the listed confidence levels. Leave empty to show
# all. Valid levels: HIGH, CONTROL_FLOW, INFERENCE, INFERENCE_FAILURE,
# UNDEFINED.
confidence=

# Disable the message, report, category or checker with the given id(s). You
# can either give multiple identifiers separated by comma (,) or put this
# option multiple times (only on the command line, not in the configuration
# file where it should appear only once). You can also use "--disable=all" to
# disable everything first and then re-enable specific checks. For example, if
# you want to run only the similarities checker, you can use "--disable=all
# --enable=similarities". If you want to run only the classes checker, but have
# no Warning level messages displayed, use "--disable=all --enable=classes
# --disable=W".
disable=raw-checker-failed,
        bad-inline-option,
        locally-disabled,
        file-ignored,
        suppressed-message,
        useless-suppression,
        deprecated-pragma,
        use-symbolic-message-instead

# Enable the message, report, category or checker with the given id(s). You can
# either give multiple identifier separated by comma (,) or put this option
# multiple time (only on the command line, not in the configuration file where
# it should appear only once). See also the "--disable" option for examples.
enable=c-extension-no-member


[REPORTS]

# Python expression which should return a score less than or equal to 10. You
# have access to the variables 'fatal', 'error', 'warning', 'refactor',
# 'convention', and 'info' which contain the number of messages in each
# category, as well as 'statement' which is the total number of statements
# analyzed. This score is used by the global evaluation report (RP0004).
evaluation=max(0, 0 if fatal else 10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10))

# Template used to display messages. This is a python new-style format string
# used to format the message information. See doc for all details.
#msg-template=

# Set the output format. Available formats are text, parseable, colorized, json
# and msvs (visual studio). You can also give a reporter class, e.g.
# mypackage.mymodule.MyReporterClass.
output-format=text

# Tells whether to display a full report or only the messages.
reports=no

# Activate the evaluation score.
score=yes


[REFACTORING]

# Maximum number of nested blocks for function / method body
max-nested-blocks=5

# Complete name of functions that never returns. When checking for
# inconsistent-return-statements if a never returning function is called then
# it will be considered as an explicit return statement and no message will be
# printed.
never-returning-functions=sys.exit,argparse.parse_error


[LOGGING]

# The type of string formatting that logging methods do. `old` means using %
# formatting, `new` is for `{}` formatting.
logging-format-style=old

# Logging modules to check that the string format arguments are in logging
# function parameter format.
logging-modules=logging


[SPELLING]

# Limits count of emitted suggestions for spelling mistakes.
max-spelling-suggestions=4

# Spelling dictionary name. Available dictionaries: none. To make it work,
# install the 'python-enchant' package.
spelling-dict=

# List of comma separated words that should be considered directives if they
# appear and the beginning of a comment and should not be checked.
spelling-ignore-comment-directives=fmt: on,fmt: off,noqa:,noqa,nosec,isort:skip,mypy:

# List of comma separated words that should not be checked.
spelling-ignore-words=

# A path to a file that contains the private dictionary; one word per line.
spelling-private-dict-file=

# Tells whether to store unknown words to the private dictionary (see the
# --spelling-private-dict-file option) instead of raising a message.
spelling-store-unknown-words=no


[MISCELLANEOUS]

# List of note tags to take in consideration, separated by a comma.
notes=FIXME,
      XXX,
      TODO

# Regular expression of note tags to take in consideration.
#notes-rgx=


[TYPECHECK]

# List of decorators that produce context managers, such as
# contextlib.contextmanager. Add to this list to register other decorators that
# produce valid context managers.
contextmanager-decorators=contextlib.contextmanager

# List of members which are set dynamically and missed by pylint inference
# system, and so shouldn't trigger E1101 when accessed. Python regular
# expressions are accepted.
generated-members=

# Tells whether missing members accessed in mixin class should be ignored. A
# class is considered mixin if its name matches the mixin-class-rgx option.
ignore-mixin-members=yes

# Tells whether to warn about missing members when the owner of the attribute
# is inferred to be None.
ignore-none=yes

# This flag controls whether pylint should warn about no-member and similar
# checks whenever an opaque object is returned when inferring. The inference
# can return multiple potential results while evaluating a Python object, but
# some branches might not be evaluated, which results in partial inference. In
# that case, it might be useful to still emit no-member and other checks for
# the rest of the inferred objects.
ignore-on-opaque-inference=yes

# List of class names for which member attributes should not be checked (useful
# for classes with dynamically set attributes). This supports the use of
# qualified names.
ignored-classes=optparse.Values,thread._local,_thread._local

# List of module names for which member attributes should not be checked
# (useful for modules/projects where namespaces are manipulated during runtime
# and thus existing member attributes cannot be deduced by static analysis). It
# supports qualified module names, as well as Unix pattern matching.
ignored-modules=

# Show a hint with possible names when a member name was not found. The aspect
# of finding the hint is based on edit distance.
missing-member-hint=yes

# The minimum edit distance a name should have in order to be considered a
# similar match for a missing member name.
missing-member-hint-distance=1

# The total number of similar names that should be taken in consideration when
# showing a hint for a missing member.
missing-member-max-choices=1

# Regex pattern to define which classes are considered mixins ignore-mixin-
# members is set to 'yes'
mixin-class-rgx=.*[Mm]ixin

# List of decorators that change the signature of a decorated function.
signature-mutators=


[VARIABLES]

# List of additional names supposed to be defined in builtins. Remember that
# you should avoid defining new builtins when possible.
additional-builtins=

# Tells whether unused global variables should be treated as a violation.
allow-global-unused-variables=yes

# List of names allowed to shadow builtins
allowed-redefined-builtins=

# List of strings which can identify a callback function by name. A callback
# name must start or end with one of those strings.
callbacks=cb_,
          _cb

# A regular expression matching the name of dummy variables (i.e. expected to
# not be used).
dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_

# Argument names that match this expression will be ignored. Default to name
# with leading underscore.
ignored-argument-names=_.*|^ignored_|^unused_

# Tells whether we should check for unused import in __init__ files.
init-import=no

# List of qualified module names which can have objects that can redefine
# builtins.
redefining-builtins-modules=six.moves,past.builtins,future.builtins,builtins,io


[FORMAT]

# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
expected-line-ending-format=

# Regexp for a line that is allowed to be longer than the limit.
ignore-long-lines=^\s*(# )?<?https?://\S+>?$

# Number of spaces of indent required inside a hanging or continued line.
indent-after-paren=4

# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
# tab).
indent-string='    '

# Maximum number of characters on a single line.
max-line-length=100

# Maximum number of lines in a module.
max-module-lines=1000

# Allow the body of a class to be on the same line as the declaration if body
# contains single statement.
single-line-class-stmt=no

# Allow the body of an if to be on the same line as the test if there is no
# else.
single-line-if-stmt=no


[SIMILARITIES]

# Comments are removed from the similarity computation
ignore-comments=yes

# Docstrings are removed from the similarity computation
ignore-docstrings=yes

# Imports are removed from the similarity computation
ignore-imports=no

# Signatures are removed from the similarity computation
ignore-signatures=no

# Minimum lines number of a similarity.
min-similarity-lines=4


[STRING]

# This flag controls whether inconsistent-quotes generates a warning when the
# character used as a quote delimiter is used inconsistently within a module.
check-quote-consistency=no

# This flag controls whether the implicit-str-concat should generate a warning
# on implicit string concatenation in sequences defined over several lines.
check-str-concat-over-line-jumps=no


[BASIC]

# Naming style matching correct argument names.
argument-naming-style=snake_case

# Regular expression matching correct argument names. Overrides argument-
# naming-style. If left empty, argument names will be checked with the set
# naming style.
#argument-rgx=

# Naming style matching correct attribute names.
attr-naming-style=snake_case

# Regular expression matching correct attribute names. Overrides attr-naming-
# style. If left empty, attribute names will be checked with the set naming
# style.
#attr-rgx=

# Bad variable names which should always be refused, separated by a comma.
bad-names=foo,
          bar,
          baz,
          toto,
          tutu,
          tata

# Bad variable names regexes, separated by a comma. If names match any regex,
# they will always be refused
bad-names-rgxs=

# Naming style matching correct class attribute names.
class-attribute-naming-style=any

# Regular expression matching correct class attribute names. Overrides class-
# attribute-naming-style. If left empty, class attribute names will be checked
# with the set naming style.
#class-attribute-rgx=

# Naming style matching correct class constant names.
class-const-naming-style=UPPER_CASE

# Regular expression matching correct class constant names. Overrides class-
# const-naming-style. If left empty, class constant names will be checked with
# the set naming style.
#class-const-rgx=

# Naming style matching correct class names.
class-naming-style=PascalCase

# Regular expression matching correct class names. Overrides class-naming-
# style. If left empty, class names will be checked with the set naming style.
#class-rgx=

# Naming style matching correct constant names.
const-naming-style=UPPER_CASE

# Regular expression matching correct constant names. Overrides const-naming-
# style. If left empty, constant names will be checked with the set naming
# style.
#const-rgx=

# Minimum line length for functions/classes that require docstrings, shorter
# ones are exempt.
docstring-min-length=-1

# Naming style matching correct function names.
function-naming-style=snake_case

# Regular expression matching correct function names. Overrides function-
# naming-style. If left empty, function names will be checked with the set
# naming style.
#function-rgx=

# Good variable names which should always be accepted, separated by a comma.
good-names=i,
           j,
           k,
           ex,
           Run,
           _

# Good variable names regexes, separated by a comma. If names match any regex,
# they will always be accepted
good-names-rgxs=

# Include a hint for the correct naming format with invalid-name.
include-naming-hint=no

# Naming style matching correct inline iteration names.
inlinevar-naming-style=any

# Regular expression matching correct inline iteration names. Overrides
# inlinevar-naming-style. If left empty, inline iteration names will be checked
# with the set naming style.
#inlinevar-rgx=

# Naming style matching correct method names.
method-naming-style=snake_case

# Regular expression matching correct method names. Overrides method-naming-
# style. If left empty, method names will be checked with the set naming style.
#method-rgx=

# Naming style matching correct module names.
module-naming-style=snake_case

# Regular expression matching correct module names. Overrides module-naming-
# style. If left empty, module names will be checked with the set naming style.
#module-rgx=

# Colon-delimited sets of names that determine each other's naming style when
# the name regexes allow several styles.
name-group=

# Regular expression which should only match function or class names that do
# not require a docstring.
no-docstring-rgx=^_

# List of decorators that produce properties, such as abc.abstractproperty. Add
# to this list to register other decorators that produce valid properties.
# These decorators are taken in consideration only for invalid-name.
property-classes=abc.abstractproperty

# Regular expression matching correct type variable names. If left empty, type
# variable names will be checked with the set naming style.
#typevar-rgx=

# Naming style matching correct variable names.
variable-naming-style=snake_case

# Regular expression matching correct variable names. Overrides variable-
# naming-style. If left empty, variable names will be checked with the set
# naming style.
#variable-rgx=


[CLASSES]

# Warn about protected attribute access inside special methods
check-protected-access-in-special-methods=no

# List of method names used to declare (i.e. assign) instance attributes.
defining-attr-methods=__init__,
                      __new__,
                      setUp,
                      __post_init__

# List of member names, which should be excluded from the protected access
# warning.
exclude-protected=_asdict,
                  _fields,
                  _replace,
                  _source,
                  _make

# List of valid names for the first argument in a class method.
valid-classmethod-first-arg=cls

# List of valid names for the first argument in a metaclass class method.
valid-metaclass-classmethod-first-arg=cls


[IMPORTS]

# List of modules that can be imported at any level, not just the top level
# one.
allow-any-import-level=

# Allow wildcard imports from modules that define __all__.
allow-wildcard-with-all=no

# Analyse import fallback blocks. This can be used to support both Python 2 and
# 3 compatible code, which means that the block might have code that exists
# only in one or another interpreter, leading to false positives when analysed.
analyse-fallback-blocks=no

# Deprecated modules which should not be used, separated by a comma.
deprecated-modules=

# Output a graph (.gv or any supported image format) of external dependencies
# to the given file (report RP0402 must not be disabled).
ext-import-graph=

# Output a graph (.gv or any supported image format) of all (i.e. internal and
# external) dependencies to the given file (report RP0402 must not be
# disabled).
import-graph=

# Output a graph (.gv or any supported image format) of internal dependencies
# to the given file (report RP0402 must not be disabled).
int-import-graph=

# Force import order to recognize a module as part of the standard
# compatibility libraries.
known-standard-library=

# Force import order to recognize a module as part of a third party library.
known-third-party=enchant

# Couples of modules and preferred modules, separated by a comma.
preferred-modules=


[DESIGN]

# List of regular expressions of class ancestor names to ignore when counting
# public methods (see R0903)
exclude-too-few-public-methods=

# List of qualified class names to ignore when counting class parents (see
# R0901)
ignored-parents=

# Maximum number of arguments for function / method.
max-args=5

# Maximum number of attributes for a class (see R0902).
max-attributes=7

# Maximum number of boolean expressions in an if statement (see R0916).
max-bool-expr=5

# Maximum number of branch for function / method body.
max-branches=12

# Maximum number of locals for function / method body.
max-locals=15

# Maximum number of parents for a class (see R0901).
max-parents=7

# Maximum number of public methods for a class (see R0904).
max-public-methods=20

# Maximum number of return / yield for function / method body.
max-returns=6

# Maximum number of statements in function / method body.
max-statements=50

# Minimum number of public methods for a class (see R0903).
min-public-methods=2


[EXCEPTIONS]

# Exceptions that will emit a warning when being caught. Defaults to
# "BaseException, Exception".
overgeneral-exceptions=BaseException,
                       Exception
```

Output of `--generate-rcfile` which just prints out the current config. Think this is just the default values?

DanielNoord: > Ah, you did say `cd /tmp/TestPackage`. Weird.

Is misread your question and thought you asked for my config. Well, there you have it 😄 

I believe `cd` into the folder is required for this "test" to work.

jacobtylerwalls: OH. This is the part I wasn't expecting:
```shell
python -m pip install -e .
```
So the test case has to do with installing the package under site-packages. I wasn't expecting that.

DanielNoord: Yeah, I guess that is required for `importlib` to figure out that `mynamespace` is indeed a namespace package (although I don't know the specifics).

jacobtylerwalls: I think it's possible it's the other way around. Installing it under site-packages with pip is enough to get astroid's `    ImportlibFinder` to find it first rather than `ExplicitNamespacePackageFinder`. And in that case, if installed with `pip`, it _is_ a regular package, not a namespace package. So we might have fixed something here if this wasn't happening before.

DanielNoord: Oh hm, that might actually be true. I just assumed changed behaviour was a regression. 

I am not completely sure though. So I'll need to play around with this a little bit more to be sure. Should find time for this this week!

astroid/interpreter/_import/util.py
jacobtylerwalls: I'll take a look, it does seem we might call this repeatedly on the same part.

jacobtylerwalls: [2697e10](https://github.com/PyCQA/astroid/pull/1536/commits/2697e10c07dc8c98fb3871ed4b7fce23e11d0460)

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,59 @@
...
+    if found_spec.submodule_search_locations is not None:
+        for search_location in found_spec.submodule_search_locations:
+            if any(
+                _is_setuptools_namespace(directory)
+                for directory in pathlib.Path(search_location).iterdir()
+                if directory.is_dir()
+            ):
+                return True
+
+    return found_spec.origin == "namespace"
```
DanielNoord, line 60: Does it make sense to do this before loading streams in `_is_setuptools_namespace`? That seems more expensive so it might be best to do this check first, then do `setuptools` and then default to `True/False`?

tests/unittest_manager.py
```diff
@@ -111,6 +110,16 @@ def test_ast_from_namespace_pkgutil(self) -> None:
...
 
+    def test_identify_old_namespace_package_protocol(self) -> None:
+        # Like the above cases, this package follows the old namespace package protocol
+        # astroid currently assumes such packages are in sys.modules, so import it
+        # pylint: disable-next=import-outside-toplevel
+        import tests.testdata.python3.data.path_pkg_resources_1.package.foo as _  # noqa
+
+        self.assertTrue(
+            util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
+        )
```
DanielNoord, line 121:
```suggestion
        assert util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
```

But this is very personal preference-y.

tests/unittest_manager.py
```diff
@@ -131,7 +140,6 @@ def test_implicit_namespace_package(self) -> None:
     def test_namespace_package_pth_support(self) -> None:
         pth = "foogle_fax-0.12.5-py2.7-nspkg.pth"
         site.addpackage(resources.RESOURCE_PATH, pth, [])
-        pkg_resources._namespace_packages["foogle"] = []
```
DanielNoord, line 143: Let's not resolve the question above as I think it contains very valuable discussion for anybody revisiting this PR later on.

That said, are these tests still meaningful without the `_namespace_packages` injection?

DanielNoord (changes requested):
Some last nitpicks. I have tested and played around with this some more and I think this might indeed fix all the issues.

Let's go ahead and get this in! 🎉

astroid/interpreter/_import/util.py
jacobtylerwalls: Yes!

tests/unittest_manager.py
jacobtylerwalls: Yeah in pytest that's better, but I guess since this one inherits from unittest I wanted to keep it this way. `unittest` output gets mangled when you do a bare assert. When those decorators for the "cachesetupmixin" go away, then maybe this doesn't need to inherit from unittest anymore.

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,59 @@
...
-        and hasattr(pkg_resources, "_namespace_packages")
-        and modname in pkg_resources._namespace_packages
-    )
+import os
+import pathlib
+from functools import lru_cache
+from importlib.util import _find_spec_from_path
+
+
+def _is_setuptools_namespace(location) -> bool:
```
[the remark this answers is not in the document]
jacobtylerwalls: Thanks, I need to specify that I'm passing `Path` objects now, too (from `iterdir`).

tests/unittest_manager.py
jacobtylerwalls: I think so. I'm not absolutely sure, but it looks like it's building an `ast` for those modules. My new tests don't do that. So I'd be reluctant to chop 'em.

astroid/interpreter/_import/util.py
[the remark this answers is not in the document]
DanielNoord: Perhaps even require it to be `Path` and let the caller handle the casting? I'm not sure. Using `os.path` on `pathlib.Path` could be avoided I guess?

tests/unittest_manager.py
DanielNoord: Yeah, even if they are not testing our `is_namespace` check anymore, they are checking if we can build these `asts` correctly. I guess that is a good test in itself.

Any regressions should probably be handled with a new test instead of expanding on these anyway.

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,64 @@
...
+    found_spec = None
+
+    # find_spec() attempts to import parent packages when given dotted paths.
+    # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
+    # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
+    # be specced from left to right.
+    working_modname = ""
+    last_parent = None
+    for component in modname.split("."):
+        if working_modname:
```
DanielNoord, line 37: One last comment just because I wonder if you know the answer:
What would be the performance impact of only assigning `component` here and adding the `.` on L47-48.

It feels a bit wasteful to have an `if` check that is only really necessary for the first iteration, but on the other hand an additional assignment would also be costly.
Just something I thought about and to which I really don't have the answer.

Anyway, such optimisations are really not what we're after (right now). Just a thought experiment 😄

DanielNoord (approved):

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,60 @@
...
+        except ValueError:
+            # executed .pth files may not have __spec__
+            return True
+        last_parent = working_modname
+
+    if found_spec is None:
+        return False
+
+    if found_spec.origin == "namespace":
+        return True
```
DanielNoord, line 50: This is uncovered and based on https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec.origin I think it should be `origin is None`.

jacobtylerwalls: Right, this changed in 3.7 and I grabbed the wrong one. See https://github.com/PyCQA/astroid/blob/24b0bc346f485f10cff145b07507057dd72629fa/astroid/interpreter/_import/spec.py#L230-L233

I will open a separate cleanup for that quoted usage.

jacobtylerwalls: Taking a look at the failed test...

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,65 @@
...
+
+    if found_spec.submodule_search_locations is not None:
+        for search_location in found_spec.submodule_search_locations:
+            if any(
+                _is_setuptools_namespace(directory)
+                for directory in pathlib.Path(search_location).iterdir()
+                if directory.is_dir()
+            ):
+                return True
```
jacobtylerwalls, line 64: Aaaand with better distinguishing between builtins and namespace packages, we ... don't even need this anymore? Highlighting in a comment in a comment in case we pop the hood on this again.

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,53 @@
...
+            found_spec = _find_spec_from_path(working_modname, last_parent)
+        except ValueError:
+            # executed .pth files may not have __spec__
+            return True
+        last_parent = working_modname
+
+    if found_spec is None:
+        return False
+
+    if found_spec.loader is BuiltinImporter:
```
jacobtylerwalls, line 50: It's possible I also need to handle `FrozenImporter`?

astroid/interpreter/_import/util.py
```diff
@@ -2,15 +2,53 @@
...
-    )
+from __future__ import annotations
+
+import pathlib
+from functools import lru_cache
+from importlib.machinery import BuiltinImporter
+from importlib.util import _find_spec_from_path
+
+
+def _is_setuptools_namespace(location: pathlib.Path) -> bool:
```
jacobtylerwalls, line 13: Should I move this back to reduce the size of the diff?

jacobtylerwalls: done

astroid/interpreter/_import/util.py
jacobtylerwalls: Ah, no I don't, because `.origin` will be `"frozen"`.

DanielNoord (approved):
🎉

DanielNoord:
@jacobtylerwalls I'm seeing a crash in one of my `pylint` primer runs. I haven't investigated, but this was my first guess. See:
https://github.com/DanielNoord/pylint/runs/6641534527?check_suite_focus=true
Seems to happen when lining `black`? Do you have any idea?

jacobtylerwalls:
Yep, this PR caused it. I'll take a look.

DanielNoord:
@jacobtylerwalls Thanks!

jacobtylerwalls:
@DanielNoord See #1575, I'm waiting on a response to the CPython issue linked there.

PR #1540, opened 2022-05-03
Infer returned value of .copy() method for collections

### LINKED ISSUE #1403: dict.copy() result is not inferred as dict
### Steps to reproduce

Run this code:
```python
import astroid

code = """
x = dict()
y = x.copy()
"""

ast = astroid.parse(code)

for name in ('x', 'y'):
  node = ast.locals[name][0]
  print(name, node.inferred())
```

### Current behavior
x [<Dict.dict l.2 at 0x7fffe852e110>]
y [Uninferable]

### Expected behavior
I expect `y` to be inferred as `dict` too.

2.4.2

timmartin:
## Description

Where Astroid can infer that the subject of a `.copy()` method call is a built-in type that supports the copy operator, we can infer that the returned value will have the same value as whatever was inferred.

This change only adds support for `list`, `dict`, `set` and `frozenset`; in theory we can apply it to anything inferred to be of type `collections.abc.MutableSequence`, but: a) that doesn't cover `set` and `frozenset` so we'd need special cases for them anyway and b) Astroid doesn't currently seem to support inferring which types implement `MutableSequence`.

`bytearray` is another type for which we could infer `.copy()` behavior, but Astroid doesn't seem to have any explicit support for it, so adding this looks like more work.

## Type of Changes

## Related Issue

Closes #1403

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,10 @@ Release date: TBA
 * Rename ``ModuleSpec`` -> ``module_type`` constructor parameter to match attribute
   name and improve typing. Use ``type`` instead.
 
+* Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``
+  and ``frozenset``.
+
+  Closes #1403
 
 What's New in astroid 2.11.5?
 =============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,7 +16,7 @@ Release date: TBA
 * Rename ``ModuleSpec`` -> ``module_type`` constructor parameter to match attribute
   name and improve typing. Use ``type`` instead.
 
-* Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``
+* Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``,
   and ``frozenset``.
 
   Closes #1403
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -5,7 +5,7 @@
 """Astroid hooks for various builtins."""
 
 from functools import partial
-from typing import Optional
+from typing import Iterator, Optional, Union
 
 from astroid import arguments, helpers, inference_tip, nodes, objects, util
 from astroid.builder import AstroidBuilder
@@ -892,12 +892,15 @@ def infer_dict_fromkeys(node, context=None):
     return _build_dict_with_elements([])
 
 
-def _looks_like_copy_method(node):
+def _looks_like_copy_method(node: nodes.Call) -> bool:
     func = node.func
     return isinstance(func, nodes.Attribute) and func.attrname == "copy"
 
 
-def _infer_copy_method(node, context=None):
+def _infer_copy_method(
+    node: nodes.Call,
+    context: Optional[InferenceContext]=None
+) -> Iterator[Union[nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]]:
     inferred = list(node.func.expr.infer(context=context))
     if all(
         isinstance(inferred_node, (nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet))
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -894,12 +894,13 @@ def infer_dict_fromkeys(node, context=None):
 
 
 def _infer_copy_method(
-    node: nodes.Call,
-    context: Optional[InferenceContext]=None
+    node: nodes.Call, context: Optional[InferenceContext] = None
 ) -> Iterator[Union[nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]]:
     inferred_orig, inferred_copy = itertools.tee(node.func.expr.infer(context=context))
     if all(
-        isinstance(inferred_node, (nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet))
+        isinstance(
+            inferred_node, (nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet)
+        )
         for inferred_node in inferred_orig
     ):
         return inferred_copy
@@ -940,5 +941,5 @@ AstroidManager().register_transform(
     nodes.Call,
     inference_tip(_infer_copy_method),
     lambda node: isinstance(node.func, nodes.Attribute)
-    and node.func.attrname == 'copy',
+    and node.func.attrname == "copy",
 )
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -897,16 +897,12 @@ def _looks_like_copy_method(node):
     return isinstance(func, nodes.Attribute) and func.attrname == "copy"
 
 
-def _has_copy_method(node):
-    return any(
-        isinstance(node, klass)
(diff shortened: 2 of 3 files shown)


astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
...
+def _looks_like_copy_method(node):
+    func = node.func
+    return isinstance(func, nodes.Attribute) and func.attrname == "copy"
+
+
+def _has_copy_method(node):
+    return any(
+        isinstance(node, klass)
+        for klass in [nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]
+    )
```
Pierre-Sassoulas, line 904:
```suggestion
    return isinstance(node, (nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet))
```
I think this work https://docs.python.org/3/library/functions.html#isinstance

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
     return _build_dict_with_elements([])
 
 
+def _looks_like_copy_method(node):
```
DanielNoord, line 895:
```suggestion
def _looks_like_copy_method(node: nodes.Call) -> bool:
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
...
+    func = node.func
+    return isinstance(func, nodes.Attribute) and func.attrname == "copy"
+
+
+def _has_copy_method(node):
```
DanielNoord, line 900:
```suggestion
def _has_copy_method(node: nodes.NodeNG) -> bool:
```

astroid/brain/brain_builtin_inference.py
DanielNoord, line 904:
```suggestion
    return isinstance(node, (nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet))
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
...
+    return any(
+        isinstance(node, klass)
+        for klass in [nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]
+    )
+
+
+def _infer_copy_method(node, context=None):
```
DanielNoord, line 907:
```suggestion
def _infer_copy_method(node: nodes.Call, context: Optional[InferenceContext]=None) -> ...:
```

I think the return value is `Iterator[Union[nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]]`?

DanielNoord (changes requested):
Thanks for working on this! Just some nitpicks about typing and the `isinstance` calls. Test seem good!

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
...
+    inferred = list(node.func.expr.infer(context=context))
```
cdce8p, line 908: With `list` you evaluate the whole iterator even if the condition would later exit early. The work is done. [itertools.tee()](https://docs.python.org/3/library/itertools.html#itertools.tee) should work here.

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +892,26 @@ def _build_dict_with_elements(elements):
...
+    if all(_has_copy_method(x) for x in inferred):
```
cdce8p, line 909: With the suggestions `_has_copy_method` is a single `isinstance` call. We could safe multiple function lookups / calls if we just inline it here.

tests/unittest_inference.py
```diff
@@ -2051,6 +2051,30 @@ def test_dict_invalid_args(self) -> None:
...
+        b = a.copy()
+        b #@
+
+        a = set([1, 2, 3])
+        b = a.copy()
+        b #@
+
+        a = frozenset([1, 2, 3])
+        b = a.copy()
+        b #@
```
cdce8p, line 2070: I would suggest to choose different variable names for each case. That could make debugging a bit easier if something were to go wrong at some point.

Could you also add a case for which we **can't** infer `copy()`?

ChangeLog
```diff
@@ -16,6 +16,10 @@ Release date: TBA
 * Rename ``ModuleSpec`` -> ``module_type`` constructor parameter to match attribute
   name and improve typing. Use ``type`` instead.
 
+* Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``
```
cdce8p, line 19:
```suggestion
* Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``,
```

cdce8p (changes requested):

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +893,21 @@ def _build_dict_with_elements(elements):
     return _build_dict_with_elements([])
 
 
+def _infer_copy_method(
+    node: nodes.Call, context: Optional[InferenceContext] = None
+) -> Iterator[Union[nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]]:
```
cdce8p, line 898:
```suggestion
) -> Iterator[nodes.NodeNG]:
```
The result is only use the check the function itself, nothing else. So it's not necessary to constrict it too much.

astroid/brain/brain_builtin_inference.py
```diff
@@ -4,8 +4,9 @@
 
 """Astroid hooks for various builtins."""
 
+import itertools
 from functools import partial
-from typing import Optional
+from typing import Iterator, Optional, Union
```
cdce8p, line 9:
```suggestion
from typing import Iterator, Optional
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +893,21 @@ def _build_dict_with_elements(elements):
     return _build_dict_with_elements([])
 
 
+def _infer_copy_method(
+    node: nodes.Call, context: Optional[InferenceContext] = None
+) -> Iterator[Union[nodes.Dict, nodes.List, nodes.Set, objects.FrozenSet]]:
+    inferred_orig, inferred_copy = itertools.tee(node.func.expr.infer(context=context))
```
cdce8p, line 899:
```suggestion
    assert isinstance(node.func, nodes.Attribute)
    inferred_orig, inferred_copy = itertools.tee(node.func.expr.infer(context=context))
```
This is already checked in `register_transform` but mypy doesn't know about it.

cdce8p (commented):
The changes look great! Just some last comments, then it's ready.

astroid/brain/brain_builtin_inference.py
```diff
@@ -892,6 +893,22 @@ def _build_dict_with_elements(elements):
     return _build_dict_with_elements([])
 
 
+def _infer_copy_method(
+    node: nodes.Call, context: Optional[InferenceContext] = None
+) -> Iterator[nodes.nodeNG]:
```
cdce8p, line 898:
```suggestion
) -> Iterator[nodes.NodeNG]:
```

cdce8p (approved):
Looks good now! Thanks @timmartin 🚀

PR #1541, opened 2022-05-03
Add typing to ``NodeNG._infer`` and associated nodes

DanielNoord:
## Steps

## Description

/CC @cdce8p 

I ran into the issue of assuming that `infer` only returns `NodeNG` when in reality it can also return `objects` etc. So I took another look at its typing. Just wondering if this would be the way to do it?

Instead of ignoring in `inference.py` we know ignore one redefinition in `nodeng.py`. I did it for two `_infer` assignments to show what it would look like. Would this be the way to do this?

## Type of Changes

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -105,14 +105,10 @@ def _infer_sequence_helper(node, context=None):
     return values
 
 
-_InferSequenceNodeT = TypeVar(
-    "_InferSequenceNodeT", bound=Union[nodes.List, nodes.Tuple, nodes.Set]
-)
-
-
 @decorators.raise_if_nothing_inferred
 def infer_sequence(
-    self: _InferSequenceNodeT, context: Optional[InferenceContext] = None
+    self: Union[nodes.List, nodes.Tuple, nodes.Set],
+    context: Optional[InferenceContext] = None,
 ) -> Iterator[Union[nodes.List, nodes.Tuple, nodes.Set]]:
     has_starred_named_expr = any(
         isinstance(e, (nodes.Starred, nodes.NamedExpr)) for e in self.elts
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -52,7 +52,9 @@ _T = TypeVar("_T")
 _BaseContainerT = TypeVar("_BaseContainerT", bound=nodes.BaseContainer)
 
 
-def infer_end(self: _T, context: InferenceContext | None = None) -> Iterator[_T]:
+def infer_end(
+    self: _T, context: InferenceContext | None = None, **kwargs: Any
+) -> Iterator[_T]:
     """Inference's end for nodes that yield themselves on inference
 
     These are objects for which inference does not have any semantic,
@@ -94,6 +96,7 @@ def _infer_sequence_helper(node, context=None):
 def infer_sequence(
     self: _BaseContainerT,
     context: InferenceContext | None = None,
+    **kwargs: Any,
 ) -> Iterator[_BaseContainerT]:
     has_starred_named_expr = any(
         isinstance(e, (nodes.Starred, nodes.NamedExpr)) for e in self.elts
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -63,11 +63,13 @@ def infer_end(
     yield self
 
 
-nodes.Module._infer = infer_end
-nodes.ClassDef._infer = infer_end
-nodes.Lambda._infer = infer_end
-nodes.Const._infer = infer_end
-nodes.Slice._infer = infer_end
+# We add ignores to all assignments to methods
+# See https://github.com/python/mypy/issues/2427
+nodes.Module._infer = infer_end  # type: ignore[assignment]
+nodes.ClassDef._infer = infer_end  # type: ignore[assignment]
+nodes.Lambda._infer = infer_end  # type: ignore[assignment]
+nodes.Const._infer = infer_end  # type: ignore[assignment]
+nodes.Slice._infer = infer_end  # type: ignore[assignment]
 
 
 def _infer_sequence_helper(node, context=None):
@@ -113,9 +115,9 @@ def infer_sequence(
         yield self
 
 
-nodes.List._infer = infer_sequence
-nodes.Tuple._infer = infer_sequence
-nodes.Set._infer = infer_sequence
+nodes.List._infer = infer_sequence  # type: ignore[assignment]
+nodes.Tuple._infer = infer_sequence  # type: ignore[assignment]
+nodes.Set._infer = infer_sequence  # type: ignore[assignment]
 
 
 def infer_map(self, context=None):
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -5,25 +5,14 @@
 """this module contains a set of functions to handle inference on astroid trees
 """
 
+from __future__ import annotations
+
 import ast
 import functools
 import itertools
 import operator
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Generator,
-    Iterable,
-    Iterator,
-    Optional,
-    Type,
-    TypeVar,
-    Union,
-)
-
-import wrapt
+from collections.abc import Callable, Generator, Iterable, Iterator
+from typing import TYPE_CHECKING, Any, TypeVar
 
 from astroid import bases, decorators, helpers, nodes, protocols, util
 from astroid.context import (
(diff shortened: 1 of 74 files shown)


astroid/inference.py
```diff
@@ -59,7 +59,15 @@
 # .infer method ###############################################################
 
 
-def infer_end(self, context=None):
+_InferEndNodeT = TypeVar(
+    "_InferEndNodeT",
+    bound=Union[nodes.Module, nodes.ClassDef, nodes.Lambda, nodes.Const, nodes.Slice],
```
cdce8p, line 64: `infer_end` would work with any object, regardless if it's a node or not. A bare TypeVar would probably be better.

```py
_T = TypeVar("_T")
```

astroid/inference.py
```diff
@@ -100,7 +106,10 @@ def _infer_sequence_helper(node, context=None):
 
 
 @decorators.raise_if_nothing_inferred
-def infer_sequence(self, context=None):
+def infer_sequence(
+    self: Union[nodes.List, nodes.Tuple, nodes.Set],
+    context: Optional[InferenceContext] = None,
+) -> Iterator[Union[nodes.List, nodes.Tuple, nodes.Set]]:
```
cdce8p, line 112: Isn't the input `self` type also the same as the return type?
I would suggest a TypeVar bound to `BaseContainer` here.

cdce8p (changes requested):
I'm not really convinced a Callback Protocol is the right solution here. From what I've read so far, they aren't designed / don't work properly for class methods.

In the end it's just something pretty complicated to work around a mypy limitation.

DanielNoord:
> I'm not really convinced a Callback Protocol is the right solution here. From what I've read so far, they aren't designed / don't work properly for class methods.
> 
> In the end it's just something pretty complicated to work around a mypy limitation.

Another solution I was thinking of is creating private sub-classes for of `NodeNG` with the `_infer` method which can then be inherited by the respective nodes. Something like `_InferEndNodeNG`. Would that be better?

DanielNoord:
> > > I'm not really convinced a Callback Protocol is the right solution here. From what I've read so far, they aren't designed / don't work properly for class methods.
> > > In the end it's just something pretty complicated to work around a mypy limitation.
> > 
> > 
> > Another solution I was thinking of is creating private sub-classes for of `NodeNG` with the `_infer` method which can then be inherited by the respective nodes. Something like `_InferEndNodeNG`. Would that be better?
> 
> Not sure. What error message are you trying to solve? I tried something like this but didn't find a way to fix `Cannot assign to a method [assignment]`.

Yeah I'm trying to solve the `assignment` issue as well as let the IDE know the `_infer` function of all nodes. Shouldn't subclassing work to resolve `[assignment]`?

cdce8p:
> Yeah I'm trying to solve the `assignment` issue as well as let the IDE know the `_infer` function of all nodes. Shouldn't subclassing work to resolve `[assignment]`?

Tbh I'm not sure I fully understand what you want to do. The whole reason the methods are assigned separately is to avoid circular dependencies. Subclassing could (in theory) work, but I suspect it would get much more complicate quickly.

Not sure the effort is worth it. Especially since you don't gain much from it with regards to type checking. Defaulting to the type signature of `NodeNG._infer` should be fine for almost all cases.

DanielNoord:
> Not sure the effort is worth it. Especially since you don't gain much from it with regards to type checking. Defaulting to the type signature of `NodeNG._infer` should be fine for almost all cases.

Well, one of the issues is that the signatures are not actually compatible as far as I can see.

`_infer` gets called with unpacking of ``kwargs``:
https://github.com/PyCQA/astroid/blob/78a486dc7ecc2567b85ba9d4a85e20639ed71aad/astroid/nodes/node_ng.py#L166-L174

However, `NodeNG._infer` only supports `context`:
https://github.com/PyCQA/astroid/blob/78a486dc7ecc2567b85ba9d4a85e20639ed71aad/astroid/nodes/node_ng.py#L596-L601

Whereas for example `Import` accepts `asname` as well:
https://github.com/PyCQA/astroid/blob/78a486dc7ecc2567b85ba9d4a85e20639ed71aad/astroid/inference.py#L252-L269

So, although we could just disable all these assignment and rely on `NodeNG` it seems as if there is at least some incompatibility between some of the sub-classes. I thought fixing the typing issues might surface these.
_I would also like to finish the typing on `NodeNG.infer()`, but to do so I want to be sure we don't miss any possible return values of `_infer`, which is why I revisited these._

cdce8p:
> Well, one of the issues is that the signatures are not actually compatible as far as I can see.
> 
> `_infer` gets called with unpacking of `kwargs`:

Good point. `NodeNG._infer` should also accept `**kwargs: Any`.

DanielNoord:
@cdce8p `pyright` seems to accept this for these two `_infer` methods and only complains about private variable usage. We would still need to disable some `mypy` messages, but at least the typing would now be correct. Is this how we should/want to proceed with this?

astroid/inference.py
```diff
@@ -57,13 +63,11 @@ def infer_end(self, context=None):
     yield self
 
 
-# We add ignores to all these assignments in this file
-# See https://github.com/python/mypy/issues/2427
-nodes.Module._infer = infer_end  # type: ignore[assignment]
-nodes.ClassDef._infer = infer_end  # type: ignore[assignment]
-nodes.Lambda._infer = infer_end  # type: ignore[assignment]
-nodes.Const._infer = infer_end  # type: ignore[assignment]
-nodes.Slice._infer = infer_end  # type: ignore[assignment]
```
cdce8p, line 66: The `type: ignore[assignment]` comments should still be necessary due to the mypy bug.

astroid/nodes/node_ng.py
```diff
@@ -38,6 +49,13 @@
 else:
     from astroid.decorators import cachedproperty as cached_property
 
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
```
cdce8p, line 56: Mypy doesn't support `Self` yet.

astroid/nodes/node_ng.py
```diff
@@ -581,7 +599,9 @@ def _infer_name(self, frame, name):
         # overridden for ImportFrom, Import, Global, TryExcept and Arguments
         pass
 
-    def _infer(self, context=None):
+    def _infer(
+        self: Self, context: InferenceContext | None = None, **kwargs: Any
+    ) -> Iterator[NodeNG | type[util.Uninferable] | bases.Instance]:
```
cdce8p, line 604:
```suggestion
    def _infer(
        self, context: InferenceContext | None = None, **kwargs: Any
    ) -> Iterator[NodeNG | type[util.Uninferable] | bases.Instance]:
```
`Self` shouldn't be used this way. Here the type can easily be inferred. Only if a second argument or return type should have the same or a derived type from `Self`. A good example would be `infer_end` if it were inside a class.

cdce8p (changes requested):
It has been a while, so trying to catch up a bit.

> @cdce8p pyright seems to accept this for these two _infer methods and only complains about private variable usage. We would still need to disable some mypy messages, but at least the typing would now be correct. Is this how we should/want to proceed with this?

I think yes, it's certainly an improvement. Even if not completely correct, it's something we can iterate upon.

--
There are also some pylint issues. We'll probably need to add `**kwargs` to all `_infer` methods now.

astroid/inference.py
DanielNoord: Yeah, sadly that ignore is also catching all incorrect assignments (difference in types).

Just wanted to show that this doesn't create any actual typing issues.

astroid/nodes/node_ng.py
DanielNoord: Do you want me to remove it? It does seem like the best fir.

astroid/nodes/node_ng.py
DanielNoord: I'll recheck but I believe `pyright` and `mypy` started complaining about this without `Self`. Because we're assigning functions to a method we need some additional typing on the first argument I think.

cdce8p: I only get the known `Cannot assign to a method` error. So it should be fine to remove `Self`.
Besides, I'm not sure how pyright would infer `Self`. Maybe as `Any` 🤷🏻‍♂️

DanielNoord: Yeah sorry, my bad. I had multiple iterations of this to make `pyright` pass.

Probably used to this fix a previous error that doesn't show anymore.

astroid/nodes/node_ng.py
```diff
@@ -38,6 +49,7 @@
 else:
     from astroid.decorators import cachedproperty as cached_property
 
+
```
DanielNoord, line 52:
```suggestion
```

DanielNoord:
@cdce8p Do you want to fix and add `kwargs` to all signatures in this PR? I'd be okay with merging in smaller batches like this. It's not as if we're breaking any CI checks anyway 😄

cdce8p:
> @cdce8p Do you want to fix and add `kwargs` to all signatures in this PR? I'd be okay with merging in smaller batches like this. It's not as if we're breaking any CI checks anyway 😄

I would only fix the ones required for pylint here.
https://github.com/PyCQA/astroid/runs/6747435148?check_suite_focus=true#step:10:26

cdce8p (approved):
Looks good. Could you update the PR title before merging?

PR #1542, opened 2022-05-04
Require Python 3.7.2

jacobtylerwalls:
## Description
Require Python 3.7.2 to run `astroid`. Partner to https://github.com/PyCQA/pylint/pull/5921.

One benefit of dropping 3.6 is to reduce risk from proposed performance-enhancing changes in #1536, where according to the complicated history of https://github.com/python/cpython/pull/5481 and its (initial?) reversion on 3.6, it is possible that the behavior might vary there.

## Type of Changes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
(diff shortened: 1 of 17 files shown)


DanielNoord:
I think we should block this until pylint 2.14 has been released. We might need to do a patch release for 2.13 still.

DanielNoord (approved):

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
The issue with python 3.11 come from main. I think it would be reasonable to temporarily skip this test for python 3.11, because the alternative seems to be making astroid understand the 3.11 code.

cdce8p (approved):
Looks good!

Regarding the 3.11 test failure: Tbh I'm not really sure what the issue is. I believe it appeared just recently (maybe with 3.11.0b1) but when testing that version locally, everything works fine. Maybe it's a linux and / or Github actions specific thing. (Only tested MacOS.)

PR #1544, opened 2022-05-04
Prevent special case for `virtualenv`'s patching of `distutils` from catching other submodules

jacobtylerwalls:
## Steps

## Description
Apparently, #1386 traded a false positive created by `virtualenv` for a false positive in `numpy.distutils`. This PR tightens the condition to pass both cases.

The regression test for `virtualenv` has to be manually triggered on PyCQA. So I [tested this](https://github.com/jacobtylerwalls/astroid/runs/6292399587?check_suite_focus=true) on my fork where I could fiddle with the workflow trigger. Or I could push this branch to PyCQA; that might allow us to trigger the workflow on a branch, but not certain.

## Type of Changes

## Related Issue

Refs PyCQA/pylint#6497

### DIFF
--- a/tests/unittest_regrtest.py
+++ b/tests/unittest_regrtest.py
@@ -81,6 +81,19 @@ multiply([1, 2], [3, 4])
         inferred = callfunc.inferred()
         self.assertEqual(len(inferred), 1)
 
+    @unittest.skipUnless(HAS_NUMPY, "Needs numpy")
+    def test_numpy_distutils(self):
+        """Special handling of virtualenv's patching of distutils shouldn't interfere
+        with numpy.distutils"""
+        node = extract_node(
+            """
+from numpy.distutils.misc_util import is_sequence
+is_sequence("ABC") #@
+"""
+        )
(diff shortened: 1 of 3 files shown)


DanielNoord:
Do you have a Windows to debug this? Seems to pass on 3.8...

tests/unittest_regrtest.py
```diff
@@ -81,6 +81,19 @@ def test_numpy_crash(self):
         inferred = callfunc.inferred()
         self.assertEqual(len(inferred), 1)
 
+    @unittest.skipUnless(HAS_NUMPY, "Needs numpy")
+    def test_numpy_distutils(self):
+        """Special handling of virtualenv's patching of distutils shouldn't interfere
+        with numpy.distutils"""
```
Pierre-Sassoulas, line 87: Do we need to define a specific version of numpy here ? I suppose numpy won't embark distutils forever.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #1550, opened 2022-05-10
qt brain: Make slot argument optional for disconnect()

The-Compiler:
## Steps

~~If this is an acceptable fix, I'll write a changelog~~ - two more questions:

- Is there a good way I can write a test for this?
- Should this go to `main` or to a maintenance branch

## Description

Discussed here:
https://github.com/PyCQA/astroid/pull/1531#issuecomment-1111963792

PyQt supports calling .disconnect() without any arguments in order to
disconnect all slots:
https://www.riverbankcomputing.com/static/Docs/PyQt6/signals_slots.html#disconnect

Strictly speaking, slot=None is a wrong call, as actually passing None
does not work:
https://github.com/python-qt-tools/PyQt5-stubs/pull/103

However, pylint/astroid does not support overloads needed to properly
express this: https://github.com/PyCQA/pylint/issues/5264

So, while this is "wrong", it's less wrong than before - without this
change, pylint expects a mandatory argument, thus raising a
false-positive here:

```python
from PyQt5.QtCore import QTimer
t = QTimer()
t.timeout.connect(lambda: None)
t.timeout.disconnect()
```

despite running fine, pylint complains:

    test.py:4:0: E1120: No value for argument 'slot' in method call (no-value-for-parameter)

while with this change, things work fine.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,8 @@ What's New in astroid 2.11.6?
 =============================
 Release date: TBA
 
+* The Qt brain now correctly treats calling ``.disconnect()`` (with no
+  arguments) on a slot as valid.
 
 What's New in astroid 2.11.5?
 =============================
--- a/astroid/brain/brain_qt.py
+++ b/astroid/brain/brain_qt.py
@@ -24,10 +24,12 @@ def _looks_like_signal(node, signal_name="pyqtSignal"):
 def transform_pyqt_signal(node: nodes.FunctionDef) -> None:
     module = parse(
         """
+    _UNSET = object()
+
     class pyqtSignal(object):
         def connect(self, slot, type=None, no_receiver_check=False):
             pass
-        def disconnect(self, slot=None):
+        def disconnect(self, slot=_UNSET):
             pass
         def emit(self, *args):
             pass
--- a/astroid/brain/brain_qt.py
+++ b/astroid/brain/brain_qt.py
@@ -27,7 +27,7 @@ def transform_pyqt_signal(node: nodes.FunctionDef) -> None:
     class pyqtSignal(object):
         def connect(self, slot, type=None, no_receiver_check=False):
             pass
-        def disconnect(self, slot):
+        def disconnect(self, slot=None):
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas:
>    Is there a good way I can write a test for this?

Check ``unittest_brain_qt.py`` there's a pytest mark for this: https://github.com/PyCQA/astroid/blob/main/tests/unittest_brain_qt.py#L17

> Should this go to main or to a maintenance branch

Let's put it on the main, then we can decide if this need to be backported to the maintenance branche or not :)

The-Compiler:
> Check `unittest_brain_qt.py` there's a pytest mark for this: ...

I've seen that file and tried to write a test inspired by `test_implicit_parameters`, but I'm lost about checking that it's valid to call this function without any parameters (or somehow checking that this `FunctionDef` has no mandatory args).

As for the fix itself: One thing we could do to avoid implying that `None` is okay is to use a `_UNSET = object()` sentinel as default value instead. I just updated the code accordingly.

> Let's put it on the main, then we can decide if this need to be backported to the maintenance branche or not :)

How does the changelog entry work? Do I add it to the earlier version here in the main branch?

jacobtylerwalls:
Thanks for the fix!

> I've seen that file and tried to write a test inspired by test_implicit_parameters, but I'm lost about checking that it's valid to call this function without any parameters (or somehow checking that this FunctionDef has no mandatory args).

You should be able to follow that test to extract the node for the `FunctionDef` representing `disconnect()`, and then drill into its `.args.kwargs` and assert something. The `#@` symbol in the code block is for jumping straight to the node you want to extract.

The-Compiler:
Updated with a test and changelog now. Not too happy with the test since it basically 1:1 checks what the implementation is, but I guess there is not much more we can do if the test should live here and not in pylint.

Pierre-Sassoulas:
Don't mind the failing test, it's due to #1551

Pierre-Sassoulas (approved):
👍

PR #1552, opened 2022-05-12
Fix ImportedModuleTests

jacobtylerwalls:
## Type of Changes

## Related Issue

Follow-up to #1399. This needed to subclass `TestCase` to actually run, and that revealed that clearing the cache before each test method was necessary. cc @tristanlatr EDIT: er, better to use new module names, and avoid clearing the cache altogether, since that's not what's being tested.

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -6628,38 +6628,60 @@ def test_inference_of_items_on_module_dict() -> None:
     builder.file_build(str(DATA_DIR / "module_dict_items_call" / "test.py"), "models")
 
 
-class ImportedModuleTests(resources.AstroidCacheSetupMixin):
-    def test_imported_module_var_inferable(self) -> None:
-        """
-        Module variables can be imported and inferred successfully as part of binary operators.
-        """
-        mod1 = parse("from top.mod import v as z\nw = [1] + z", module_name="top")
-        parse("v = [2]", module_name="top.mod")
-        w_val = mod1.body[-1].value
-        i_w_val = next(w_val.infer())
-        assert i_w_val != util.Uninferable
-        assert i_w_val.as_string() == "[1, 2]"
+def test_imported_module_var_inferable() -> None:
+    """
+    Module variables can be imported and inferred successfully as part of binary operators.
+    """
+    mod1 = parse(
+        textwrap.dedent(
+            """
+    from top1.mod import v as z
+    w = [1] + z
+    """
+        ),
+        module_name="top1",
+    )
+    parse("v = [2]", module_name="top1.mod")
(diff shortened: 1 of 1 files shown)


tests/unittest_inference.py
```diff
@@ -6627,12 +6628,15 @@ def test_inference_of_items_on_module_dict() -> None:
     builder.file_build(str(DATA_DIR / "module_dict_items_call" / "test.py"), "models")
 
 
-class ImportedModuleTests(resources.AstroidCacheSetupMixin):
+class ImportedModuleTests(unittest.TestCase, resources.AstroidCacheSetupMixin):
```
jacobtylerwalls, line 6631: my fault!

jacobtylerwalls (commented):
I'll conform these to the way Marc did them in #1553

tests/unittest_inference.py
```diff
@@ -6627,40 +6627,39 @@ def test_inference_of_items_on_module_dict() -> None:
...
-        parse("__dunder_var__ = ['v']", module_name="top.mod")
-        w_val = mod1.body[-1].value
-        i_w_val = next(w_val.infer())
-        assert i_w_val != util.Uninferable
-        assert i_w_val.as_string() == "['w', 'v']"
+def test_imported_module_var_inferable() -> None:
+    """
+    Module variables can be imported and inferred successfully as part of binary operators.
+    """
+    mod1 = parse("from top1.mod import v as z\n" + "w = [1] + z", module_name="top1")
```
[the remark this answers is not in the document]
Pierre-Sassoulas:
```suggestion
    mod1 = parse("""\
from top1.mod import v as z
w = [1] + z
""", module_name="top1"
)
```

```suggestion
    mod1 = parse(textwrap.dedent("""
    from top1.mod import v as z
    w = [1] + z
    """, module_name="top1"
    )
```
(Black would modify this too, but you get the idea)

Pierre-Sassoulas (approved):
I like the switch to pytest style, it's not astroid style for legacy reason but we need to start somewhere if we want good thing 😄

tests/unittest_inference.py
```diff
@@ -6628,38 +6628,60 @@ def test_inference_of_items_on_module_dict() -> None:
     builder.file_build(str(DATA_DIR / "module_dict_items_call" / "test.py"), "models")
 
 
-class ImportedModuleTests(resources.AstroidCacheSetupMixin):
```
cdce8p, line 6631: Just a note, you don't need to remove test classes. Even with `pytest`, they can be helpful to group test cases. I believe the only important things to remember are the classname must begin with `Test...` and you can use `@staticmethod` if you don't need `self`.

PR #1553, opened 2022-05-12
Update pylint to 2.14.0b1

cdce8p:
## Description
* Fix `implicit-str-concat` and `duplicate-value` errors
* Remove `uselesss-suppression`
* Remove `unrecognizedoption`

### DIFF
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -160,7 +160,7 @@ class CallSite:
         """
         if name in self.duplicated_keywords:
             raise InferenceError(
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (approved):
👌

cdce8p:
Wasn't sure if we want to use the beta version here but I guess it will work.
Pushed a small commit https://github.com/PyCQA/astroid/pull/1553/commits/2148be05dcbf1146c55486315e56c07d9940e681 with changes which weren't possible before (one useless-suppression and the removal of `no-self-use`).

Issue #1551, opened 2022-05-11
Test issue with 3.11


cdce8p:
Managed to track down the test issue we are seeing with `3.11.0b1` a bit. Seems like it's partially caused by another test case which is why I wasn't able to reproduce it locally at first. The following command can be used to recreate the issue

```
pytest \
    tests/unittest_brain.py::ModuleExtenderTest::test_extension_modules \
    tests/unittest_object_model.py::ExceptionModelTest::test_valueerror_py3
```

In particular it seems like the `unittest` module extender is partially responsible. Not sure though if it's a false positive or a real issue.
https://github.com/PyCQA/astroid/blob/0bf711f84d2de55bf777b45cec2449ce776a669a/astroid/brain/brain_unittest.py#L28-L31

So far I haven't figured out what exactly is going on here. It must have been a change in `3.11.0b1` though since `3.11.0a7` was ok. Maybe you two have an idea @DanielNoord @jacobtylerwalls?

/CC: @Pierre-Sassoulas

https://docs.python.org/3.11/whatsnew/changelog.html#python-3-11-0-beta-1

DanielNoord:
Haven't looked at this myself, but one user also reports this on the latest 3.8:

https://github.com/PyCQA/astroid/commit/78a486dc7ecc2567b85ba9d4a85e20639ed71aad#commitcomment-73321532

Not sure if that is the same bug or if it is actually valid, but that might help narrow this down?

jacobtylerwalls:
see https://github.com/python/cpython/pull/28045

cdce8p:
> see [python/cpython#28045](https://github.com/python/cpython/pull/28045)

That looked promising. However, I did bisect it to https://github.com/python/cpython/pull/92202
Need to take a closer look tomorrow.

PR #1558, opened 2022-05-13
Fix 3.11 test

cdce8p:
## Description
Not the ideal solution I was hoping for tbh, but it should work.
As suspected, the test failure is causes by https://github.com/python/cpython/pull/92202. In particular, inside the `__exit__` method there is the following assignment
```py
def __exit__(self, typ, value, traceback):
    ...
    exc.__traceback__ = traceback
```
Previously, astroid couldn't infer `__traceback__` and so defaulted to our fallback method.
https://github.com/PyCQA/astroid/blob/fd102d45a3552c00d951a07b02c49d278a7a69a2/astroid/interpreter/objectmodel.py#L644-L648
Now, with we have an assignment, which later gets inferred as `Uninferable`. However, that means we don't use the fallback anymore. I haven't quite figured out why an assignment in `contextlib` is applied to `__builtins__.BaseException` but that might be related the initial bootstrapping 🤷🏻‍♂️

Anyway, the test change should work and a fallback to `Uninferable` isn't the end of the world IMO.

Closes: #1551

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,9 @@ Release date: TBA
 
   Closes #1403
 
+* Fix test for Python ``3.11``. In some instances ``err.__traceback__`` will
+  be uninferable now.
+
 What's New in astroid 2.11.6?
 =============================
 Release date: TBA
--- a/tests/unittest_object_model.py
+++ b/tests/unittest_object_model.py
@@ -9,6 +9,7 @@ import pytest
 
 import astroid
 from astroid import builder, nodes, objects, test_utils, util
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (approved):
Make sense, I'm not sure we'll even get regression in pylint for this.

cdce8p:
> Make sense, I'm not sure we'll even get regression in pylint for this.

I don't think so either. We can keep it in mind though if something does come up.

--
/CC: @DanielNoord @jacobtylerwalls Just FYI

PR #1560, opened 2022-05-13
Pin Python 3.11.0b1 [ci]

cdce8p:
## Description
Pin `3.11.0b1` for CI runs. Initially we added `3.11-dev` as CI version. That helped a lot with addressing all open issues. However now that all have been fixed, it might be better to pin a specific beta / rc version. That way the CI pipeline can't break accidentally with the next update. We would have full control when we update and can address issues as they pop up.

The obvious downside of course is that version updates would be manual.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11.0-beta.1"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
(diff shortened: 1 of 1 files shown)


cdce8p:
Interesting 🤔

Python 3.11.0b1, Linux: `7:09min`
PYthon 3.11.0b1, Windows: `0:33min`

Pierre-Sassoulas (approved):
Approving for the cleanup. I would prefer keeping -dev because I'm ready to merge failing pipeline if the issue is obviously only linked to 3.11, and because there's a balance to achieve between fixing 3.11.0-b1 only issues and having the pipeline break when a new release is out. I prefer the latter because we'll be fixing problem we'll still have in the end.

cdce8p:
> I would prefer keeping -dev because I'm ready to merge failing pipeline if the issue is obviously only linked to 3.11, and because there's a balance to achieve between fixing 3.11.0-b1 only issues and having the pipeline break when a new release is out. I prefer the latter because we'll be fixing problem we'll still have in the end.

It does make sense.

> Python 3.11.0b1, Linux: `7:09min`
> Python 3.11.0b1, Windows: `0:33min`

I tracked down the performance issue to `coverage`, or more specifically changes in `3.11` which make coverage slower. The increase for astroid already seems like a lot, but pylint will be much worse unfortunately. I really hope they can address it.

https://github.com/nedbat/coveragepy/issues/1287#issuecomment-1126711320

--
I will close the PR for now. We can always do the cleanup later. The separation would allow us to exclude coverage for `3.11` and furthermore, the windows tests start up earlier since they don't have to wait for `3.11` to finish.

PR #1563, opened 2022-05-14
Avoid a global state in `utils.builtin_lookup` and avoid reinstantiating `TransformVisitor`

### LINKED ISSUE #1559: Pollution test cases
While debugging #1551, I used [dectect-test-pollution](https://github.com/asottile/detect-test-pollution) to isolate the test case.

The tool does have a fuzzy mode which can find other pollution test cases. Unfortunately, it didn't even make it past round `1` 😕
It seems like our test suite doesn't clean up properly and is somewhat reliant on a global state. For now it's fine to continue as is though, at some point, these issues might come back to haunt us.

**Install**
```
pip install detect-test-pollution
```

**Run with** (Python 3.10)
```
detect-test-pollution --tests tests/ --fuzz --runs 10
```

If someone would like to take a look, that would be greatly appreciated 🙏🏻

/CC: @jacobtylerwalls and @DanielNoord I saw you two worked on caching recently, maybe this is something you could take a look at?

/CC: @Pierre-Sassoulas

--
#### The first case I saw

**Failing test**: `tests/unittest_brain.py::TestIsinstanceInference::test_isinstance_int_true`
**Pollution test**: `tests/unittest_manager.py::ClearCacheTest::test_brain_plugins_reloaded_after_clearing_cache`
```
pytest \
    tests/unittest_manager.py::ClearCacheTest::test_brain_plugins_reloaded_after_clearing_cache \
    tests/unittest_brain.py::TestIsinstanceInference::test_isinstance_int_true 
```

jacobtylerwalls:
## Description
**1**
The global state of `utils.builtin_lookup()` introduced in #1460 could be stale if the builtins module had been rebuilt by `AstroidManager.clear_cache()`.
**2**
Noticed while testing (necessary to get the tests to pass): reinstantiating `TransformVisitor` in `clear_cache()` could lead to to diverging registries of transforms.

The bug from **1** can be reproduced on prior versions by just calling `clear_cache` at the beginning of `TestIsinstanceInference.test_isinstance_int_true`. That's a public API, so I added a changelog and an explicit regression test.

The bug (**2**) in `clear_cache` itself was an unreleased issue with my changes in #1528, so I didn't doc it.

## Type of Changes

## Related Issue

Fixes #1559

### DIFF
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -370,7 +370,7 @@ class ClearCacheTest(unittest.TestCase):
 
     def test_builtins_inference_after_clearing_cache_manually(self) -> None:
         # Not recommended to manipulate this, so we detect it and call clear_cache() instead
-        astroid.MANAGER.brain["astroid_cache"] = {}
+        astroid.MANAGER.brain["astroid_cache"].clear()
         isinstance_call = astroid.extract_node("isinstance(1, int)")
         inferred = next(isinstance_call.infer())
         self.assertIs(inferred.value, True)
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -362,6 +362,12 @@ class ClearCacheTest(unittest.TestCase):
         inferred = next(format_call.infer())
         self.assertIsInstance(inferred, Const)
 
+    def test_builtins_inference_after_clearing_cache(self) -> None:
+        astroid.MANAGER.clear_cache()
+        isinstance_call = astroid.extract_node("isinstance(1, int)")
+        inferred = next(isinstance_call.infer())
+        self.assertIs(inferred.value, True)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -318,7 +318,7 @@ class BorgAstroidManagerTC(unittest.TestCase):
         self.assertIs(built, second_built)
 
 
-class ClearCacheTest(unittest.TestCase, resources.AstroidCacheSetupMixin):
+class ClearCacheTest(unittest.TestCase):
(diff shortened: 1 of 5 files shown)


tests/unittest_manager.py
```diff
@@ -318,7 +318,7 @@ def test_borg(self) -> None:
         self.assertIs(built, second_built)
 
 
-class ClearCacheTest(unittest.TestCase, resources.AstroidCacheSetupMixin):
+class ClearCacheTest(unittest.TestCase):
```
jacobtylerwalls, line 321: I'd like to get out of the usage of this mixin. I think it masks tests that don't clean up after each other (in this case, the problems were upstream in other tests). We can keep fuzzing to find the problems.

tests/unittest_manager.py
```diff
@@ -362,6 +362,19 @@ def test_brain_plugins_reloaded_after_clearing_cache(self) -> None:
...
         self.assertIsInstance(inferred, Const)
 
+    def test_builtins_inference_after_clearing_cache(self) -> None:
+        astroid.MANAGER.clear_cache()
+        isinstance_call = astroid.extract_node("isinstance(1, int)")
+        inferred = next(isinstance_call.infer())
+        self.assertIs(inferred.value, True)
+
+    def test_builtins_inference_after_clearing_cache_manually(self) -> None:
+        # Not recommended to manipulate this, so we detect it and call clear_cache() instead
```
jacobtylerwalls, line 372: In trying to cover the missed lines I just decided to support this, since it's an easy mistake to make.

tests/unittest_manager.py
Pierre-Sassoulas: Thank you for understanding the problem well enough so that we can stop fuzzing around 😄

PR #1564, opened 2022-05-14
Fix coverage performance issue with Python 3.11

cdce8p:
## Description
Pin coverage to `~=6.3.3`.
Changelog: https://github.com/nedbat/coveragepy/releases/tag/6.3.3


cdce8p:
There is still some performance regression with `3.11.0b1`, but it's much better than before.

**Previously**
Python 3.11: `344s`

**Now**
Python 3.10: `100s`
Python 3.11: `130s`

Pierre-Sassoulas (approved):

PR #1565, opened 2022-05-15
Add a bound to the inference tips cache

### LINKED ISSUE #1150: Automatic cache clearing with ``lru_cache``'s ``maxsize``
Following https://github.com/PyCQA/astroid/pull/1139, it's not possible to clear the cache. We did not use ``lru_cache`` because the current implementation only uses the first argument as key (because the other one is context?), and lru_cache does not allow that.

In order to use ``lru_cache`` we need to ask ourselves if is it right that inference tip functions always take exactly two arguments -- a node and a context? If so, what is context exactly? Does it remain constant during an inference session? Is it fine to hash it via id()?

See original comment by @superbobry [here](https://github.com/PyCQA/astroid/pull/1139#discussion_r693555590) and [here](https://github.com/PyCQA/astroid/pull/1139#issuecomment-905906350)

### Current behavior

The cache can be cleared with an API.

### Expected behavior

Automatic clearing when the cache increase above a set value (configurable ?).

jacobtylerwalls:
## Description
Add a bound to the size of the inference tips cache.

I explored making this user-configurable, but I don't really see a compelling use case. It's also difficult because of the inaccessibility of manipulating the `inference_tip` module at runtime: `inference_tip()` is imported over the module name.

## Type of Changes

## Related Issue

Closes #1150

### DIFF
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -44,6 +44,9 @@ def _inference_tip_cached(
         _cache[func, node] = None
         result = _cache[func, node] = list(func(*args, **kwargs))
         assert result
+        # Django can be linted without exceeding this value; next value is 9296
+        if _cache.__sizeof__() > 4680:
+            clear_inference_tip_cache()
     return iter(result)
 
 
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -41,12 +41,11 @@ def _inference_tip_cached(
         if result is None:
             raise UseInferenceDefault()
     except KeyError:
+        if len(_cache) > 127:
+            clear_inference_tip_cache()
         _cache[func, node] = None
         result = _cache[func, node] = list(func(*args, **kwargs))
         assert result
-        # Django can be linted without exceeding this value; next value is 9296
-        if _cache.__sizeof__() > 4680:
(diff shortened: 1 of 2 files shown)


astroid/inference_tip.py
```diff
@@ -41,6 +41,8 @@ def _inference_tip_cached(
         if result is None:
             raise UseInferenceDefault()
     except KeyError:
+        if len(_cache) > 127:
```
Pierre-Sassoulas, line 44: If we get a single cache miss we clear the entire cache ? This does not feel right, I would expect that we clear the oldest entry and add the new one, in a kind of first in first out queue ?

jacobtylerwalls: Easier than I thought to just make this an LRU-style cache.

DanielNoord (approved):

cdce8p:
This PR causes a regression in Home Assistant
```
Error: homeassistant/components/daikin/config_flow.py:88:15: E0712:
    Catching an exception which doesn't inherit from Exception: HTTPForbidden (catching-non-exception)
```

Discovered in https://github.com/cdce8p/ha-core/pull/11, with current pylint and astroid `main`.

PR #1570, opened 2022-05-16
Revert "Add a bound to the inference tips cache (#1565)"

jacobtylerwalls:
This reverts commit 95bbd5ed58da9e56da8d705e1893a804787f1998 (#1565).

@cdce8p reported a [regression](https://github.com/PyCQA/astroid/pull/1565#issuecomment-1127586743) with that commit, and after moving the `popitem()` call out of the `except`, and also trying without `--jobs=n`, I saw no improvement. So reverting.

Despite that, I feel like it's a similar class of issue to the several issues in `pylint` with `--jobs`: we need a better understanding on how the global states are managed and interact with `--init-hook` and the rest. E.g. PyCQA/pylint#4874

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,10 +24,6 @@ Release date: TBA
 
   Refs PyCQA/pylint#5113
 
-* Add a bound to the inference tips cache.
-
-  Closes #1150
-
 * Infer the return value of the ``.copy()`` method on ``dict``, ``list``, ``set``,
   and ``frozenset``.
 
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import typing
-from collections import OrderedDict
 from collections.abc import Iterator
 
 import wrapt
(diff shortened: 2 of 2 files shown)


DanielNoord:
> we need a better understanding on how the global states are managed and interact with `--init-hook` and the rest.

It's not really my interest atm, but I think it would be better if we could remove most of those global states. I believe the reason why we do use those states is to 1) avoid some duplicate initialisation and 2) make combination of results easier. But I think we could just as well initialise a new linter for every process without storing it on some global object.

Too bad this didn't work out, it seemed like it should have worked!

DanielNoord (approved):

DanielNoord:
Btw, just one thought: did you try to increase the cache size as well? The error Marc reported seems to deal with some inheritance patterns. I have seen earlier regressions where inheritance patterns which were too large tripped up the cache as the first real result was removed due to the number of inferences needed to get the second real result. I don't think that is happening here, but it is worth a try?

jacobtylerwalls:
I didn't, on the assumption that it would just mask the underlying issue. We also don't have any actual users asking for this feature ATM.

cdce8p (approved):

cdce8p:
> I didn't, on the assumption that it would just mask the underlying issue. We also don't have any actual users asking for this feature ATM.

I agree. It might work for Home Assistant but other, bigger projects might still experience issues. Better to revert it for now.

PR #1567, opened 2022-05-15
Detect builtins module on PyPy 3.9

jacobtylerwalls:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Related Issue

Closes #1562

PR #1572, opened 2022-05-22
Permit subscripting additional builtin classes

jacobtylerwalls:
## Description
Treat other classes such as `enumerate` and `staticmethod` identically to list, dict, tuple, set, frozenset.

```python
>>> enumerate[2]  # not sure why, but...
enumerate[2]
>>> type(enumerate[2])
<class 'types.GenericAlias'>
```
## Type of Changes

## Related Issue
Noticed while working on #1567.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,7 +22,7 @@ Release date: TBA
 
   Closes #1559
 
-* On Python versions >= 3.9, ``astroid`` now understands (dubiously) subscripting
+* On Python versions >= 3.9, ``astroid`` now understands subscripting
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
LGTM

DanielNoord:
Not sure if we should call this "dubious". Seems to be a conscious decision, see:
https://github.com/python/cpython/issues/83662 and https://github.com/python/cpython/pull/19421

Strangely I don't see `enumerate` discussed in the PEP so perhaps this is indeed "dubious" I don't know.

DanielNoord (approved):

PR #1575, opened 2022-05-29
Provide first component of dotted path to namespace searches

jacobtylerwalls:
## Steps

## Type of Changes

## Related Issue
Fixes (unreleased) crash reported [here](https://github.com/PyCQA/astroid/pull/1536#issuecomment-1140386630). ~My current theory is that this function should only be using the top-level package name as the search path.~. EDIT: no, I'm back to thinking it's better to just catch the `KeyError` and return False.

### DIFF
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -18,14 +18,20 @@ def is_namespace(modname: str) -> bool:
     # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
     # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
     # be specced from left to right.
-    components = modname.split(".")
-    for i in range(1, len(components) + 1):
-        working_modname = ".".join(components[:i])
+    processed_components = []
+    last_parent = None
+    for component in modname.split("."):
+        processed_components.append(component)
+        working_modname = ".".join(processed_components)
         try:
-            found_spec = _find_spec_from_path(working_modname, components[0])
+            found_spec = _find_spec_from_path(working_modname, last_parent)
         except ValueError:
             # executed .pth files may not have __spec__
             return True
+        except KeyError:
+            # https://github.com/python/cpython/issues/93334
+            return False
+        last_parent = working_modname
 
     if found_spec is None:
         return False
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -18,21 +18,18 @@ def is_namespace(modname: str) -> bool:
     # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
     # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
     # be specced from left to right.
-    processed_components = []
-    last_parent = None
-    for component in modname.split("."):
-        processed_components.append(component)
-        working_modname = ".".join(processed_components)
+    components = modname.split(".")
+    for i in range(1, len(components) + 1):
+        working_modname = ".".join(components[:i])
         try:
-            found_spec = _find_spec_from_path(working_modname, last_parent)
+            # Search under the highest package name
+            # Only relevant if package not already on sys.path
+            # See https://github.com/python/cpython/issues/89754 for reasoning
+            # Otherwise can raise bare KeyError: https://github.com/python/cpython/issues/93334
+            found_spec = _find_spec_from_path(working_modname, components[0])
         except ValueError:
             # executed .pth files may not have __spec__
             return True
-        except KeyError:
-            # https://github.com/python/cpython/issues/93334
-            # TODO: remove when minimum python version has the fix
-            return False
-        last_parent = working_modname
 
     if found_spec is None:
         return False
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -18,17 +18,14 @@ def is_namespace(modname: str) -> bool:
     # That's unacceptable here, so we fallback to _find_spec_from_path(), which does
     # not, but requires instead that each single parent ('astroid', 'nodes', etc.)
     # be specced from left to right.
-    processed_components = []
-    last_parent = None
-    for component in modname.split("."):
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
The alternative I'm looking at is just catching `KeyError` -- this might be a limitation of `importlib`, and I want to see first if there are any CPython issue discussions about it.

jacobtylerwalls:
Will for feedback on https://github.com/python/cpython/issues/93334

tests/testdata/python3/data/parent_of_homonym/doc/README.md
```diff
@@ -0,0 +1,2 @@
+This submodule should have the same name as a module in the root of the project
```
jacobtylerwalls, line 1:
```suggestion
This submodule should have the same name as a directory in the root of the project
```

tests/testdata/python3/data/parent_of_homonym/doc/README.md
```diff
@@ -0,0 +1,2 @@
+This submodule should have the same name as a directory in the root of the project
+directory that is _NOT_ a python module. For this reason, it's currently called "doc".
```
jacobtylerwalls, line 2:
```suggestion
that is _NOT_ a python module. For this reason, it's currently called "doc".
```

tests/unittest_manager.py
```diff
@@ -120,6 +120,11 @@ def test_identify_old_namespace_package_protocol(self) -> None:
             util.is_namespace("tests.testdata.python3.data.path_pkg_resources_1")
         )
 
+    def test_homonoym_with_non_module(self) -> None:
```
jacobtylerwalls, line 123:
```suggestion
    def test_submodule_homonym_with_non_module(self) -> None:

**Comment by DanielNoord** [comment]
We don't want to wait for response from `cpython`?

**Comment by Pierre-Sassoulas** [comment]
Seeing the label on the CPython issue we'll need a "temporary" patch for 3.7, 3.8, 3.9.

**Review by Pierre-Sassoulas — APPROVED** [review]

File: astroid/interpreter/_import/util.py [hunk 4]
```diff
@@ -28,6 +28,9 @@ def is_namespace(modname: str) -> bool:
         except ValueError:
             # executed .pth files may not have __spec__
             return True
+        except KeyError:
+            # https://github.com/python/cpython/issues/93334
```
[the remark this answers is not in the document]
jacobtylerwalls: Added a comment. Honestly it may never be fixed, the labels don't actually block releases.

jacobtylerwalls:
@DanielNoord yeah, I wasn't very confident of the correct behavior here but reading python/CPython#89754 suggests that explicitly giving the topmost parent package is probably correct. I just didn't want to change it in case it could miss someone's namespace package, but I realized this is only relevant if it's not on `sys.path` already, and that's on the user to take care of.

astroid/interpreter/_import/util.py
```diff
@@ -18,17 +18,18 @@ def is_namespace(modname: str) -> bool:
...
+    components = modname.split(".")
+    for i in range(1, len(components) + 1):
+        working_modname = ".".join(components[:i])
         try:
-            found_spec = _find_spec_from_path(working_modname, last_parent)
+            # Search under the highest package name
+            # Only relevant if package not already on sys.path
+            # See https://github.com/python/cpython/issues/89754 for reasoning
+            # Otherwise can raise bare KeyError: https://github.com/python/cpython/issues/93334
+            found_spec = _find_spec_from_path(working_modname, components[0])
```
DanielNoord, line 29: @jacobtylerwalls Just thinking about this again. We cached this function because of `importlib` being slow in https://github.com/PyCQA/astroid/pull/1536#discussion_r864481144. (My very first review comment there).

What if we try to get the spec of:
``a.b.c.d.e.f.g`` and then ``a.b.c.d.e.f.h``? We don't benefit from any of the caching right?

tests/testdata/python3/data/parent_of_homonym/doc/README.md
```diff
@@ -0,0 +1,2 @@
+This submodule should have the same name as a directory in the root of the project that
```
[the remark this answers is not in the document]
Pierre-Sassoulas: There is definitely some cleanup to do in the tests directory, there is also the issue of legacy functional tests. https://github.com/PyCQA/pylint/blob/main/tests/test_func.py#L117

Pierre-Sassoulas (approved):
I don't see anything wrong with this but it's very easy to break something when touching namespace / imports, and if we break something we're going to get an issue with 10+ thumbs up an hour, so I'm still a little nervous 😄

astroid/interpreter/_import/util.py
jacobtylerwalls: True, but I don't know if we ever confirmed that this use of importlib was slow. It's slow if importlib is actually importing the parent modules (why we did the #1409 to revert) but the whole point of this approach was to not do that, so I'm not sure it's an issue.

jacobtylerwalls:
> I'm still a little nervous

I hear that! We could do another beta for pylint 2.15.

Pierre-Sassoulas:
> I hear that! We could do another beta for pylint 2.15.

How do you feel about astroid 2.12 in pylint 2.14 ? Possible, desirable ? Should we delay this particular MR until astroid 2.13 ? Or were you thinking of using 2.12 only in pylint 2.15 ?

jacobtylerwalls:
> How do you feel about astroid 2.12 in pylint 2.14 ? Possible, desirable ? Should we delay this particular MR until astroid 2.13 ? Or were you thinking of using 2.12 only in pylint 2.15 ?

I guess I'd hope to just tidy up the changelog for 2.14 and no more functional changes, since we already had a healthy beta period.

jacobtylerwalls:
Rats,I missed it by a few seconds. I found more KeyErrors. I'm going to revert to the other approach.

jacobtylerwalls:
no worries, it's my fault for not immediately marking as draft!

PR #1579, opened 2022-05-31
Special case `__main__` in `is_namespace()`

jacobtylerwalls:
## Type of Changes

## Related Issue

Glad that this one was not a head scratcher. `__main__` is a [special case](https://docs.python.org/3/reference/import.html?highlight=__main__%20__spec__#main-spec) where `__spec__` might be None.

Resolves primer failure in pylint. Caused very recently (unreleased) in #1536, so not documented.

### DIFF
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -34,8 +34,8 @@ def is_namespace(modname: str) -> bool:
                 working_modname, path=last_submodule_search_locations
             )
         except ValueError:
-            # executed .pth files may not have __spec__
-            return True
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
👍

DanielNoord (approved):

DanielNoord:
Still seeing a crash on `pandas`:
https://github.com/PyCQA/pylint/runs/6669595502?check_suite_focus=true

Do we need to special case ``_path__`` as well?

jacobtylerwalls:
See #1581 for the resolution.

PR #1580, opened 2022-05-31
Add partial typing to NodeNG.infer

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -7,11 +7,12 @@ from __future__ import annotations
 import pprint
 import sys
 import warnings
-from collections.abc import Iterator
+from collections.abc import Generator, Iterator
 from functools import singledispatch as _singledispatch
 from typing import TYPE_CHECKING, ClassVar, Tuple, Type, TypeVar, Union, cast, overload
 
-from astroid import decorators, util
+from astroid import bases, decorators, util
+from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidError,
     InferenceError,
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Wow, I'm surprised we did not do that earlier, that's a pretty central function 😄 Thanks !

astroid/nodes/node_ng.py
```diff
@@ -124,7 +125,9 @@ def __init__(
         E.g. ClassDef, FunctionDef.
         """
 
-    def infer(self, context=None, **kwargs):
+    def infer(
+        self, context: InferenceContext | None = None, **kwargs
```
cdce8p, line 129:
```suggestion
        self, context: InferenceContext | None = None, **kwargs: Any
```

astroid/nodes/node_ng.py
```diff
@@ -7,11 +7,12 @@
 import pprint
 import sys
 import warnings
-from collections.abc import Iterator
+from collections.abc import Generator, Iterator
 from functools import singledispatch as _singledispatch
 from typing import TYPE_CHECKING, ClassVar, Tuple, Type, TypeVar, Union, cast, overload
```
cdce8p, line 12:
```suggestion
from typing import (
    TYPE_CHECKING,
    Any,
    ClassVar,
    Tuple,
    Type,
    TypeVar,
    Union,
    cast,
    overload,
)
```

cdce8p (commented):
Looks good. We should just make sure to type `**kwargs` as well, even if it's only with `Any`.

cdce8p (approved):

PR #1589, opened 2022-06-04
Add typing to function in ``raw_building``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -165,7 +165,9 @@ def register_arguments(func, args=None):
             register_arguments(func, arg.elts)
 
 
-def object_build_class(node, member, localname):
+def object_build_class(
+    node: nodes.Module | nodes.ClassDef, member: type, localname: str
+) -> nodes.ClassDef:
     """create astroid for a living class object"""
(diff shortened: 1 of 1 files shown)


astroid/raw_building.py
```diff
@@ -314,12 +330,14 @@ def inspect_build(
...
-    def object_build(self, node, obj):
+    def object_build(
+        self, node: nodes.Module | nodes.ClassDef, obj: types.ModuleType | type
+    ) -> None:
         """recursive method which create a partial ast from real objects
         (only function, class, and method are handled)
         """
         if obj in self._done:
-            return self._done[obj]
+            return None
```
Pierre-Sassoulas, line 340: Could you explain the reasoning ?

DanielNoord: It's never used as a return value. The final `return` in this method just returns `None`. So this is an early exit check to see if we should build this object or if we have already done so.
For some reason it was decided to also return the already finished object, but that's not necessary for anything and unnecessarily complicates the typing/signature.

Pierre-Sassoulas (approved):

PR #1590, opened 2022-06-04
Infer instances of builtins created by `__new__()`

jacobtylerwalls:
## Steps

## Description
Before, instances of builtins created with `__new__()`, e.g. `int.__new__(int, 42)` were inferred as `Uninferable`.
Now, a `Const` node is created, or at the very least, an `Instance`.

## Type of Changes

## Related Issue

Ref https://github.com/PyCQA/pylint/pull/6822#discussion_r888859802

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -372,23 +372,37 @@ class UnboundMethod(Proxy):
         on ``object.__new__`` will be of type ``object``,
         which is incorrect for the argument in general.
         If no context is given the ``object.__new__`` call argument will
-        correctly inferred except when inside a call that requires
+        be correctly inferred except when inside a call that requires
         the additional context (such as a classmethod) of the boundnode
         to determine which class the method was called from
         """
 
-        # If we're unbound method __new__ of builtin object, the result is an
+        # If we're unbound method __new__ of a builtin, the result is an
         # instance of the class given as first argument.
-        if (
-            self._proxied.name == "__new__"
-            and self._proxied.parent.frame(future=True).qname() == "builtins.object"
-        ):
-            if caller.args:
-                node_context = context.extra_context.get(caller.args[0])
-                infer = caller.args[0].infer(context=node_context)
-            else:
-                infer = []
-            return (Instance(x) if x is not Uninferable else x for x in infer)
+        if self._proxied.name == "__new__":
+            qname = self._proxied.parent.frame(future=True).qname()
+            # Avoid checking builtins.type: _infer_type_new_call() does more validation
+            if qname.startswith("builtins.") and qname != "builtins.type":
+                if caller.args:
+                    # pylint: disable-next=import-outside-toplevel; circular import
+                    from astroid.nodes import Const, const_factory
+
+                    # Attempt to create a constant
+                    if len(caller.args) > 1:
+                        value = None
+                        if isinstance(caller.args[1], Const):
+                            value = caller.args[1].value
+                        inferred_arg = next(caller.args[1].infer(), None)
+                        if isinstance(inferred_arg, Const):
+                            value = inferred_arg.value
+                        if value is not None:
+                            return (const_factory(value),)
+
+                    node_context = context.extra_context.get(caller.args[0])
+                    infer = caller.args[0].infer(context=node_context)
+                else:
+                    infer = []
+                return (Instance(x) if x is not Uninferable else x for x in infer)
         return self._proxied.infer_call_result(caller, context)
 
     def bool_value(self, context=None):
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -383,27 +383,32 @@ class UnboundMethod(Proxy):
             qname = self._proxied.parent.frame(future=True).qname()
             # Avoid checking builtins.type: _infer_type_new_call() does more validation
             if qname.startswith("builtins.") and qname != "builtins.type":
-                if caller.args:
-                    # pylint: disable-next=import-outside-toplevel; circular import
-                    from astroid.nodes import Const, const_factory
-
-                    # Attempt to create a constant
-                    if len(caller.args) > 1:
-                        value = None
-                        if isinstance(caller.args[1], Const):
-                            value = caller.args[1].value
-                        inferred_arg = next(caller.args[1].infer(), None)
-                        if isinstance(inferred_arg, Const):
-                            value = inferred_arg.value
-                        if value is not None:
-                            return (const_factory(value),)
-
-                    node_context = context.extra_context.get(caller.args[0])
-                    infer = caller.args[0].infer(context=node_context)
-                else:
-                    infer = []
-                return (Instance(x) if x is not Uninferable else x for x in infer)
+                return self._infer_builtin_new(caller, context)
         return self._proxied.infer_call_result(caller, context)
+
+    def _infer_builtin_new(self, caller, context):
+        # pylint: disable-next=import-outside-toplevel; circular import
+        from astroid.nodes import Const, const_factory
+
+        if not caller.args:
+            yield from []
+            return
+        # Attempt to create a constant
+        if len(caller.args) > 1:
+            value = None
+            if isinstance(caller.args[1], Const):
+                value = caller.args[1].value
+            else:
+                inferred_arg = next(caller.args[1].infer(), None)
+                if isinstance(inferred_arg, Const):
+                    value = inferred_arg.value
+            if value is not None:
+                yield const_factory(value)
+
+        node_context = context.extra_context.get(caller.args[0])
+        infer = caller.args[0].infer(context=node_context)
+
+        yield from (Instance(x) if x is not Uninferable else x for x in infer)
 
     def bool_value(self, context=None):
         return True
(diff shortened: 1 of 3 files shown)


tests/unittest_inference.py
```diff
@@ -3768,6 +3768,32 @@ class MetaBook(type):
         ]
         self.assertEqual(titles, ["Catch 22", "Ubik", "Grimus"])
 
+    def test_builtin_new(self):
```
Pierre-Sassoulas, line 3771:
```suggestion
    def test_builtin_new(self) -> None:
```

astroid/bases.py
```diff
@@ -372,23 +372,37 @@ def infer_call_result(self, caller, context):
...
-            self._proxied.name == "__new__"
-            and self._proxied.parent.frame(future=True).qname() == "builtins.object"
-        ):
-            if caller.args:
-                node_context = context.extra_context.get(caller.args[0])
-                infer = caller.args[0].infer(context=node_context)
-            else:
-                infer = []
-            return (Instance(x) if x is not Uninferable else x for x in infer)
+        if self._proxied.name == "__new__":
```
Pierre-Sassoulas, line 382: Maybe we could create a private function for everything under ``self._proxied.name == "__new__":`` ?

Pierre-Sassoulas (approved):
LGTM

astroid/bases.py
```diff
@@ -372,25 +372,44 @@ def infer_call_result(self, caller, context):
...
+            qname = self._proxied.parent.frame(future=True).qname()
+            # Avoid checking builtins.type: _infer_type_new_call() does more validation
+            if qname.startswith("builtins.") and qname != "builtins.type":
+                return self._infer_builtin_new(caller, context)
         return self._proxied.infer_call_result(caller, context)
 
+    def _infer_builtin_new(self, caller, context):
```
DanielNoord, line 389: Are we able to add typing here?

cdce8p (changes requested):

astroid/bases.py
```diff
@@ -338,6 +342,12 @@ def getitem(self, index, context=None):
         return next(method.infer_call_result(self, new_context), None)
 
 
+if TYPE_CHECKING:
+    from astroid.nodes import Call, Const
+
+    _BuiltinNewT = TypeVar("_BuiltinNewT", Type[Uninferable], Const, Instance)
```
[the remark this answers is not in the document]
jacobtylerwalls: `Instance` doesn't yet exist at the top of the file. But I suppose this is all moot if I just go with `Union` so 👍🏻

astroid/bases.py
```diff
@@ -372,25 +382,46 @@ def infer_call_result(self, caller, context):
...
+                inferred_arg = next(caller.args[1].infer(), None)
+                if isinstance(inferred_arg, Const):
+                    value = inferred_arg.value
+            if value is not None:
+                yield const_factory(value)
+
+        node_context = context.extra_context.get(caller.args[0])
+        infer = caller.args[0].infer(context=node_context)
+
+        yield from (Instance(x) if x is not Uninferable else x for x in infer)
```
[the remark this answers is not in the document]
jacobtylerwalls: I didn't know if that was acceptable style to mix `yield` and `return ...` but happy to leave it as return 👍🏻

astroid/bases.py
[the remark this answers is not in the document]
cdce8p: For typing, you can always fall back to a string. `"Instance"`. As long as it's not evaluated at runtime, it should work.
The `Union` does resolve too, of course.

astroid/bases.py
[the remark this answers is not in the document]
cdce8p: As long as it works 🤷🏻‍♂️ Type checkers should complain otherwise, especially since you added type annotations.

astroid/bases.py
```diff
@@ -372,25 +382,46 @@ def infer_call_result(self, caller, context):
...
+        if self._proxied.name == "__new__":
+            qname = self._proxied.parent.frame(future=True).qname()
+            # Avoid checking builtins.type: _infer_type_new_call() does more validation
+            if qname.startswith("builtins.") and qname != "builtins.type":
+                return self._infer_builtin_new(caller, context)
         return self._proxied.infer_call_result(caller, context)
 
+    def _infer_builtin_new(
+        self, caller: Call, context: InferenceContext
+    ) -> GenT[_BuiltinNewT, None, None]:
```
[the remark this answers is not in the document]
jacobtylerwalls: The prior version of this depended on the context being not None. Should I add an assert?

astroid/bases.py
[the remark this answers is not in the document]
jacobtylerwalls: Interesting, I get a test failure with `return` that I haven't investigated.

astroid/bases.py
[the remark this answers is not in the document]
cdce8p: > The prior version of this depended on the context being not None. Should I add an assert?

Maybe this method is different and always called with a context 🤔

I just debugged the test case you added to look at the call stack, there it's called from `infer_call`. If the `context` is `None`, `copy_context` would create a new one. Let's leave it as is, with a mandatory context parameter. We can always change it later.
https://github.com/PyCQA/astroid/blob/c2135324b5fcb045000fbaee52e6c3f19b02c1f5/astroid/inference.py#L216-L234

astroid/bases.py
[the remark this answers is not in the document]
cdce8p: > Interesting, I get a test failure with `return` that I haven't investigated.

Might be my mistake. I overlooked that it's a generator method now. Thus `return` will raise a `StopIteration` with the generator object as argument, instead of returning the new generator. So `yield from` is probably correct.

astroid/bases.py
```diff
@@ -5,8 +5,10 @@
 """This module contains base classes and functions for the nodes and some
 inference utils.
 """
+from __future__ import annotations
 
 import collections
+from typing import TYPE_CHECKING
```
cdce8p, line 11: Not strictly necessary, but nice for syntax highlighting in VS Code.
```suggestion
import collections
import collections.abc
from typing import TYPE_CHECKING
```

astroid/bases.py
```diff
@@ -372,25 +377,50 @@ def infer_call_result(self, caller, context):
...
+                return self._infer_builtin_new(caller, context)
         return self._proxied.infer_call_result(caller, context)
 
+    def _infer_builtin_new(
+        self,
+        caller: nodes.Call,
+        context: InferenceContext,
+    ) -> collections.abc.Generator[
+        nodes.Const | Instance | type[Uninferable], None, None
+    ]:
```
cdce8p, line 400: Would it make sense to add the same type hints to `infer_call_result` above?

astroid/bases.py
```diff
@@ -372,25 +377,50 @@ def infer_call_result(self, caller, context):
...
+        # pylint: disable-next=import-outside-toplevel; circular import
+        from astroid.nodes import Const, const_factory
```
cdce8p, line 402:
```suggestion
        from astroid import nodes
```
To keep it consistent.

tests/unittest_inference.py
```diff
@@ -3768,6 +3768,33 @@ class MetaBook(type):
...
+    def test_builtin_new() -> None:
+        ast_node = extract_node("int.__new__(int, 42)")
+        inferred = next(ast_node.infer())
+        assert isinstance(inferred, nodes.Const)
+        assert inferred.value == 42
+
+        ast_node2 = extract_node("int.__new__(int)")
+        inferred2 = next(ast_node2.infer())
+        assert isinstance(inferred2, Instance)
+        assert not isinstance(inferred2, nodes.Const)
```
cdce8p, line 3781: Isn't the second assert redundant? If it is an `Instance`, it's not a `Const`.

tests/unittest_inference.py
```diff
@@ -3768,6 +3768,33 @@ class MetaBook(type):
...
+        assert inferred._proxied is inferred._proxied
```
cdce8p, line 3782: This is always true.

cdce8p (changes requested):

tests/unittest_inference.py
jacobtylerwalls: Actually `Const` subclasses `Instance`. Trying to do it all in one assertion with `issubclass` seemed less straightforward to me.

tests/unittest_inference.py
jacobtylerwalls: hah, good spot!  meant "inferred2".

astroid/bases.py
jacobtylerwalls: Not that easily, because we immediately hit the main blocker for adding typing in `astroid` to my knowledge:

```shell
astroid/bases.py:388: error: "None" has no attribute "name"  [attr-defined]
astroid/bases.py:389: error: "None" has no attribute "parent"  [attr-defined]
astroid/bases.py:393: error: "None" has no attribute "infer_call_result"  [attr-defined]
```

cdce8p (approved):

astroid/bases.py
cdce8p: > Not that easily, because we immediately hit the main blocker for adding typing in `astroid` to my knowledge:

Let's leave it as is then, at least for now.

jacobtylerwalls:
Thanks for the review and for getting to the bottom of the accurate types!

PR #1593, opened 2022-06-05
Get argument info from method descriptors

DanielNoord:
## Steps

## Description

Counter PR to https://github.com/PyCQA/astroid/pull/1368.

<s>Work in progress as I try to resolve the pylint issues.</s>

<s>Blocked on https://github.com/PyCQA/astroid/pull/2006</s>

## Type of Changes


astroid/raw_building.py
```diff
@@ -194,16 +210,25 @@ def object_build_class(
...
     args: list[str] = []
     defaults: list[Any] = []
     posonlyargs: list[str] = []
     kwonlyargs: list[str] = []
+    kwonlydefaults: list[Any] = []
+
+    try:
+        signature = inspect.signature(member)
+    except (ValueError, RuntimeError):
+        # Raising RuntimeError is undocumented, see:
```
tristanlatr, line 229: Did not know that, thanks

tristanlatr:
Looking at the code, it looks like it works as expected.

A note from the older PR that worth fixing though: Change the Note in the docstring of Arguments.args: https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.Arguments.args saying that we now get infos from function defined in c-extensions.

The tests failure looks weird thought, no idea why Iterator is not recognized as being subscriptable?

astroid/nodes/node_classes.py
```diff
@@ -914,12 +914,14 @@ def get_children(self):
...
+            yield from self.defaults
         yield from self.kwonlyargs
 
-        for elt in self.kw_defaults:
-            if elt is not None:
-                yield elt
+        if self.kw_defaults is not None:
+            for elt in self.kw_defaults:
+                if elt is not None:
+                    yield elt
```
cdce8p, line 924:
```suggestion
        for elt in self.kw_defaults or ():
             if elt is not None:
                 yield elt
```

tests/unittest_builder.py
```diff
@@ -949,7 +956,60 @@ def test_arguments_of_signature() -> None:
...
+    )
+    package_path = c_module_invalid_text_signature / "mymod"
+
+    # build extension
+    try:
+        cwd = os.getcwd()
+        code, outstr = subprocess.getstatusoutput(
+            f"cd {c_module_invalid_text_signature} && python3 setup.py build_ext --inplace"
+        )
+        os.chdir(cwd)
```
cdce8p, line 994: I don't like that this is done in the repo file tree. Cleanup will always be difficult. Even the `cwd` reset here isn't safe if `subprocess.getstatusoutput` throws an exception.

Maybe take a look at the `tempdir` contextmanager from `pylint -> lint/unittest_lint.py` and also [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).

cdce8p (commented):
_Not a full review, just some things I noticed._

tests/unittest_builder.py
Pierre-Sassoulas: There's [a pytest fixture named tmpdir](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture) that would be useful here. What is done in pylint is a little dated but refactoring everything is not a top priority.

DanielNoord:
@cdce8p Could you confirm that this slows down `pylint`? I think my local environment is showing this, but I'm not sure it isn't something else.

DanielNoord:
Closing as there doesn't seem to be too much interest in this.

PR #1596, opened 2022-06-05
Move TreeRebuilder overloads behind TYPE_CHECKING guard

cdce8p:
## Description
Slight performance improvement. By moving all overloads for `visit` behind `if TYPE_CHECKING`, we can ignore them at runtime. For 2-3 overloads that might not make a difference but here it could be useful.


Pierre-Sassoulas (approved):
Nice 👍 I'll trust that you only indented the overload to the right because the diff is hardcore 😄

PR #1597, opened 2022-06-06
Create and use ``InferenceResult``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import collections
 import collections.abc
-from collections.abc import Iterator
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 from astroid import decorators
@@ -34,7 +34,7 @@ helpers = lazy_import("helpers")
 manager = lazy_import("manager")
 
 if TYPE_CHECKING:
-    from astroid import nodes
+    from astroid import nodes, util
 
 
 # TODO: check if needs special treatment
@@ -123,10 +123,10 @@ class Proxy:
 
 
 def _infer_stmts(
-    stmts: list[InferenceResult],
+    stmts: Sequence[nodes.NodeNG | type[util.Uninferable] | Instance],
     context: InferenceContext | None,
-    frame: nodes.LocalsDictNodeNG | None = None,
-) -> Iterator[InferenceResult]:
+    frame: nodes.NodeNG | Instance | None = None,
+) -> collections.abc.Generator[InferenceResult, None, None]:
     """Return an iterator on statements inferred by each statement in *stmts*."""
     inferred = False
     if context is not None:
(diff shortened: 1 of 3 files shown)


astroid/bases.py
```diff
@@ -120,7 +122,11 @@ def infer(self, context=None):
         yield self
 
 
-def _infer_stmts(stmts, context, frame=None):
+def _infer_stmts(
+    stmts: list[InferenceResult],
```
cdce8p, line 126: Even if technically correct, this doesn't look right. Why should we call a function which takes `InferenceResults` as inputs only to also return `InferenceResults`. The Union is probably better. One other thing, maybe `Sequence` instead of `list`?

astroid/bases.py
```diff
@@ -120,7 +122,11 @@ def infer(self, context=None):
...
+    context: InferenceContext | None,
+    frame: nodes.LocalsDictNodeNG | None = None,
```
cdce8p, line 128: Are you sure it's `LocalsDictNodeNG`? AFAICT `frame` can at minimum also be a `BaseInstance` object. Maybe something else too.

astroid/bases.py
```diff
@@ -120,7 +122,11 @@ def infer(self, context=None):
...
+) -> Iterator[InferenceResult]:
```
cdce8p, line 129: Have been thinking about this again. We currently mix the return types quite a bit. `NodesNG.infer` returns a `Generator` and `NodeNG._infer` an `Interator`. In practice, it doesn't make much of a difference, but it might make sense to at least be consistent.

**Suggestion**
`-> Generator[InferenceResult, None, None]` for all `*infer*` methods / functions. Brains can continue to return an `Iterator`.

cdce8p (changes requested):

astroid/bases.py
DanielNoord: So `InferenceResult` as input and the `Union` of `InferenceResult` as return value?

Updated to `Sequence`.

astroid/bases.py
DanielNoord: You're right. Changing!

astroid/bases.py
cdce8p: > So `InferenceResult` as input and the `Union` of `InferenceResult` as return value?

Switched. `stmts: Union[...]` and `-> InferenceResult`. Result and return annotation does make sense IMO.

astroid/bases.py
```diff
@@ -120,7 +122,11 @@ def infer(self, context=None):
         yield self
 
 
-def _infer_stmts(stmts, context, frame=None):
+def _infer_stmts(
+    stmts: Sequence[nodes.NodeNG | type[Uninferable] | Instance],
+    context: InferenceContext | None,
+    frame: nodes.NodeNG | Instance | None = None,
```
cdce8p, line 128: I haven't checked that again, but it seems to be fine, at least for now. We can come back to it  later if we need to.

cdce8p (approved):

PR #1598, opened 2022-06-06
Infer members of ``Enums`` as instances of the ``Enum`` they belong to

### LINKED ISSUE #744: Type of enum value being incorrectly inferred
**Note**: this content is roughly equivalent to that of an still unanswered [question I posted at SO](https://stackoverflow.com/questions/59829094/incorrect-inference-of-enum-values-type-with-astroid-the-engine-behind-py).

``` py
from enum import Enum

class Whatever(Enum):
    Foo = 1
    Bar = 2

def f():
    x = Whatever.Foo
```

Consider the program above an a call to `infer()` on expression `Whatever.Foo`.

### Current behavior
I get a (proxied) `ClassDef` for a `Foo` "type".

### Expected behavior
I'd expect a `ClassDef` for `Whatever`, which is the actual type -- according to the [language specification](https://docs.python.org/3/library/enum.html#creating-an-enum), that says that *the type of an enumeration member is the enumeration it belongs to.*

astroid version: 2.3.3

DanielNoord:
## Steps

## Description

Closes #744

Also partially refs https://github.com/PyCQA/astroid/issues/142.

We need some way to store attributes that are specific to an individual instance. I'm not sure if this is the best way so I'm open to feedback!

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1367,6 +1367,11 @@ Release date: 2022-07-09
 
   Closes pylint-dev/pylint#5776
 
+* Members of ``Enums`` are now correctly inferred as instances of the ``Enum`` they belong
+  to instead of instances of their own class.
+
+  Closes #744
+
 * Rename ``ModuleSpec`` -> ``module_type`` constructor parameter to match attribute
   name and improve typing. Use ``type`` instead.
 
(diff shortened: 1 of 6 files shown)


DanielNoord:
A decision was made to revert the change in:
https://github.com/python/cpython/issues/93910

So, this is still valid for `3.11`.

It's blocked by https://github.com/PyCQA/astroid/pull/1689 which is a bug fix that is necessary for the change here but I think this can already be reviewed.

DanielNoord:
Will not pursue this in the foreseeable future.

PR #1606, opened 2022-06-08
Add ``__new__`` and ``__call__`` to``ObjectModel`` and ``ClassModel``

DanielNoord:
## Steps

## Description

/CC @jacobtylerwalls 

Part of https://github.com/PyCQA/astroid/pull/1519.

This was what I was talking about in https://github.com/PyCQA/astroid/pull/1519#discussion_r890105397. ``__new__`` and ``__call__`` are part of the models for `object` and `classes`. Thus, instead of adding them as ``attr___ne__`` I think it makes sense to try and define them.

One thing I wonder is whether this should indeed be properties. But looking at other methods I think they should? See for example:
https://github.com/PyCQA/astroid/blob/01db68b1b37fdfa4d7899fa1ba71015b33efb34b/astroid/interpreter/objectmodel.py#L288-L289

## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import collections
 import collections.abc
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 from astroid import decorators
@@ -25,6 +26,7 @@ from astroid.exceptions import (
     InferenceError,
     NameInferenceError,
 )
+from astroid.typing import InferenceResult
 from astroid.util import Uninferable, lazy_descriptor, lazy_import
 
 objectmodel = lazy_import("interpreter.objectmodel")
@@ -120,7 +122,11 @@ class Proxy:
         yield self
 
 
-def _infer_stmts(stmts, context, frame=None):
+def _infer_stmts(
+    stmts: Sequence[nodes.NodeNG | type[Uninferable] | Instance],
+    context: InferenceContext | None,
+    frame: nodes.NodeNG | Instance | None = None,
+) -> collections.abc.Generator[InferenceResult, None, None]:
     """Return an iterator on statements inferred by each statement in *stmts*."""
     inferred = False
     if context is not None:
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -10,9 +10,8 @@ from __future__ import annotations
 import collections
 import collections.abc
 from collections.abc import Sequence
-from typing import TYPE_CHECKING
 
-from astroid import decorators
+from astroid import decorators, nodes
 from astroid.const import PY310_PLUS
 from astroid.context import (
     CallContext,
@@ -32,10 +31,6 @@ from astroid.util import Uninferable, lazy_descriptor, lazy_import
 objectmodel = lazy_import("interpreter.objectmodel")
 helpers = lazy_import("helpers")
 manager = lazy_import("manager")
-nodes = lazy_import("nodes")
-
-if TYPE_CHECKING:
-    from astroid import nodes
 
 
 # TODO: check if needs special treatment
@@ -416,9 +411,6 @@ class UnboundMethod(Proxy):
     ) -> collections.abc.Generator[
         nodes.Const | Instance | type[Uninferable], None, None
     ]:
-        # pylint: disable-next=import-outside-toplevel; circular import
-        from astroid import nodes
-
         if not caller.args:
             return
         # Attempt to create a constant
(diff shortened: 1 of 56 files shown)


DanielNoord:
> Thanks for pushing forward on this. I haven't found time to do a review yet, but I did quickly check it against [PyCQA/pylint#6402](https://github.com/PyCQA/pylint/pull/6402) and this doesn't pass its test case. Was that expected?

Yeah! Ultimately we should merge #1519 as well.

However, since `__new__` and `__init__` are used so ubiquitously I thought there would be benefits to defining them rather than making them `attr___ne__`. I think for the other dunders most tests will automatically pass if we define them on `ObjectModel` instead of `FunctionModel`. (Although I haven't checked).

astroid/bases.py
```diff
@@ -32,6 +32,7 @@
 objectmodel = lazy_import("interpreter.objectmodel")
 helpers = lazy_import("helpers")
 manager = lazy_import("manager")
+nodes = lazy_import("nodes")
```
jacobtylerwalls, line 35: I didn't find this to be necessary earlier today in another PR. WDYT?

astroid/interpreter/objectmodel.py
```diff
@@ -118,6 +118,19 @@ def lookup(self, name):
             return getattr(self, IMPL_PREFIX + name)
         raise AttributeInferenceError(target=self._instance, attribute=name)
 
+    @property
+    def attr___new__(self):
+        """Calling cls.__new__(cls) on an object returns an instance of that object.
+
+        Instance is either an instance or a class definition of the instance to be
+        created.
+        """
```
jacobtylerwalls, line 127: I'm wondering if this obviates most of `UnboundMethod.infer_call_result`? Or no?

DanielNoord:
```python
import astroid

mod = astroid.parse(
    """
class A:
    ...

A().__new__(A)
A.__new__(A)
    """
)
list(mod.body[-1].value.infer())
list(mod.body[-2].value.infer())
```

With:
```diff
diff --git a/astroid/bases.py b/astroid/bases.py
index e41964341..dd1e9d2c1 100644
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -401,6 +401,7 @@ class UnboundMethod(Proxy):
             qname = self._proxied.parent.frame(future=True).qname()
             # Avoid checking builtins.type: _infer_type_new_call() does more validation
             if qname.startswith("builtins.") and qname != "builtins.type":
+                print("CALLED")
                 return self._infer_builtin_new(caller, context)
         return self._proxied.infer_call_result(caller, context)
 
diff --git a/astroid/interpreter/objectmodel.py b/astroid/interpreter/objectmodel.py
index d120895aa..65be429ce 100644
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -127,6 +127,7 @@ class ObjectModel:
         """
         # TODO: Use isinstance instead of try ... except after _instance has typing
         try:
+            print("CALLED")
             return self._instance._proxied.instantiate_class()
         except AttributeError:
             return self._instance.instantiate_class()
```

Shows that this is used for bound methods. Both are being called in other instances I think.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2298,7 +2298,12 @@ def infer_call_result(self, caller, context=None):
         try:
             metaclass = self.metaclass(context=context)
             if metaclass is not None:
-                dunder_call = next(metaclass.igetattr("__call__", context))
+                # Only get __call__ if it's defined locally for the metaclass.
+                # Otherwise we will find ObjectModel.__call__ which will
+                # return an instance of the metaclass. Instantiating the class is
+                # handled later.
+                if "__call__" in metaclass.locals:
+                    dunder_call = next(metaclass.igetattr("__call__", context))
```
jacobtylerwalls, line 2306: Should there be a default value for next?

jacobtylerwalls (commented):
Looks really good! One last Q

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Caught by `StopIteration` on L2307.

jacobtylerwalls: Thanks, didn't see it!

jacobtylerwalls (approved):

PR #1610, opened 2022-06-09
Avoid performing `**` operations on values greater than 1e5

jacobtylerwalls:
## Description
Avoid actually calculating `**` operations on values greater than 1e5.

## Type of Changes

## Related Issue

Closes PyCQA/pylint#6745

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -62,6 +62,11 @@ Release date: TBA
 
   Closes PyCQA/pylint#5783
 
+* Avoid inferring the results of ``**`` operations involving values greater than ``1e5``
+  to avoid expensive computation.
+
+  Closes PyCQA/pylint#6745
+
 * Fix test for Python ``3.11``. In some instances ``err.__traceback__`` will
   be uninferable now.
 
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -113,6 +113,13 @@ for _KEY, _IMPL in list(BIN_OP_IMPL.items()):
 def const_infer_binary_op(self, opnode, operator, other, context, _):
     not_implemented = nodes.Const(NotImplemented)
     if isinstance(other, nodes.Const):
+        if (
+            operator == "**"
+            and isinstance(self, nodes.Const)
+            and (self.value > 1e5 or other.value > 1e5)
+        ):
+            yield not_implemented
+            return
         try:
             impl = BIN_OP_IMPL[operator]
             try:
--- a/tests/unittest_protocols.py
+++ b/tests/unittest_protocols.py
@@ -269,6 +269,12 @@ class ProtocolTests(unittest.TestCase):
         )
         parsed.accept(Visitor())
 
+    @staticmethod
+    def test_uninferable_exponents() -> None:
+        """Attempting to calculate the result is prohibitively expensive."""
+        parsed = extract_node("15 ** 20220609")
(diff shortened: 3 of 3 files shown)


Pierre-Sassoulas (commented):
Thank you for working on this ! I thought about that issues a little and I was wondering about a time out with a really small threshold (0.01s?) and the possibility to relaunch with a flag if you really want the result. Knowing that something will take a long time to calculate is a NP problem, seeing it took longer than x is a P problem. How long does `9999 ** 9999 **` takes for exemple ? How long does any mathematical calculation actually takes ? I don't think we want to had à bunch of conditions like that for each possible formula. What do you think?

jacobtylerwalls:
> I don't think we want to add a bunch of conditions like that for each possible formula.

Oh I agree, I just didn't think we were likely to run into any additional cases. Exponents seemed like the only case that could blow up.

> I was wondering about a time out with a really small threshold (0.01s?)

Interesting, I'm not aware of a means to interrupt a call like this, do you have a suggestion?

Pierre-Sassoulas:
>  Exponents seemed like the only case that could blow up.

A lot of things have a tendency to surprise us when pylint is ran on a lot of unexpected code 😄 I don't have an example of calculation that would take a long time but I think it will always be possible to do so. Maybe it's time to open a cop/robber question on code golf stackexchange with one team trying to catch pathological cases and the other one trying to make the infer slow 😄 ! But maybe it's enough to handle this case only. 

> I'm not aware of a means to interrupt a call like this, do you have a suggestion?

Actually doing it crossplatform and multiprocessus compatible is not that easy. (I was thinking of ``signal`` for that but I checked and it turns out it only works on unix). Do we need to handle multiprocess for that ? It's possible that we're always inferring on a single thread. We have a ``max_inference`` option where we just return uninferable if we recurse more than 100 time (or 500 I don't remember). Having a time limit for inference would be somewhat similar and "solve" all pathological cases at once (we had some other the years, especially with pandas/numpy genuinely large recursion).

Pierre-Sassoulas:
@cdce8p or @hippo91 do you have an opinion on a timeout option for expansive inference calls ? To sum up the discussion, the idea would be to cut inference taking too long after a threshold time, like we cut inference that recurse above a threshold of calls currently.

jacobtylerwalls:
> the idea would be to cut inference taking too long after a threshold time

I love the idea, but are we aware of a way to accomplish this on Windows?

hippo91:
@Pierre-Sassoulas i think it is a good idea to limit the time spent in inference process but i would like to hear @cdce8p opinion on this...

cdce8p:
> @cdce8p or @hippo91 do you have an opinion on a timeout option for expansive inference calls ? To sum up the discussion, the idea would be to cut inference taking too long after a threshold time, like we cut inference that recurse above a threshold of calls currently.

I'm skeptical if it's truly worth the effort. My understanding is that canceling recursive calls above a threshold is comparatively easy. Any timeout based solution does add additional complexity, even if we can avoid threading / multiprocessing, which needs to be maintained. Additionally, it will make debugging more difficult and might even lead to non-deterministic results which in itself can be a nightmare. In the end we probably have to ask ourselves if it's worth it or if there are better / easier alternatives.

For instance, I do like this PR. It will provide a speedup in some edge cases without adding much complexity.

DanielNoord (approved):

PR #1612, opened 2022-06-10
Add a second test for simple augmented assignments

tristanlatr:
## Description

This PR adds a test for with two augmented assignments, I believe there is a flaw in the logic, so creating a failing test for it.

## Type of Changes

| ✓   | Adds a test          |

## Related Issue

#192

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -1414,6 +1414,20 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         self.assertEqual(len(inferred), 1)
         self.assertIsInstance(inferred[0], nodes.Const)
         self.assertEqual(inferred[0].value, 3)
+    
+    def test_augassign_multi(self) -> None:
+        code = """
+            a = 1
+            a += 1
+            a += 1
+            print (a)
+        """
+        ast = parse(code, __name__)
+        inferred = list(test_utils.get_name_node(ast, "a").infer())
+
+        self.assertEqual(len(inferred), 1)
+        self.assertIsInstance(inferred[0], nodes.Const)
+        self.assertEqual(inferred[0].value, 3)
 
     def test_nonregr_func_arg(self) -> None:
         code = """


tristanlatr:
Weird, the test does not pass locally with astroid-2.11.5. 

Did you fixed the #192 without closing the issue recently ?

Pierre-Sassoulas:
Thank you for the regression test and the issue triaging 👍 It's entirely possible that some issues are not up to date and should be closed. Would you mind adding the example from the issue with lists maybe it was the problem and not integer ?

tristanlatr:
> Would you mind adding the example from the issue with lists maybe it was the problem and not integer ?

The exact same test case fails locally. I'll investigate more.

tristanlatr:
Sorry it wasn't exactly the same test case:

```python
   def test_augassign_multi_expr(self) -> None:
        code = """
            a = 1
            a += 1
            a += 1
            a
        """
        ast = parse(code, __name__)
        inferred = list(ast.body[1].infer())
        self.assertIsInstance(inferred[0], nodes.Const)
>       self.assertEqual(inferred[0].value, 3)
E       AssertionError: 2 != 3
```

It's because we are inferring the `Expr` element, not sure what's the expected inference of Expr? Maybe this is not an issue? But the behaviour is somewhat weird.

tristanlatr:
Nevertheless, I believe the `igetattr()` method does not handle augmented assignments very well. I'll try to provide some more test cases for this.

tristanlatr:
So here's the weird behaviour:

```python
    def test_except_assign_after_block_overwritten_getattr(self) -> None:
        """When a variable is assigned in an except clause, it is not returned
        when it is reassigned and used after the except block.
        """
        code = """
            try:
                1 / 0
            except ZeroDivisionError:
                x = 10
            except NameError:
                x = 100
            x = 1000
            print(x)
        """
        astroid = builder.parse(code)
        stmts = astroid.getattr('x')
>       self.assertEqual(len(stmts), 1)
E       AssertionError: 3 != 1
```

Shouldn't it return only the last assignment?

DanielNoord:
I believe this works like intended:
```console
❯ cat test.py
from astroid import builder, extract_node

code = """
    try:
        1 / 0
    except ZeroDivisionError:
        x = 10
    except NameError:
        x = 100
    x = 1000
    x
"""
astroid = builder.parse(code)
stmts = astroid.getattr("x")

print(stmts)
node = extract_node(code)
print(list(node.infer()))
❯ python test.py
[<AssignName.x l.5 at 0x1021e8090>, <AssignName.x l.7 at 0x102289650>, <AssignName.x l.8 at 0x1022b6d10>]
[<Const.int l.8 at 0x1022b5d90>]
```

`getattr` returns all assignments in a module, whereas inferring the value at the end of the flow correctly infers what that value should be.

DanielNoord:
I think the issue should probably be resolved by taking all/the last definition into consideration. For `__get__` on `ClassDef's` the last definition should probably be enough while `getattr` should probably consider all assignments. That would require some control flow knowledge in some cases, but would avoid returning incorrect values.

PR #1615, opened 2022-06-11
[RFC] Yield self when inferring `GeneratorExp`

jacobtylerwalls:
## Description
```python
import astroid
code = """
def get_generator():
    return (x for x in range(0))
x = get_generator()  #@
if x:
  pass
"""
x = astroid.extract_node(code)
x.value.inferred()
```
Before: `[Uninferable]`
Now: `[<GeneratorExp l.3 at 0x103e80880>]`

## Type of Changes

## Related Issue
See https://github.com/PyCQA/pylint/issues/6909#issuecomment-1152723978

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -43,6 +43,7 @@ objects = util.lazy_import("objects")
 
 
 _FunctionDefT = TypeVar("_FunctionDefT", bound=nodes.FunctionDef)
+_GeneratorExpT = TypeVar("_GeneratorExpT", bound=nodes.GeneratorExp)
 
 
 # .infer method ###############################################################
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Looks good, but looking at the pipeline it seems we're going to get crashes in code that never had to deal with generator expressions before and only handle Uninferable right now.

jacobtylerwalls:
Yeah, I was looking at that. Most of the other uses of `pytype()` in pylint are guarded against there being no pytype, there's even a util for it. That one place could be changed to use that util.

However, I realize this is falling afoul of the same issue in https://github.com/PyCQA/astroid/pull/1475#discussion_r868142345, so I'll be closing. There should be another way in pylint to deal with  https://github.com/PyCQA/pylint/issues/6909#issuecomment-1152723978.

PR #1617, opened 2022-06-12
Infer modulo operations on strings

### LINKED ISSUE #151: Infer string interpolation with ``%``

---

Not sure how important this is, but we could understand string interpolation, both `msg % something`, as well as `msg.format(something)`. Currently, for the second form we are returning an empty string, which is wrong, while the first form we're returning an YES object.

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/151

DanielNoord:
## Steps

## Description

Closes #151

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,7 +26,7 @@ Release date: TBA
 
   Closes #104
 
-* Module (``%``) operations on strings are now correctly inferred.
+* Old style string formatting (using ``%`` operators) is now correctly inferred.
 
   Closes #151
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,10 @@ Release date: TBA
 
   Closes #104
 
+* Module (``%``) operations on strings are now correctly inferred.
+
+  Closes #151
+
 * Adds missing enums from ``ssl`` module.
 
   Closes PyCQA/pylint#3691
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -562,7 +562,7 @@ def _is_not_implemented(const):
     return isinstance(const, nodes.Const) and const.value is NotImplemented
 
 
-def _infer_string_modulo(
+def _infer_old_style_string_formatting(
     instance: nodes.Const, other: nodes.NodeNG, context: InferenceContext
 ) -> tuple[type[util.Uninferable] | nodes.Const]:
     """Infer the result of '"string" % ...'."""
@@ -602,7 +602,7 @@ def _invoke_binop_inference(instance, opnode, op, other, context, method_name):
     context.callcontext.callee = method
 
     if isinstance(instance, nodes.Const) and op == "%":
-        return iter(_infer_string_modulo(instance, other, context))
+        return iter(_infer_old_style_string_formatting(instance, other, context))
 
     try:
         inferred = next(method.infer(context=context))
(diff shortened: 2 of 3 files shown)


ChangeLog
```diff
@@ -26,6 +26,10 @@ Release date: TBA
 
   Closes #104
 
+* Module (``%``) operations on strings are now correctly inferred.
```
jacobtylerwalls, line 29: I think we should call it `%-formatting` to avoid sounding like the mathematical modulo operator.

tests/unittest_inference.py
```diff
@@ -6797,5 +6797,113 @@ def test_function_def_cached_generator() -> None:
...
+            "My name is %(fname)s, I'm %(age)s" % {"fsname": fsname, "age": age}
+            """,
+            """
+            fname = "Daniel"
+            age = 12
+            "My name is %0s, I'm %(age)s" % (fname, age)
+            """,
+        ],
+    )
+    def test_string_modulation_uninferable(self, format_string: str) -> None:
```
jacobtylerwalls, line 6894: Same

DanielNoord: I can't use `%` in a function name. So `percentage_formatting` then?

Pierre-Sassoulas: How about ``c_style_formatting`` ``old_string_formatting`` or ``interpolation_formatting`` ? (I checked the python doc to see how it's called : https://docs.python.org/3/tutorial/inputoutput.html#old-string-formatting)

astroid/inference.py
```diff
@@ -562,12 +562,48 @@ def _is_not_implemented(const):
...
+        values = other.value
+    else:
+        return (util.Uninferable,)
+
+    try:
+        const_node = nodes.const_factory(instance.value % values)
+    except (TypeError, KeyError):
+        return (util.Uninferable,)
+
+    return (const_node,)
```
Pierre-Sassoulas, line 594:
```suggestion
        return (nodes.const_factory(instance.value % values),)
    except (TypeError, KeyError):
        return (util.Uninferable,)
```

Pierre-Sassoulas (approved):

astroid/inference.py
```diff
@@ -562,12 +562,48 @@ def _is_not_implemented(const):
...
+    if isinstance(other, nodes.Tuple):
+        inferred_positional = [helpers.safe_infer(i, context) for i in other.elts]
+        if all(isinstance(i, nodes.Const) for i in inferred_positional):
+            values = tuple(i.value for i in inferred_positional)
+    elif isinstance(other, nodes.Dict):
+        values: dict[Any, Any] = {}
+        for pair in other.items:
+            key = helpers.safe_infer(pair[0], context)
+            if not isinstance(key, nodes.Const):
+                return (util.Uninferable,)
```
Pierre-Sassoulas, line 579: This line is not covered, I guess it's not worth it ?

astroid/inference.py
```diff
@@ -562,12 +562,48 @@ def _is_not_implemented(const):
     return isinstance(const, nodes.Const) and const.value is NotImplemented
 
 
+def _infer_old_style_string_formatting(
+    instance: nodes.Const, other: nodes.NodeNG, context: InferenceContext
```
jacobtylerwalls, line 566: I thought InferenceContext was usually optional.

jacobtylerwalls (approved):

astroid/inference.py
DanielNoord: Covered now!

astroid/inference.py
DanielNoord: Due to `callee` this is not the case. See L602 just before the `isinstance`. Such operations can't happen when `context` is `None`.

astroid/inference.py
DanielNoord: Done!

PR #1618, opened 2022-06-13
Release branch 2.11.6

Pierre-Sassoulas:


DanielNoord:
> @DanielNoord do you know how I can fix the cache issue here ? I tried to increase to above what we have on main or relaunching, or adding the new pypy version identifier commit from main, but it's not working.

There is a general outage of caches on Github. See https://www.githubstatus.com.

Pierre-Sassoulas:
Merging as the fail is due to useless suppression that we don't want to fix on a maintenance branch

Issue #713, opened 2019-10-21
Incorrect inference of ancestors of Enum class with six.with_metaclass


tkukushkin:
### Steps to reproduce
```python
import astroid

code = '''import six
from enum import Enum, EnumMeta

class FooMeta(EnumMeta):
    pass

class Foo(six.with_metaclass(FooMeta, Enum)):
    bar = 1'''

module = astroid.parse(code)
print(list(module.body[3].ancestors()))
```

### Current behavior
Prints `[<ClassDef.NoneType l.0 at 0x102bb4d68>, <ClassDef.object l.0 at 0x102bbe1d0>]`

### Expected behavior
Prints `[<ClassDef.Enum l.10 at 0x10d0617b8>, <ClassDef.object l.0 at 0x10c4650f0>]`

With another simple example without enums works ok:

```python
import astroid

code = '''import six

class FooMeta(type):
    pass

class BaseFoo(object):
    pass

class Foo(six.with_metaclass(FooMeta, BaseFoo)):
    bar = 1'''

module = astroid.parse(code)
print(list(module.body[3].ancestors()))
```
Prints: `[<ClassDef.BaseFoo l.6 at 0x1091531d0>, <ClassDef.object l.0 at 0x10853d128>]`

### ``python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"`` output
2.3.2

This behaviour breaks pylint with exception:
```
Traceback (most recent call last):
  File "bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "lib/python3.7/site-packages/pylint/__init__.py", line 19, in run_pylint
    Run(sys.argv[1:])
  File "lib/python3.7/site-packages/pylint/lint.py", line 1394, in __init__
    linter.check(args)
  File "lib/python3.7/site-packages/pylint/lint.py", line 801, in check
    self._do_check(files_or_modules)
  File "lib/python3.7/site-packages/pylint/lint.py", line 938, in _do_check
    self.check_astroid_module(ast_node, walker, rawcheckers, tokencheckers)
  File "lib/python3.7/site-packages/pylint/lint.py", line 1018, in check_astroid_module
    walker.walk(ast_node)
  File "lib/python3.7/site-packages/pylint/utils.py", line 1162, in walk
    self.walk(child)
  File "lib/python3.7/site-packages/pylint/utils.py", line 1162, in walk
    self.walk(child)
[...]

PR #1622, opened 2022-06-15
Revert #893: Avoid setting a Call as a base for classes from `six.with_metaclass()`

jacobtylerwalls:
## Steps

## Description
#893 intended to fix a false positive for `unused-import` of `six` in a pattern like `class Thing(six.with_metaclass(meta, ancestor))` by making the `nodes.Call` containing `six.with_metaclass` into a base. Then, this violated assumptions elsewhere, leading to the crash in PyCQA/pylint#5935.

#893 was itself a partial reversion of #841, a fix for #713. #713 is essentially the same issue as PyCQA/pylint#5935. So what I'm suggesting is that we revert to #841, and then find another way to deal with the `unused-import` in pylint, if at all.

(I'd much rather have a false positive than a crash. And the class of false positives is larger than anything to do with `six.with_metaclass` anyway: see PyCQA/pylint#1630. So we shouldn't treat `six` as more special.)

Let's not backport this, to give us time to make a decision about what to do in pylint with PyCQA/pylint#1630.

## Type of Changes

## Related Issue

Refs https://github.com/PyCQA/pylint/issues/5935

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,10 @@ Release date: TBA
 
 * ``astroid`` now requires Python 3.7.2 to run.
 
+* Avoid setting a Call as a base for classes created using ``six.with_metaclass()``.
+
+  Refs PyCQA/pylint#5935
+
 * Fix detection of builtins on ``PyPy`` 3.9.
 
 * Fix ``re`` brain on Python ``3.11``. The flags now come from ``re._compile``.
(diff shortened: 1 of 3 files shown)


hippo91:
@jacobtylerwalls @Pierre-Sassoulas i'm ok with this PR. 
It is indeed much preferable to have a false positive than a crash.
Sorry for #893. A bad answer to the initial problem.

Pierre-Sassoulas (approved):

PR #1624, opened 2022-06-16
Only initialize importers when they are required

DanielNoord:
## Steps

## Description

Due to some refactoring in the `spec.py` module I was able to obtain significant performance boosts for `pylint`.
For the `sentry` primer job I got local speedups of **-15%** run time.

There are two related changes here:
1) Only create instances of `Finders` when they are needed. If we can find a module with the first `finder` in `_SPEC_FINDERS` it us unnecessary to instantiate the others. Since this is a costly process we should do this inside the `for` loop.
2) Iterate over `sys.path_importer_cache` in `_search_zip`. This iteration is very costly (still about 10% of runtime). But by doing it within `_search_zip` we only do it when necessary.

I think there is a good chance this will also speed up `home-assistant` since this performance decreases grew with the number of imported modules on `path`.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,6 +44,8 @@ Release date: TBA
 * On Python versions >= 3.9, ``astroid`` now understands subscripting
   builtin classes such as ``enumerate`` or ``staticmethod``.
 
+* Speed up discovery of modules by initializing importers only when they are required.
+
 * Fixed inference of ``Enums`` when they are imported under an alias.
 
   Closes PyCQA/pylint#5776
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,8 +44,6 @@ Release date: TBA
 * On Python versions >= 3.9, ``astroid`` now understands subscripting
   builtin classes such as ``enumerate`` or ``staticmethod``.
 
-* Speed up discovery of modules by initializing importers only when they are required.
-
 * Fixed inference of ``Enums`` when they are imported under an alias.
 
   Closes PyCQA/pylint#5776
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -211,13 +211,20 @@ class ZipFinder(Finder):
 
     def __init__(self, path: Sequence[str]) -> None:
         super().__init__(path)
-        self._zipimporters = _precache_zipimporters(path)
+        for entry_path in path:
+            if entry_path not in sys.path_importer_cache:
+                try:
+                    sys.path_importer_cache[entry_path] = zipimport.zipimporter(
+                        entry_path
+                    )
+                except zipimport.ZipImportError:
+                    continue
 
     def find_module(
         self, modname, module_parts: Sequence[str], processed, submodule_path
     ):
         try:
-            file_type, filename, path = _search_zip(module_parts, self._zipimporters)
+            file_type, filename, path = _search_zip(module_parts)
         except ImportError:
             return None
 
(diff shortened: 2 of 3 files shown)


astroid/interpreter/_import/spec.py
```diff
@@ -283,79 +290,53 @@ def _cached_set_diff(left, right):
...
-        try:
-            pic[entry_path] = zipimport.zipimporter(entry_path)
-        except zipimport.ZipImportError:
-            continue
-    return {
-        key: value
-        for key, value in pic.items()
-        if isinstance(value, zipimport.zipimporter)
-    }
+# @lru_cache
```
DanielNoord, line 293:
```suggestion
```

DanielNoord:
I can't reproduce those `pylint` warnings locally...

astroid/interpreter/_import/spec.py
```diff
@@ -211,13 +211,20 @@ class ZipFinder(Finder):
 
     def __init__(self, path: Sequence[str]) -> None:
         super().__init__(path)
-        self._zipimporters = _precache_zipimporters(path)
+        for entry_path in path:
```
Pierre-Sassoulas, line 214: It seems we're still pre-caching, maybe we could add to cache only when required with a subclass and a property ?

```python
class CachedZipImporters:

    def __init__(self):
        self.cache = {}

    def __getattr__(path):
        if path not in self.cache:
            self.cache[path] = zipimport.zipimporter(path)
        return self.cache[path]
```
This way we cache only when we actually need the value ?

Pierre-Sassoulas (commented):
This is 🔥 , nice catch, the performance increase sounds amazing !

astroid/interpreter/_import/spec.py
```diff
@@ -276,86 +283,53 @@ def _is_setuptools_namespace(location: pathlib.Path) -> bool:
...
-                return (
-                    ModuleType.PY_ZIPMODULE,
-                    os.path.abspath(filepath) + os.path.sep + os.path.sep.join(modpath),
-                    filepath,
+            elif not importer.find_module(os.path.sep.join(modpath)):
+                raise ImportError(
+                    "No module named %s in %s/%s"
+                    % (".".join(modpath[1:]), filepath, modpath)
                 )
+            # import code; code.interact(local=locals())
```
jacobtylerwalls, line 313:
```suggestion
```

ChangeLog
```diff
@@ -44,6 +44,8 @@ Release date: TBA
 * On Python versions >= 3.9, ``astroid`` now understands subscripting
   builtin classes such as ``enumerate`` or ``staticmethod``.
 
+* Speed up discovery of modules by initializing importers only when they are required.
+
 * Fixed inference of ``Enums`` when they are imported under an alias.
```
jacobtylerwalls, line 49:
```suggestion
* Fixed inference of ``Enums`` when they are imported under an alias.
```

jacobtylerwalls (approved):
Terrific!

ChangeLog
DanielNoord: You don't want to add a changelog?

jacobtylerwalls: True, I don't think we don't have to document refactors.

cdce8p:
> I think there is a good chance this will also speed up `home-assistant` since this performance decreases grew with the number of imported modules on `path`.

Just a quick feedback. I haven't seen any performance differences with `home-assistant` unfortunately.

DanielNoord:
> Just a quick feedback. I haven't seen any performance differences with `home-assistant` unfortunately.

Is this locally or in CI? In CI I'm also not seeing any real difference due to the runtimes being all over the place.
Locally my traces did show significant improvements, calculated as percentage of time spent in `spec.py` as percentage of overall time.

But I agree, the benefit seems lower than those traces initially showed. Anyway, only creating instances when necessary will always be better although it might not be 10/15% 😓

cdce8p:
> Is this locally or in CI?

CI. My laptop isn't fast enough to run pylint over the whole Home Assistant codebase. Additionally I don't even have all dependencies installed 😅 Github Actions is just too convenient.

> In CI I'm also not seeing any real difference due to the runtimes being all over the place.

True, although to be fair it's possible to see larger trends over time. Especially a 10-15% improvement.

PR #1625, opened 2022-06-16
Add strict typing to ``spec.py``

DanielNoord:
## Steps

## Description

This should help with future refactoring.

## Type of Changes


DanielNoord:
If possible I'd also like to add basic typing to `modutils` in `2.12` as we use it throughout `pylint` and would benefit from it. `modutils` only uses `astroid.const` and `astroid._import` so we should be able to get pretty far after this PR has landed.

Pierre-Sassoulas (approved):
Nice !

PR #1627, opened 2022-06-16
Small performance boosts

DanielNoord:
## Steps

## Description

Kept playing around with some profiles. This is a very minor **-0.5%** boost, but that's probably still worth it.

## Type of Changes

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -174,17 +174,15 @@ class NodeNG:
             yield from context.inferred[key]
             return
 
-        generator = self._infer(context=context, **kwargs)
         results = []
 
         # Limit inference amount to help with performance issues with
         # exponentially exploding possible results.
-        limit = AstroidManager().max_inferable_values
-        for i, result in enumerate(generator):
+        limit = AstroidManager.max_inferable_values
+        for i, result in enumerate(self._infer(context=context, **kwargs)):
             if i >= limit or (context.nodes_inferred > context.max_inferred):
-                uninferable = util.Uninferable
-                results.append(uninferable)
-                yield uninferable
+                results.append(util.Uninferable)
+                yield util.Uninferable
                 break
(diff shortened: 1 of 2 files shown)


astroid/nodes/node_ng.py
```diff
@@ -174,17 +174,15 @@ def infer(
...
         # exponentially exploding possible results.
-        limit = AstroidManager().max_inferable_values
-        for i, result in enumerate(generator):
+        limit = AstroidManager.max_inferable_values
+        for i, result in enumerate(self._infer(context=context, **kwargs)):
             if i >= limit or (context.nodes_inferred > context.max_inferred):
-                uninferable = util.Uninferable
-                results.append(uninferable)
-                yield uninferable
+                results.append(util.Uninferable)
```
Pierre-Sassoulas, line 184: I don't understand why we yield and also add the value to a results list ? Shouldn't it be one or the other ?

DanielNoord: Results is being used later to store the results to the cache.

Pierre-Sassoulas (approved):

astroid/transforms.py
```diff
@@ -32,9 +32,6 @@ def _transform(self, node: NodeNG) -> NodeNG:
         transformed node.
         """
         cls = node.__class__
-        if cls not in self.transforms:
-            # no transform registered for this class of node
-            return node
 
         transforms = self.transforms[cls]
```
Pierre-Sassoulas, line 36: Is it safe to do that after removing the condition ?

DanielNoord: Yes, `transforms` is a `defaultdict` so the `if` statement is completely unnecessary.

_Performance benefit is minimal but I found it when looking at this method. Might give us a 0.00001% boost 😄_

Pierre-Sassoulas (approved):

PR #1629, opened 2022-06-17
Add typing to ``Dict._infer``

DanielNoord:
## Steps

## Type of Changes

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -41,15 +41,13 @@ if TYPE_CHECKING:
 # Prevents circular imports
 objects = util.lazy_import("objects")
 
-
+_T = TypeVar("_T")
+_BaseContainerT = TypeVar("_BaseContainerT", bound=nodes.BaseContainer)
+_NodesDictT = TypeVar("_NodesDictT", bound=nodes.Dict)
 _FunctionDefT = TypeVar("_FunctionDefT", bound=nodes.FunctionDef)
 
 
 # .infer method ###############################################################
-
-
-_T = TypeVar("_T")
-_BaseContainerT = TypeVar("_BaseContainerT", bound=nodes.BaseContainer)
 
 
 def infer_end(
@@ -120,7 +118,9 @@ nodes.Tuple._infer = infer_sequence  # type: ignore[assignment]
 nodes.Set._infer = infer_sequence  # type: ignore[assignment]
 
 
-def infer_map(self, context=None):
+def infer_map(
+    self: _NodesDictT, context: InferenceContext | None = None
+) -> Iterator[_NodesDictT | nodes.Dict]:
     if not any(isinstance(k, nodes.DictUnpack) for k, _ in self.items):
         yield self
     else:


astroid/inference.py
```diff
@@ -120,7 +118,9 @@ def infer_sequence(
 nodes.Set._infer = infer_sequence  # type: ignore[assignment]
 
 
-def infer_map(self, context=None):
+def infer_map(
+    self: _NodesDictT, context: InferenceContext | None = None
+) -> Iterator[_NodesDictT | nodes.Dict]:
```
DanielNoord, line 123: Is being explicit like this desired? Or should we just return `Iterator[nodes.Dict]`?

DanielNoord:
> Does it make sense to add types to `_infer_map` and `_update_with_replacement` here as well? With that, you could also fix the typo here:
> 
> https://github.com/PyCQA/astroid/blob/faca7915bf56ba128e198e1d483da83fd0bc4ba6/astroid/inference.py#L142
> 
> 
> => `duplicate`

I wasn't sure about `nodes.Dict.items` being correctly typed. But looking at the relevant code it might actually be correct:
https://github.com/PyCQA/astroid/blob/faca7915bf56ba128e198e1d483da83fd0bc4ba6/astroid/nodes/node_classes.py#L2266

astroid/inference.py
```diff
@@ -130,7 +131,10 @@ def infer_map(self, context=None):
         yield new_seq
 
 
-def _update_with_replacement(lhs_dict, rhs_dict):
+def _update_with_replacement(
+    lhs_dict: dict[InferenceResult, InferenceResult],
```
DanielNoord, line 135: @cdce8p I know you previously said you don't like using this alias in argument annotations, but I wonder if we should reconsider that.

I don't think it makes a lot of sense to create the alias in every module that uses inference results and it also helps a bit with readability if we just use it consistently.
We could also import the alias in `pylint` and build its typing on top of `astroid's`

astroid/inference.py
```diff
@@ -130,7 +131,10 @@ def infer_map(self, context=None):
         yield new_seq
```
DanielNoord, line 131: L129 shows that the current typing of `nodes.Dict.items` is indeed incorrect as I initially feared 😓 

`helpers.safe_infer` returns `InferenceContext | None`. In `_infer_map` we make sure to only pass `InferenceResult` to `_update_with_replacement` but that also means that the returning dictionaries are made up of `InferenceResult` and not `NodeNG`.
This also seems correct as `Instance` should be in `.items` for:
```python
class A: ...
b = {1: A()}
```

DanielNoord: I created https://github.com/PyCQA/astroid/pull/1636 to show how that affects the inference and return types here.

astroid/inference.py
```diff
@@ -153,7 +158,9 @@ def _update_with_replacement(lhs_dict, rhs_dict):
     return dict(string_map.values())
 
 
-def _infer_map(node, context):
+def _infer_map(
+    node: nodes.Dict, context: InferenceContext | None
+) -> dict[InferenceResult, InferenceResult]:
     """Infer all values based on Dict.items"""
     values = {}
```
cdce8p, line 165:
```suggestion
    values: dict[InferenceResult, InferenceResult] = {}
```

astroid/inference.py
cdce8p: I agree, the alias does make sense here.

astroid/inference.py
cdce8p: I typed `Dict.items` mainly based of the `TreeRebuilder` which explains why I only expected `dict[NodeNG, NodeNG]`. You're right though that it should be `dict[InferenceResult, InferenceResult]` instead.

astroid/inference.py
DanielNoord: Follow up:
Do we want to start importing this in `pylint` as well? Or do we still consider `astroid/typing.py` private and define `InferenceResult` in `pylint/typing.py` as well?
I'm not sure how `mypy` and `pyright` handle this. Are they able to infer that `pylint.InferenceResult` and `astroid.InferenceResult` would be the same?

astroid/inference.py
cdce8p: Let's update `Dict.items` here too, then.

astroid/inference.py
cdce8p: > Do we want to start importing this in `pylint` as well?

For `safe_infer`?

> Or do we still consider `astroid/typing.py` private and define `InferenceResult` in `pylint/typing.py` as well?

I would recommend it pre se, but `InferenceResult` should be fine.

> I'm not sure how `mypy` and `pyright` handle this. Are they able to infer that `pylint.InferenceResult` and `astroid.InferenceResult` would be the same?

They should be, once we have added `py.typed` to astroid at least. Type checkers will resolve the union and compare each type individually.

cdce8p (approved):

astroid/inference.py
DanielNoord: > For `safe_infer`?

Yeah and in other places. There are methods which get an `inferred` from their callee, which might require the full scope of `InferenceResult` because they haven't been narrowed down before the call.
 
> I would recommend it pre se, but `InferenceResult` should be fine.

Let's redefine. Saves potential headaches. I think with the addition of `Proxy` we have covered all case so we shouldn't need to update it.

> They should be, once we have added `py.typed` to astroid at least. Type checkers will resolve the union and compare each type individually.

once... 😄

PR #1632, opened 2022-06-17
Make ``LookupMixIn`` inherit from ``NodeNG``

DanielNoord:
## Steps

## Description

Fixing typing issues one message at a time.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/scoped_nodes/mixin.py
+++ b/astroid/nodes/scoped_nodes/mixin.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
 _T = TypeVar("_T")
 
 
-class LocalsDictNodeNG(node_classes.LookupMixIn, node_classes.NodeNG):
+class LocalsDictNodeNG(node_classes.LookupMixIn):
     """this class provides locals handling common to Module, FunctionDef
     and ClassDef nodes, including a dict like interface for direct access
     to locals information
@@ -49,6 +49,24 @@ class LocalsDictNodeNG(node_classes.LookupMixIn, node_classes.NodeNG):
         :rtype: Module or FunctionDef or ClassDef or Lambda or GenExpr
         """
         return self
+
+    def scope_lookup(self, node, name: str, offset: int = 0):
+        """Lookup where the given variable is assigned.
+
+        :param node: The node to look for assignments up to.
+            Any assignments after the given node are ignored.
+        :type node: NodeNG
+
+        :param name: The name of the variable to find assignments for.
+
+        :param offset: The line offset to filter statements up to.
+
+        :returns: This scope node and the list of assignments associated to the
+            given name according to the scope where it has been found (locals,
+            globals or builtin).
+        :rtype: tuple(str, list(NodeNG))
+        """
+        raise NotImplementedError
 
     def _scope_lookup(self, node, name, offset=0):
         """XXX method for interfacing the scope lookup"""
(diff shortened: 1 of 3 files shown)


astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -50,6 +50,24 @@ def scope(self: _T) -> _T:
         """
         return self
 
+    def scope_lookup(self, node, name: str, offset: int = 0):
+        """Lookup where the given variable is assigned.
+
+        :param node: The node to look for assignments up to.
+            Any assignments after the given node are ignored.
+        :type node: NodeNG
```
Pierre-Sassoulas, line 58:
```suggestion
```

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -50,6 +50,24 @@ def scope(self: _T) -> _T:
         """
         return self
 
+    def scope_lookup(self, node, name: str, offset: int = 0):
```
Pierre-Sassoulas, line 53:
```suggestion
    def scope_lookup(self, node: NodeNG, name: str, offset: int = 0) -> tuple(str, list(NodeNG):
```

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -50,6 +50,24 @@ def scope(self: _T) -> _T:
...
+        """Lookup where the given variable is assigned.
+
+        :param node: The node to look for assignments up to.
+            Any assignments after the given node are ignored.
+        :type node: NodeNG
+
+        :param name: The name of the variable to find assignments for.
+
+        :param offset: The line offset to filter statements up to.
+
```
Pierre-Sassoulas, line 63:
```suggestion
        :param offset: The line offset to filter statements up to.
```

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -50,6 +50,24 @@ def scope(self: _T) -> _T:
...
+        :returns: This scope node and the list of assignments associated to the
+            given name according to the scope where it has been found (locals,
+            globals or builtin).
+        :rtype: tuple(str, list(NodeNG))
```
Pierre-Sassoulas, line 67:
```suggestion
```

Pierre-Sassoulas (changes requested):
I guess it make sense to make the mixin inherit from the base class for node as they will always be used in a child class of it.

astroid/nodes/scoped_nodes/mixin.py
DanielNoord: I'm not sure this is actually true. Wherever `node: NodeNG` I have become real suspicious that it should be `node: NodeNG | Instance`.
I'll investigate it later, `mypy` at least passes now.

That's also why I let the docstring in.

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -50,6 +50,24 @@ def scope(self: _T) -> _T:
...
+        """
+        raise NotImplementedError
```
Pierre-Sassoulas, line 69:
```suggestion
        raise NotImplementedError  # pragma: no cover
```

Pierre-Sassoulas (approved):

astroid/nodes/scoped_nodes/mixin.py
DanielNoord: I'll add this to the last of standard uncovered lines.

PR #1634, opened 2022-06-17
Fix small typing issues

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -361,7 +361,11 @@ class InspectBuilder:
         except AttributeError:
             # in jython, java modules have no __doc__ (see #109562)
             node = build_module(modname)
-        node.file = node.path = os.path.abspath(path) if path else path
+        if path is None:
+            node.path = node.file = path
+        else:
+            node.path = [os.path.abspath(path)]
+            node.file = node.path[0]
         node.name = modname
         self._manager.cache_module(node)
         node.package = hasattr(module, "__path__")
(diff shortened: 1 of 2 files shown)


astroid/raw_building.py
```diff
@@ -361,7 +361,11 @@ def inspect_build(
         except AttributeError:
             # in jython, java modules have no __doc__ (see #109562)
             node = build_module(modname)
-        node.file = node.path = os.path.abspath(path) if path else path
+        if path is None:
+            node.path = node.file = path
+        else:
+            node.path = [os.path.abspath(path)]
```
Pierre-Sassoulas, line 367: This line look like a change in behavior.

Pierre-Sassoulas: Hmm.. and we might not crash outright because a string is iterable... I wonder if it will magically fix issues. Did you try the result in pylint ?

DanielNoord: `pylint` test suite still passes.

I think it's also logical that we never use this: why would we ever use `builtins.path`? We don't really need it to import anything and in the rare case we do want to check the path of the `builtins` module `sys` offers way better attributes and methods to determine the location of all builtin modules.

DanielNoord: I haven't actually tested this, but I wouldn't be surprised if `path` is actually `None` here and we never create the list. It's just that `mypy` complains about the assignment of a `str` without it ever occurring.

Pierre-Sassoulas (approved):

PR #1636, opened 2022-06-19
Type safe_infer, Proxy.infer and fix InferenceResult

DanielNoord:
## Steps

## Description

See also https://github.com/PyCQA/astroid/pull/1629.

## Type of Changes

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -143,7 +143,7 @@ def object_issubclass(node, class_or_seq, context=None):
 
 # TODO: PY310: Remove the disable after InferenceResult no longer uses Union
 def safe_infer(  # pylint: disable=consider-alternative-union-syntax
-    node: Union[bases.Proxy, nodes.NodeNG],
+    node: Union[nodes.NodeNG, bases.Proxy],
     context: Union[InferenceContext, None] = None,
 ) -> Union[InferenceResult, None]:
     """Return the inferred value for the given node.
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -7,7 +7,7 @@ Various helper utilities.
 """
 
 
-from typing import Union
+from typing import Optional, Union
 
 from astroid import bases, manager, nodes, raw_building, util
 from astroid.context import CallContext, InferenceContext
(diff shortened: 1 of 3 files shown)


astroid/typing.py
```diff
@@ -41,4 +41,4 @@ class AstroidManagerBrain(TypedDict):
     _transform: transforms.TransformVisitor
 
 
-InferenceResult = Union["nodes.NodeNG", "type[util.Uninferable]", "bases.Instance"]
+InferenceResult = Union["nodes.NodeNG", "type[util.Uninferable]", "bases.Proxy"]
```
jacobtylerwalls, line 44: Just skimming quickly -- do you mind explaining this part?

DanielNoord: See L124 of `bases.py`. We initially thought that `Instance` was the "lowest" type of the `bases` that could be returned by `_infer` but it turns out it is `Proxy`.

Any inference in `astroid` can thus either return a node, a proxy or be uninferable.

jacobtylerwalls: Sounds good. Looks like `Proxy` could be an abstract class someday?

DanielNoord: > Sounds good. Looks like `Proxy` could be an abstract class someday?

Probably. I don't think we ever instantiate one directly, but I haven't checked.

astroid/helpers.py
```diff
@@ -138,7 +141,11 @@ def object_issubclass(node, class_or_seq, context=None):
     return _object_type_is_subclass(node, class_or_seq, context=context)
 
 
-def safe_infer(node, context=None):
+# TODO: PY310: Remove the disable after InferenceResult no longer uses Union
+def safe_infer(  # pylint: disable=consider-alternative-union-syntax
```
cdce8p, line 145: What has the `Union` in `InferenceResult` to do with the syntax here?

DanielNoord: This crashes without the `Union`. `pylint` doesn't recognise this. See:
https://github.com/PyCQA/astroid/runs/6956094603?check_suite_focus=true

Or at least, I assumed it's because of the both `Union` and `|`.

astroid/helpers.py
```diff
@@ -138,7 +141,11 @@ def object_issubclass(node, class_or_seq, context=None):
...
+    node: Union[bases.Proxy, nodes.NodeNG],
```
DanielNoord, line 146:
```suggestion
    node: Union[nodes.NodeNG, bases.Proxy],
```

astroid/helpers.py
cdce8p: Try adding `from __future__ import annotations`

DanielNoord: 😓

cdce8p (changes requested):

cdce8p (approved):

PR #1637, opened 2022-06-20
Correct typing of ``relative_to_absolute_name``

DanielNoord:
## Steps

## Description

Working on https://github.com/PyCQA/pylint/pull/6982 showed that this typing was incorrect.

It originated from https://github.com/PyCQA/astroid/pull/1211 which was passed on the typing in the docstring. However, this method can also be called with `None` values, as the `do_import_module` method shows.

## Type of Changes

## Related Issue


Pierre-Sassoulas (approved):
👍

PR #1640, opened 2022-06-20
Move ``Statement`` into ``_base_nodes``

DanielNoord:
## Steps

## Type of Changes

## Related Issue


DanielNoord:
Many of the `MixIns` actually should inherit `Statement`. I think `Statement` itself should also be considered a base node as it is very similar to all other `Mixins`/base nodes: it adds 2/3 new methods to nodes.

However, we use it in `pylint` as well for various typing stuff so I think it's best to re-export it as a public class and make that explicit in the docstring.

Pierre-Sassoulas (approved):

PR #1643, opened 2022-06-20
Remove remaining references to old Mixin base nodes

DanielNoord:
## Steps

## Description

Final PR 🎉 

## Type of Changes

## Related Issue


DanielNoord:
> 👍 We almost always forget to upgrade the doc it seems. There's a lot of documentation work to do in pylint but astroid's doc is just... completely forgotten.

Some of the `_base_nodes` are also not in this list. But since they are now private I'm not sure we should even list them here.
I just wanted to make sure I had touched all references to `mixins` so that the clean up in `3.0` is quick and easy.

PR #1648, opened 2022-06-21
Fix typing of ``Dict`` with ``SuccessfulInferenceResult``

DanielNoord:
## Steps

## Description

@cdce8p I noticed a mistake. I'll annotate.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -33,7 +33,7 @@ from astroid.exceptions import (
 )
 from astroid.interpreter import dunder_lookup
 from astroid.manager import AstroidManager
-from astroid.typing import InferenceErrorInfo, InferenceResult
+from astroid.typing import InferenceErrorInfo, SuccessfulInferenceResult
 
 if TYPE_CHECKING:
     from astroid.objects import Property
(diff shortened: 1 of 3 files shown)


astroid/inference.py
```diff
@@ -174,10 +174,11 @@ def _infer_map(
             values = _update_with_replacement(values, unpack_items)
         else:
             key = helpers.safe_infer(name, context=context)
-            value = helpers.safe_infer(value, context=context)
-            if any(not elem for elem in (key, value)):
+            safe_value = helpers.safe_infer(value, context=context)
+            if any(not elem for elem in (key, safe_value)):
```
DanielNoord, line 178: L178 makes sure that `safe_value` on L181 is only `type[NodeNG] | type[Proxy]` as `Uninferable` is also filtered here.

This is crucial as that means that `Uninferable` will never end up in `Dict.items` as an `InferenceError` will be raised.

Sorry for the churn, I didn't know `Uninferable` did that and didn't recognise `mypy` warning us about it.

Issue #1649, opened 2022-06-21
Items in ``nodes.Dict`` are only inferred if there is a `DictUnpack` in them.


DanielNoord:
```python
import astroid

dict1, dict2 = astroid.extract_node(
    """
class A:
    ...
class B:
    ...

b = {A(): 1}
dict1 = {B(): 1} #@
dict2 = {**b, B(): 1} #@
"""
)


print(list(dict1.value.infer())[0].items)
print(list(dict2.value.infer())[0].items)
```

Creates:
```console
[(<Call l.8 at 0x10bb87510>, <Const.int l.8 at 0x10bb86990>)]
[(<Instance of .A at 0x4491606736>, <Const.int l.7 at 0x10bb86710>), (<Instance of .B at 0x4491794192>, <Const.int l.9 at 0x10bb87690>)]
```

Caused by:
https://github.com/PyCQA/astroid/blob/6e36cd88ebc181a35e148b6418dd6b898e7b361a/astroid/inference.py#L126-L127

But should probably be fixed in:
https://github.com/PyCQA/astroid/blob/6e36cd88ebc181a35e148b6418dd6b898e7b361a/astroid/nodes/node_classes.py#L5394-L5399

I have a WIP PR for `const_factory` though which I'd like to merge before trying to infer `elts`/`value` here.

Issue #1479, opened 2022-03-17
Rename of branches is incorrect


DanielNoord:
@Pierre-Sassoulas I think you might have done this, but I'm not sure.

Some time ago all branch names have been updated. This also included the `2.0` branch into `maintenance/2.0.x`. However, this was not the `2.0.x` branch but rather a WIP branch for a complete overhaul of `astroid`, which has sadly failed.
See for example this issue https://github.com/PyCQA/astroid/issues/679, which refers to the branch and now links to the incorrectly named branch.

Since I did not do the change I do not know if there are any other branches that have been renamed incorrectly, but I would ask the person who did so to revisit the renaming and see if there are any other issues.

Pierre-Sassoulas:
Yeap, that's me 😄

Pierre-Sassoulas:
Renamed to ``2.0.experimental``

Pierre-Sassoulas:
Yeah maybe all of them but the current one ? I was keeping them because they were there , but I think you're right. (And we really don't want to maintain astroid 2.1)

Pierre-Sassoulas:
I don't know what we should do with the 2.0 experimental branch now.

DanielNoord:
Let's keep it around. There is no tag for it and it is good reference to look into for some open issues + what other solutions to common problems have been thgouht of.

PR #1650, opened 2022-06-21
Add typing to ``const_factory``

DanielNoord:
## Steps

## Description

This looks like a big chance, but I tested and there is no change in functionality. It's just all a bit more explicit.

I have thought about `overloading` it so `value: str -> Const`, but that's not actually true as `str` also covers all subclasses of `str`. That's not what this function does.

I have also tested if we should just create a node if `value` is a subclass of `str` (or any other `CONST_CLS` but then `sys.version_info` gets inferred as `Tuple`, while it is an instance of `sys.version_info`. Which itself is a subclass of `tuple`.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -51,9 +51,7 @@ def _is_const(value):
     return isinstance(value, tuple(CONST_CLS))
 
 
-_ContainerNodesT = TypeVar(
-    "_ContainerNodesT", bound=Union["List", "Set", "Tuple", "Dict"]
-)
+_ContainerNodesT = TypeVar("_ContainerNodesT", "List", "Set", "Tuple", "Dict")
 _NodesT = TypeVar("_NodesT", bound=NodeNG)
 
 AssignedStmtsPossibleNode = Union["List", "Tuple", "AssignName", "AssignAttr", None]
@@ -5371,7 +5369,7 @@ def _two_step_initialization(
 ) -> _ContainerNodesT:
     instance = cls()
     instance.postinit(value)
-    return instance  # type: ignore[return-value]
+    return instance
 
 
 def _dict_initialization(cls: type[Dict], value: list[Any]) -> Dict:
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -5374,10 +5374,6 @@ def _two_step_initialization(
     return instance  # type: ignore[return-value]
 
 
-def _dict_initialization(cls: type[Dict], value: list[Any]) -> Dict:
-    return _two_step_initialization(cls, value)
-
-
 def const_factory(value: Any) -> List | Set | Tuple | Dict | Const | EmptyNode:
     """Return an astroid node for a python value."""
     assert not isinstance(value, NodeNG)
@@ -5395,8 +5391,6 @@ def const_factory(value: Any) -> List | Set | Tuple | Dict | Const | EmptyNode:
     # each element to an AST. This is fixed in 2.0
     # and this approach is a temporary hack.
     initializer_cls = CONST_CLS[value.__class__]
-    if issubclass(initializer_cls, (List, Set, Tuple)):
+    if issubclass(initializer_cls, (Dict, List, Set, Tuple)):
         return _two_step_initialization(initializer_cls, [])  # type: ignore[return-value, type-var]
-    if issubclass(initializer_cls, Dict):
-        return _dict_initialization(initializer_cls, [])
     return Const(value)
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -5392,12 +5392,9 @@ def const_factory(value: Any) -> List | Set | Tuple | Dict | Const | EmptyNode:
     # or a mapping, in order to avoid transforming
     # each element to an AST. This is fixed in 2.0
     # and this approach is a temporary hack.
-    if isinstance(value, list):
-        return _two_step_initialization(List, [])
-    if isinstance(value, set):
-        return _two_step_initialization(Set, [])
-    if isinstance(value, tuple):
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
This look pretty good ! I'm wondering if we could also get rid of the ``_two_step_initialization`` function ?

DanielNoord:
> This look pretty good ! I'm wondering if we could also get rid of the `_two_step_initialization` function ?

Well, it makes it easier to at some point tackle #1649. It seems to be the setup to it.

We could look on the `maintenance/2.0` branch (which you still need to re-rename #1479 😄) to see what they did there. The previous comment says that they have fixed the issue there.

astroid/nodes/node_classes.py
```diff
@@ -5344,71 +5347,59 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
     # or a mapping, in order to avoid transforming
     # each element to an AST. This is fixed in 2.0
     # and this approach is a temporary hack.
+    if isinstance(value, list):
+        return _two_step_initialization(List, [])
+    if isinstance(value, set):
+        return _two_step_initialization(Set, [])
+    if isinstance(value, tuple):
+        return _two_step_initialization(Tuple, [])
     if isinstance(value, (list, set, tuple, dict)):
```
cdce8p, line 5403:
```suggestion
    if isinstance(value, dict):
```

astroid/nodes/node_classes.py
```diff
@@ -5344,71 +5347,59 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
-    try:
-        initializer_cls = CONST_CLS[value.__class__]
-        initializer = _CONST_CLS_CONSTRUCTORS[initializer_cls]
-        return initializer(initializer_cls, elts)
-    except (KeyError, AttributeError):
-        node = EmptyNode()
-        node.object = value
-        return node
+        return _dict_initialization(Dict, [])
+    return Const(value)
```
cdce8p, line 5405: We do get a mapping from `CONST_CLS`. Maybe we could use that to speed up the if block here. Would something like this work?
```py
    initializer_cls = CONST_CLS[value.__class__]
    if isinstance(initializer_cls, (List, Set, Tuple)):
        return _two_step_initialization(initializer_cls, [])
    if isinstance(initializer_cls, Dict):
        return _dict_initialization(initializer_cls, [])
    return Const(value)
```

DanielNoord: I added this in the last commit but used `issubclass` as we're comparing classes. This doesn't pass `pyright` though and I don't really understand why. `mypy` is also having trouble with it.
Could you perhaps have a quick look locally?

cdce8p: > I added this in the last commit but used `issubclass` as we're comparing classes.

👍🏻 


Revert the TypeVar change, i.e. use `bound=Union["List", "Set, "Tuple", "Dict"]`. That will at least fix the pyright errors. Mypy somehow tries to find a common ancestor in `BaseContainer` and passes that along. Probably a bug.

DanielNoord: This also required re-adding the initial ignore. Let me know what you think!

astroid/nodes/node_classes.py
```diff
@@ -5344,71 +5347,56 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
+
+    # TODO: We pass an empty list as elements for a sequence
+    # or a mapping, in order to avoid transforming
+    # each element to an AST. This is fixed in 2.0
+    # and this approach is a temporary hack.
+    initializer_cls = CONST_CLS[value.__class__]
+    if issubclass(initializer_cls, (List, Set, Tuple)):
+        return _two_step_initialization(initializer_cls, [])  # type: ignore[return-value, type-var]
+    if issubclass(initializer_cls, Dict):
+        return _dict_initialization(initializer_cls, [])
```
cdce8p, line 5401: I'm wondering, `_dict_initialization` just delegates to `_two_step_initialization` anyway. Can the `if issubclass` checks be merged?

astroid/nodes/node_classes.py
cdce8p: > This also required re-adding the initial ignore. Let me know what you think!

Not ideal, but that seems to be the best possible typing.

astroid/nodes/node_classes.py
DanielNoord: Removed.

astroid/nodes/node_classes.py
```diff
@@ -5344,71 +5347,50 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
         node.object = value
         return node
+
+    # TODO: We pass an empty list as elements for a sequence
+    # or a mapping, in order to avoid transforming
+    # each element to an AST. This is fixed in 2.0
+    # and this approach is a temporary hack.
+    initializer_cls = CONST_CLS[value.__class__]
+    if issubclass(initializer_cls, (Dict, List, Set, Tuple)):
+        return _two_step_initialization(initializer_cls, [])  # type: ignore[return-value, type-var]
```
cdce8p, line 5395: We could probably inline the `_two_step_initialization` function, now that it's only called once. Maybe that resolves the type errors.

cdce8p:
> @cdce8p This is actually not tested by the way, although it is covered.
> 
> In an early iteration I actually removed the `EmptyNode` and both `pylint` and `astroid` passed. For example, inferring `sys.version_info` as `Tuple` instead of `Uninferable` doesn't break any of the tests. Should we add a test for this? It felt weird to add a test to show that we aren't able to infer something correctly.

I just used the branch to run pylint over Home Assistant without issues. That's why I added the label.
Tbh I'm not too familiar with that part of the codebase, so I don't know at the moment if tests would make sense. For now I just looked at the changes itself to determine that it's only a refactoring.

PR #1652, opened 2022-06-21
Fix type issue in TreeRebuilder added with change to Dict.items

cdce8p:
## Description
Add additional type hints to fix issue with `Dict.items` changed in #1629 and #1648.

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -22,6 +22,7 @@ from astroid.const import IS_PYPY, PY38, PY38_PLUS, PY39_PLUS, Context
 from astroid.manager import AstroidManager
 from astroid.nodes import NodeNG
 from astroid.nodes.utils import Position
+from astroid.typing import SuccessfulInferenceResult
 
 if sys.version_info >= (3, 8):
     from typing import Final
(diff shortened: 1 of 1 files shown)


astroid/rebuilder.py
```diff
@@ -1086,7 +1087,9 @@ def visit_dict(self, node: ast.Dict, parent: NodeNG) -> nodes.Dict:
             end_col_offset=getattr(node, "end_col_offset", None),
             parent=parent,
         )
-        items = list(self._visit_dict_items(node, parent, newnode))
+        items: list[tuple[SuccessfulInferenceResult, SuccessfulInferenceResult]] = list(
```
DanielNoord, line 1090: For now this correct, but at some point we'll likely want to change `_visit_dict_items` as I think it is also related to https://github.com/PyCQA/astroid/issues/1649.

I'm fine with adding this here, but shouldn't we just update that method and "future-proof" it?

cdce8p: How would you "future-proof" `_visit_dict_items`? The TreeRebuild doesn't do inference, so the return type should be correct with `Generator[tuple[NodeNG, NodeNG], None, None]`.

The annotation is only necessary since lists are invariant.

DanielNoord: Ah yeah, it was late and I wasn't thinking right. The inference of `Dict.items` should happen at a later stage.

DanielNoord (approved):

PR #1655, opened 2022-06-23
Fix typing of ``_infer`` to allow a return value in the ``Generator``

DanielNoord:
## Steps

## Description

Ref. https://github.com/PyCQA/astroid/pull/1653.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import collections
 import collections.abc
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any
 
 from astroid import decorators
 from astroid.const import PY310_PLUS
@@ -26,7 +26,7 @@ from astroid.exceptions import (
     InferenceError,
     NameInferenceError,
 )
-from astroid.typing import InferenceResult
+from astroid.typing import InferenceErrorInfo, InferenceResult
 from astroid.util import Uninferable, lazy_descriptor, lazy_import
 
 objectmodel = lazy_import("interpreter.objectmodel")
@@ -36,7 +36,6 @@ manager = lazy_import("manager")
 if TYPE_CHECKING:
     from astroid import nodes
 
-_ProxyT = TypeVar("_ProxyT", bound="Proxy")
 
 # TODO: check if needs special treatment
 BOOL_SPECIAL_METHOD = "__bool__"
@@ -119,9 +118,9 @@ class Proxy:
             return self.__dict__[name]
         return getattr(self._proxied, name)
 
-    def infer(
-        self: _ProxyT, context: InferenceContext | None = None, **kwargs: Any
-    ) -> collections.abc.Generator[_ProxyT, None, None]:
+    def infer(  # type: ignore[return]
+        self, context: InferenceContext | None = None, **kwargs: Any
+    ) -> collections.abc.Generator[InferenceResult, None, InferenceErrorInfo | None]:
         yield self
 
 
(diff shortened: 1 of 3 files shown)


astroid/bases.py
```diff
@@ -119,9 +118,9 @@ def __getattr__(self, name):
             return self.__dict__[name]
         return getattr(self._proxied, name)
 
-    def infer(
-        self: _ProxyT, context: InferenceContext | None = None, **kwargs: Any
-    ) -> collections.abc.Generator[_ProxyT, None, None]:
+    def infer(  # type: ignore[return]
```
DanielNoord, line 121: `mypy` complains about wanting a `return None` statement whereas `pylint` complains about `useless-return`. Who is right here?

astroid/nodes/node_ng.py
```diff
@@ -593,7 +593,7 @@ def _infer_name(self, frame, name):
 
     def _infer(
         self, context: InferenceContext | None = None, **kwargs: Any
-    ) -> Generator[InferenceResult, None, None]:
+    ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
```
DanielNoord, line 596: We already defined this for `FunctionDef._infer`

astroid/objects.py
```diff
@@ -333,5 +333,5 @@ def infer_call_result(self, caller=None, context=None):
 
     def _infer(
         self: _T, context: InferenceContext | None = None, **kwargs: Any
-    ) -> Iterator[_T]:
+    ) -> Generator[_T, None, None]:
```
DanielNoord, line 336: I thought I would fix this at the same time as well as it is somewhat related.

astroid/bases.py
Pierre-Sassoulas: I don't think you **need** an explicit return at the end of a generator, so I think mypy is wrong... But they say in their doc:

> You should give a statically typed function an explicit None return type even if it doesn’t return a value, as this lets mypy catch additional type errors:

So you must be explicit to help mypy it seems.

DanielNoord: I think that refers to:
```python
def func(x: str):
    print(x)
```

Which should have a `-> None` as well.

cdce8p (approved):

astroid/bases.py
DanielNoord: Yeah I'll create a small reproducer in the coming days.

DanielNoord: Reported at https://github.com/python/mypy/issues/13019.

DanielNoord: @cdce8p See `mypy` issue. This is working as intended. Should we update `pylint` to allow `return None` for annotated functions?

cdce8p: > But the real question is what should the default value be. mypy is used a lot so probably the one mypy chose to reduce friction.

I would agree. Both checks (mypy and pylint) are opinionated. The only thing for us is to make sure they don't collide if a users chooses to enable either one.

> Should we update `pylint` to allow `return None` for annotated functions?

Would it make sense to go one step further and disable `useless-return` entirely by default?

Pierre-Sassoulas: > Would it make sense to go one step further and disable useless-return entirely by default?

Yeap, we don't need an option, my bad.

DanielNoord: We should really tweak the default values of `pylint`.

Pierre-Sassoulas: >We should really tweak the default values of pylint.

Agreed. I'm thinking about applying 3512 but there's a lot of other issues to handle and I never have the time. This is probably one of the big pain point of pylint right now though. The issue that take us a lot of time can be very niche in very specific cases sometime but default values are really freaking important. (Also invalid-name for < 3 char value. I get bitten by it for new project myself, first thing I need to do is always to disable this warning, someone trying pylint always talk to me about it, it's not a good look).

PR #1656, opened 2022-06-23
Add brain for numpy core module ``einsumfunc``.

mbyrnepr2:
Refs PyCQA/pylint#5821

## Steps

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -71,7 +71,7 @@ Release date: TBA
 
 * Add brain for numpy core module ``einsumfunc``.
 
-  Refs PyCQA/pylint#5821
+  Closes PyCQA/pylint#5821
 
 * Infer the ``DictUnpack`` value for ``Dict.getitem`` calls.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -69,6 +69,10 @@ Release date: TBA
 * Fix test for Python ``3.11``. In some instances ``err.__traceback__`` will
   be uninferable now.
 
+* Add brain for numpy core module ``einsumfunc``.
+
+  Refs PyCQA/pylint#5821
+
 * Infer the ``DictUnpack`` value for ``Dict.getitem`` calls.
 
   Closes #1195
--- a/astroid/brain/brain_numpy_core_einsumfunc.py
+++ b/astroid/brain/brain_numpy_core_einsumfunc.py
@@ -12,7 +12,7 @@ from astroid.builder import parse
 from astroid.manager import AstroidManager
 
 
-def numpy_core_einsumfunc_transform():
+def numpy_core_einsumfunc_transform() -> nodes.Module:
     return parse(
         """
     def einsum(*operands, out=None, optimize=False, **kwargs):
--- a/astroid/brain/brain_numpy_core_einsumfunc.py
+++ b/astroid/brain/brain_numpy_core_einsumfunc.py
@@ -1,0 +1,26 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/PyCQA/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/astroid/blob/main/CONTRIBUTORS.txt
+
+"""
+Astroid hooks for numpy.core.einsumfunc module:
+https://github.com/numpy/numpy/blob/main/numpy/core/einsumfunc.py
+"""
+
+from astroid.brain.helpers import register_module_extender
+from astroid.builder import parse
+from astroid.manager import AstroidManager
+
+
+def numpy_core_einsumfunc_transform():
+    return parse(
+        """
+    def einsum(*operands, out=None, optimize=False, **kwargs):
+        return numpy.ndarray([0, 0])
+    """
+    )
+
+
+register_module_extender(
+    AstroidManager(), "numpy.core.einsumfunc", numpy_core_einsumfunc_transform
+)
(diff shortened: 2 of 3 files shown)


tests/unittest_brain_numpy_core_einsumfunc.py
```diff
@@ -0,0 +1,80 @@
...
+
+    HAS_NUMPY = True
+except ImportError:
+    HAS_NUMPY = False
+
+from astroid import builder
+
+
+@unittest.skipUnless(HAS_NUMPY, "This test requires the numpy library.")
+class BrainNumpyCoreFromNumericTest(unittest.TestCase):
```
DanielNoord, line 22: Could you rewrite these test without using `unittest`? I think we're trying to only add `pytest` style tests but have decided that rewriting old tests is not worth it (atm..)

tests/unittest_brain_numpy_core_einsumfunc.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Test the numpy core einsumfunc brain module
+    """
```
DanielNoord, line 25:
```suggestion
    """Test the numpy core einsumfunc brain module."""
```

At some point we should turn on `pydocstringformatter` here as well.

ChangeLog
```diff
@@ -69,6 +69,10 @@ Release date: TBA
 * Fix test for Python ``3.11``. In some instances ``err.__traceback__`` will
   be uninferable now.
 
+* Add brain for numpy core module ``einsumfunc``.
+
+  Refs PyCQA/pylint#5821
```
DanielNoord, line 74:
```suggestion
  Closes PyCQA/pylint#5821
```

astroid/brain/brain_numpy_core_einsumfunc.py
```diff
@@ -0,0 +1,26 @@
...
+Astroid hooks for numpy.core.einsumfunc module:
+https://github.com/numpy/numpy/blob/main/numpy/core/einsumfunc.py
+"""
+
+from astroid.brain.helpers import register_module_extender
+from astroid.builder import parse
+from astroid.manager import AstroidManager
+
+
+def numpy_core_einsumfunc_transform():
```
DanielNoord, line 15:
```suggestion
def numpy_core_einsumfunc_transform() -> nodes.Module:
```

DanielNoord (changes requested):
Not sure why this landed in PR limbo, but I thought I'd start with a review.

More general question: why are we putting al `numpy` brains into separate files? Seems like a bit of a waste, right? Or am I missing something?

mbyrnepr2:
> Not sure why this landed in PR limbo, but I thought I'd start with a review.
> 
> More general question: why are we putting al `numpy` brains into separate files? Seems like a bit of a waste, right? Or am I missing something?

Thanks!

I don't know the story behind this but to my mind it seems more maintainable to have them as separate modules, especially if if they increase in size over time. Perhaps we could create a separate sub-package `astroid/brain/numpy/` to collect them in one place?

Or perhaps your point is more related to performance @DanielNoord?

DanielNoord:
> I don't know the story behind this but to my mind it seems more maintainable to have them as separate modules, especially if they increase in size over time. 

Yeah I think a refactor might be beneficial as I'm not sure this is actually more maintainable. For example, the brain you added could have been added to https://github.com/PyCQA/astroid/blob/main/astroid/brain/brain_numpy_core_fromnumeric.py.
That would have removed 17 unnecessary lines from this diff in the brain file alone.

Same probably goes for test files.

mbyrnepr2:
Hi @DanielNoord I think you have good ideas. As a way forward may I suggest you create a new issue to capture the ideas behind your refactoring suggestion? I'm sure there are knowledgable folks who are aware of the reasoning behind the existing structure who can provide input & help move that forward.

The current issue can move forward independently & according to the existing structure; if the refactoring issue comes to fruition then it wouldn't be much extra work to incorporate einsumfunc brain into the new system.

What do you think?

DanielNoord:
I'm fine with creating a new issue, but I'm not sure if there is anybody around anymore that knows the original rationale.

Why don't approach this from what we would want ourselves. If you think keeping these separate is better that's fine with me, but if you think it's better to merge them then let's do so now!

mbyrnepr2:
Mainly I think for now I'd like to follow the existing structure in order to stick to the scope of fixing the Pylint issue; hopefully I have time to finish that much this week.
Regarding the refactor itself - I'm in two minds because I think the structure now is helpful with the 1:1 of brain to numpy module mapping; but perhaps I am missing something.

tests/unittest_brain_numpy_core_einsumfunc.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.mark.skipif(not HAS_NUMPY, reason="This test requires the numpy library.")
+@pytest.mark.parametrize(
+    "method, args",
+    [
+        ("einsum", "ii, np.arange(25).reshape(5, 5)"),
+    ],
+)
+def test_numpy_function_calls_inferred_as_ndarray(method, args):
```
DanielNoord, line 37:
```suggestion
def test_numpy_function_calls_inferred_as_ndarray(method: str, args: str) -> None:
```

DanielNoord (changes requested):

tests/unittest_brain_numpy_core_einsumfunc.py
mbyrnepr2: This isn't relevant if we remove the parametrize decorator.

DanielNoord (approved):

DanielNoord:
Pushed some final changes. Thanks @mbyrnepr2 😄

PR #1657, opened 2022-06-24
Type ``_infer`` of Attribute, AssignAttr, Global and Subscript

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -345,7 +345,6 @@ def infer_attribute(
             continue
 
         context = copy_context(context)
-
         old_boundnode = context.boundnode
         try:
             context.boundnode = owner
@@ -371,7 +370,7 @@ nodes.AssignAttr.infer_lhs = infer_attribute
 def infer_global(
     self: nodes.Global, context: InferenceContext | None = None, **kwargs: Any
 ) -> Generator[InferenceResult, None, None]:
-    if not context or context.lookupname is None:
+    if context is None or context.lookupname is None:
         raise InferenceError(node=self, context=context)
     try:
         return bases._infer_stmts(self.root().getattr(context.lookupname), context)
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -334,7 +334,12 @@ def infer_import_from(
 nodes.ImportFrom._infer = infer_import_from  # type: ignore[assignment]
 
 
-def infer_attribute(self, context=None):
+@decorators.raise_if_nothing_inferred
+def infer_attribute(
+    self: nodes.Attribute | nodes.AssignAttr,
+    context: InferenceContext | None = None,
+    **kwargs: Any,
+) -> Generator[InferenceResult, None, InferenceErrorInfo]:
     """infer an Attribute node by using getattr on the associated object"""
     for owner in self.expr.infer(context):
         if owner is util.Uninferable:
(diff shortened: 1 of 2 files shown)


cdce8p (approved):

Pierre-Sassoulas (approved):
LGTM, nice cleanup.

astroid/inference.py
```diff
@@ -346,10 +344,7 @@ def infer_attribute(
             yield owner
             continue
 
-        if not context:
-            context = InferenceContext()
-        else:
-            context = copy_context(context)
+        context = copy_context(context)
```
cdce8p, line 348:
```suggestion
```
Empty line

astroid/inference.py
```diff
@@ -358,20 +358,20 @@ def infer_attribute(self, context=None):
...
 
 
 @decorators.raise_if_nothing_inferred
 @decorators.path_wrapper
-def infer_global(self, context=None):
-    if context.lookupname is None:
+def infer_global(
+    self: nodes.Global, context: InferenceContext | None = None, **kwargs: Any
+) -> Generator[InferenceResult, None, None]:
+    if not context or context.lookupname is None:
```
cdce8p, line 374:
```suggestion
    if context is None or context.lookupname is None:
```

cdce8p (approved):

cdce8p:
Somehow, this changes causes a pylint failure with Home Assistant.
https://github.com/cdce8p/ha-core/runs/7053862874?check_suite_focus=true#step:8:93

I'm currently running a test to see if a revert fixes the issue.-
https://github.com/cdce8p/ha-core/runs/7053960620?check_suite_focus=true

DanielNoord:
Could it be that we actually depend on the crash of the missing `context`? And that we shouldn't create a new context there?

cdce8p:
> Could it be that we actually depend on the crash of the missing `context`? And that we shouldn't create a new context there?

Could be. Will need to investigate that. For now, I would suggest to revert it completely. That does fix it https://github.com/cdce8p/ha-core/runs/7053960620?check_suite_focus=true

cdce8p:
**Update**
Not sure why, but I narrowed it down to the decorator changes: https://github.com/PyCQA/astroid/commit/eb1b58e59a90dcd20f8180e8436109545ddbd8c2
The context modifications seem to be fine.

DanielNoord:
> **Update**
> Not sure why, but I narrowed it down to the decorator changes: [eb1b58e](https://github.com/PyCQA/astroid/commit/eb1b58e59a90dcd20f8180e8436109545ddbd8c2)
> The context modifications seem to be fine.

Should we re-commit without the decorator changes?

cdce8p:
> > **Update**
> > Not sure why, but I narrowed it down to the decorator changes: [eb1b58e](https://github.com/PyCQA/astroid/commit/eb1b58e59a90dcd20f8180e8436109545ddbd8c2)
> > The context modifications seem to be fine.
> 
> Should we re-commit without the decorator changes?

👍🏻 I can test Home Assistant before we merge it, just to be sure.

DanielNoord:
Do you have any idea as to why this would break? Shouldn't the decorators work as decorators as well?

cdce8p:
> Do you have any idea as to why this would break? Shouldn't the decorators work as decorators as well?

Just found this commit: https://github.com/PyCQA/astroid/commit/0a8a75db30da060a24922e05048bc270230f5bad
I think that explains it. Missed it myself during the review.

PR #1659, opened 2022-06-24
Type ``_infer`` of binary operation nodes

DanielNoord:
## Steps

## Description

Sorry for the size of the PR. Wanted to catch all `_infer's` that use `_filter_operation_errors` and `nodes.Compare` was in between them.

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -599,7 +599,7 @@ def infer_unaryop(
     return InferenceErrorInfo(node=self, context=context)
 
 
-nodes.UnaryOp._infer_unaryop = _infer_unaryop  # type: ignore[assignment]
+nodes.UnaryOp._infer_unaryop = _infer_unaryop
 nodes.UnaryOp._infer = infer_unaryop  # type: ignore[assignment]
 
 
@@ -877,7 +877,7 @@ def infer_binop(
     )
 
 
-nodes.BinOp._infer_binop = _infer_binop  # type: ignore[assignment]
+nodes.BinOp._infer_binop = _infer_binop
 nodes.BinOp._infer = infer_binop  # type: ignore[assignment]
 
 COMPARE_OPS: dict[str, Callable[[Any, Any], bool]] = {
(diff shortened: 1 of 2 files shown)


astroid/inference.py
```diff
@@ -510,18 +510,28 @@ def _infer_boolop(
 # UnaryOp, BinOp and AugAssign inferences
```
Pierre-Sassoulas, line 511:
```suggestion
InferenceResultOrFail = InferenceResult | util.BadUnaryOperationMessage
InferenceFunction = Callable[
        [_T, InferenceContext | None],
        Generator[InferenceResultOrFail, None, None],
    ]

```
Could simplify some of the other typing ?

DanielNoord: Not sure it makes sense to create an alias for something that is used twice. Most of different combinations of the same thing. We could use a generic, but not sure if that's more readable.

Pierre-Sassoulas: There's actually two proposal here, the ``InferenceResultOrFail = InferenceResult | util.BadUnaryOperationMessage`` might be used quite a lot

DanielNoord: It's only used twice as it is specific to `Unary` operations.

We could make the `Callable` into a alias, but I don't mind it in its current position. The file is quite large anyway.

Pierre-Sassoulas (approved):

astroid/nodes/node_classes.py
```diff
@@ -1343,7 +1343,9 @@ def postinit(
     """
 
     # This is set by inference.py
-    def _infer_augassign(self, context=None):
+    def _infer_augassign(
+        self, context: InferenceContext | None = None
+    ) -> Generator[InferenceResult | util.BadBinaryOperationMessage, None, None]:
         raise NotImplementedError
```
cdce8p, line 1349: I'm wondering, IMO it would make sense to change these to type definitions similar to the once for `assigned_stmts` and `infer_lhs`. They are set anyway, so there is no point in assigning a dummy function for it in the first place. What do you think?

DanielNoord: Fine with me!

cdce8p: > Fine with me!

Let's try it then!

cdce8p (approved):
@DanielNoord I moved some type definitions. I think it's useful to always start with `__init__` and `postinit` and put the `ClassVar[Callable[...]]` thereafter. Basically where they have been previously.

Fell free to merge it.

PR #1661, opened 2022-06-24
Type ``_infer`` of the last nodes

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -1053,7 +1053,6 @@ def infer_empty_node(
 nodes.EmptyNode._infer = infer_empty_node  # type: ignore[assignment]
 
 
-
 def _populate_context_lookup(call, context):
     # Allows context to be saved for later
     # for inference inside a function
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -1005,7 +1005,7 @@ nodes.AugAssign._infer = infer_augassign
 def infer_arguments(
     self: nodes.Arguments, context: InferenceContext | None = None, **kwargs: Any
 ) -> Generator[InferenceResult, None, None]:
-    if not context or context.lookupname is None:
+    if context is None or context.lookupname is None:
         raise InferenceError(node=self, context=context)
     return protocols._arguments_infer_argname(self, context.lookupname, context)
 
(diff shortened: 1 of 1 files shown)


astroid/inference.py
```diff
@@ -1025,13 +1030,15 @@ def infer_assign(self, context=None):
     return bases._infer_stmts(stmts, context)
 
 
-nodes.AssignName._infer = infer_assign
-nodes.AssignAttr._infer = infer_assign
+nodes.AssignName._infer = infer_assign  # type: ignore[assignment]
+nodes.AssignAttr._infer = infer_assign  # type: ignore[assignment]
```
Pierre-Sassoulas, line 1034: Forgot to say it on previous merge request but I guess this is done in order to fight against circular import. Did you try to define this function directly in ``AssignAttr`` ? For some function I think it's done by force of habits more than necessity. This would simplify the code a lot and this kind of refactor worked in the past (although it's a bet and uncover big issue sometime).

DanielNoord: No I haven't tried this. However, it would be very difficult to reuse functions like we do here.

The best refactor would be to create Mixins/Base classes with the current inference functions that can then be inherited, but that's something for 2043 😄

cdce8p: Let's not do that, at least not for a while. True, it might be possible to move some methods but that can just lead to more issues down the road. What if for example we  want to split up `node_classes.py` and suddenly we added a new circular import just because of it.

Furthermore, although that might just be might opinion, I find it quite useful to have most of the inference methods in one place. Especially considering the nodes file is huge.

Pierre-Sassoulas: Right, I did not think of this reusing problem. It was used once in the  example I saw.

astroid/inference.py
```diff
@@ -1047,7 +1054,9 @@ def infer_empty_node(self, context=None):
 
 
 @decorators.raise_if_nothing_inferred
-def infer_index(self, context=None):
+def infer_index(
+    self: nodes.Index, context: InferenceContext | None = None, **kwargs: Any
+) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
     return self.value.infer(context)
```
cdce8p, line 1060: The `index` node has been deprecated a while ago and is no longer used. The `value` attribute doesn't exist anymore 😅 We can probably remove the `infer_index` method completely.

cdce8p (changes requested):

astroid/inference.py
DanielNoord: Yeah I saw that, but I wondered if I should keep it.

Shall I remove it?

cdce8p: Let's just remove the `infer_index` method since it wouldn't work anyway. We can keep the `Index` class around a bit longer until `3.0`. It hasn't hurt so far.

cdce8p (approved):

PR #1662, opened 2022-06-25
`hashlib`: Add support for `usedforsecurity` keyword

jacobtylerwalls:
## Steps

## Description
The `usedforsecurity` argument was new in Python 3.9, so this just brings the `hashlib` brain up to date. It has a default, so I think having it in the brain will be harmless in versions below 3.9. (I don't see other brains with interpreter-version checks.)

Suggesting a backport only if we end up merging and backporting #1660--otherwise no need.

## Type of Changes

## Related Issue
Closes PyCQA/pylint#6017


ChangeLog
```diff
@@ -77,6 +77,9 @@ What's New in astroid 2.11.7?
 =============================
 Release date: TBA
 
+* Added support for ``usedforsecurity`` keyword to ``hashlib`` constructors.
+
+  Refs PyCQA/pylint#6017
```
jacobtylerwalls, line 82: On second thought, I don't see the point of retesting this in pylint -- we don't need to bloat the `unexpected-keyword-arg` tests with a verification of every stdlib keyword.
```suggestion
  Closes PyCQA/pylint#6017
```

Pierre-Sassoulas (approved):

cdce8p:
> The `usedforsecurity` argument was new in Python 3.9, so this just brings the `hashlib` brain up to date. It has a default, so I think having it in the brain will be harmless in versions below 3.9. (I don't see other brains with interpreter-version checks.)

Not entirely. With it being unconditionally, we can't detect if the keyword argument was added when run Python 3.8. In the other brain modules we do have version dependent checks for that, usually combined with some kind of sting formatting. E.g.
https://github.com/PyCQA/astroid/blob/8e7174a99a9b9dde848632f99b0498f8d67999f3/astroid/brain/brain_ssl.py#L13-L26

jacobtylerwalls:
Ah, sorry! I'll do a quick followup. I was searching for `sys....` in the brain directories instead of the version constants. 🤦🏻‍♂️

PR #1663, opened 2022-06-25
Fix inference of living container elements

DanielNoord:
## Steps

## Type of Changes


astroid/nodes/node_classes.py
```diff
@@ -5377,6 +5377,30 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
 }
 
 
+def _create_basic_elements(value: Any, node: List | Set | Tuple) -> list[NodeNG]:
+    """Create a list of nodes to function as the elements of a new node."""
+    elements: list[NodeNG] = []
+    for elem in value:
+        elem_node = const_factory(elem)
+        elem_node.parent = node
+        elements.append(elem_node)
```
Pierre-Sassoulas, line 5386:
```suggestion
    for element in value:
        element_node = const_factory(element )
        element_node .parent = node
        elements.append(element_node )
```

Pierre-Sassoulas (approved):
👍

astroid/nodes/node_classes.py
```diff
@@ -5377,6 +5377,30 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
 }
 
 
+def _create_basic_elements(value: Any, node: List | Set | Tuple) -> list[NodeNG]:
```
cdce8p, line 5380:
```suggestion
def _create_basic_elements(value: Iterable[Any], node: List | Set | Tuple) -> list[NodeNG]:
```

astroid/nodes/node_classes.py
```diff
@@ -5377,6 +5377,30 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
+    elements: list[NodeNG] = []
+    for element in value:
+        element_node = const_factory(element)
+        element_node.parent = node
+        elements.append(element_node)
+    return elements
+
+
+def _create_dict_items(
+    values: dict[Any, Any], node: Dict
```
cdce8p, line 5391:
```suggestion
    values: Mapping[Any, Any], node: Dict
```

astroid/nodes/node_classes.py
```diff
@@ -5389,13 +5413,13 @@ def const_factory(value: Any) -> List | Set | Tuple | Dict | Const | EmptyNode:
         node.object = value
         return node
 
-    # TODO: We pass an empty list as elements for a sequence
-    # or a mapping, in order to avoid transforming
-    # each element to an AST. This is fixed in 2.0
-    # and this approach is a temporary hack.
     initializer_cls = CONST_CLS[value.__class__]
```
cdce8p, line 5416:
```suggestion
    instance: List | Set | Tuple | Dict
    initializer_cls = CONST_CLS[value.__class__]
```
The type definition for `instance` helps avoid mypy errors in the second `if` block. Without it `instance` would be limited to `List | Set | Tuple`.

ChangeLog
```diff
@@ -78,6 +78,9 @@ Release date: TBA
 
   Closes #1403
 
+* Fixed inference of elements of living container objects such as tuples and sets in the
+  ``sys`` and ``ssl`` modules.
```
cdce8p, line 82: Does this mean we don't need the imports in the `ssl` brain anymore?
https://github.com/PyCQA/astroid/blob/8e7174a99a9b9dde848632f99b0498f8d67999f3/astroid/brain/brain_ssl.py#L71-L108

astroid/nodes/node_classes.py
DanielNoord: Ah, `pyright`  thought this was okay so I didn't notice.

ChangeLog
DanielNoord: Sadly it doesn't. Those imports are used to construct `Enums`, which aren't covered by this.

There are some `tuples` in `_ssl` which we do infer now without the brain, but since nobody actually opened an issue about this I don't think anybody ever noticed.

cdce8p: 👌🏻

cdce8p (approved):
Let's merge #1666 first.

PR #1665, opened 2022-06-25
Refs #1662: Condition `usedforsecurity` support on Py3.9+

jacobtylerwalls:
## Description
See https://github.com/PyCQA/astroid/pull/1662#issuecomment-1166271441, the better way to do this is to condition the support for this keyword on Python 3.9+.

## Type of Changes

## Related Issue

Refs #1662


Pierre-Sassoulas (approved):

cdce8p (changes requested):
Thanks for the quick followup @jacobtylerwalls!
Can you update `blake2b_signature` and `blake2s_signature`, too?

cdce8p (approved):
Thanks @jacobtylerwalls 🐬

PR #1584, opened 2022-06-01
Move ``3.11`` to standard test job

DanielNoord:
## Steps

- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Related Issue

PR #1389, opened 2022-02-10
Add `Try` node

cdce8p:
## Description
Depends on #1388

In the Python AST `TryExcept` and `TryFinally` have long been replaced by `Try`. This PR adds a new `Try` node to do the same in astroid.

To test this change in pylint, add this in `pylinter.py` (after the imports).
```py
from astroid.manager import AstroidManager
AstroidManager.tree_rev = 2
```

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

PR #1670, opened 2022-06-27
Improve packaging [PEP 517 + 621]

cdce8p:
## Description
Similar to https://github.com/PyCQA/pylint/pull/7076

* Use isolated build environments [PEP 517](https://peps.python.org/pep-0517/) with `python -m build` instead of deprecated `python setup.py sdist bdist_wheel`
* Move project metadata from `setup.cfg` to `pyproject.toml` [PEP 621](https://peps.python.org/pep-0621/)
* Cleanup `MANIFEST.in` -> entries haven't been used.
* Add `README.rst` to `MANIFEST.in`. Required to be able to build the `wheel` from `sdist` and still include the readme text in the project metadata.

**Metadata diff**
```diff
-Author: Python Code Quality Authority
-Author-email: code-quality@python.org
+Author-email: Python Code Quality Authority <code-quality@python.org>
...
-Home-page: https://github.com/PyCQA/astroid
+Project-URL: Source Code, https://github.com/PyCQA/astroid
...
+Project-URL: Docs, https://pylint.pycqa.org/projects/astroid/en/latest/
...
-Platform: UNKNOWN
```

**Notes**
* `README.rst` should be included in `sdist` to build `wheel` successfully. Without `MANIFEST.in`, `README.rst` would be added automatically. Let's add it manually, just in case something else is added to `MANIFEST.in` later.
* `Platform` isn't used currently / anymore
* `Home-page` was defined with `url`. I've moved it to `project.urls` and renamed it to `Source Code`.
* Added `project.url` for `Docs`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,55 @@
+[build-system]
+requires = ["setuptools~=62.6", "wheel~=0.37.1"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name        = "astroid"
+license     = {text = "LGPL-2.1-or-later"}
+description = "An abstract syntax tree for Python with inference support."
+readme      = "README.rst"
+authors     = [
+    {name = "Python Code Quality Authority", email = "code-quality@python.org"}
+]
+keywords    = ["static code analysis", "python", "abstract syntax tree"]
+classifiers = [
+    "Development Status :: 6 - Mature",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: GNU Lesser General Public License v2 (LGPLv2)",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "Topic :: Software Development :: Quality Assurance",
+    "Topic :: Software Development :: Testing",
+]
+requires-python = ">=3.7.2"
+dependencies = [
+    "lazy_object_proxy>=1.4.0",
+    "wrapt>=1.11,<2",
+    "typed-ast>=1.4.0,<2.0;implementation_name=='cpython' and python_version<'3.8'",
+    "typing-extensions>=3.10;python_version<'3.10'",
+]
+dynamic = ["version"]
+
+[project.urls]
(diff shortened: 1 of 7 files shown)


Pierre-Sassoulas (approved):
Great !

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: GNU Lesser General Public License v2 (LGPLv2)",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
DanielNoord, line 26: Should we add `3.11` as well?

setup.cfg
```diff
@@ -63,6 +24,11 @@ known_first_party = astroid
 include_trailing_comma = True
 skip_glob = tests/testdata
 
+[flake8]
```
DanielNoord, line 27: We have a `.flake8` file, shouldn't this go there?

DanielNoord (changes requested):

pyproject.toml
Pierre-Sassoulas: Hmm nice catch, I think we forgot to add it in https://github.com/PyCQA/astroid/pull/1584/

Pierre-Sassoulas: Hmm nice catch, I think we forgot to add it in https://github.com/PyCQA/astroid/pull/1584/

```suggestion
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.11",
```

cdce8p: > Should we add `3.11` as well?

We don't support all `3.11` features yet. Especially support for `TryStar` is still missing. Tracking issue #1516
I should probably continue the work on #1389 soon. With that `TryStar` should be fairly strait forward.

Pierre-Sassoulas: I think we should add the 3.11 metadata before supporting all 3.11 features, because we're supporting all the old features and if we don't do that, then there's no astroid **at all** for 3.11 which is a lot worse than having astroid without some 3.11 exclusive features.

cdce8p: > I think we should add the 3.11 metadata before supporting all 3.11 features, because we're supporting all the old features and if we don't do that, then there's no astroid **at all** for 3.11 which is a lot worse than having astroid without some 3.11 exclusive features.

We don't block anyone from installing astroid with Python 3.11.
However, the way I understand the classifier is "All of the features (or at least the major ones) for version x.xx are supported." That's why I would recommend to wait a bit longer before adding it.

cdce8p:
After the issue with dependabot in pylint, I've added back `setup.py` here. At least for now, it seems like it's still required.

Pierre-Sassoulas (approved):

DanielNoord (approved):

PR #1671, opened 2022-06-28
Avoid inferring all subscripted names in `pathlib` brain

### LINKED ISSUE #1667: 25% performance regression on `main`
#1442 adds about 10-25% time to a pylint run, depending on how much inference would have taken place otherwise, because it infers every single name.

On my machine, with `cProfile`:
`astroid` lints in 46.2s on main
`astroid` lints in 36.6s with reverting #1442

Are we okay with that? If so, should we short-circuit before inferring in the pathlib brain if the interpreter is below 3.10? (#1442 only has to do with 3.10 behavior.)

jacobtylerwalls:
## Steps

## Type of Changes

## Related Issue
Fixes #1667 by reducing the amount of inference done by #1442

### DIFF
--- a/tests/unittest_brain_pathlib.py
+++ b/tests/unittest_brain_pathlib.py
@@ -26,6 +26,23 @@ def test_inference_parents() -> None:
     assert inferred[0].qname() == "pathlib._PathParents"
 
 
+def test_inference_parents_subscript_index() -> None:
+    """Test inference of ``pathlib.Path.parents``, accessed by index."""
+    path = astroid.extract_node(
+        """
+    from pathlib import Path
+
+    current_path = Path().resolve()
+    current_path.parents[2]  #@
+    """
+    )
+
+    inferred = path.inferred()
+    assert len(inferred) == 1
+    assert isinstance(inferred[0], bases.Instance)
+    assert inferred[0].qname() == "pathlib.Path"
+
+
 def test_inference_parents_subscript_slice() -> None:
     """Test inference of ``pathlib.Path.parents``, accessed by slice."""
     name_node = astroid.extract_node(


Pierre-Sassoulas (approved):

astroid/brain/brain_pathlib.py
```diff
@@ -18,6 +19,9 @@
 
 
 def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
+    """Infer subscripted names on Python 3.10+, which supports slicing PurePath.parents"""
+    if not PY310_PLUS:
+        return False
```
cdce8p, line 24: Better to wrap the `register_transform` call below in a version check. That way we don't even call the predicate function for `< 3.10`.

https://github.com/PyCQA/astroid/blob/91533b7b8a5f7951c2b590add0a73ad3bc113a07/astroid/brain/brain_typing.py#L416-L419

cdce8p (commented):
This is a regression for `path_parents[2]`, or am I overlooking something?

A general question, do you know what exactly is causing the slowdown? Is the predicate or inference function evaluated too often? It could also be the inference call in the predicate. Maybe test one run without it.

jacobtylerwalls:
Yep, I profiled it, and it's just the sheer fact of inferring so many names we don't need to infer below 3.10.

cdce8p:
> Yep, I profiled it, and it's just the sheer fact of inferring so many names

Isn't everything not `xxx.parents[...]` skipped?

> we don't need to infer below 3.10.

Not entirely true, is it? We do lose some things. Otherwise, you wouldn't need to add the pytest `skipif`.

jacobtylerwalls:
> Isn't everything not xxx.parents[...] skipped?

That's only for attributes. Names are unconditionally inferred.

> Not entirely true, is it? We do lose some things. Otherwise, you wouldn't need to add the pytest skipif.

Right, we definitely lose the specific inference, but I thought the idea behind the contribution was handling slicing pathlib parents in 3.10+, and I wouldn't think the performance tradeoff of inferring every name in a subscript would be worth it for an abstract benefit.

astroid/brain/brain_pathlib.py
```diff
@@ -18,6 +19,7 @@
 
 
 def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
+    """Infer subscripted names on Python 3.10+, which supports slicing PurePath.parents"""
```
jacobtylerwalls, line 22:
```suggestion
```

jacobtylerwalls:
I'm on board with that!

cdce8p (approved):
Looks good! The performance improvement is also noticeable for Home Assistant.

PR #1678, opened 2022-06-29
Add typing to ``metaclass`` methods

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -46,6 +46,7 @@ from astroid.nodes import Arguments, Const, NodeNG, _base_nodes, node_classes
 from astroid.nodes.scoped_nodes.mixin import ComprehensionScope, LocalsDictNodeNG
 from astroid.nodes.scoped_nodes.utils import builtin_lookup
 from astroid.nodes.utils import Position
+from astroid.typing import SuccessfulInferenceResult
 
 if sys.version_info >= (3, 8):
     from functools import cached_property
@@ -2810,7 +2811,9 @@ class ClassDef(
             return builtin_lookup("type")[1][0]
         return None
 
-    def declared_metaclass(self, context=None):
+    def declared_metaclass(
+        self, context: InferenceContext | None = None
+    ) -> SuccessfulInferenceResult | None:
         """Return the explicit declared metaclass for the current class.
 
         An explicit declared metaclass is defined
@@ -2821,7 +2824,6 @@ class ClassDef(
 
         :returns: The metaclass of this class,
             or None if one could not be found.
-        :rtype: NodeNG or None
         """
         for base in self.bases:
             try:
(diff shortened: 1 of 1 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2810,7 +2811,9 @@ def implicit_metaclass(self):
             return builtin_lookup("type")[1][0]
         return None
 
-    def declared_metaclass(self, context=None):
+    def declared_metaclass(
+        self, context: InferenceContext | None = None
+    ) -> SuccessfulInferenceResult | None:
```
DanielNoord, line 2816: Or do we want to assert the `infer()` to make sure we only return certain types? It doesn't make a lot of sense for a `metaclass` to be an `UnboundMethod` (for example), which it could be with the current typing and lack of assertions.

cdce8p: From what I can tell or at least in the astroid tests. Only `nodes.ClassDef | None` is ever returned. I.e. it could actually make sense to restrict the return type here. Especially since `metaclass` could be called independently of `infer`.

I'm not sure how we can add an `assert` here though without changing the return generator expression. Might be better to use `type: ignore` or `cast` instead.

DanielNoord: My idea was to change the `next` call in the return statement to only allow ClassDef to pass through. That should work right?

cdce8p: Ideally we should get a warning / error if somehow something else would be returned. Otherwise we might mask an issue that is entirely preventable.

I was also thinking about wether it should be `ClassDef` or `NodeNG`. Currently, I would lean towards `NodeNG`. It might not be as precise, but users can do a simple `isinstance` check afterwards. Or do you think `ClassDef` would be better?

DanielNoord: > Ideally we should get a warning / error if somehow something else would be returned. Otherwise we might mask an issue that is entirely preventable.

I don't really have any idea on how to allow this. Do you want to log something when this occurs?

> I was also thinking about wether it should be `ClassDef` or `NodeNG`. Currently, I would lean towards `NodeNG`. It might not be as precise, but users can do a simple `isinstance` check afterwards. Or do you think `ClassDef` would be better?

Yeah I think that's better. Especially considering plugins might also set meta classes and might not be as strict as we are.

DanielNoord: I have added the `NodeNG` check.

Pierre-Sassoulas (approved):
LGTM. I'm ok with merging without the coverage, it seems coverall is down.

DanielNoord:
@cdce8p Gentle ping as this is blocking some other typing work

DanielNoord:
Merging as I think this is still a good change!

PR #1682, opened 2022-07-02
Prevent creating `Instance` that proxies another `Instance` when inferring `__new__(cls)`

jacobtylerwalls:
## Steps

## Description
See discussion at https://github.com/PyCQA/pylint/issues/7109#issuecomment-1172909800. When astroid thinks `__new__(cls)` is being called with an instance either due to user error or because of inference ambiguity (e.g. the library code in the OP used an `isinstance` guard that astroid doesn't cope with), we now raise `InferenceError` instead of creating a bogus Instance-proxied-to-Instance. Callers then transform this to `Uninferable` as appropriate. (See #782 for a precedent.)

## Type of Changes

## Related Issue
Refs PyCQA/pylint#7109

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -10,9 +10,9 @@ from __future__ import annotations
 import collections
 import collections.abc
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
-from astroid import decorators
+from astroid import decorators, nodes
 from astroid.const import PY310_PLUS
 from astroid.context import (
     CallContext,
@@ -32,9 +32,6 @@ from astroid.util import Uninferable, lazy_descriptor, lazy_import
 objectmodel = lazy_import("interpreter.objectmodel")
 helpers = lazy_import("helpers")
 manager = lazy_import("manager")
-
-if TYPE_CHECKING:
-    from astroid import nodes
 
 
 # TODO: check if needs special treatment
@@ -293,6 +290,11 @@ class Instance(BaseInstance):
     # pylint: disable=unnecessary-lambda
     special_attributes = lazy_descriptor(lambda: objectmodel.InstanceModel())
 
+    def __init__(self, proxied: nodes.ClassDef) -> None:
+        if not isinstance(proxied, nodes.ClassDef):
+            raise TypeError
+        super().__init__(proxied)
+
     def __repr__(self):
         return "<Instance of {}.{} at 0x{}>".format(
             self._proxied.root().name, self._proxied.name, id(self)
@@ -415,9 +417,6 @@ class UnboundMethod(Proxy):
     ) -> collections.abc.Generator[
         nodes.Const | Instance | type[Uninferable], None, None
     ]:
-        # pylint: disable-next=import-outside-toplevel; circular import
-        from astroid import nodes
-
         if not caller.args:
             return
         # Attempt to create a constant
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -10,9 +10,9 @@ from __future__ import annotations
 import collections
 import collections.abc
 from collections.abc import Sequence
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
-from astroid import decorators, nodes
+from astroid import decorators
 from astroid.const import PY310_PLUS
 from astroid.context import (
     CallContext,
@@ -32,6 +32,9 @@ from astroid.util import Uninferable, lazy_descriptor, lazy_import
 objectmodel = lazy_import("interpreter.objectmodel")
 helpers = lazy_import("helpers")
 manager = lazy_import("manager")
+
+if TYPE_CHECKING:
+    from astroid import nodes
 
 
 # TODO: check if needs special treatment
@@ -290,11 +293,6 @@ class Instance(BaseInstance):
     # pylint: disable=unnecessary-lambda
     special_attributes = lazy_descriptor(lambda: objectmodel.InstanceModel())
 
-    def __init__(self, proxied: nodes.ClassDef) -> None:
-        if not isinstance(proxied, nodes.ClassDef):
-            raise TypeError
-        super().__init__(proxied)
-
     def __repr__(self):
         return "<Instance of {}.{} at 0x{}>".format(
             self._proxied.root().name, self._proxied.name, id(self)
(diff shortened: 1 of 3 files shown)


astroid/bases.py
```diff
@@ -436,7 +436,14 @@ def _infer_builtin_new(
         node_context = context.extra_context.get(caller.args[0])
         infer = caller.args[0].infer(context=node_context)
 
-        yield from (Instance(x) if x is not Uninferable else x for x in infer)  # type: ignore[misc,arg-type]
+        for inferred in infer:
+            if isinstance(inferred, type(Uninferable)):  # isinstance helps mypy
+                yield inferred
+            if isinstance(inferred, Instance):
+                raise InferenceError
```
Pierre-Sassoulas, line 443: If I understand correctly a single instance of type ``Instance`` raise an InferanceError and void every value in infer ? Why don't we yield ``Uninferable`` instead ?

Pierre-Sassoulas (approved):
LGTM, just a question for my information :)

astroid/bases.py
jacobtylerwalls: Yes, it would prevent reaching any subsequent values (not prior values). We could go either way, but this seemed more consistent with the rest of the system to me.

Examples:
- #782
- https://github.com/PyCQA/astroid/blob/3621e2e7d68653d66cbf770e6dcb61ba541117f1/astroid/arguments.py#L176-L187
- pylint's `safe_infer_call_result`
- the various `try/except` around `infer_call_result` that catch InferenceError instead of checking for Uninferable: https://github.com/PyCQA/pylint/blob/01ea163c48d759f86d75dfb47e00796b7409abe1/pylint/checkers/classes/class_checker.py#L1184-L1187

Pierre-Sassoulas: Thank you !

astroid/bases.py
```diff
@@ -436,7 +436,14 @@ def _infer_builtin_new(
         node_context = context.extra_context.get(caller.args[0])
         infer = caller.args[0].infer(context=node_context)
 
-        yield from (Instance(x) if x is not Uninferable else x for x in infer)  # type: ignore[misc,arg-type]
+        for inferred in infer:
+            if isinstance(inferred, type(Uninferable)):  # isinstance helps mypy
```
DanielNoord, line 440:
```suggestion
            if inferred is Uninferable:
```

Let's fix all those `is Uninferable` statements with #1680.

astroid/bases.py
```diff
@@ -436,7 +436,14 @@ def _infer_builtin_new(
...
+                yield inferred
+            if isinstance(inferred, Instance):
+                raise InferenceError
+            if isinstance(inferred, (nodes.ClassDef, nodes.Lambda, Proxy)):
```
DanielNoord, line 444: I thought about this some more, and this (currently) can't be correct as well.

We expect the `proxy` of an `Instance` to be `ClassDef`, see its typing. If this crashes without also allowing `Lambda` or `Proxy` we should fix that typing, if not this should only accept `ClassDef`.

Somewhat related, I think it would be good to add an `__init__` to `Instance` which narrows down the type of `proxied`. Would you be willing to do that in this PR as well?

astroid/bases.py
```diff
@@ -436,7 +436,14 @@ def _infer_builtin_new(
...
+                yield Instance(inferred)
+            raise InferenceError
```
DanielNoord, line 446: If we got for the `isinstance(inferred, nodes.ClassDef)` variant we should probably switch those `if` statements around and remove the `Instance` check: we can just raise if it is not `ClassDef`.

DanielNoord (changes requested):

jacobtylerwalls:
This is going to be too difficult to backport; bumping to 2.12.

astroid/bases.py
jacobtylerwalls: Are you suggesting to block this until #1680? That could be a long way off. Or just to silence the mypy error?

jacobtylerwalls: Ah, right, we're not running mypy. So just letting it fail, then?

DanielNoord: let if fail 😄

astroid/bases.py
jacobtylerwalls: Thanks, I should have noticed I was getting the type from the wrong parent!

astroid/bases.py
jacobtylerwalls: Sure, if we're not backporting that's on the table 👍🏻

astroid/bases.py
```diff
@@ -293,6 +290,11 @@ class Instance(BaseInstance):
     # pylint: disable=unnecessary-lambda
     special_attributes = lazy_descriptor(lambda: objectmodel.InstanceModel())
 
+    def __init__(self, proxied: nodes.ClassDef) -> None:
+        if not isinstance(proxied, nodes.ClassDef):
+            raise TypeError
```
DanielNoord, line 295:
```suggestion
```

That _should_ be unreachable 😄

astroid/bases.py
```diff
@@ -436,7 +435,12 @@ def _infer_builtin_new(
         node_context = context.extra_context.get(caller.args[0])
         infer = caller.args[0].infer(context=node_context)
 
-        yield from (Instance(x) if x is not Uninferable else x for x in infer)  # type: ignore[misc,arg-type]
+        for inferred in infer:
```
DanielNoord, line 438: I can't suggest it, but can we skip the assignment on L436 and just iterate directly?

DanielNoord (changes requested):
Last two nits!

astroid/bases.py
jacobtylerwalls: I don't follow -- I thought you asked for this guard? Otherwise the constructor is pointless.

DanielNoord: No, the constructor is helpful for type checkers. `mypy` doesn't recognise that `Instance._proxied: ClassDef` (just a couple lines above) means that for `Instance(x)` `x` needs to be `ClassDef`. By making the constructor explicit we will get type warnings from `mypy` (when we turn it on) if we create an `Instance` without making sure `x` is `ClassDef`.

jacobtylerwalls: gotcha!

DanielNoord (approved):
Thanks @jacobtylerwalls!

jacobtylerwalls:
Thanks for the initial digging and the review! ✨

Pierre-Sassoulas:
There's two issues left in 2.12, including this one that is mergeable, should we even release 2.11.7 ?

DanielNoord:
I'd like to get the `is None` in `2.12` if possible. It was moved to `2.13` recently, but the last reviews don't look like they should take too long to resolve.

DanielNoord:
This has created an issue for `nodes.Const` as it can no longer call the `__init__` of `Instance`. We set the `._proxied` of `Const` during bootstrapping.

I'm looking into a solution.

PR #1686, opened 2022-07-05
Fix order of attribute look ups for Super

### LINKED ISSUE #1684: Investigate behavior changes from #1606
Placeholder issue for investigating behavior changes reported in pylint primer after #1606: https://github.com/PyCQA/pylint/pull/7118#issuecomment-1173141895

I know @DanielNoord is already aware, but just raising visibility to make sure we don't release 2.12 until we have a chance to debug/revert.

DanielNoord:
## Steps

## Description

Closes #1684

## Type of Changes

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -3808,6 +3808,50 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         with pytest.raises(InferenceError):
             next(ast_node5.infer())
 
+        ast_nodes6 = extract_node(
+            """
+        class A:  pass
+        class B(A):  pass
+        class C: pass
+        A.__new__(A)  #@
+        A.__new__(B)  #@
+        B.__new__(A)  #@
+        B.__new__(B)  #@
+        C.__new__(A)  #@
+        """
+        )
+        instance_A1 = next(ast_nodes6[0].infer())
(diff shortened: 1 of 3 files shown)


tests/unittest_inference.py
```diff
@@ -3808,6 +3808,50 @@ class A:  pass
         with pytest.raises(InferenceError):
             next(ast_node5.infer())
 
+        ast_nodes6 = extract_node(
```
jacobtylerwalls, line 3811: This test doesn't fail on `main` for me. 🤔

DanielNoord: Yeah, I was just writing some tests for `__new__` and though it would be good to keep this in as a some sort of regression tests.

DanielNoord: I've struggling to do so. I'll give it another try. I can probably hack something together with `getattr("__new__")` on `super()`.

jacobtylerwalls: No leads from the primer examples (too complicated)?

DanielNoord: Not really. Those fails are dependent on what `pylint` does with the information it gets from `astroid`. Using `getattr` we can test that information here.

jacobtylerwalls (approved):
Great, thanks!

PR #1687, opened 2022-07-06
Add ``__init__`` to the ``ObjectModel`` and return ``BoundMethods``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

Blocked by https://github.com/PyCQA/astroid/pull/1686
Precedes https://github.com/PyCQA/astroid/pull/1519

@jacobtylerwalls Please only consider the last commit. This is the next step in "completing" our `ObjectModel`. Since all objects have an `__init__` that normally returns `None` we can define ``attr___init__`` on ``ObjectModel``.

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -28,7 +28,7 @@ import os
 import pprint
 import types
 from functools import lru_cache
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import astroid
 from astroid import bases, nodes, util
@@ -63,6 +63,14 @@ def _dunder_dict(instance, attributes):
 
     obj.postinit(list(zip(keys, values)))
     return obj
+
+
+def _get_bound_node(model: ObjectModel) -> Any:
+    # TODO: Use isinstance instead of try ... except after _instance has typing
+    try:
+        return model._instance._proxied
+    except AttributeError:
+        return model._instance
 
 
 class ObjectModel:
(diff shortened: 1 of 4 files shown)


DanielNoord:
After testing on `pylint` I found that there were still issues with the previous implementation.

I think now I (finally) fixed those.

I was mistaken to think that the other `attr___` methods are correct. Rather than returning the value directly these methods should return the `BoundMethod` of the method that is going to be called. The `functools` brain uses a somewhat similar trick:
https://github.com/PyCQA/astroid/blob/58750a6f982c20fd7224ac396c95ee8adbbfa34f/astroid/brain/brain_functools.py#L52-L55

We will infer the result of these `BoundMethods` succesfully by calling `infer_call_result` on them, but for some cases we need the `BoundMethod` itself. For example for the following code `A().__init__ #@`. That shouldn't be a `nodes.Const`, but the `BoundMethod` that will return `nodes.Const`.

astroid/interpreter/objectmodel.py
```diff
@@ -119,17 +121,41 @@ def lookup(self, name):
...
-        created.
-        """
         # TODO: Use isinstance instead of try ... except after _instance has typing
         try:
-            return self._instance._proxied.instantiate_class()
+            bound = self._instance._proxied
         except AttributeError:
-            return self._instance.instantiate_class()
+            bound = self._instance
+        return bases.BoundMethod(proxy=node, bound=bound)
```
jacobtylerwalls, line 158: This looks the same as for `attr___new__`: how would you feel about factoring it out?

jacobtylerwalls (commented):
Looks good!

astroid/interpreter/objectmodel.py
DanielNoord: I'll add a little helper function.

jacobtylerwalls (approved):
🎉

PR #1692, opened 2022-07-09
Merge ``Maintenance/2.11.x`` into ``main``

DanielNoord:
## Steps

## Type of Changes

## Related Issue


ChangeLog
```diff
@@ -106,7 +106,7 @@ Release date: TBA
 
   Closes #1195
 
-What's New in astroid 2.11.7?
+What's New in astroid 2.11.8?
 =============================
 Release date: TBA
```
Pierre-Sassoulas, line 112: We need to remove everything inside 2.11.8 so it's empty.

DanielNoord: Ah!

Pierre-Sassoulas: You're becoming an expert in patch release 😄

DanielNoord: Man, this is horrible.. 😄 

Should this be squashed or merge committed?

Pierre-Sassoulas: You were right to create a merge request because I would not have caught that otherwise, but I generally merge, then amend the merge commit (the issue with duplication of the patch note is always there). Check with a ``git diff origin/main``, then simply push the merge commit of maintenance/2.x on main. So I don't know exactly how github handle this, but I'd say it's probably "create a merge commit" ? The current state of the branch would add a "fix changelog" commit that we won't want in the history.

Pierre-Sassoulas: Maybe you can still interactive rebase with --preserve-merge so fix changelog is squashed in the merge commit

DanielNoord: Wouldn't squashing also be right? Otherwise there are two `hashlib: Add support for usedforsecurity keyword (#1662)` (and the other 2) commits on `main`.

Pierre-Sassoulas: This took me forever but here's how I think the final result should look like with main in red and maintenance in orange. If we squash it's not obvious that the two branches are linked. [IMG_20220709_162140.jpg](https://user-images.githubusercontent.com/5493666/178109878-4ab119b4-6ca4-4e2f-b7cb-b5912fe974ac.jpg)

DanielNoord: Okay, I'll close this PR then because that requires another order.

PR #1693, opened 2022-07-09
Merge maintenance into ``main``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -106,7 +106,7 @@ Release date: TBA
 
   Closes #1195
 
-What's New in astroid 2.11.7?
+What's New in astroid 2.11.8?
 =============================
 Release date: TBA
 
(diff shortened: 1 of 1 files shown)


DanielNoord:
@Pierre-Sassoulas They way you showed can't be done without pushing directly to `main`. You need to resolve merge conflicts before you can merge on Github so you'll always get a merge commit on the branch. Then you can't rebase due to conflicts so you get another merge commit. I think this is the best we can do.

ChangeLog
```diff
@@ -106,7 +106,7 @@ Release date: TBA
 
   Closes #1195
 
-What's New in astroid 2.11.7?
+What's New in astroid 2.11.8?
```
Pierre-Sassoulas, line 109: The changelog got duplicated again.

Pierre-Sassoulas (changes requested):
Do you have the right to push to main directly ? That's what I usually do after a ``git diff origin/main`` (where I check that nothing was modified except the release date and the new changelog for 2.11.8). If you can't then we can also use "create a merge commit" from this this is not the end of the world 😄 

Thank you for releasing astroid, much appreciated !

DanielNoord:
> Do you have the right to push to main directly ? That's what I usually do after a `git diff origin/main` (where I check that nothing was modified except the release date and the new changelog for 2.11.8). If you can't then we can also use "create a merge commit" from this this is not the end of the world 😄
> 
> Thank you for releasing astroid, much appreciated !

I think so. I'll push to `main` and I'll make sure to fix that changelog.

Pierre-Sassoulas:
Congratulation on your first release !! 🎉

PR #1694, opened 2022-07-09
Release ``2.12``

DanielNoord:
## Steps

## Type of Changes

## Related Issue

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -161,6 +161,11 @@ Contributors
 - Alphadelta14 <alpha@alphaservcomputing.solutions>
 - Alexander Presnyakov <flagist0@gmail.com>
 - Ahmed Azzaoui <ahmed.azzaoui@engie.com>
+- nathannaveen <42319948+nathannaveen@users.noreply.github.com>
+- adam-grant-hendry <59346180+adam-grant-hendry@users.noreply.github.com>
+- Deepyaman Datta <deepyaman.datta@utexas.edu>
+- Batuhan Taskaya <isidentical@gmail.com>
+- Alexander Scheel <alexander.m.scheel@gmail.com>
 
 Co-Author
 ---------
--- a/ChangeLog
+++ b/ChangeLog
@@ -115,12 +115,6 @@ Release date: TBA
 
   Refs PyCQA/pylint#7109
 
-
-What's New in astroid 2.11.8?
-=============================
-Release date: TBA
-
-
 What's New in astroid 2.11.7?
 =============================
 Release date: 2022-07-09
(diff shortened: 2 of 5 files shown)


CONTRIBUTORS.txt
```diff
@@ -1,3 +1,9 @@
+# This file is autocompleted by 'contributors-txt',
```
DanielNoord, line 1: I can't use `tbump` without this popping up...

CONTRIBUTORS.txt
```diff
@@ -1,3 +1,9 @@
...
+# using the configuration in '/Users/daniel/DocumentenLaptop/Programming/Github/astroid/script/.contributors_aliases.json'.
+# Do not add new persons manually and only add information without
+# using '-' as the line first character.
+# Please verify that your change are stable if you modify manually.
+
```
DanielNoord, line 6:
```suggestion
```

DanielNoord:
@Pierre-Sassoulas I would be okay with rebasing like this. Not being able to use `tbump` is really annoying 😓

CONTRIBUTORS.txt
```diff
@@ -173,3 +173,8 @@ under this name, or we did not manage to find their commits in the history.
 - alain lefroy
 - Mark Gius
 - jarradhope
+- nathannaveen <42319948+nathannaveen@users.noreply.github.com>
+- adam-grant-hendry <59346180+adam-grant-hendry@users.noreply.github.com>
+- Deepyaman Datta <deepyaman.datta@utexas.edu>
+- Batuhan Taskaya <isidentical@gmail.com>
+- Alexander Scheel <alexander.m.scheel@gmail.com>
```
Pierre-Sassoulas, line 180: There's a bug in contributor-txt, could you move those to the contributors section ? (I did not fix this because I was the only one affected so far, guess now I should fix it 😄 )

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas:
>  Not being able to use tbump is really annoying

I did not understand the problem with tbump, could you explain  ?

DanielNoord:
> 

Since `tbump` runs `contributor-txt` it will always auto commit the errors in that program.

Pierre-Sassoulas:
> Since tbump runs contributor-txt it will always auto commit the errors in that program.

Ha yes I understand. There's always some kind of interactive rebase to have a proper commit in the end right now, I did not invest a lot of time to fix it because releasing is not something we do that often and I had less issue with it then you, I don't know exactly why. The reordering of contributor by commit number is deactivated too, I'm doing it manually for frequent contributor.

Pierre-Sassoulas (approved):

PR #1701, opened 2022-07-10
Only infer module operations on string ``Const`` nodes

### LINKED ISSUE #1700: Crash in modulo expression
### Steps to reproduce
```python
class BatchSequence:
    def __init__(self):
        self.batch_size = 0

    def __len__(self):
        return 20 % self.batch_size
```

### Current behavior
```
  File "/Users/.../astroid/astroid/decorators.py", line 108, in wrapped
  File "/Users/.../astroid/astroid/inference.py", line 480, in _filter_operation_errors
    for result in infer_callable(self, context):
  File "/Users/.../astroid/astroid/inference.py", line 817, in _infer_binop
    yield from _infer_binary_operation(lhs, rhs, self, context, _get_binop_flow)
  File "/Users/.../astroid/astroid/inference.py", line 768, in _infer_binary_operation
    results = list(method())
  File "/Users/.../astroid/astroid/inference.py", line 607, in _invoke_binop_inference
    return iter(_infer_old_style_string_formatting(instance, other, context))
  File "/Users/.../astroid/astroid/inference.py", line 594, in _infer_old_style_string_formatting
    return (nodes.const_factory(instance.value % values),)
ZeroDivisionError: integer division or modulo by zero
```
### Expected behavior
Regression in 957d3d686879c1775e3e257955ba3b5eb929ec6c
Found in pylint primer (the old primer with `keras`, so I didn't see it yesterday when it was blocked on stdlib failures)
@DanielNoord would you be able to take a look for 2.12.2?

DanielNoord:
## Steps

## Description

Closes #1700

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,9 @@ What's New in astroid 2.12.2?
 =============================
 Release date: TBA
 
+* Fixed crash in modulo operations for divisions by zero.
 
+  Closes #1700
 
 What's New in astroid 2.12.1?
 =============================
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -659,7 +659,11 @@ def _invoke_binop_inference(instance, opnode, op, other, context, method_name):
     method = methods[0]
     context.callcontext.callee = method
 
-    if isinstance(instance, nodes.Const) and op == "%":
+    if (
+        isinstance(instance, nodes.Const)
+        and isinstance(instance.value, str)
+        and op == "%"
+    ):
         return iter(_infer_old_style_string_formatting(instance, other, context))
 
     try:
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -6948,6 +6948,7 @@ class TestOldStyleStringFormatting:
             """
             "My name is %s, I'm %s" % ((fname,)*2)
             """,
+            """20 % 0""",
         ],
     )
     def test_old_style_string_formatting_uninferable(self, format_string: str) -> None:


Pierre-Sassoulas (approved):

jacobtylerwalls (approved):

Pierre-Sassoulas:
Should we release 2.12.2 right now ?

jacobtylerwalls:
Let me get a clean run of both primer suites locally. I should have done that yesterday.

DanielNoord:
Do we need this for the upgrade to `2.12.x`? If it's just a crash fix we might want to wait to see we can get `2.12.1` to work with `pylint` `2.15.x`.

jacobtylerwalls:
It was found in the old pylint primer and was caused in astroid 2.12.0, so it's necessary if we want all the checks to pass when upgrading to 2.12.0. But if you mean something like "wait and make sure this is the only blocker" i.e. want to wait for the first report from Pierre on his [`sys.path` patching PR](https://github.com/PyCQA/pylint/pull/7153#issuecomment-1179726046) that's fine too.

DanielNoord:
> It was found in the old pylint primer and was caused in astroid 2.12.0, so it's necessary if we want all the checks to pass when upgrading to 2.12.0-1. But if you mean something like "wait and make sure this is the only blocker" i.e. want to wait for the first report from Pierre on his [`sys.path` patching PR](https://github.com/PyCQA/pylint/pull/7153#issuecomment-1179726046) that's fine too.

Yeah. I'd say let's make sure we get all fixes necessary for the `2.12.x` upgrade into `2.12.2`. Nothing inherently bad with releasing a number of versions, but it only costs us more time.

PR #1702, opened 2022-07-10
Prevent applying boolean ops to `NotImplemented`

jacobtylerwalls:
## Steps

## Description
Prevents this DeprecationWarning:

```DeprecationWarning: NotImplemented should not be used in a boolean context```

Reproducer:
``` python -Werror -m pylint ipaddress```
## Type of Changes

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -1117,6 +1117,12 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
         first = next(ast_node.infer())
         self.assertEqual(first, util.Uninferable)
 
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
+    def test_binary_op_not_used_in_boolean_context(self) -> None:
+        ast_node = extract_node("not NotImplemented")
+        first = next(ast_node.infer())
+        self.assertIsInstance(first, nodes.Const)
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/unittest_inference.py [hunk 1]
```diff
@@ -1117,6 +1117,12 @@ def __radd__(self, other): return NotImplemented
         first = next(ast_node.infer())
         self.assertEqual(first, util.Uninferable)
 
+    @pytest.mark.filterwarnings("error::DeprecationWarning")
```
Pierre-Sassoulas, line 1120: Do we need to filter that ? I thought the fix was precisely to remove the deprecation warning ?

Pierre-Sassoulas (approved):

tests/unittest_inference.py
jacobtylerwalls: Good Q, "filter" is a little misleading here. It's taking DeprecationWarning and raising it to `error`. Revert the fix and this test will fail.

jacobtylerwalls: Otherwise the test would pass anyway, since `pytest` doesn't fail tests when there are deprecationwarnings.

Pierre-Sassoulas: We might as well change the global configuration, we often have to so "fix deprecation warning" merge request both for astroid and pylint.

PR #1703, opened 2022-07-11
[teyit] Add teyit to the pre-commit hooks

Pierre-Sassoulas:
## Steps

## Type of Changes

## Related Issue

We have a lot of non-pytest legacy test class in astroid so teyit feels appropriate.

### DIFF
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -4543,7 +4543,7 @@ class HasattrTest(unittest.TestCase):
         )
         inferred = next(node.infer())
         self.assertIsInstance(inferred, nodes.Const)
-        self.assertIs(inferred.value, False)
+        self.assertFalse(inferred.value)
 
 
 class BoolOpTest(unittest.TestCase):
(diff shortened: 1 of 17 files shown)


tests/unittest_inference.py
```diff
@@ -4543,7 +4543,7 @@ def test_lambda(self) -> None:
         )
         inferred = next(node.infer())
         self.assertIsInstance(inferred, nodes.Const)
-        self.assertIs(inferred.value, False)
+        self.assertFalse(inferred.value)
```
Pierre-Sassoulas, line 4546: This one feels wrong because we already check with a specialized function and ``inferred.value`` could be falsey.

DanielNoord: Yeah, this is incorrect, see: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertFalse.

Makes me wonder whether we should really use this.

I have looked into https://github.com/dannysepler/pytestify and similar tools but thought it wasn't worth the churn to apply them.

Pierre-Sassoulas: Migrating to pytest's assertions actually make more sense.

PR #1714, opened 2022-07-16
Fix crash in `ExplicitNamespacePackageFinder`

### LINKED ISSUE #1708: Astroid 2.12 breaks sphinx autoapi (at least for Airflow)
### Steps to reproduce

1. Checkout apache airlfow (https://github.com/apache/airflow)
2. Install airflow with doc extras `pip install ".[doc]"`
3. Make sure to use latest astroid version `pip install --upgrade astroid` (2.12.1 should be installed)
4. export PYTHONPATH=$(pwd)
5. ./docs/build_docs.py --package-filter apache-airflow --one-pass-only --docs-only
6. docs build fails with 111  errors similar to:
```
------------------------------ Error  95 --------------------
  File "/usr/local/lib/python3.7/site-packages/astroid/nodes/node_ng.py", line 182, in infer
------------------------------ Error  96 --------------------
  File "/usr/local/lib/python3.7/site-packages/astroid/nodes/scoped_nodes/scoped_nodes.py", line 1842, in _is_metaclass
```
7. downgrade astroid to 2.11.7: `pip install astroid==2.11.7`
8. ./docs/build_docs.py --package-filter apache-airflow --one-pass-only --docs-only
9. Observe the build succeeds.

NOTE. The above might fail because of libenchant could not be installed (especially on MacOS silicon) 
    You can reproduce it easily using our CI docker image - with all dependencies needed installed:

1. docker run -it ghcr.io/apache/airflow/main/ci/python3.7:2af19f16a4d94e749bbf6c7c4704e02aac35fc11
2. ./docs/build_docs.py --package-filter apache-airflow --one-pass-only --docs-only
3. docs build fails with 111  errors similar to:
```
------------------------------ Error  95 --------------------
  File "/usr/local/lib/python3.7/site-packages/astroid/nodes/node_ng.py", line 182, in infer
------------------------------ Error  96 --------------------
  File "/usr/local/lib/python3.7/site-packages/astroid/nodes/scoped_nodes/scoped_nodes.py", line 1842, in _is_metaclass
```
4. downgrade astroid to 2.11.7: `pip install astroid==2.11.7`
5. ./docs/build_docs.py --package-filter apache-airflow --one-pass-only --docs-only
6. Observe the build succeeds.

Also downgrading to 2.12.0 leads to failures.
Example failure in CI: https://github.com/apache/airflow/runs/7291353339?check_suite_focus=true#step:9:6115

### Current behavior

Version 2.12.*  of astroid causes sphinx autoapi to fail
Version 2.11.7 of astroid and before works.

### Expected behavior

* Version 2.12.* of astroid should work with sphinx autoapi 

```
python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"
2.12.1
```

jacobtylerwalls:
## Steps

## Description
Some weirdness with `_SixMetaPathImporter` demonstrated that `astroid.interpreter._import.util.is_namespace` needs to check the presence of `submodule_search_locations` before reporting something to be a namespace package.

## Type of Changes

## Related Issue
Closes #1708

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,11 +12,9 @@ What's New in astroid 2.12.3?
 =============================
 Release date: TBA
 
-
-* Fixed crash in ``ExplicitNamespacePackageFinder``.
+* Fixed crash in ``ExplicitNamespacePackageFinder`` involving ``_SixMetaPathImporter``.
 
   Closes #1708
-
 
 * Fix unhandled `FutureWarning` from pandas import in cython modules
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,7 +20,6 @@ Release date: TBA
 
   Closes #1717
 
-
 What's New in astroid 2.12.2?
 =============================
 Release date: 2022-07-12
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,10 @@ What's New in astroid 2.12.3?
 =============================
 Release date: TBA
 
+* Fixed crash in ``ExplicitNamespacePackageFinder``.
+
+  Closes #1708
+
 
 What's New in astroid 2.12.2?
 =============================
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -201,12 +201,7 @@ class ExplicitNamespacePackageFinder(ImportlibFinder):
         if processed:
             modname = ".".join(processed + [modname])
         if util.is_namespace(modname) and modname in sys.modules:
-            try:
-                submodule_path = sys.modules[modname].__path__
-            except AttributeError:
-                submodule_path = sys.modules[
-                    modname
-                ].__spec__.submodule_search_locations
+            submodule_path = sys.modules[modname].__path__
             return ModuleSpec(
                 name=modname,
                 location="",
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -201,7 +201,12 @@ class ExplicitNamespacePackageFinder(ImportlibFinder):
         if processed:
             modname = ".".join(processed + [modname])
         if util.is_namespace(modname) and modname in sys.modules:
-            submodule_path = sys.modules[modname].__path__
+            try:
+                submodule_path = sys.modules[modname].__path__
+            except AttributeError:
+                submodule_path = sys.modules[
+                    modname
+                ].__spec__.submodule_search_locations
             return ModuleSpec(
                 name=modname,
                 location="",
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -72,7 +72,8 @@ def is_namespace(modname: str) -> bool:
         if found_spec and found_spec.submodule_search_locations:
             last_submodule_search_locations = found_spec.submodule_search_locations
 
-    if found_spec is None:
-        return False
-
-    return found_spec.origin is None
+    return (
+        found_spec is not None
+        and found_spec.submodule_search_locations is not None
+        and found_spec.origin is None
+    )
--- a/tests/unittest_modutils.py
+++ b/tests/unittest_modutils.py
@@ -17,6 +17,7 @@ from pathlib import Path
 from xml import etree
 from xml.etree import ElementTree
 
+import pytest
 from pytest import CaptureFixture, LogCaptureFixture
 
 import astroid
@@ -24,6 +25,13 @@ from astroid import modutils
 from astroid.interpreter._import import spec
 
 from . import resources
+
+try:
+    import six  # pylint: disable=unused-import
+
+    HAS_SIX = True
+except ImportError:
+    HAS_SIX = False
 
 
 def _get_file_from_object(obj) -> str:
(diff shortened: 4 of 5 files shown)


Pierre-Sassoulas (approved):
LGTM !

astroid/interpreter/_import/spec.py
```diff
@@ -201,7 +201,12 @@ def find_module(
         if processed:
             modname = ".".join(processed + [modname])
         if util.is_namespace(modname) and modname in sys.modules:
-            submodule_path = sys.modules[modname].__path__
+            try:
```
DanielNoord, line 204: Do we understand why `__path__` isn't available? 

https://docs.python.org/3/reference/import.html#path__
`If the module is a package (either regular or namespace), the module object’s __path__ attribute must be set.`

We probably shouldn't crash, but is this "normal" behaviour?

https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec
`Thus it is possible to update the module’s __path__ at runtime, and this will not be automatically reflected in __spec__.submodule_search_locations.`

`__path__` not being available seems to indicate that this isn't a package and thus perhaps we shouldn't return a `ModuleSpec` with a `PY_NAMESPACE` type?

Pierre-Sassoulas:
It looks pretty similar to an issue we have with pypy in pylint's main https://github.com/PyCQA/pylint/runs/7428964268?check_suite_focus=true

astroid/interpreter/_import/spec.py
jacobtylerwalls: Just dashing off a quick reply that when I looked into it I thought it was because `urllib3` was doing magic with `six`, so I wasn't concerned, but I can try to get a real answer :-)

jacobtylerwalls: Coming back to this, I guess I don't want to take on the burden of understanding `_SixMetaPathImporter` inside and out -- there are four bugs filed against it anyway, so I wouldn't trust it to work in all scenarios, esp. if libraries like `urllib3` are doing things on top of it. I'd rather just catch the error until someone can demonstrate that `astroid` is at fault.

Pierre-Sassoulas: Sometime the code is too dynamic to be reasonably understood by astroid. Understanding this kind of dynamic code would be very good and reduce the use of brain and workarounds but I agree with @jacobtylerwalls. Ready to release the next patch version of astroid as soon as we merge this :)

DanielNoord: Note that this is still on my radar but I haven't been around much. Should be able to get to this in a 10 days or so. My main question is whether the check for `__path__` should be included in the `is_namespace` function.

jacobtylerwalls: Thanks Daniel, looking again, I think you have a point. I pushed a new commit to fix this in `is_namespace()`. ModuleSpec has `submodule_search_locations` as a better name than `__path__` per [PEP 451](https://peps.python.org/pep-0451/#attributes).

tests/unittest_modutils.py
```diff
@@ -439,5 +447,14 @@ def test_is_module_name_part_of_extension_package_whitelist_success(self) -> Non
         )
 
 
+@pytest.mark.skipif(not HAS_URLLIB3, reason="This test requires urllib3.")
+def test_file_info_from_modpath__SixMetaPathImporter() -> None:
+    pytest.raises(
+        ImportError,
```
jacobtylerwalls, line 453: So we no longer have a finder that will find the module created by `_SixMetaPathImporter`, but that's better than crashing while trying to find it with `ExplicitNamespacePackageFinder` (or incorrectly finding it with it!)

ChangeLog
```diff
@@ -12,10 +12,15 @@ What's New in astroid 2.12.3?
...
 
+* Fixed crash in ``ExplicitNamespacePackageFinder`` involving ``_SixMetaPathImporter``.
+
+  Closes #1708
+
 * Fix unhandled `FutureWarning` from pandas import in cython modules
 
   Closes #1717
 
+
```
Pierre-Sassoulas, line 23:
```suggestion
```

Pierre-Sassoulas (approved):
👍

astroid/interpreter/_import/util.py
```diff
@@ -72,7 +72,8 @@ def is_namespace(modname: str) -> bool:
         if found_spec and found_spec.submodule_search_locations:
             last_submodule_search_locations = found_spec.submodule_search_locations
 
-    if found_spec is None:
-        return False
-
-    return found_spec.origin is None
+    return (
+        found_spec is not None
```
DanielNoord, line 76: Should we use `last_submodule_search_locations` here?

jacobtylerwalls: I don't know, I suspect (without looking too closely) that could create false positives if a portion of the path was a valid namespace package but the rest of the path wasn't. That variable is really just a hack to get around the KeyError described in the comment, I wouldn't trust it to represent more than that.

DanielNoord (approved):

Pierre-Sassoulas:
I won't be able to release astroid until Friday, feel free to release if you want @jacobtylerwalls :)

potiuk:
Cool. Looking forward to it :)

PR #1722, opened 2022-07-26
Remove str instance in model of BaseException.attrs

mymedia2:
Do not require first exception argument to be a string. The word "usually" does not imply an obligation.
https://docs.python.org/3/library/exceptions.html#BaseException.args

## Steps

## Type of Changes

## Related Issue

Refs: PyCQA/pylint#7233

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -672,10 +672,7 @@ class InstanceModel(ObjectModel):
 class ExceptionInstanceModel(InstanceModel):
     @property
     def attr_args(self):
-        message = node_classes.Const("")
-        args = node_classes.Tuple(parent=self._instance)
-        args.postinit((message,))
-        return args
+        return node_classes.Tuple(parent=self._instance)
 
     @property
     def attr___traceback__(self):


Pierre-Sassoulas:
All tests are still passing on pylint current main branch.

Pierre-Sassoulas (approved):
I created https://github.com/PyCQA/pylint/pull/7236 in pylint, it fails without this, it succeed with it, thank you for fixing the issue, 👍

astroid/interpreter/objectmodel.py
```diff
@@ -672,10 +672,7 @@ def attr___dict__(self):
 class ExceptionInstanceModel(InstanceModel):
     @property
     def attr_args(self):
```
DanielNoord, line 674:
```suggestion
    def attr_args(self) -> nodes.Tuple:
```

astroid/interpreter/objectmodel.py
```diff
@@ -672,10 +672,7 @@ def attr___dict__(self):
...
-        message = node_classes.Const("")
-        args = node_classes.Tuple(parent=self._instance)
-        args.postinit((message,))
-        return args
+        return node_classes.Tuple(parent=self._instance)
```
DanielNoord, line 675:
```suggestion
        return nodes.Tuple(parent=self._instance)
```

DanielNoord (changes requested):
I think it makes more sense to add the test in `astroid`, that's where the change is and also is a better place for a test in case we change this later on.

@mymedia2 Could you add a test similar to the one Pierre added in this PR please?

DanielNoord:
Superseded by https://github.com/PyCQA/astroid/pull/1749

PR #1743, opened 2022-08-17
Fix crash with inference of type-annotated Enum classes

mbyrnepr2:
Fix crash with inference of type-annotated Enum classes where the member has no value.

## Steps

## Type of Changes

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -379,7 +379,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 continue
 
             inferred_return_value = None
-            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)) and stmt.value:
+            if stmt.value is not None:
                 if isinstance(stmt.value, nodes.Const):
                     if isinstance(stmt.value.value, str):
                         inferred_return_value = repr(stmt.value.value)
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -379,7 +379,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 continue
 
             inferred_return_value = None
-            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
+            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)) and stmt.value:
                 if isinstance(stmt.value, nodes.Const):
                     if isinstance(stmt.value.value, str):
                         inferred_return_value = repr(stmt.value.value)
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -379,7 +379,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 continue
 
             inferred_return_value = None
-            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
+            if isinstance(stmt, nodes.Assign) or isinstance(stmt, nodes.AnnAssign) and stmt.value:
                 if isinstance(stmt.value, nodes.Const):
                     if isinstance(stmt.value.value, str):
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -379,7 +379,11 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 continue
 
             inferred_return_value = None
-            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
+            if (
+                isinstance(stmt, nodes.Assign)
+                or isinstance(stmt, nodes.AnnAssign)
+                and stmt.value
+            ):
                 if isinstance(stmt.value, nodes.Const):
```
Pierre-Sassoulas, line 387: Nitpick but I would do this (it feels simpler to me):
```suggestion
            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
                if stmt.value and isinstance(stmt.value, nodes.Const):
```

tests/unittest_scoped_nodes.py
```diff
@@ -2515,6 +2515,25 @@ class Veg(Enum):
     assert inferred_member_value.value == "sweet"
 
 
+@pytest.mark.parametrize("annotation", ["bool", "dict", "int", "str"])
```
Pierre-Sassoulas, line 2518: 👍

Pierre-Sassoulas (approved):
Thanks !

astroid/brain/brain_namedtuple_enum.py
mbyrnepr2: 👍

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -380,7 +380,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
 
             inferred_return_value = None
             if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
-                if isinstance(stmt.value, nodes.Const):
+                if stmt.value and isinstance(stmt.value, nodes.Const):
```
Pierre-Sassoulas, line 383:
```suggestion
                if isinstance(stmt.value, nodes.Const):
```

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -380,7 +380,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
 
             inferred_return_value = None
             if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
```
Pierre-Sassoulas, line 382:
```suggestion
            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)) and stmt.value:
```

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -379,7 +379,7 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 continue
 
             inferred_return_value = None
-            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)):
+            if isinstance(stmt, (nodes.AnnAssign, nodes.Assign)) and stmt.value:
```
cdce8p, line 382: `stmt` will always be of type `AnnAssign` or `Assign`. If it isn't, `continue` is triggered a few lines above. Thus, only checking for `stmt.value` is enough.
https://github.com/PyCQA/astroid/blob/492a4bfde15fdd4e14de7a93a160a9fafa6ecf46/astroid/brain/brain_namedtuple_enum.py#L371-L382
```suggestion
            if stmt.value is not None:
```

cdce8p (changes requested):
Thanks for the quick follow up @mbyrnepr2!

astroid/brain/brain_namedtuple_enum.py
mbyrnepr2: Hats off that makes perfect sense!

cdce8p (approved):
Just double checked Home Assistant and the crash is indeed fixed.
Thanks again @mbyrnepr2 🚀

PR #1745, opened 2022-08-21
Bump ``wrapt`` to ``1.14``

DanielNoord:
## Steps

## Description

Ref. https://github.com/PyCQA/astroid/issues/1516#issuecomment-1220984764.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,7 +30,7 @@ Release date: TBA
 
   Closes #1737
 
-* Bumped minimum requirement of ``wrapt`` to 1.14.
+* Bumped minimum requirement of ``wrapt`` to 1.14 on Python 3.11.
 
 * Don't add dataclass fields annotated with ``KW_ONLY`` to the list of fields.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,8 @@ Release date: TBA
 
   Closes #1737
 
+* Bumped minimum requirement of ``wrapt`` to 1.14.
+
 What's New in astroid 2.12.2?
 =============================
 Release date: 2022-07-12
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
I think we should differentiate between Python 3.11 and other interpreters. We are compatible with wrap lower than 1.14 after all, let's not create dependency issues fort user if we can easily avoid them.

DanielNoord:
But `1.14` supports `3.5+`? Wouldn't we unnecessarily complicate our own dependency for gains that we do not know about?

Pierre-Sassoulas:
> But 1.14 supports 3.5+?

The issue is not the version of the interpreter but the version required by all the other libs that can be installed by someone wanting to use pylint. There's probably a dependency on pypi or in an internal pypi repository somewhere that ask for wrapt < 1.14 (maybe for good reason, maybe without good reason but...). pylint/astroid would be impossible to install alongside it.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = [
 requires-python = ">=3.7.2"
 dependencies = [
     "lazy_object_proxy>=1.4.0",
-    "wrapt>=1.11,<2",
+    "wrapt>=1.14,<2",
```
Pierre-Sassoulas, line 36:
```suggestion
    "wrapt>=1.14,<2;python_version>='3.11'",
    "wrapt>=1.11,<2;python_version<'3.11'",
```

Pierre-Sassoulas (changes requested):
The diff looks bad, but I don't know what I did wrong.

ChangeLog
```diff
@@ -30,6 +30,8 @@ Release date: TBA
 
   Closes #1737
 
+* Bumped minimum requirement of ``wrapt`` to 1.14.
```
DanielNoord, line 33:
```suggestion
* Bumped minimum requirement of ``wrapt`` to 1.14 on Python 3.11.
```

Pierre-Sassoulas (approved):

PR #1746, opened 2022-08-21
Don't add ``KW_ONLY`` fields to dataclass fields

DanielNoord:
## Steps

## Description

Refs. https://github.com/PyCQA/pylint/issues/5767

/CC @jacobtylerwalls 

How should we deal with this. 

```python
import dataclasses


@dataclasses.dataclass
class A:
    _: dataclasses.KW_ONLY
    data: str


print(A.__dataclass_fields__)
```

Returns:
```python
❯ python test.py
{'data': Field(name='data',type=<class 'str'>,default=<dataclasses._MISSING_TYPE object at 0x107b04c50>,default_factory=<dataclasses._MISSING_TYPE object at 0x107b04c50>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=True,_field_type=_FIELD)}
```

But on < 3.10 `KW_ONLY` doesn't exist so this will crash. How should we deal with that? We can't infer `KW_ONLY` and thus don't know if it should be added to the fields.

## Type of Changes

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -19,9 +19,9 @@ import sys
 from collections.abc import Generator
 from typing import Tuple, Union
 
-from astroid import bases, context, inference_tip
+from astroid import bases, context, helpers, inference_tip
 from astroid.builder import parse
-from astroid.const import PY39_PLUS
+from astroid.const import PY39_PLUS, PY310_PLUS
 from astroid.exceptions import (
     AstroidSyntaxError,
     InferenceError,
@@ -409,10 +409,12 @@ def _is_class_var(node: NodeNG) -> bool:
 
 def _is_keyword_only_sentinel(node: NodeNG) -> bool:
     """Return True if node is the KW_ONLY sentinel."""
-    return any(
+    if not PY310_PLUS:
+        return False
+    inferred = helpers.safe_infer(node)
+    return (
         isinstance(inferred, bases.Instance)
         and inferred.qname() == "dataclasses._KW_ONLY_TYPE"
-        for inferred in node.infer()
     )
 
 
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -19,7 +19,7 @@ import sys
 from collections.abc import Generator
 from typing import Tuple, Union
 
-from astroid import context, inference_tip
+from astroid import bases, context, inference_tip
 from astroid.builder import parse
 from astroid.const import PY39_PLUS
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_dataclasses.py
```diff
@@ -404,6 +407,15 @@ def _is_class_var(node: NodeNG) -> bool:
     )
 
 
+def _is_keyword_only_sentinel(node: NodeNG) -> bool:
+    """Return True if node is the KW_ONLY sentinel."""
+    return any(
+        isinstance(inferred, bases.Instance)
+        and inferred.qname() == "dataclasses._KW_ONLY_TYPE"
+        for inferred in node.infer()
```
jacobtylerwalls, line 415: Do we not need to catch any exceptions?

jacobtylerwalls (changes requested):
Thanks for picking this up. I find it confusing to remove a legitimate field from the list of fields just because it has been marked keyword-only. Shouldn't we leave the fields alone and instead fix the constructor generated by `_generate_dataclass_init`? That was my hope behind marking the tickets as duplicates (having not looked at the code until just now).

DanielNoord:
Hmm, the field itself isn't keyword only though. See the output of the example in the OP. It's more like the `/` you can use in normal signatures, it changes the behaviour of other fields that come after it. But it seems as if the field which is annotated as `KW_ONLY` doesn't get added to the dictionary of fields of the class.

astroid/brain/brain_dataclasses.py
DanielNoord: Yeah, we should probably use `safe_infer` here. Good call.

Although, this depends on what we want to do with failing inference on < 3.10. Perhaps we want to catch it and still return `True`?

jacobtylerwalls: Does this condition need to check `if PY310_PLUS`?

DanielNoord: Yeah I don't really know. If we make this interpreter dependent then pylint will produce different output on different interpreters. Is that how we usually handle this?
Normally this would crash anyway as there is no KW_ONLY on lower versions in `dataclasses`.

jacobtylerwalls: Yes, that's how we normally handle python feature differences.

DanielNoord: Okay, then this becomes pretty easy to fix. I'll update the PR :)

jacobtylerwalls (approved):

PR #1747, opened 2022-08-22
Don't rely on the module cache when "importing self"

DanielNoord:
## Steps

## Description
Relevant historic commits are:
https://github.com/PyCQA/astroid/commit/6d59ad07d722d01e458aaf8fd14fd7dfc7ebaa6e
https://github.com/PyCQA/astroid/commit/9684d0fb05c4721876c2b7274b6af3f3122ff62d

Basically, by not relying on the cache we trigger a new import system and check and rightfully end up with the `stdlib` `math` module instead of the current module. I tested with `pylint` and test suite there still passes.

Refs: https://github.com/PyCQA/pylint/issues/7289 and https://github.com/PyCQA/pylint/issues/5151, but mostly the latter.

## Type of Changes


tests/unittest_manager.py
```diff
@@ -325,6 +325,24 @@ def hook(modname: str):
...
+        module = self.manager.ast_from_file(str(math_file))
+
+        # Change the cache key and module name to mimick importing the test file
+        # from the root/top level. This creates a clash between math.py and stdlib math.
+        self.manager.astroid_cache["math"] = self.manager.astroid_cache.pop(module.name)
+        module.name = "math"
+
+        # Infer the 'import math' statement
+        stdlib_math = next(module.body[1].value.args[0].infer())
+        assert self.manager.astroid_cache["math"] != stdlib_math
```
DanielNoord, line 344: Note that the old `math` stays in place in the cache. This is how we handle all cache clashes so I think this is also good in this case?

tests/resources.py
```diff
@@ -20,6 +20,10 @@ def find(name: str) -> str:
     return os.path.normpath(os.path.join(os.path.dirname(__file__), DATA_DIR, name))
 
 
+def find_in_testdata(name: str) -> Path:
+    return (Path(__file__).parent / "testdata" / name).resolve()
```
jacobtylerwalls, line 24: Why is the new test file the only one "loose" outside testdata/python3/data? That makes this new lookup unnecessary.

tests/unittest_manager.py
```diff
@@ -325,6 +325,24 @@ def hook(modname: str):
             with self.assertRaises(AstroidBuildingError):
                 self.manager.ast_from_module_name("foo.bar.baz")
 
+    def test_same_name_import_module(self) -> None:
+        """test inference of an import statement with the same name as the module.
+
+        see https://github.com/pycqa/pylint/issues/7289.
```
jacobtylerwalls, line 331: This seems like it's fixing PyCQA/pylint#5151, not PyCQA/pylint#7289, which is about modules that _do not_ end in `.py`. This test case ends in `.py`.

ChangeLog
```diff
@@ -6,6 +6,9 @@ What's New in astroid 2.13.0?
 =============================
 Release date: TBA
 
+* Fixed importing self when a module had the same name as an installed package.
```
jacobtylerwalls, line 9: I think this could be made a little more clear (what is importing self?)

jacobtylerwalls (commented):
Had some time to do a quick first pass.

tests/resources.py
```diff
@@ -20,6 +20,10 @@ def find(name: str) -> str:
     return os.path.normpath(os.path.join(os.path.dirname(__file__), DATA_DIR, name))
 
 
+def find_in_testdata(name: str) -> Path:
```
Pierre-Sassoulas, line 23:
```suggestion
def construct_absolute_path_in_testdata(name: str) -> Path:
```
For me the verb "find" imply searching for it by walking the directory tree. I would expect the give "foo.py" and get back the file in ``testdata/bar/`` called ``foo.py``

tests/unittest_manager.py
```diff
@@ -325,6 +325,24 @@ def hook(modname: str):
...
+        """
+        data_dir = resources.find_in_testdata("import_conflicting_names")
+        math_file = data_dir / "math.py"
```
Pierre-Sassoulas, line 334:
```suggestion
        math_file = resources.construct_absolute_path_in_testdata("import_conflicting_names/math.py")
```
I think this would work ?

Pierre-Sassoulas (changes requested):
Nice MR, I like the typing update too :)

astroid/manager.py
```diff
@@ -147,9 +147,18 @@ def _can_load_extension(self, modname: str) -> bool:
             modname, self.extension_package_whitelist
         )
 
-    def ast_from_module_name(self, modname, context_file=None):
-        """given a module name, return the astroid object"""
-        if modname in self.astroid_cache:
+    def ast_from_module_name(
+        self,
+        modname: str | None,
+        context_file: str | None = None,
```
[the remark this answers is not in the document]
DanielNoord: Other stuff down the "call tree" only accepts `str | None`. Notably `file_info_from_modpath` and many of the `spec` finding functions.

tests/resources.py
DanielNoord: I changed it to use `get_testdata_path`. The other one is a bit long.

tests/resources.py
DanielNoord: I think it makes sense to try and move data in `testdata` instead of `python3`. I tried moving everything some time ago but a lof of tests are dependent on the current directory structure. That said, I think adding new data in `testdata` instead of `testdata/python3` makes more sense. I'll happily move if you want to keep using `python3` though.

tests/unittest_manager.py
DanielNoord: I mean, they are both issues of the same type. Basically any file that conflicts with `same-named file and external module` will create this issue.

Want me to add them both? Or change to `5151`?

DanielNoord:
Ready for review again! @Pierre-Sassoulas @jacobtylerwalls 😄

tests/unittest_manager.py
jacobtylerwalls: I'm not sure they're the same issue. I thought timmartin demonstrated there was nothing wrong on #5151.

jacobtylerwalls: In any case, we need a test with a file not ending in .py if we're claiming to fix it! :D

tests/resources.py
jacobtylerwalls: Yeah I'm -0 on moving things around, I just don't see the point. And I'm -0.83333 on having tests in two different places.

DanielNoord: Changed to be in one place.

Pierre-Sassoulas (approved):

ChangeLog
```diff
@@ -6,6 +6,11 @@ What's New in astroid 2.13.0?
 =============================
 Release date: TBA
 
+* Fixed importing of modules that have the same name as the file that is importing.
+  A file called ``math.py`` will now correctly ``import math`` from the standard library
+  instead of trying to import itself.
```
jacobtylerwalls, line 11: Up to you, if you like this better.
```suggestion
  ``astroid`` will now correctly view a file called ``math.py`` as importing math from the
    standard library when it calls ``import math`` instead of treating it as importing itself.
```

tests/unittest_manager.py
```diff
@@ -325,6 +325,23 @@ def hook(modname: str):
...
 
+    def test_same_name_import_module(self) -> None:
+        """Test inference of an import statement with the same name as the module.
+
+        See https://github.com/PyCQA/pylint/issues/5151.
+        """
+        math_file = resources.find("data/import_conflicting_names/math.py")
+        module = self.manager.ast_from_file(math_file)
+
+        # Change the cache key and module name to mimick importing the test file
```
jacobtylerwalls, line 336:
```suggestion
        # Change the cache key and module name to mimic importing the test file
```

jacobtylerwalls (changes requested):
The test passes for me on `main`. Would you have the opportunity to create a test that fails before the fix?

ChangeLog
DanielNoord: What about:
```suggestion
  ``astroid`` will now correctly handle an ``import math`` statement in a file called ``math.py``
  by relying on the import system.
```

DanielNoord:
> The test passes for me on `main`. Would you have the opportunity to create a test that fails before the fix?

Are you sure?

```diff
diff --git a/tests/unittest_manager.py b/tests/unittest_manager.py
index 9cf02aff8..fe1f22642 100644
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -325,6 +325,22 @@ class AstroidManagerTest(
             with self.assertRaises(AstroidBuildingError):
                 self.manager.ast_from_module_name("foo.bar.baz")
 
+    def test_same_name_import_module(self) -> None:
+        """Test inference of an import statement with the same name as the module.
+        See https://github.com/PyCQA/pylint/issues/5151.
+        """
+        math_file = resources.find("data/import_conflicting_names/math.py")
+        module = self.manager.ast_from_file(math_file)
+
+        # Change the cache key and module name to mimic importing the test file
+        # from the root/top level. This creates a clash between math.py and stdlib math.
+        self.manager.astroid_cache["math"] = self.manager.astroid_cache.pop(module.name)
+        module.name = "math"
+
+        # Infer the 'import math' statement
+        stdlib_math = next(module.body[1].value.args[0].infer())
+        assert self.manager.astroid_cache["math"] != stdlib_math
+
 
 class BorgAstroidManagerTC(unittest.TestCase):
     def test_borg(self) -> None:

```
+ the `math.py` file

Fails for me with `pytest -k test_same_name_import_module` on:
```python
>       assert self.manager.astroid_cache["math"] != stdlib_math
E       assert <Module.math l.0 at 0x110670210> != <Module.math l.0 at 0x110670210>
```

DanielNoord:
Oh never mind. That doesn't actually fail. You just can't `Module != Module`. I'll have a look today.

ChangeLog
jacobtylerwalls: I like it.

jacobtylerwalls (approved):
Yeah, I did something horribly wrong when testing it, I guess. Sorry!

emcd:
This PR may cause breakage with the use of `importlib.import_module` on PyPy 3.8. Please see #1755 for reference.

PR #1752, opened 2022-08-23
Fix setuptools>=64 import hooks

DanielNoord:
## Steps

## Description

This needs a test. Refs: https://github.com/PyCQA/pylint/issues/7306 and https://github.com/pypa/setuptools/issues/3518.

## Type of Changes

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -14,6 +14,7 @@ import pathlib
 import sys
 import zipimport
 from collections.abc import Iterator, Sequence
+from importlib.abc import MetaPathFinder
 from pathlib import Path
 from typing import Any, NamedTuple
 
@@ -41,6 +42,13 @@ class ModuleType(enum.Enum):
     PY_SOURCE = enum.auto()
     PY_ZIPMODULE = enum.auto()
     PY_NAMESPACE = enum.auto()
+
+
+_MetaPathFinderModuleTypes: dict[str, ModuleType] = {
+    # Finders created by setuptools editable installs
+    "_EditableFinder": ModuleType.PY_SOURCE,
+    "_EditableNamespaceFinder": ModuleType.PY_NAMESPACE,
+}
 
 
 class ModuleSpec(NamedTuple):
@@ -349,7 +357,7 @@ def _find_spec_with_path(
     module_parts: list[str],
     processed: list[str],
     submodule_path: Sequence[str] | None,
-) -> tuple[Finder, ModuleSpec]:
+) -> tuple[Finder | MetaPathFinder, ModuleSpec]:
     for finder in _SPEC_FINDERS:
         finder_instance = finder(search_path)
         spec = finder_instance.find_module(
@@ -358,6 +366,28 @@ def _find_spec_with_path(
         if spec is None:
             continue
         return finder_instance, spec
+
+    # Support for custom finders
+    for meta_finder in sys.meta_path:
+        spec = meta_finder.find_spec(modname, submodule_path)
+        if spec:
+            try:
+                module_type = _MetaPathFinderModuleTypes[
+                    meta_finder.__name__  # type: ignore[attr-defined]
+                ]
+            except KeyError:
+                # If we don't recognise the finder, we assume it's a regular module
+                module_type = ModuleType.PY_SOURCE
+            return (  # type: ignore[return-value]
+                meta_finder,
+                ModuleSpec(
+                    spec.name,
+                    module_type,
+                    spec.origin,
+                    spec.origin,
+                    spec.submodule_search_locations,
+                ),
+            )
 
     raise ImportError(f"No module named {'.'.join(module_parts)}")
 
@@ -394,7 +424,7 @@ def find_spec(modpath: list[str], path: Sequence[str] | None = None) -> ModuleSp
             _path, modname, module_parts, processed, submodule_path or path
         )
         processed.append(modname)
-        if modpath:
+        if modpath and isinstance(finder, Finder):
             submodule_path = finder.contribute_to_path(spec, processed)
 
         if spec.type == ModuleType.PKG_DIRECTORY:
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -398,7 +398,8 @@ def _find_spec_with_path(
         if not hasattr(meta_finder, "find_spec"):
             continue
 
-        if spec := meta_finder.find_spec(modname, submodule_path):
+        spec = meta_finder.find_spec(modname, submodule_path)
+        if spec:
             return (
                 meta_finder,
                 ModuleSpec(
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -387,7 +387,7 @@ def _find_spec_with_path(
         meta_finder_name = meta_finder.__class__.__name__
         if meta_finder_name not in _MetaPathFinderModuleTypes:
             # Setuptools>62 creates its EditableFinders dynamically and have
-            # "type" as their __clas__.__name__. We check __name__ as well
+            # "type" as their __class__.__name__. We check __name__ as well
             # to see if we can support the finder.
             try:
                 meta_finder_name = meta_finder.__name__
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -373,9 +373,9 @@ def _find_spec_with_path(
         if spec:
             try:
                 module_type = _MetaPathFinderModuleTypes[
-                    meta_finder.__name__  # type: ignore[attr-defined]
+                    meta_finder.__name__  # type: ignore[attr-defined], caught by AttributeError
                 ]
-            except KeyError:
+            except (KeyError, AttributeError):
                 # If we don't recognise the finder, we assume it's a regular module
                 module_type = ModuleType.PY_SOURCE
             return (  # type: ignore[return-value]
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -38,7 +38,7 @@ class _MetaPathFinder(Protocol):
         path: Sequence[str] | None,
         target: types.ModuleType | None = ...,
     ) -> importlib.machinery.ModuleSpec | None:
-        ...
+        ...  # pragma: no cover
 
 
 class ModuleType(enum.Enum):
@@ -248,7 +248,6 @@ class ZipFinder(Finder):
         super().__init__(path)
         for entry_path in path:
             if entry_path not in sys.path_importer_cache:
-                # pylint: disable=no-member
                 try:
                     sys.path_importer_cache[entry_path] = zipimport.zipimporter(  # type: ignore[assignment]
                         entry_path
@@ -332,7 +331,6 @@ def _is_setuptools_namespace(location: pathlib.Path) -> bool:
 
 def _get_zipimporters() -> Iterator[tuple[str, zipimport.zipimporter]]:
     for filepath, importer in sys.path_importer_cache.items():
-        # pylint: disable-next=no-member
         if isinstance(importer, zipimport.zipimporter):
             yield filepath, importer
 
@@ -383,15 +381,24 @@ def _find_spec_with_path(
 
     # Support for custom finders
     for meta_finder in sys.meta_path:
-        spec = meta_finder.find_spec(modname, submodule_path)
-        if spec:
-            try:
-                module_type = _MetaPathFinderModuleTypes[
-                    meta_finder.__name__  # type: ignore[attr-defined] # Caught by AttributeError
-                ]
-            except (KeyError, AttributeError):
-                # If we don't recognise the finder, we assume it's a regular module
-                module_type = ModuleType.PY_SOURCE
+        # See if we support the customer import hook of the meta_finder
+        try:
+            meta_finder_name: str = meta_finder.__name__  # type: ignore[attr-defined]
+        except AttributeError:
+            continue
+        if meta_finder_name not in _MetaPathFinderModuleTypes:
+            continue
+
+        module_type = _MetaPathFinderModuleTypes[meta_finder_name]
+
+        # Meta path finders are supposed to have a find_spec method since
+        # Python 3.4. However, some third-party finders do not implement it.
+        # PEP302 does not refer to find_spec as well.
+        # See: https://github.com/PyCQA/astroid/pull/1752/
+        if not hasattr(meta_finder, "find_spec"):
+            continue
+
+        if spec := meta_finder.find_spec(modname, submodule_path):
             return (
                 meta_finder,
                 ModuleSpec(
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -12,9 +12,9 @@ import importlib.util
 import os
 import pathlib
 import sys
+import types
 import zipimport
 from collections.abc import Iterator, Sequence
-from importlib.abc import MetaPathFinder
 from pathlib import Path
 from typing import Any, NamedTuple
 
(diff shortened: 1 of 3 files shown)


DanielNoord:
@jacobtylerwalls This fixes the test that started failing after https://github.com/PyCQA/astroid/pull/1714

Would you mind taking a look at the PR and the linked issues/discussions? I'm hesitant to go forward, but this actually seems like it fixes some issues.

jacobtylerwalls:
I'd like to provide input, but the fall is going to be very busy for me. I'm not sure when I'll have the time to take a look. I'm very hesitant on doing anything additional touching the import system for 2.15.

DanielNoord:
> I'd like to provide input, but the fall is going to be very busy for me. I'm not sure when I'll have the time to take a look. I'm very hesitant on doing anything additional touching the import system for 2.15.

Let's definitely not put this in `2.15`. However, since this is going to affect myself as well when I update my local setuptools installation at some point I'd like to get this on `astroid` `main` at least. I'll try and see if I can get some more context in the meantime.

DanielNoord:
@Pierre-Sassoulas Would you be okay with putting this in the next release? This will be necessary for `pylint` not to be incompatible with `setuptools>=64` and I'd like to get feedback on it from users before everybody is on `>=64`.

Pierre-Sassoulas (changes requested):
Sure but I think this need a rebase the changelog is not in 2.14.0

DanielNoord:
> Sure but I think this need a rebase the changelog is not in 2.14.0

Yeah I will prepare the PR for final review then. Thanks!

Pierre-Sassoulas:
On closer inspection (seeing Jacob's comment) this feel like something a little risky if we want to release 2.16.0 soon and not do another beta release.  Maybe we should do an rc release ?

DanielNoord:
> On closer inspection (seeing Jacob's comment) this feel like something a little risky if we want to release 2.16.0 soon and not do another beta release. Maybe we should do an rc release ?

Yeah I don't know. Based on the tests I think this should be fine, but we have been wrong before. If anything we can just quickly revert the change and release a patch `astroid` release? There is no code that builds upon this (both in `astroid` or `pylint`) so I don't foresee many merge issues when back porting.

Pierre-Sassoulas:
I'm afraid 😄 Let's wait for a review by Jacob, forcing a namespace change last minute can ruin the effort we made with two beta releases and make our lives difficult (when we lived without this for months).

jacobtylerwalls:
Sorry for not prioritizing this yet. @DanielNoord would you be able to write a (short) summary of why you believe this is the correct approach? That would help me get into this. (There has been more discussion on the linked PyPA issue.)

DanielNoord:
The idea is that we actually have a quite an advanced import system compared to other tools and are able to follow imports (somewhat successfully).
However, when installing editable installs with `pip` a new type of `_SPEC_FINDERS` is returned. We don't have support for those but we can quite easily add them by telling our import system that we should handle them as normal (namespace) packages.
I chose to use the `__name__` attribute of the finders as we then don't build in dependencies on `pip` and `setuptools` etc. but can just infer the type of package from the finder's name.

The test case that now succeeds is notable as it shows that with this PR we actually are able to infer more imports than we previously could.

astroid/interpreter/_import/spec.py
```diff
@@ -394,7 +424,7 @@ def find_spec(modpath: list[str], path: Sequence[str] | None = None) -> ModuleSp
             _path, modname, module_parts, processed, submodule_path or path
         )
         processed.append(modname)
-        if modpath:
+        if modpath and isinstance(finder, Finder):
```
jacobtylerwalls, line 427: No tests fail when I revert this change.

DanielNoord: This is because of the typing change of `_find_spec_with_path`. `MetaPathFinder` is not a subclass of `Finder`, see:
https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder

astroid/interpreter/_import/spec.py
```diff
@@ -359,6 +367,28 @@ def _find_spec_with_path(
...
+                ]
+            except (KeyError, AttributeError):
+                # If we don't recognise the finder, we assume it's a regular module
+                module_type = ModuleType.PY_SOURCE
+            return (  # type: ignore[return-value]
+                meta_finder,
+                ModuleSpec(
+                    spec.name,
+                    module_type,
+                    spec.origin,
```
jacobtylerwalls, line 386: Should we set location to `""` or `None` rather than `spec.origin`?

astroid/interpreter/_import/spec.py
```diff
@@ -359,6 +367,28 @@ def _find_spec_with_path(
             continue
         return finder_instance, spec
 
+    # Support for custom finders
+    for meta_finder in sys.meta_path:
+        spec = meta_finder.find_spec(modname, submodule_path)
```
jacobtylerwalls, line 372: I think we have to guard against `find_spec` being absent. It was only recently added to `six._SixMetaPathImporter`, see https://github.com/benjaminp/six/pull/352.

astroid/interpreter/_import/spec.py
jacobtylerwalls: I'm still confused: won't this cause a behavior change across versions? On Python 3.9 this condition will match; on Py3.10 it won't.

astroid/interpreter/_import/spec.py
```diff
@@ -349,7 +357,7 @@ def _find_spec_with_path(
     module_parts: list[str],
     processed: list[str],
     submodule_path: Sequence[str] | None,
-) -> tuple[Finder, ModuleSpec]:
+) -> tuple[Finder | MetaPathFinder, ModuleSpec]:
```
jacobtylerwalls, line 360: I'm not so sure we can limit the type of the return object now. The `six._SixMetaPathImporter` inherits directly from `object`:

```py
(Pdb) meta_finder.__class__.mro()
[<class 'urllib3.packages.six._SixMetaPathImporter'>, <class 'object'>]
```

jacobtylerwalls (changes requested):
Thanks for the description, that really helped focus my review. This is looking really good. One requested change on guarding against a missing `find_spec` and then some questions.

astroid/interpreter/_import/spec.py
DanielNoord: Hm, this is an issue within `six` then..

https://github.com/python/typeshed/blob/b9270da7b083f08a86e4366b3cf284e627a63873/stdlib/sys.pyi#L45

The `stdlib` is typed with a protocol to signal that this should indeed inherit from the `abc`. I'll make us use a `Protocol` as well.

jacobtylerwalls: I think part of the issue is that PEP 302 doesn't mention `find_spec`, so there could be code in the wild without it.

jacobtylerwalls:
Also, I noticed two failures in pylint with this change, including one crash and one `astroid-error`:

```py
pytest -k private-import

Exception on node <ImportFrom l.33 at 0x1144f4790> in file '/Users/jacobwalls/pylint/tests/functional/ext/private_import/private_import.py'
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 91, in walk
    callback(astroid)
  File "/Users/.../pylint/pylint/checkers/imports.py", line 562, in visit_importfrom
    self._add_imported_module(node, f"{imported_module.name}.{name}")
  File "/Users/.../pylint/pylint/checkers/imports.py", line 894, in _add_imported_module
    elif not astroid.modutils.is_standard_module(importedmodname):
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../astroid/astroid/modutils.py", line 528, in is_standard_module
    filename = file_from_modpath([modname])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../astroid/astroid/modutils.py", line 334, in file_from_modpath
    return file_info_from_modpath(modpath, path, context_file).location
  File "/Users/.../astroid/astroid/modutils.py", line 385, in file_info_from_modpath
    return _spec_from_modpath(modpath, path, context)
  File "/Users/.../astroid/astroid/modutils.py", line 595, in _spec_from_modpath
    found_spec = spec.find_spec(modpath, path)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../astroid/astroid/interpreter/_import/spec.py", line 423, in find_spec
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../astroid/astroid/interpreter/_import/spec.py", line 372, in _find_spec_with_path
    spec = meta_finder.find_spec(modname, submodule_path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/_pytest/assertion/rewrite.py", line 92, in find_spec
    if self._early_rewrite_bailout(name, state):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/_pytest/assertion/rewrite.py", line 193, in _early_rewrite_bailout
    path = PurePath(*parts).with_suffix(".py")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pathlib.py", line 694, in with_suffix
    raise ValueError("%r has an empty name" % (self,))
ValueError: PurePosixPath('.') has an empty name

pytest -k non_init_parent_called

E       AssertionError: Wrong results for file "non_init_parent_called":
E       
E       Expected in testdata:
E          6: import-error
E         14: non-parent-init-called
E         22: no-member
E         27: no-member
E         50: no-member
E       
E          1: astroid-error
```

astroid/interpreter/_import/spec.py
DanielNoord: See my last commit, would that be an okay solution?

astroid/interpreter/_import/spec.py
DanielNoord: Why? Isn't that "sort of" right?

astroid/interpreter/_import/spec.py
DanielNoord: Yeah it might, it is just really hard to make this type safe..
`_MetaPathFinder` doesn't have this method so we can't safely call it.

We could use `hasattr` but that also feels unsatisfactory..

DanielNoord:
> ```python
> PurePosixPath
> ```

Feels like we should expand "." to the current working directory? 😓
That sounds like more issues..

We can switch that logic around? Only use the `meta_finders` in `sys.meta_path` that we know we support?

jacobtylerwalls:
Interesting that the failure is coming from `pytest` itself.

DanielNoord:
> Interesting that the failure is coming from `pytest` itself.

They probably have their own `MetaPathFinder` to handle all the `fixture` magic which we then try to use to locate an import.
That's why I'm debating on moving the check for known `MetaPathFinders` to above `line 372, in _find_spec_with_path`. That should resolve this I think.

astroid/interpreter/_import/spec.py
jacobtylerwalls: I think it's supposed to be `None` for namespace packages and a string for others:
https://github.com/jacobtylerwalls/astroid/commit/3ff65be6c9e3a4480cd9338a68f418cb578e347b

So we could switch on the `module_type` and change to `None` -> `""` if PY_SOURCE. But I'm interested in your opinion.

astroid/interpreter/_import/spec.py
jacobtylerwalls: Ah. I think we can't let type-checking considerations cause behavior differences.

astroid/interpreter/_import/spec.py
jacobtylerwalls: I guess I still don't understand, an importer without a find_spec method, such as [older vendored versions of six](https://sourcegraph.com/github.com/cronyo/cronyo/-/blob/cronyo/vendor/urllib3/packages/six.py?subtree=true) will still raise an uncaught AttributeError on `spec = meta_finder.find_spec(modname, submodule_path)`.

jacobtylerwalls:
> That's why I'm debating on moving the check for known MetaPathFinders to above line 372, in _find_spec_with_path. That should resolve this I think.

I'd be curious to see the pylint test results: I worry about giving new priority to the builtin importers, if that's what you're suggesting:

```py
>>> sys.meta_path
[<_distutils_hack.DistutilsMetaFinder object at 0x104452070>, <class '_frozen_importlib.BuiltinImporter'>, <class '_frozen_importlib.FrozenImporter'>, <class '_frozen_importlib_external.PathFinder'>]
```

astroid/interpreter/_import/spec.py
DanielNoord: The issue is that `find_module` has been deprecated since 3.4:
https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_module

Anyway, I'll keep the typing and use a `hasattr`. That seems the most "safe" while also correct.

astroid/interpreter/_import/spec.py
DanielNoord: It feels like this mimics `has_location`, see https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec.has_location.

Looking for `.location` it seems like only (frozen) builtins don't have their `location` set? And we use it in `ast_from_module_name` so we need it to be set to whatever `origin` is?

astroid/interpreter/_import/spec.py
DanielNoord: I'm dumb. `Finder != Finder`. The `Finder` in the `isinstance` check is actually our own implementation of `Finder`, which has `contribute_to_path`. `MetaPathFinders` don't so we can't call the method here.

Previously all `Finders` were of our own type but with this change this is no longer the case so we need to explicitly check this.

DanielNoord:
@jacobtylerwalls See my last change, this is what I meant with changing the order of checks. We still rely on our own 4 `Finders` initially, but if that doesn't work we start trying `MetaPathFinders` for which we now stuff works.

This means that we would need to add new support by hand, but it is much less error prone.

`pylint` test suite passed with https://github.com/PyCQA/astroid/pull/1752/commits/f889b5d6bb49278d9736b799172f19b3f09631df.

DanielNoord:
If we agree that this is the way to go I'll add support for `SixMetaPathImporter` so the test case that I "unfailed" in this PR will pass again.

astroid/interpreter/_import/spec.py
jacobtylerwalls: This is moot now that we're not setting `PY_SOURCE`. I was just wondering about the 0.05% of cases where we hit `KeyError` and assumed `PY_SOURCE` and then still set location to `None` instead of `""`, but it looks like after latest commits we aren't doing setting `PY_SOURCE`.

jacobtylerwalls (approved):
Looks good, thanks! (A test would be nice.)

DanielNoord:
> Looks good, thanks! (A test would be nice.)

I think the `six` test serves as a good test for the general logic. We can now import and find that module where we previously couldn't as we support the `_SixMetaPathImporter` now.
As for `setuptools` I don't think there is a good way to test it without introducing a dependency on it, which I don't really want.

I have tested this with https://github.com/jooola/pylint-import-error-bug to confirm that this branch does indeed fix the issue mentioned in the linked issue.

astroid/interpreter/_import/spec.py
```diff
@@ -382,12 +384,17 @@ def _find_spec_with_path(
...
         # See if we support the customer import hook of the meta_finder
-        try:
-            meta_finder_name: str = meta_finder.__name__  # type: ignore[attr-defined]
-        except AttributeError:
-            continue
+        meta_finder_name = meta_finder.__class__.__name__
         if meta_finder_name not in _MetaPathFinderModuleTypes:
-            continue
+            # Setuptools>62 creates its EditableFinders dynamically and have
+            # "type" as their __clas__.__name__. We check __name__ as well
```
jacobtylerwalls, line 390:
```suggestion
            # "type" as their __class__.__name__. We check __name__ as well
```

Pierre-Sassoulas (approved):
👍

jacobtylerwalls (approved):
Great!

> I think the six test serves as a good test for the general logic. We can now import and find that module where we previously couldn't as we support the _SixMetaPathImporter now.

Indeed. (Sorry, I thought you had said you were about to revert the changes to that test.)

DanielNoord:
> Great!
> 
> > I think the six test serves as a good test for the general logic. We can now import and find that module where we previously couldn't as we support the _SixMetaPathImporter now.
> 
> Indeed. (Sorry, I thought you had said you were about to revert the changes to that test.)

Yeah, that was unclear. My bad!

Thanks for the reviews all. I'm skipping the coverage check on this one!

PR #1756, opened 2022-08-27
Prevent first-party imports from being resolved to `site-packages`

jacobtylerwalls:
## Description
Regression in 5067f08.

## Type of Changes

## Related Issue
Closes PyCQA/pylint#7365

### DIFF
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -13,6 +13,8 @@ from importlib.util import _find_spec_from_path  # type: ignore[attr-defined]
 
 @lru_cache(maxsize=4096)
 def is_namespace(modname: str) -> bool:
+    from astroid.modutils import EXT_LIB_DIRS  # pylint: disable=import-outside-toplevel
+
     if modname in sys.builtin_module_names:
         return False
 
@@ -70,6 +72,11 @@ def is_namespace(modname: str) -> bool:
 
         # Update last_submodule_search_locations
         if found_spec and found_spec.submodule_search_locations:
+            if any(
+                location in EXT_LIB_DIRS
+                for location in found_spec.submodule_search_locations
+            ):
+                return False
             last_submodule_search_locations = found_spec.submodule_search_locations
 
     return (
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -15,7 +15,12 @@ from astroid.const import IS_PYPY
 
 @lru_cache(maxsize=4096)
 def is_namespace(modname: str) -> bool:
-    from astroid.modutils import EXT_LIB_DIRS  # pylint: disable=import-outside-toplevel
+    from astroid.modutils import (  # pylint: disable=import-outside-toplevel
+        EXT_LIB_DIRS,
+        STD_LIB_DIRS,
+    )
+
+    STD_AND_EXT_LIB_DIRS = STD_LIB_DIRS.union(EXT_LIB_DIRS)
 
     if modname in sys.builtin_module_names:
         return False
@@ -75,7 +80,7 @@ def is_namespace(modname: str) -> bool:
         # Update last_submodule_search_locations
         if found_spec and found_spec.submodule_search_locations:
             if any(
-                location in EXT_LIB_DIRS
+                any(location.startswith(lib_dir) for lib_dir in STD_AND_EXT_LIB_DIRS)
                 for location in found_spec.submodule_search_locations
             ):
                 return False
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
LGTM but let's wait for another review :)

Pierre-Sassoulas:
@DanielNoord could you check this please ? I'm thinking of releasing astroid 2.14.5 once it's merged.

astroid/interpreter/_import/util.py
```diff
@@ -72,6 +74,11 @@ def is_namespace(modname: str) -> bool:
 
         # Update last_submodule_search_locations
         if found_spec and found_spec.submodule_search_locations:
+            if any(
+                location in EXT_LIB_DIRS
+                for location in found_spec.submodule_search_locations
+            ):
+                return False
```
Pierre-Sassoulas, line 81: Look like this line is not covered, is there something wrong with the tests ?

jacobtylerwalls: Yes! b89ef27871d01a28e6e5ece6458836e0e8580b4f

(After testing my test, I butchered it when trying to "simplify" it further.)

jacobtylerwalls:
Test passes on macOS for me locally but fails on CI because of a discrepancy between:
'/home/runner/work/astroid/astroid/venv/lib/python3.7/site-packages' found by `sysconfig.get_path("purelib")`
and
'/opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages' found by python's import system

astroid/interpreter/_import/util.py
```diff
@@ -15,6 +15,13 @@
 
 @lru_cache(maxsize=4096)
 def is_namespace(modname: str) -> bool:
+    from astroid.modutils import (  # pylint: disable=import-outside-toplevel
+        EXT_LIB_DIRS,
```
DanielNoord, line 19: Can we move these to `const` to remove any import cycles? We could also move the union there.

astroid/interpreter/_import/util.py
DanielNoord: Can we add a comment about this. This function is becoming very intricate really quickly and I want to avoid the guessing we're doing for the old namespace detection in a couple of years from now.

DanielNoord (changes requested):
Change itself looks good, just some stylistic and trivial changes! Thanks @jacobtylerwalls!

dalcinl:
@jacobtylerwalls Success! 

Rather than fighting with local reproduction, I fired up a test on the original place of failure, that is, GitHub Actions.

I force-installed this PR [the following way](https://github.com/mpi4py/mpi4py/runs/8059039904?check_suite_focus=true#step:6:1). The pylint run step [is now clean](https://github.com/mpi4py/mpi4py/runs/8059039904?check_suite_focus=true#step:10:1).

Many thanks for the prompt address of this issue! I really appreciate it.

astroid/interpreter/_import/util.py
jacobtylerwalls: I considered that, but there's too much logic in `modutils` that would have to be moved. My preference would be to move `is_namespace` into `modutils`, but I'd prefer to do that in a followup commit that wouldn't be backported. What do you think?

DanielNoord: I think it would be good to back port this. We never know how long it takes until we release a minor version and we might want to make more fixes here.

I'd be happy to make a back port PR 😄

Pierre-Sassoulas: I opened #1759 we just need to merge this before release 2.12.5.

jacobtylerwalls: > I think it would be good to back port this

Ah, we might be miscommunicating! I'm saying yes, we should definitely backport the crash fix, but also that I wouldn't want to block that on refactoring the cyclic import out. There's a lot of logic in modutils that would have to be moved, and it's all public, so I wouldn't want to break backwards compatibility on that in a patch release.

So I was hoping to merge and backport this in its current form, and leave the refactor for another day. How does that sound?

DanielNoord: Let's merge this as is, but let's also back port the refactor. Like I said: with `astroid` the time in between minor versions can be quite long and we might need to fix something and release a patch version after the refactor. To avoid future merge issues I would also merge any refactor to this particular piece of code.

DanielNoord (approved):

astroid/interpreter/_import/util.py
Pierre-Sassoulas: I agree with @DanielNoord about cherry-picking the refactors. Case in point the 2.12 maintenance branch has been live for 2 months now. Cherry-picking things after a refactor without the refactor is a lot harder than just cherry-picking the refactor too. (We do need to make sure our refactor are just refactor, but we never had any problem with this so far).

PR #1783, opened 2022-09-15
Give a created Arguments node a parent

DanielNoord:
## Steps

## Description

This can lead to crashes when asking for the `__repr__` of an `Arguments` node as that assumes it has a `parent` attribute, like it should.

## Type of Changes

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -795,7 +795,9 @@ class PropertyModel(ObjectModel):
     """Model for a builtin property"""
 
     def _init_function(self, name):
-        args = nodes.Arguments()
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Should we add a regression test ?

DanielNoord:
> Should we add a regression test ?

I don't think we can. I only found this as I was using the debugger and iterated over all nodes in memory.

We can only write a test for the `Arguments` created in this specific piece of code, but I don't think we would ever revert this without good reason thus making the test invalid again.
We also can't create a test that iterates over all nodes ever created and see if they call `postinit` correctly.. I think this is just something we have to deal with due to how `postinit` works.

Pierre-Sassoulas (approved):

PR #1787, opened 2022-09-17
Partial typing of the ``infer_binary_op`` path

DanielNoord:
## Steps

## Description

Quite difficult to follow, but I think this should be correct. Let me know if something is wrong still!

## Type of Changes

### DIFF
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -11,8 +11,8 @@ from __future__ import annotations
 import collections
 import itertools
 import operator as operator_mod
-from collections.abc import Callable, Generator
-from typing import Any
+from collections.abc import Callable, Generator, Iterator, Sequence
+from typing import Any, TypeVar
 
 from astroid import arguments, bases, decorators, helpers, nodes, util
 from astroid.const import Context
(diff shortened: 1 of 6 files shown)


astroid/protocols.py
```diff
@@ -118,12 +125,18 @@ def _infer_unary_op(obj: Any, op: str) -> ConstFactoryResult:
...
+    operator: str,
+    other: InferenceResult,
+    context: InferenceContext,
+    _: SuccessfulInferenceResult,
+) -> Generator[ConstFactoryResult | type[util.Uninferable], None, None]:
     not_implemented = nodes.Const(NotImplemented)
     if isinstance(other, nodes.Const):
         if (
             operator == "**"
-            and isinstance(self, nodes.Const)
```
Pierre-Sassoulas, line 126: The little optimization like this are a nice cherry on the typing cake.

nickdrozd (approved):
Tested with `mypy --strict`; LGTM!

Pierre-Sassoulas (approved):
All right let's merge, there's no reason to risk a conflict, we can always fix it later if Marc find something.

PR #1790, opened 2022-09-18
Don't set flake8 max-complexity and max-line-length and also ignore them

cclauss:
## Steps

- [ ] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description
Flake8 code `C901` measures code complexity so don't `ignore` it and also set `max-complexity`.  Instead set `max-complexity` high enough so that tests pass.  This will make contributors and reviewers think twice before adding even more complexity.

Flake8 code `B950` and `E501` measure line length so don't `ignore` them and also set `max-line-length`.  Instead set `max-line-length` high enough so that tests pass.  This will make contributors and reviewers think twice before adding even longer lines.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |
| ✓   | :sparkles: New feature |
| ✓   | :hammer: Refactoring   |
| ✓   | :scroll: Docs          |

PR #1791, opened 2022-09-18
Saner high values for flake8's line length and complexity checks

Pierre-Sassoulas:
## Type of Changes

## Related Issue

Inspired by #1790

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -240,7 +240,7 @@ def _forbid_class_getitem_access(node: ClassDef) -> None:
     def full_raiser(origin_func, attr, *args, **kwargs):
         """
         Raises an AttributeInferenceError in case of access to __class_getitem__ method.
-        Otherwise just call origin_func.
+        Otherwise, just call origin_func.
         """
         if attr == "__class_getitem__":
             raise AttributeInferenceError("__class_getitem__ access is not allowed")
@@ -248,8 +248,9 @@ def _forbid_class_getitem_access(node: ClassDef) -> None:
 
     try:
         node.getattr("__class_getitem__")
-        # If we are here, then we are sure to modify object that do have __class_getitem__ method (which origin is one the
-        # protocol defined in collections module) whereas the typing module consider it should not
+        # If we are here, then we are sure to modify an object that does have
+        # __class_getitem__ method (which origin is the protocol defined in
+        # collections module) whereas the typing module considers it should not.
         # We do not want __class_getitem__ to be found in the classdef
         partial_raiser = partial(full_raiser, node.getattr)
         node.getattr = partial_raiser
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -672,7 +672,7 @@ nodes.NamedExpr.assigned_stmts = named_expr_assigned_stmts
 
 
 @decorators.yes_if_nothing_inferred
-def starred_assigned_stmts(
+def starred_assigned_stmts(  # noqa: C901
     self: nodes.Starred,
     node: node_classes.AssignedStmtsPossibleNode = None,
     context: InferenceContext | None = None,
@@ -823,7 +823,6 @@ def starred_assigned_stmts(
         last_lookup = lookup_slice
 
         for element in itered:
-
             # We probably want to infer the potential values *for each* element in an
             # iterable, but we can't infer a list of all values, when only a list of
             # step values are expected:
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,10 +14,7 @@ extend-ignore =
     E203, # Incompatible with black see https://github.com/psf/black/issues/315
     W503, # Incompatible with black
     B901 # Combine yield and return statements in one function
-max-complexity = 83
-max-line-length = 120
-# per-file-ignores =
-#    astroid/brain/brain_numpy_core_multiarray.py: E501, B950
+max-line-length = 110
 select = B,C,E,F,W,T4,B9
 # Required for flake8-typing-imports (v1.12.0)
 # The plugin doesn't yet read the value from pyproject.toml
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,9 @@ extend-ignore =
     W503, # Incompatible with black
     B901 # Combine yield and return statements in one function
 max-complexity = 83
-max-line-length = 138
+max-line-length = 120
+# per-file-ignores =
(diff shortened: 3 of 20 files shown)


setup.cfg
```diff
@@ -30,8 +30,10 @@ extend-ignore =
     E203, # Incompatible with black see https://github.com/psf/black/issues/315
     W503, # Incompatible with black
     B901 # Combine yield and return statements in one function
-max-complexity = 83
-max-line-length = 138
+max-complexity = 36
+max-line-length = 120
+per-file-ignores =
+   astroid/rebuilder.py: C901
```
cclauss, line 36: Why not do this with a in-line comment: `# noqa: C901` instead of ignoring the whole file?
Doing so makes any editor of that file think about decreasing, not increasing complexity.

astroid/brain/brain_typing.py
```diff
@@ -240,16 +240,17 @@ def _forbid_class_getitem_access(node: ClassDef) -> None:
...
         """
         if attr == "__class_getitem__":
             raise AttributeInferenceError("__class_getitem__ access is not allowed")
         return origin_func(attr, *args, **kwargs)
 
     try:
         node.getattr("__class_getitem__")
-        # If we are here, then we are sure to modify object that do have __class_getitem__ method (which origin is one the
-        # protocol defined in collections module) whereas the typing module consider it should not
+        # If we are here, then we are sure to modify object that do have
```
DanielNoord, line 251: While we are here, do we want to translate this into actual English? 😄

setup.cfg
```diff
@@ -30,8 +30,10 @@ extend-ignore =
     E203, # Incompatible with black see https://github.com/psf/black/issues/315
     W503, # Incompatible with black
     B901 # Combine yield and return statements in one function
-max-complexity = 83
-max-line-length = 138
+max-complexity = 36
+max-line-length = 120
```
DanielNoord, line 34: Can we get this lower? Or is this a sane current max?

setup.cfg
DanielNoord: Yeah I think `in-line` makes sense.

tests/unittest_brain.py
```diff
@@ -1473,7 +1473,8 @@ def test_collections_object_not_yet_subscriptable_2(self):
 
     @test_utils.require_version(minver="3.9")
     def test_collections_object_subscriptable_3(self):
-        """With python39 ByteString class of the colletions module is subscritable (but not the same class from typing module)"""
+        """With python39 ByteString class of the collections module is subscritable
```
DanielNoord, line 1476: This also seems to be a poor translation of some other language

astroid/brain/brain_typing.py
Pierre-Sassoulas: What would you suggest ? 😄

setup.cfg
Pierre-Sassoulas: There's a lot of over 100+, let's try 110.

astroid/brain/brain_typing.py
Pierre-Sassoulas: (Maybe it's because I'm french like the original writer but it make sense to me)

DanielNoord: I mean `to modify object` seems to be missing a `the`? or should it be `objects`? The tenses seem to be a bit messed up and honestly I don't even understand what it is trying to say 😄

Pierre-Sassoulas:
```suggestion
        # If we are here, then we are sure to modify objects that do have
```

tests/unittest_brain.py
Pierre-Sassoulas:
```suggestion
        """With python39 the ByteString class of the collections module is subscritable
```

astroid/brain/brain_typing.py
```diff
@@ -240,16 +240,17 @@ def _forbid_class_getitem_access(node: ClassDef) -> None:
...
+        # __class_getitem__ method (which origin is one the protocol defined in
+        # collections module) whereas the typing module consider it should not.
```
DanielNoord, line 253:
```suggestion
        # If we are here, then we are sure to modify an object that does have
        # __class_getitem__ method (which origin is the protocol defined in
        # collections module) whereas the typing module considers it should not.
```

I think? I really don't understand the sentence after `(which ...`

astroid/protocols.py
```diff
@@ -672,7 +672,7 @@ def named_expr_assigned_stmts(
 
 
 @decorators.yes_if_nothing_inferred
-def starred_assigned_stmts(
+def starred_assigned_stmts(  # noqa: C901
```
DanielNoord, line 675: Should we maybe just set the limit a little higher? We seem to be adding a lot of diables.

tests/unittest_brain.py
```diff
@@ -1984,7 +1985,8 @@ class Derived(typing.Hashable, typing.Iterator[int]):
         )
 
     def test_typing_object_notsubscriptable_3(self):
-        """Until python39 ByteString class of the typing module is not subscritable (whereas it is in the collections module)"""
+        """Until python39 ByteString class of the typing module is not
+        subscritable (whereas it is in the collections' module)"""
```
DanielNoord, line 1989:
```suggestion
        subscriptable (whereas it is in the collections' module)"""
```

astroid/protocols.py
cclauss: Agreed.  Let's `noqa` the three or four most complex functions and then set max-complexity above the other.

Pierre-Sassoulas: It means we can introduce some new functions that will be too complex in the future. There isn't that much complex functions to noqa (and I did noqa them all already). This is the reasoning we had when we did not filter the whole protocol file but added a noqa instead.

tests/unittest_brain.py
```diff
@@ -1473,7 +1473,7 @@ def test_collections_object_not_yet_subscriptable_2(self):
 
     @test_utils.require_version(minver="3.9")
     def test_collections_object_subscriptable_3(self):
-        """With python39 ByteString class of the collections module is subscritable
+        """With python39 the ByteString class of the collections module is subscritable
```
cclauss, line 1476:
```suggestion
        """With Python 3.9 the ByteString class of the collections module is subscritable
```
The executable is `python3.9` with the dot being significant.

DanielNoord (approved):

PR #1795, opened 2022-09-18
tests: pyupgrade --py37-plus **/*.py

cclauss:
## Steps

- [x] Write a good description of what the PR does.

## Type of Changes
Upgrade files in `tests/` to use Python 3.7 and greater syntax using https://github.com/asottile/pyupgrade

### DIFF
--- a/tests/testdata/python3/data/SSL1/Connection1.py
+++ b/tests/testdata/python3/data/SSL1/Connection1.py
@@ -1,4 +1,3 @@
-
 class Connection:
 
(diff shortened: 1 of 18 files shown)


Pierre-Sassoulas:
Thank you for opening the pull request. We already use pyupgrade (see https://github.com/PyCQA/astroid/blob/main/.pre-commit-config.yaml#L34) but this is the test data directory, we need to be able to analyse old but valid syntax without crashing.

PR #1797, opened 2022-09-19
Use pylint-config minimal option to move pylint config into pyproject.toml

cclauss:
## Steps

## Description
```
pipx install git+https://github.com/PyCQA/pylint.git
pylint-config --rcfile=pylintrc generate --interactive .
* Do you want a minimal configuration without comments or default values?
    (y)es or (n)o  (default=no) y
git rm pylintrc
```
Sorting longer arrays like plugins and disable makes duplicates jump out like `pylint.extensions.code_style`
## Type of Changes


Pierre-Sassoulas:
@cclauss this branch was the one we should keep, right ? I think we can remove old values that are not the default and it's good to go (thank you!)

DanielNoord:
I self requested a review because there were some comments that shouldn't be removed, for example a `TODO` that can probably actually be fixed.
I don't think `pylint-config` will ever be able to notice this so we will still need to validate if there aren't any useful comments in the old file.

cclauss:
I think that this process should be a way to test PyCQA/pylint#7484 as well as the new pylint-config minimal approach.  If we merge a PR that has a lot of manual modifications then that testing opportunity is gone.  I will leave it to pylint experts to complete this process.

Pierre-Sassoulas:
Some values are not the default, but we need to remove them anyway because the default is good enough (astroid did not need the "CVS" option for more than 10+ years for example).

PR #1801, opened 2022-09-20
Handle empty ``modname`` in ``ast_from_module_name``

DanielNoord:
## Steps

## Description

Based on the typing we support this, but currently this would crash with an `AttributeError` in one of the called functions. This handles this a little bit nicer, even though we never actually call this like this.

## Type of Changes

### DIFF
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -155,6 +155,8 @@ class AstroidManager:
         use_cache: bool = True,
     ) -> nodes.Module:
         """Given a module name, return the astroid object."""
+        if modname is None:
+            raise AstroidBuildingError("No module name given.")
         # Sometimes we don't want to use the cache. For example, when we're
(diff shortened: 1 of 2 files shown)


DanielNoord:
@Pierre-Sassoulas Gentle ping as I think you might have missed this in my spam of `astroid` PRs 😅

Pierre-Sassoulas (commented):
Should we forbid this in astroid 3.0 and deprecate now then change the typing ? This feel "temporary".

DanielNoord:
I mean it doesn't matter too much? It's a relatively quick check and I think there are edge cases we don't currently hit in which `modname` is not set correctly and can be `None`. I think we can just leave as is.

Pierre-Sassoulas (approved):
Fair enough :)

PR #1777, opened 2022-09-10
Stop detecting modules compiled by `cffi` as namespace packages

jacobtylerwalls:
## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description

The `is_namespace()` changes in 2.12 broke introspection of modules compiled by `cffi`, which as quoted in the issue, patch `sys.modules`. The only reason we are looking at `sys.modules` is to support the edge case of old `pkg_resources`-style namespace packages. The principle of false negatives are better than letting things crash suggests that we should just loosen the check to top level modnames ("pylint" rather than "pylint.lint"), because that allows the `cffi` modules to not crash without failing the `pkg_resources` tests. (Another way of saying: I'm not aware of any false negatives we're causing with `pkg_resources`-style packages, since the tests pass, but I can't rule it out.)

No tests--did not want to introduce a test dependency on `cffi`.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Related Issue

Closes #1776
Closes PyCQA/pylint#7399

@dalcinl, does this resolve your issue?

PR #1802, opened 2022-09-21
Improve is_namespace() check for modules where `__spec__` is None

skshetry:
## Steps

## Description

See https://stackoverflow.com/a/42962529.

Let's take the following contents as an example:
```python
import celery.result

```

From #1777, astroid started to use `processed_components` for namespace check. In the above case, the `modname` is `celery.result`, it first checks for `celery` and then `celery.result`. Before that PR, it'd always check for `celery.result`.

~~But if you have imported it as `celery.result`,
`sys.modules["celery"].__spec__` is going to be `None`, and hence the function will return True, but below where we try to load get `submodule_path`/`__path__` for `celery.result` will fail as it is not a package. Also celery is not a namespace package.~~

[`celery` is recreating module to make it lazily load](https://github.com/celery/celery/blob/34533ab44d2a6492004bc3df44dc04ad5c6611e7/celery/__init__.py#L150), which does not have `__spec__` set. Reading through [Python's docs](https://docs.python.org/3/reference/import.html#spec__), it seems that `__spec__` can be set to None, so it seems like it's not a thing that we can depend upon for namespace checks.

---

The `celery.result` gets imported for me when `pylint-pytest` plugin tries to load fixtures, but this could happen anytime if any plugin imports packages. In that case, `importlib.util._find_spec_from_path("celery")` will raise ValueError since it's already in `sys.modules` and does not have a spec.

## Type of Changes

## Related Issue

Fixes https://github.com/PyCQA/pylint/issues/7488.

### DIFF
--- a/astroid/interpreter/_import/util.py
+++ b/astroid/interpreter/_import/util.py
@@ -52,8 +52,11 @@ def is_namespace(modname: str) -> bool:
                 # Check first fragment of modname, e.g. "astroid", not "astroid.interpreter"
                 # because of cffi's behavior
                 # See: https://github.com/PyCQA/astroid/issues/1776
+                mod = sys.modules[processed_components[0]]
                 return (
-                    sys.modules[processed_components[0]].__spec__ is None
+                    mod.__spec__ is None
+                    and getattr(mod, "__file__", None) is None
+                    and hasattr(mod, "__path__")
                     and not IS_PYPY
                 )
             except KeyError:
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,10 @@ Release date: TBA
 
   Refs PyCQA/pylint#5151
 
+* Improve detection of namespace packages for the modules with ``__spec__`` set to None.
+
+  Closes PyCQA/pylint#7488.
+
 
 What's New in astroid 2.12.11?
 ==============================
--- a/tests/unittest_manager.py
+++ b/tests/unittest_manager.py
@@ -144,6 +144,15 @@ class AstroidManagerTest(
         finally:
             astroid_module.__spec__ = original_spec
 
+    def test_module_unexpectedly_spec_is_none(self) -> None:
+        astroid_module = sys.modules["astroid"]
+        original_spec = astroid_module.__spec__
+        astroid_module.__spec__ = None
+        try:
+            self.assertFalse(util.is_namespace("astroid"))
+        finally:
+            astroid_module.__spec__ = original_spec
+
     def test_implicit_namespace_package(self) -> None:
         data_dir = os.path.dirname(resources.find("data/namespace_pep_420"))
         contribute = os.path.join(data_dir, "contribute_to_namespace")


DanielNoord:
Thanks for that through investigation! I don't have the time for a full review right now, but this would definitely need a test in order not to regress in the future. It also needs a small entry in the changelog of the next patch release.

skshetry:
>  this would definitely need a test in order not to regress in the future. It also needs a small entry in the changelog of the next patch release.

Thanks, I have added a changelog and a test. I was wrong before, the `celery` module does not have `__spec__` set due to which this fails. See https://github.com/celery/celery/blob/34533ab44d2a6492004bc3df44dc04ad5c6611e7/celery/__init__.py#L150 where they recreate module to make it lazily load, which does not have `__spec__` set. I added a test for a similar scenario.

DanielNoord:
We probably should support this but we should probably open an issue against `celery` for this. I think it's a bit up to semantics but the [documentation for `__spec__`](https://docs.python.org/3/reference/import.html#spec__) does seem to imply that it should be set where possible.

skshetry:
I was/am a bit hesitant to create a PR due to unclear semantics. My understanding is `__spec__` can be `None`, so I believe it's pylint/astroid responsibility to handle here. Anyway, I can try proposing a PR and see.

DanielNoord:
The PEP is really unclear:
https://peps.python.org/pep-0451/#module-objects

> Other than adding `__spec__`, none of the import-related module attributes will be changed or deprecated, though some of them could be; any such deprecation can wait until Python 4.

Similarly the docs state, https://docs.python.org/3/reference/import.html#spec__:
> The `__spec__` attribute must be set to the module spec that was used when importing the module.`

The `must` and `add` seem to imply some sort of requirement but I think this is somewhat relaxed due to backward compatibility concerns.

I did a deep dive into `typeshed` and how it types `__spec__`.
First commit:
https://github.com/python/typeshed/commit/05e02c188f4757ec859b4f2ba8ef17a0a4d9389b
Second commit:
https://github.com/python/typeshed/commit/43c34067701b48cd931769742b4c55511324dad4

The first commit is linked to https://github.com/python/typeshed/pull/321, which is part of https://github.com/python/typeshed/issues/189.

There is a crucial reply in that thread by Brett:

> I'm going to fix this issue step-by-step to make sure I'm doing it correctly, so I'm starting w/ `importlib.__init__` and I will build from there.
> 
> I also noticed that the `types` module has the wrong details for modules, so I will fix that as part of this issue as well (`__package__`, `__loader__`, and `__spec__` as initially set to `None` [as of Python 3.4](https://docs.python.org/3/whatsnew/3.4.html); `__doc__` and `__name__` have always existed).

My (perhaps somewhat biased) conclusion is that `__spec__` is `ModuleType | None` because it was only introduced in 3.4. However, the language with which the documentation and PEP refer to it seem to imply that ideally it should be set. I completely agree that `astroid` should correctly deal with those cases where it isn't, but I think the evidence here makes a compelling case for `celery` to at least consider adding it as well. Where possible it seems best practice to make it available.

skshetry:
Great points. Thank you for the great research, and all the help. :) 

I have opened https://github.com/celery/celery/pull/7773.

jacobtylerwalls (approved):
Thanks for the PR and investigation @skshetry, I appreciate having another pair of eyes on a very difficult bit of `astroid`'s code.

I agree we need to make `astroid` more robust against unexpected inputs. Your change looks sensible (and for this branch you edited, all we are trying to do is make sure that we still pass the test cases for old-style namespace packages such as `test_identify_old_namespace_package_protocol`).

astroid/interpreter/_import/util.py
```diff
@@ -52,8 +52,11 @@ def is_namespace(modname: str) -> bool:
                 # Check first fragment of modname, e.g. "astroid", not "astroid.interpreter"
                 # because of cffi's behavior
                 # See: https://github.com/PyCQA/astroid/issues/1776
+                mod = sys.modules[processed_components[0]]
                 return (
-                    sys.modules[processed_components[0]].__spec__ is None
+                    mod.__spec__ is None
```
skshetry, line 57: I wonder if we even need to check `mod.__spec__ is None`, but I don't fully understand all different styles of namespace packages. The test does pass even if we remove this.

DanielNoord (approved):

PR #1806, opened 2022-09-21
Finish typing of ``astroid.manager.py``

DanielNoord:
## Steps

## Description

Quite a lot, but this finishes most of this file 😄 

## Type of Changes

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -263,6 +263,7 @@ class AstroidBuilder(raw_building.InspectBuilder):
 
 
 def build_namespace_package_module(name: str, path: Sequence[str]) -> nodes.Module:
+    # TODO: Typing: Remove the cast to list and just update typing to accept Sequence
     return nodes.Module(name, path=list(path), package=True)
 
 
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import os
 import textwrap
 import types
+from collections.abc import Sequence
 from tokenize import detect_encoding
 
 from astroid import bases, modutils, nodes, raw_building, rebuilder, util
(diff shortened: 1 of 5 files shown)


DanielNoord:
@Pierre-Sassoulas This is slightly dependent on https://github.com/PyCQA/astroid/pull/1801 btw. Merging that PR will remove some of the `mypy` issues that turn up once this is merged. But since `mypy` is turned off anyway the order doesn't matter right now 😄

Pierre-Sassoulas:
Let's keep our option to cover the missing line later, if we add a pragma we'll never do it.

DanielNoord:
> Let's keep our option to cover the missing line later, if we add a pragma we'll never do it.

Then the current iteration makes sense. The pragma I did add does make sense: we really should never something that is not a `type` and doesn't have a `__class__` attribute. If we get something like that we are in some very unpythonic, constructed object and we should probably just raise that exception. I can't even begin to think of a test to cover it.

astroid/builder.py
```diff
@@ -261,8 +262,8 @@ def delayed_assattr(self, node):
             pass
 
 
-def build_namespace_package_module(name: str, path: list[str]) -> nodes.Module:
-    return nodes.Module(name, path=path, package=True)
+def build_namespace_package_module(name: str, path: Sequence[str]) -> nodes.Module:
+    return nodes.Module(name, path=list(path), package=True)
```
Pierre-Sassoulas, line 266: Should we type as sequence in Module constructor too ? I'm not sure of the performance implications of this cast.

Pierre-Sassoulas (commented):
Happy to see the day we can enable mypy is closer yet !

astroid/builder.py
DanielNoord: Shouldn't be too bad I think.

I'll add a TODO. I think this cascades to a lot of other things and even potentially to `pylint`. Because `mypy` can't be turned on right now it's quite difficult to do the updating everywhere were needed.

DanielNoord:
```suggestion
    # TODO: Typing: Remove the cast to list and just update typing to accept Sequence
    return nodes.Module(name, path=list(path), package=True)
```

Pierre-Sassoulas: Right this could be huge, let's wait for proper type check before going crazy.

Pierre-Sassoulas (approved):

PR #1812, opened 2022-09-30
Fix regression in the creation of the ``__init__`` of dataclasses

DanielNoord:
## Steps

## Description

Closes https://github.com/PyCQA/pylint/issues/7434

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,7 +25,7 @@ What's New in astroid 2.12.11?
 ==============================
 Release date: TBA
 
-* Fixed a regression in the creation of the ``__init__`` of ``dataclassess`` with
+* Fixed a regression in the creation of the ``__init__`` of dataclasses with
   multiple inheritance.
 
   Closes PyCQA/pylint#7434
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,7 +25,10 @@ What's New in astroid 2.12.11?
 ==============================
 Release date: TBA
 
+* Fixed a regression in the creation of the ``__init__`` of ``dataclassess`` with
+  multiple inheritance.
 
+  Closes PyCQA/pylint#7434
 
 What's New in astroid 2.12.10?
 ==============================
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -181,9 +181,12 @@ def _find_arguments_from_base_classes(
     node: nodes.ClassDef, skippable_names: set[str]
 ) -> tuple[str, str]:
     """Iterate through all bases and add them to the list of arguments to add to the init."""
-    prev_pos_only = ""
-    prev_kw_only = ""
-    for base in node.mro():
+    pos_only_store: dict[str, tuple[str | None, str | None]] = {}
+    kw_only_store: dict[str, tuple[str | None, str | None]] = {}
+    # See TODO down below
+    # all_have_defaults = True
+
+    for base in reversed(node.mro()):
         if not base.is_dataclass:
             continue
         try:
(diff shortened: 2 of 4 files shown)


DanielNoord:
I expected the coverage to be `sup-optimal`. I think it's okay to merge this right now. If we start using this method more then the coverage (and potential bugs) will come naturally. You just don't have `*args` and `**kwargs` in `dataclasses`.

Pierre-Sassoulas:
I'm way over my head with this MR, I'm going to pass unless there's really no other reviewer.

ChangeLog
```diff
@@ -25,7 +25,10 @@ What's New in astroid 2.12.11?
 ==============================
 Release date: TBA
 
+* Fixed a regression in the creation of the ``__init__`` of ``dataclassess`` with
```
jacobtylerwalls, line 28:
```suggestion
* Fixed a regression in the creation of the ``__init__`` of dataclasses with
```

tests/unittest_brain_dataclasses.py
```diff
@@ -991,6 +992,47 @@ class ChildWithMixedParents(BaseParent, NotADataclassParent):
...
+    """
+    )
+
+    first_init: bases.UnboundMethod = next(first.infer())
+    assert [a.name for a in first_init.args.args] == ["self", "required", "optional"]
+    assert [a.value for a in first_init.args.defaults] == [False]
+
+    # TODO: This should be an assertion, but the dataclass brain is a transform
+    # which currently can't return an Uninferable correctly.
+    assert not next(impossible.infer()) is Uninferable
```
jacobtylerwalls, line 1034: I'm having trouble following what "should be an assertion" means. You mean the assertion should have the opposite result? If so we should extract it into a test case and add an expected fail marker.

jacobtylerwalls (commented):
Just an initial skim.

tests/unittest_brain_dataclasses.py
DanielNoord: Yeah I thought about extracting but this example kinda belongs with the other test. Can we `xfail` within a function? (That's basically what I'm doing with the `assert not` here).

jacobtylerwalls: I'm not sure xfailing within a function helps, because you only get one result per (parameterized) test. That result should be xfailed. It's okay with me if you want to xfail the entire test.

DanielNoord: I can separate this test out, but doesn't that achieve the same as this `assert not`? The `# TODO` or `xfail` serves as a reminder to look into it and the comment or docstring explains that this is actually supposed to be the inverse of the current situation?

jacobtylerwalls: I didn't find the `TODO` clear. And I think future developers will rely on the assertion (e.g. with git blame) to understand what the correct behavior was understood to be at the time. So I'm -1 on having an assert that asserts the incorrect behavior. Also, the xfail is useful because it creates an xpass when it's finally fixed, which can we can use as an alert to clean things up. (I think there are some xpasses currently we should do something about, as it happens...)

jacobtylerwalls (approved):

astroid/brain/brain_dataclasses.py
```diff
@@ -282,7 +297,7 @@ def _generate_dataclass_init(
         params_string += ", "
 
     if prev_kw_only:
-        params_string += "*, " + prev_kw_only + ", "
+        params_string += "*, " + prev_kw_only
```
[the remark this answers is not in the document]
DanielNoord: `prev_kw_only` now takes care of this itself in the new helper function so there is not need to add it here anymore.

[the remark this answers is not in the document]
jacobtylerwalls: Thanks, that explains it. Merge ho!

PR #1834, opened 2022-10-16
Fix `getattr` inference with empty annotation assignments

cdce8p:
## Description
Don't exit with `AttributeInferenceError` for annotation assignments if parent class contains valid assignment.

```py
class Parent:
    attr: int = 2

class Child(Parent):
    attr: int
```

## Type of Changes

## Related Issue
Refs https://github.com/PyCQA/pylint/issues/7631

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -46,6 +46,7 @@ from astroid.nodes import Arguments, Const, NodeNG, _base_nodes, node_classes
 from astroid.nodes.scoped_nodes.mixin import ComprehensionScope, LocalsDictNodeNG
 from astroid.nodes.scoped_nodes.utils import builtin_lookup
 from astroid.nodes.utils import Position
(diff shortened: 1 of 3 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2525,7 +2526,12 @@ def instantiate_class(self) -> bases.Instance:
             pass
         return bases.Instance(self)
 
-    def getattr(self, name, context=None, class_context=True):
+    def getattr(
+        self,
+        name: str,
+        context: InferenceContext | None = None,
+        class_context: bool = True,
+    ) -> list[NodeNG]:
```
DanielNoord, line 2534: Are you sure about this? I belive `locals` is actually `SuccesfulInferenceResult` but typed incorrectly currently...

DanielNoord: See comment in other issue. This is not always the case and I think the typing should be updated here.

PR #1837, opened 2022-10-17
Add typing for `set_local`

cdce8p:
Ref https://github.com/PyCQA/astroid/pull/1834#pullrequestreview-1143601977
/CC: @DanielNoord

### DIFF
--- a/astroid/nodes/scoped_nodes/mixin.py
+++ b/astroid/nodes/scoped_nodes/mixin.py
@@ -6,7 +6,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, TypeVar
+from typing import TYPE_CHECKING, TypeVar, overload
 
 from astroid.filter_statements import _filter_stmts
 from astroid.nodes import node_classes, scoped_nodes
@@ -88,23 +88,21 @@ class LocalsDictNodeNG(node_classes.LookupMixIn):
         # self is at the top level of a module, or is enclosed only by ClassDefs
         return builtin_lookup(name)
 
-    def set_local(self, name, stmt):
+    def set_local(self, name: str, stmt: nodes.NodeNG) -> None:
         """Define that the given name is declared in the given statement node.
 
         .. seealso:: :meth:`scope`
 
         :param name: The name that is being defined.
-        :type name: str
 
         :param stmt: The statement that defines the given name.
-        :type stmt: NodeNG
         """
         # assert not stmt in self.locals.get(name, ()), (self, stmt)
         self.locals.setdefault(name, []).append(stmt)
 
     __setitem__ = set_local
 
-    def _append_node(self, child):
+    def _append_node(self, child: nodes.NodeNG) -> None:
         """append a child, linking it in the tree"""
         # pylint: disable=no-member; depending by the class
         # which uses the current class as a mixin or base class.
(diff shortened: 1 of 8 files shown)


Pierre-Sassoulas (approved):

DanielNoord:
As I said before, `locals` is not always only `NodeNG`. See for example:
https://github.com/PyCQA/astroid/blob/dabffad7e84c93f9d6dd8f0eb67e63a30e5e098b/astroid/brain/brain_io.py#L15-L21

We might need to fix that brain or just accept `SuccesfulInferenceResult` here. I believe there are also other places where this is the case as I looked into typing this previously and was held back by the inconsistencies.

cdce8p:
Thanks for the followup @DanielNoord! I should have checked it more carefully.
However, it's not as simple as it might seem. `instance` in the example you referenced is actually a subtype of `NodeNG`.
https://github.com/PyCQA/astroid/blob/aae90e6480c3fb9da95447587e639af6ad297689/astroid/nodes/node_ng.py#L613-L623

The only example I found now is with `nodes.Dict.items` which creeps in the `SuccessfulInferenceResult` typing.
https://github.com/PyCQA/astroid/blob/aae90e6480c3fb9da95447587e639af6ad297689/astroid/nodes/scoped_nodes/scoped_nodes.py#L2282-L2285

I'll update the PR.

cdce8p:
**Update**
Seems like `instantiate_class` can actually return an `Instance`. It's just not annotated in the base class.
https://github.com/PyCQA/astroid/blob/aae90e6480c3fb9da95447587e639af6ad297689/astroid/nodes/scoped_nodes/scoped_nodes.py#L2516-L2527

DanielNoord:
Ah yeah, I knew there was something with `Dict.items` but couldn't find it. Thanks for taking another look.

Like I said, I'm not even sure if it is correct or we should actually reject types like `Instance`.

I do think that updating this will cause multiple trickle down issues in other places as I think many return values depend on `locals`. However, it also doesn't seem weird that in many cases where we expect a node it can actually be a `SuccessfulInferenceResult`.

cdce8p:
> Like I said, I'm not even sure if it is correct or we should actually reject types like `Instance`.
> 
> I do think that updating this will cause multiple trickle down issues in other places as I think many return values depend on `locals`. However, it also doesn't seem weird that in many cases where we expect a node it can actually be a `SuccessfulInferenceResult`.

Since it so widely used in brain modules, I suspect any code dealing with it needs to handle `Instance` already. We'll see it over time, probably.

_Just a note, I've only updated `locals` itself. The typing for the `set_local` methods can stay the same as far as I can see._

astroid/nodes/scoped_nodes/mixin.py
```diff
@@ -26,7 +27,7 @@ class LocalsDictNodeNG(node_classes.LookupMixIn):
 
     # attributes below are set by the builder module or by raw factories
 
-    locals: dict[str, list[nodes.NodeNG]] = {}
+    locals: dict[str, list[SuccessfulInferenceResult]] = {}
```
DanielNoord, line 30: Do we want to update all other places where we define `local`?

DanielNoord (commented):
One final question!

Pierre-Sassoulas (approved):

astroid/nodes/scoped_nodes/mixin.py
cdce8p: Pushed an update. AFAICT there wasn't that much to deal with. Added `type: ignore` in a few places where `node.locals` would only return `node.NodeNG`. Those could theoretical be `cast` as well. Let me know what you think.

DanielNoord (approved):

PR #1841, opened 2022-10-19
Release branch 2.12.12

Pierre-Sassoulas:


ChangeLog
```diff
@@ -16,16 +16,42 @@ Release date: TBA
...
+
 What's New in astroid 2.12.11?
 ==============================
 Release date: 2022-10-10
 
+* Improve detection of namespace packages for the modules with ``__spec__`` set to None.
+
+  Closes PyCQA/pylint#7488.
+
+* Fixed a regression in the creation of the ``__init__`` of dataclasses with
```
DanielNoord, line 51: Seems like this was still in the wrong release and thus not in the changelog?

Pierre-Sassoulas: I used the release note from GitHub as a source of truth to fix the release note, if they were wrong then this is also wrong. (I had a conflict when cherry-picking the fix from main)

PR #1844, opened 2022-10-20
Fix uncaught ValueError on inference of __getitem__ with slice(..., ..., 0)

### LINKED ISSUE #1843: Uncaught ValueError on inferring __getitem__ with nonsensical slices
### Steps to reproduce

```python
import astroid
node = astroid.extract_node("[][::0]")
print(next(node.infer()))
```

### Current behavior

```python
  File "repro.py", line 3, in <module>
    print(next(node.infer()))
  File ".../astroid/nodes/node_ng.py", line 169, in infer
    yield from self._infer(context=context, **kwargs)
  File ".../astroid/decorators.py", line 140, in raise_if_nothing_inferred
    yield next(generator)
  File ".../astroid/decorators.py", line 109, in wrapped
  File ".../astroid/inference.py", line 440, in infer_subscript
    assigned = value.getitem(index_value, context)
  File ".../astroid/nodes/node_classes.py", line 3499, in getitem
    return _container_getitem(self, self.elts, index, context=context)
  File ".../astroid/nodes/node_classes.py", line 232, in _container_getitem
    new_cls.elts = elts[index_slice]
ValueError: slice step cannot be zero
```

Originally spotted in `pylint` with a minimal reproducer like:

```python
for _ in [][::0]:
    pass
```

### Expected behavior

`Uninferable`

```
(venv) $ python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"
2.13.0-dev0
(venv) $ git rev-parse HEAD
1a074cf66fc4fdb6714fac7afb14f9e7457cd704
```

nelfin:
## Steps

## Description

See #1843 for a reproducer.

## Type of Changes

## Related Issue

Closes #1843

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,9 @@ Release date: TBA
 
   Refs PyCQA/pylint#7706
 
+* Catch ``ValueError`` when indexing some builtin containers and sequences during inference.
+
+  Closes #1843
 
 What's New in astroid 2.12.12?
 ==============================
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -27,6 +27,7 @@ from astroid.exceptions import (
     AstroidError,
     AstroidIndexError,
     AstroidTypeError,
+    AstroidValueError,
     AttributeInferenceError,
     InferenceError,
     NameInferenceError,
@@ -441,6 +442,7 @@ def infer_subscript(
             except (
                 AstroidTypeError,
                 AstroidIndexError,
+                AstroidValueError,
                 AttributeInferenceError,
                 AttributeError,
             ) as exc:
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -22,6 +22,7 @@ from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidIndexError,
     AstroidTypeError,
+    AstroidValueError,
     InferenceError,
     NoDefault,
     ParentMissingError,
@@ -234,6 +235,13 @@ def _container_getitem(instance, elts, index, context=None):
             return new_cls
         if isinstance(index, Const):
             return elts[index.value]
+    except ValueError as exc:
+        raise AstroidValueError(
+            message="Slice {index!r} cannot index container",
+            node=instance,
+            index=index,
+            context=context,
+        ) from exc
     except IndexError as exc:
         raise AstroidIndexError(
             message="Index {index!s} out of range",
@@ -2030,6 +2038,10 @@ class Const(_base_nodes.NoChildrenNode, Instance):
         try:
             if isinstance(self.value, (str, bytes)):
                 return Const(self.value[index_value])
+        except ValueError as exc:
+            raise AstroidValueError(
+                f"Could not index {self.value!r} with {index_value!r}"
+            ) from exc
         except IndexError as exc:
             raise AstroidIndexError(
                 message="Index {index!r} out of range",
--- a/tests/unittest_inference.py
+++ b/tests/unittest_inference.py
@@ -5316,6 +5316,16 @@ def test_slice_inference_in_for_loops_not_working() -> None:
         assert inferred == util.Uninferable
 
 
+def test_slice_zero_step_does_not_raise_ValueError() -> None:
+    node = extract_node("x = [][::0]; x")
+    assert next(node.infer()) == util.Uninferable
+
+
+def test_slice_zero_step_on_str_does_not_raise_ValueError() -> None:
+    node = extract_node('x = ""[::0]; x')
+    assert next(node.infer()) == util.Uninferable
+
+
 def test_unpacking_starred_and_dicts_in_assignment() -> None:
     node = extract_node(
         """


astroid/inference.py
```diff
@@ -441,6 +442,7 @@ def infer_subscript(
             except (
                 AstroidTypeError,
                 AstroidIndexError,
+                AstroidValueError,
```
nelfin, line 445: Thoughts on simplifying this to `AstroidError`?

tests/unittest_inference.py
```diff
@@ -5316,6 +5316,11 @@ def test_slice_inference_in_for_loops_not_working() -> None:
         assert inferred == util.Uninferable
 
 
+def test_slice_zero_step_does_not_raise_ValueError() -> None:
+    node = extract_node("[][::0]")
+    assert helpers.safe_infer(node) is None
```
nelfin, line 5321: Preference on this integration with `safe_infer` vs isolation (and thus coupling to `InferenceError` behaviour instead of `Uniniferable`?)

nelfin:
https://github.com/PyCQA/astroid/actions/runs/3289097305/jobs/5420261690 seems spurious

astroid/inference.py
Pierre-Sassoulas: AstroidValueError make sense imo as we reraise a ValueError.

Pierre-Sassoulas:
Sorry for the cache miss, we have an issue right now, maybe Marc fixed it on main but I'm not certain of it.

tests/unittest_inference.py
DanielNoord: I'd prefer testing that this returns `Uninferable`.

nelfin: It does not return `Uninferable`, it raises `InferenceError`

astroid/inference.py
nelfin: Sorry, I meant simplifying all three lines (since they all derive from `AstroidError`)

nelfin: I'm happy to leave as-is

Pierre-Sassoulas: Hmm sorry I missed that. Yeah I think it's pretty safe to simplify the ``except`` (It's probably going to avoid problem in the future too).

tests/unittest_inference.py
DanielNoord: Oh, yeah of course. Let's test assigning this to a var then and inferring the var

nelfin: Like this?

```diff
 def test_slice_zero_step_does_not_raise_ValueError() -> None:
-    node = extract_node("[][::0]")
-    with pytest.raises(InferenceError):
-        next(node.infer())
+    node = extract_node("x = [][::0]; x")
+    assert next(node.infer()) == util.Uninferable
```

astroid/inference.py
nelfin: No need to apologise, when I look at it now it's not clear that the fact that I highlighted just the three lines is significant. It was clear to me when I highlighted them because that was an intentional action, but it's not clear when viewing the PR since it just looks like context. When I said "this", I should have said "these three lines".

tests/unittest_inference.py
DanielNoord: Yeah!

astroid/inference.py
nelfin: I think I'll leave it, `AstroidError` feels too broad. Maybe that means there's a "missing" exception type that could be the base of `Astroid{Index,Value,Type}Error` etc.? Maybe they should derive from `ResolveError` instead?

Pierre-Sassoulas: That make sense, we don't want to hide genuine problem in astroid by catching an exception that is too general. Maybe we leave it like this for now and we do a refactor in another MR ?

astroid/nodes/node_classes.py
```diff
@@ -234,6 +235,13 @@ def _container_getitem(instance, elts, index, context=None):
             return new_cls
         if isinstance(index, Const):
             return elts[index.value]
+    except ValueError as exc:
+        raise AstroidValueError(
+            message="Slice {index!r} cannot index container",
+            node=instance,
+            index=index,
+            context=context,
+        ) from exc
```
cdce8p, line 244: I think it would make more sense to catch it even earlier in `_infer_slice`. `0` is always an invalid value for `step`. So we shouldn't infer a `slice` object if we encounter it.
https://github.com/PyCQA/astroid/blob/4acf5785c54f4ccb8f41402991f6ddc5d8b28b89/astroid/nodes/node_classes.py#L211-L223

This would also have the benefit of fixing the other call to `_infer_slice` as well.
https://github.com/PyCQA/astroid/blob/4acf5785c54f4ccb8f41402991f6ddc5d8b28b89/astroid/nodes/node_classes.py#L2022-L2023

To reproduce, run pylint for
```py
for _ in ""[::0]:
    pass
```

tests/unittest_inference.py
```diff
@@ -5316,6 +5316,11 @@ def test_slice_inference_in_for_loops_not_working() -> None:
         assert inferred == util.Uninferable
 
 
+def test_slice_zero_step_does_not_raise_ValueError() -> None:
+    node = extract_node("x = [][::0]; x")
+    assert next(node.infer()) == util.Uninferable
```
cdce8p, line 5321: Please add a test for the other failure too.
Either via `infer` or a direct call to `Const.getitem`.

cdce8p (changes requested):

cdce8p (changes requested):
Could you also add a small changelog entry?

astroid/nodes/node_classes.py
nelfin: Thanks, this is a cleaner solution. I'll get to it today

nelfin: Looks like I've got a pretty generous definition of "today" these days. I think I originally considered this but stopped because it's not actually an error to _create_ a `slice` object with a step of 0, just an error to use it to index the built-in sequences:

```python
>>> slice(None, 10, 0)
slice(None, 10, 0)
>>> [][slice(None, 10, 0)]
ValueError: slice step cannot be zero
```

Authors can choose to handle extended indexes however they'd like, though the only example that comes to mind is `numpy.nd_grid`: https://numpy.org/doc/stable/reference/generated/numpy.mgrid.html
```python
>>> np.mgrid[0:3, 0:3]
array([[[0, 0, 0],
        [1, 1, 1],
        [2, 2, 2]],

       [[0, 1, 2],
        [0, 1, 2],
        [0, 1, 2]]])
```

or `range` supports sub-ranges through slices:
```python
>>> range(20)
range(0, 20)
>>> range(20)[::3]
range(0, 20, 3)
>>> range(20)[:5]
range(0, 5)
```

It turns out that slice indices don't even have to be integers as I've just learned, https://docs.python.org/3/reference/datamodel.html#index-67:
> ... start is the lower bound; stop is the upper bound; step is the step value; each is None if omitted. These attributes can have any type.

Pylint has tests for these cases (such as the string indices), but at least one of them is incorrect and will need to be changed: https://github.com/PyCQA/pylint/blob/main/tests/functional/i/invalid/invalid_slice_index.py#L25-L28
```python
# Valid indices
def function4():
    """integers used as indices"""
    return TESTLIST[0:0:0] # no error
```

I think the appropriate fix is probably:
1. Does the LHS have a custom `__getitem__` method? If so, we're probably in the uninferable realm
2. Try to defer the computed indexes to `slice.indices` since this is the canonical source of the raised `Type` and `ValueError`s:

```python
>>> [][:10:0]
ValueError: slice step cannot be zero
>>> slice(None, 10, 0).indices(0)
ValueError: slice step cannot be zero
>>> (1, 2, 3)["a":"z"]
TypeError: slice indices must be integers or None or have an __index__ method
>>> slice('a', 'z').indices(26)
TypeError: slice indices must be integers or None or have an __index__ method
```

I'm leaning towards adding the `except ValueError` case in `Const.getitem` similarly what was added here in `_container_getitem`. We can refactor the duplication out (now or later) since the distinction isn't really between "const" and "container", it's between "indexables" and "sequences" https://docs.python.org/3/reference/datamodel.html#index-15.

nelfin:
It occurs to me that these should probably return something useful for the Pylint `invalid-sequence-index` checks. @Pierre-Sassoulas, @cdce8p, @DanielNoord, what do you think? Returning `Uninferable` makes these lines ignored instead of a useful pylint warning:


```python
# fixtures/slices-pylint.py 
for _ in [][::0]:
    pass

for _ in []["a"]:
    pass

for _ in []["a":"z"]:
    pass

for _ in ""[::0]:
    pass

for _ in ""["a"]:
    pass

for _ in ""["a":"z"]:
    pass


class Custom:
    def __getitem__(self, indices):
        return "wow"

for _ in Custom()[::0]:
    pass
```
```bash
$ pylint fixtures/slices-pylint.py
************* Module slices-pylint
fixtures/slices-pylint.py:4:9: E1126: Sequence index is not an int, slice, or instance with __index__ (invalid-sequence-index)
fixtures/slices-pylint.py:7:12: E1127: Slice index is not an int, None, or instance with __index__ (invalid-slice-index)
fixtures/slices-pylint.py:7:16: E1127: Slice index is not an int, None, or instance with __index__ (invalid-slice-index)
fixtures/slices-pylint.py:13:9: E1126: Sequence index is not an int, slice, or instance with __index__ (invalid-sequence-index)
```

astroid/nodes/node_classes.py
cdce8p: > I think I originally considered this but stopped because it's not actually an error to _create_ a `slice` object with a step of 0, just an error to use it to index the built-in sequences:

Good point!

> Pylint has tests for these cases (such as the string indices), but at least one of them is incorrect and will need to be changed: https://github.com/PyCQA/pylint/blob/main/tests/functional/i/invalid/invalid_slice_index.py#L25-L28
> 
> ```python
> # Valid indices
> def function4():
>     """integers used as indices"""
>     return TESTLIST[0:0:0] # no error
> ```
> 
> I think the appropriate fix is probably:
> 
> 1. Does the LHS have a custom `__getitem__` method? If so, we're probably in the uninferable realm
> 2. Try to defer the computed indexes to `slice.indices` since this is the canonical source of the raised `Type` and `ValueError`s:

Pylint does most of that already. It only emits `invalid-slice-index` for a set of known objects which expect an int-like type. We would just need to special-case the `0` step error. https://github.com/PyCQA/pylint/blob/v2.15.5/pylint/checkers/typecheck.py#L1769-L1778

> I'm leaning towards adding the `except ValueError` case in `Const.getitem` similarly what was added here in `_container_getitem`.

Sounds good.

tests/unittest_inference.py
```diff
@@ -5316,6 +5316,17 @@ def test_slice_inference_in_for_loops_not_working() -> None:
...
+
+
+def test_slice_zero_step_on_str_does_not_raise_ValueError() -> None:
+    # TODO: move these to SliceTest?
```
cdce8p, line 5325:
```suggestion
```

ChangeLog
```diff
@@ -25,6 +25,10 @@ Release date: TBA
 
   Refs PyCQA/pylint#3941
 
+* Catch ``ValueError`` when indexing some builtin sequences during inference
```
cdce8p, line 28:
```suggestion
* Catch ``ValueError`` when indexing some builtin containers and sequences during inference.
```

cdce8p (commented):
> It occurs to me that these should probably return something useful for the Pylint `invalid-sequence-index` checks. Returning `Uninferable` makes these lines ignored instead of a useful pylint warning.

The `Uninferable` isn't the issue here. The `invalid-sequence-index` check doesn't call `node.getitem`. Fixing the false-negatives should be possible in pylint alone from what I can tell.

--
Left some last comments. After that I think we can merge this one.

cdce8p (approved):
Thanks, @nelfin 👍🏻

PR #1852, opened 2022-10-24
Replace "St" by "Street" in LICENSE

mtelka:
... to match the license text at https://www.gnu.org/licenses/old-licenses/lgpl-2.1-standalone.html

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,7 +3,7 @@
                        Version 2.1, February 1999
 
  Copyright (C) 1991, 1999 Free Software Foundation, Inc.
-	51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
Thank you for creating the merge request but this is already correctly detected by github (https://github.com/PyCQA/astroid/blob/main/LICENSE), and other services checking licences. Any change in this file will make a lot of person check it and additional work for a lot of person including me. So I'm not going to change it unless there's a really good reason.

mtelka:
The main reason for this PR was to make sure the astroid license could be automatically matched to a SPDX license (either [LGPL-2.1-only](https://spdx.org/licenses/LGPL-2.1-only.html) or [LGPL-2.1-or-later](https://spdx.org/licenses/LGPL-2.1-or-later.html)) using their [Matching Guidelines](https://spdx.dev/license-list/matching-guidelines/).  According to the Guidelines the FSF address should be taken literally.  It is neither omitable text nor replaceable text.  Please see https://spdx.org/licenses/ for more details.

Thank you.

PR #1857, opened 2022-10-29
Use Python 3.9 for virtualenv-15-windows-test job

jacobtylerwalls:
## Steps

## Description

virtualenv 15.1.0 cannot be installed on Python 3.10+, so the release test that uses virtualenv 15.1.0 broke when the workflow was changed to use Python 3.10 in #1704.

[Failing run](https://github.com/PyCQA/astroid/actions/runs/3312658937) that I manually triggered earlier this week.

[Example passing run](https://github.com/jacobtylerwalls/astroid/actions/runs/3351386322/jobs/5552801026) on my fork with these changes.

When we eventually drop Python 3.9 as a runtime for pylint, we can remove this test.

## Type of Changes

### DIFF
--- a/.github/workflows/release-tests.yml
+++ b/.github/workflows/release-tests.yml
@@ -21,7 +21,8 @@ jobs:
         id: python
         uses: actions/setup-python@v4.3.0
         with:
-          python-version: ${{ env.DEFAULT_PYTHON }}
+          # virtualenv 15.1.0 cannot be installed on Python 3.10+
+          python-version: 3.9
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👍 Just wondering for how long it's valuable to manually test this old version of virtualenv. Probably until 3.9 is not maintained anymore seeing we don't launch it very often. (I wanted to check the prevalence of virtualenv 20.16 vs 15.x but pypistat do not have this kind of info and I don't remember where I've seen it before.)

PR #1860, opened 2022-11-02
CI improvements

cdce8p:
## Description
Similar changes to https://github.com/PyCQA/pylint/pull/7651

* Remove `restore-keys`. Especially on Windows there are issues with reusing old cache entries and trying to install newer versions. Those are just skipped if the old one still satisfies the requirement. It's easy / fast enough to create a whole new environment if something changes.
* Add `check-latest: true` for `setup-python` action. This will prevent cache restore issue when runners use different Python patch versions. https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#check-latest-version
* Add `KEY_PREFIX` env variable to more easily identify the key prefix in each workflow.
* Replace `setup.cfg` with `pyproject.toml` for file hash

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,7 @@ jobs:
         uses: actions/setup-python@v4.3.0
         with:
           python-version: ${{ env.DEFAULT_PYTHON }}
+          check-latest: true
       - name: Generate partial Python venv restore key
         id: generate-python-key
         run: >-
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
I'm going to trust you on this one, I'm not an expert. The new hash key does make much more sense than manually upgrading the cache indeed. I wonder why I did not think of this when upgrading astroid in pylint,

cdce8p:
> The new hash key does make much more sense than manually upgrading the cache indeed. I wonder why I did not think of this when upgrading astroid in pylint,

You might want to take a look at https://github.com/PyCQA/pylint/pull/7651 then. It does simplify update astroid in pylint. The primer test failure there is expected.

PR #1864, opened 2022-11-06
Fix `ClassDef.igetattr` on multiple attributes (`not-an-iterable` false-positive)

### LINKED ISSUE #1015: false not-an-iterable for class method with @overload (2.5.7 regression)

Rogdham:
## Steps

## Description

This is an attempt at fixing #1015; however I really don't know `astroid`'s code base so this is more like a shot in the dark, which needs a careful review!

In a few words:

```py
class A:
    def foo(self): ...
    def foo(self):
        yield

for _ in A().foo():
    pass
```

Only the first implementation returning `...` is considered, resulting in `pylint` complaining:

```
E1133: Non-iterable value A().foo() is used in an iterating context (not-an-iterable)
```

This is especially annoying when we define `@overload` variants for typing purposes.

## Type of Changes

## Related Issue

Closes #1015

Ping @nelfin @belm0 who are probably interested in this.

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2649,12 +2649,13 @@ class ClassDef(
             # to the attribute happening *after* the attribute's definition (e.g. AugAssigns on lists)
             if len(attributes) > 1:
                 first_attr, attributes = attributes[0], attributes[1:]
-                first_scope = first_attr.scope()
-                attributes = [first_attr] + [
-                    attr
-                    for attr in attributes
-                    if attr.parent and attr.parent.scope() == first_scope
-                ]
+                if first_attr.parent:
+                    first_scope = first_attr.parent.scope()
+                    attributes = [first_attr] + [
+                        attr
+                        for attr in attributes
+                        if attr.parent and attr.parent.scope() == first_scope
+                    ]
 
             for inferred in bases._infer_stmts(attributes, context, frame=self):
                 # yield Uninferable object instead of descriptors when necessary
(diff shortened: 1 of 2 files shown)


belm0:
thank you for working on it

I think it will need unit test coverage.  It's good to write the test and verify that it fails before applying the fix.

Rogdham:
Thank you for the feedback, I have added a test and confirmed that it failed before the fix.

jacobtylerwalls:
Thanks @Rogdham, did you give any thought to the [comment](https://github.com/PyCQA/astroid/pull/1173#discussion_r709651765) on the previous attempt?

Rogdham:
Dammit I completely missed #1173 for some reason, this would have same me some headache :grimacing: 

Since their fix looks the same to me, should I just close this PR? They seems to know what they are talking about way more than I do!

jacobtylerwalls:
I'd see if the author responds soon, but since it's been over a year, feel free to pick up where they left off later this month.

nelfin:
I never did figure out how to address those comments within the scope of that change and then I spent some time away from contributing to open source for a while. Feel free to cherry-pick those tests if you think they will help. If I remember correctly and understand this change, the comment that `Klass.foo` should only be inferred as `Generator` would necessitate a change here as `len(attrs)` should be 1 in that case, not 2.

jacobtylerwalls:
Superseded in #1173

PR #1867, opened 2022-11-14
Add `Try` node

cdce8p:
## Description
Replaces both #1388 and #1389. I first tried implementing the migration plan laid out in https://github.com/PyCQA/astroid/pull/1388#issuecomment-1126947919. However, the new `Try` node not being part of the "official" tree caused a lot of issues.
* The most difficult part would have been to set the `parent` attributes on all child nodes correctly. Since it needs to point to the immediate parent, it would have been different depending if a child was part of the "old" `TryExcept` or the new `Try` node. That would basically mean that we couldn't reuse the child nodes and instead need to create new ones.
* With that, it's not clear how we can resolve inference to (only) use the new ones while they are effectively still hidden behind the special `try_node` attribute.
* All tree visitors in pylint would still use the old child nodes or might even emit all errors twice if both the old and new ones are visited.

I reached the conclusion that it might be better to do a breaking change now, instead of trying to work around a solution which would likely create new issues and would still lead to a breaking change down the line.

--
This PR replaces the old `TryExcept` and `TryFinally` nodes with the new `Try` node. Python replaced the old nodes a long time ago in 3.3. The new node simplifies things a bit and even helps generate better line numbers for pylint errors in try-except blocks. It will also make it much easier to add support for `TryStar` nodes which would just need to inherit from `Try`. https://github.com/PyCQA/pylint/issues/7703

--
As this PR is a (major) breaking change, I suggest it should be the start for astroid `3.0` once merged. I'm going to create a planning issue for `3.0`. However, even with the breaking change, the work required to update is manageable. I'll also open a pylint PR to update our code base.
* https://github.com/PyCQA/pylint/pull/7767
* https://github.com/PyCQA/astroid/issues/1868

--
### What's left to do
I'll also mention it in the `3.0` issue, but I think we should create a separate docs page just describing the breaking changes and how to update plugins. That can probably be done separately. If someone else would like to take the lead on this, I would appreciate any help. Some notes
* `TryExcept` and `TryFinally` removed
* Update `visit_*` and `leave_*` methods


DanielNoord:
@cdce8p Should this be reviewed? Or should we wait until we are sure we can start merging `3.0` PRs?

cdce8p:
> @cdce8p Should this be reviewed? Or should we wait until we are sure we can start merging `3.0` PRs?

The change itself is done. Might only need to resolve merge conflicts and update the changelog when we start with `3.0`. If you have some spare time, feel free to take a look. Just don't merge it quite yet.

astroid/nodes/node_classes.py
```diff
@@ -3231,40 +3173,62 @@ def __init__(
...
+            self.orelse = orelse
+        if finalbody:
             self.finalbody = finalbody
 
-    def block_range(self, lineno: int) -> tuple[int, int]:
-        """Get a range from the given line number to where this node ends.
-
-        :param lineno: The line number to start the range at.
+    def _infer_name(self, frame, name):
+        return name
```
Pierre-Sassoulas, line 3202: It would be nice to cover this line in tests.

Pierre-Sassoulas (commented):
It's not a work in progress anymore, right ? I think it's going to simplify a lot of checks in pylint 👍

astroid/nodes/node_classes.py
```diff
@@ -3110,97 +3110,33 @@ def get_children(self):
...
-
-    _astroid_fields = ("body", "finalbody")
-    _multi_line_block_fields = ("body", "finalbody")
+    _astroid_fields = ("body", "handlers", "orelse", "finalbody")
+    _multi_line_block_fields = ("body", "handlers", "orelse", "finalbody")
 
     def __init__(
         self,
+        *,
         lineno: int | None = None,
```
DanielNoord, line 3135: I think we should follow the new pattern here and make all of these not have a default.

cdce8p: Done. Although I'm (still) not fully convinced it's actually better 🤷🏻‍♂️

cdce8p:
> It's not a work in progress anymore, right ? I think it's going to simplify a lot of checks in pylint 👍

The PR is basically ready at this point. We could merge it as is.
The real work will be to update pylint after that. I do already have a PR with most of the required changes but we'll definitely need to double check these https://github.com/pylint-dev/pylint/pull/7767.

Once we merge it, I'll create another alpha release just for it.

Pierre-Sassoulas (dismissed):

jacobtylerwalls (approved):
Great! Eager to merge this after we release 3.0.0a7 and can devote an alpha release to just this change.

Pierre-Sassoulas:
Let's do it ! :)

Pierre-Sassoulas (approved):

PR #1869, opened 2022-11-14
Use stable astroid for pylint CI check

cdce8p:
## Description
At the moment, we use `pip install -e .` for the pylint CI check. That only works as long as there aren't any breaking changes.

Instead we should probably just use the latest `released` version which usually is the one the pinned pylint version ships with.

Required for #1867


Pierre-Sassoulas (approved):
Make sense, thank you.

PR #1876, opened 2022-11-19
Add some type annotations

nickdrozd:
## Description

This PR adds some type annotations. A lot of them are straightforward, but some are a little tricky.

## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -206,7 +206,7 @@ class BaseInstance(Proxy):
 
     special_attributes = None
 
-    def display_type(self):
+    def display_type(self) -> str:
         return "Instance of"
 
     def getattr(self, name, context=None, lookupclass=True):
@@ -639,7 +639,7 @@ class AsyncGenerator(Generator):
     def pytype(self) -> Literal["builtins.async_generator"]:
         return "builtins.async_generator"
 
-    def display_type(self):
+    def display_type(self) -> str:
         return "AsyncGenerator"
 
     def __repr__(self):
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -482,7 +482,7 @@ class BoundMethod(UnboundMethod):
             return 0
         return 1
 
-    def is_bound(self):
+    def is_bound(self) -> bool:
         return True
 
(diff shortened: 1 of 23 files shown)


DanielNoord (changes requested):
@nickdrozd Would you mind splitting this PR up? The typing for astroid is so broken that we need to double double double check every change to make, see https://github.com/PyCQA/astroid/pull/1837 which was just the annotation of one attribute.

We need to check every single change quite thoroughly so really small PRs helps with reviewing 😄

nickdrozd:
@DanielNoord Sure, split up along what lines? Also, is Mypy running in CI?

Pierre-Sassoulas:
> Also, is Mypy running in CI?

We aim to. There's still a lot of error right now.  You can modify this line to run mypy in pre-commit: https://github.com/PyCQA/astroid/blob/main/.pre-commit-config.yaml#L91

DanielNoord:
Mypy is not running in CI, which makes reviewing even harder. We need to check every file by hand.

Easiest is to just group similar changes to in a PR. It is a bit tedious, but for example: `def looks_like_typing_typevar_or_newtype(node) -> bool:` and the `-> str` changes are very easy to review which helps with decreasing the review burden of this PR if they can get split out in a separate PR.

nickdrozd:
Okay, I reworked the PR to be easier to review. Different types of changes are grouped in different commits, so I suggest reading through it that way.

Here are the number of Mypy warnings before and after these changes in non-strict and strict modes:

```
Non-Strict
  Before
    277
  After
    279

Strict
  Before
    1742
  After
    1632
```

The number of warnings went up a by 2 in non-strict. That could be because some of the changes are wrong, or it could be existing errors that are exposed by these changes.

Pierre-Sassoulas:
The bool commit seems pretty reasonable, would you mind separating it in a New PR ? Being able to review it once and merge instead of multiple time if we have a comment on another part of the PR will definitely make everything faster. Generally when a PR goes stale it's because it can't be reviewed fast. Maintainers might have dozen of 10mn review time per day (say waiting for the teapot to boil) but one 1h review time per week (you have to concentrate on the review for a full hour and actually allocate time for it).

PR #1878, opened 2022-11-20
Add some str annotations

nickdrozd:
## Type of Changes

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -206,7 +206,7 @@ class BaseInstance(Proxy):
 
     special_attributes = None
 
-    def display_type(self):
+    def display_type(self) -> str:
         return "Instance of"
 
     def getattr(self, name, context=None, lookupclass=True):
@@ -317,12 +317,12 @@ class Instance(BaseInstance):
     def __init__(self, proxied: nodes.ClassDef | None) -> None:
         super().__init__(proxied)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "<Instance of {}.{} at 0x{}>".format(
             self._proxied.root().name, self._proxied.name, id(self)
         )
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"Instance of {self._proxied.root().name}.{self._proxied.name}"
 
     def callable(self):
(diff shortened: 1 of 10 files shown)


astroid/nodes/as_string.py
```diff
@@ -50,15 +51,17 @@ def _docs_dedent(self, doc_node: Const | None) -> str:
             self.indent, doc_node.value.replace("\n", DOC_NEWLINE)
         )
 
-    def _stmt_list(self, stmts, indent=True):
+    def _stmt_list(self, stmts: list, indent: bool = True) -> str:
         """return a list of nodes to string"""
-        stmts = "\n".join(nstr for nstr in [n.accept(self) for n in stmts] if nstr)
-        if indent:
-            return self.indent + stmts.replace("\n", "\n" + self.indent)
+        stmts_str: str = "\n".join(
```
nickdrozd, line 56: Mypy doesn't like variables changed to different types.

Pierre-Sassoulas: Let's use two variable name instead. This is a pattern that is used a lot in old python but it's very error prone, imo it dying with mypy's adoption is a good thing).

Pierre-Sassoulas (approved):
👌

astroid/bases.py
```diff
@@ -620,16 +620,16 @@ def callable(self):
     def pytype(self) -> Literal["builtins.generator"]:
         return "builtins.generator"
 
-    def display_type(self):
+    def display_type(self) -> str:
```
DanielNoord, line 623: We could have used `Literals` here. Not sure if it makes sense to change them now though...

Pierre-Sassoulas: Yeah, we have a lot to do before starting to polish this to use litteral imo :D

DanielNoord: For future PRs I do think we can start using them, if they make sense. Sometimes it's useful to know the specific string a function returns 😄

cdce8p: > For future PRs I do think we can start using them, if they make sense. Sometimes it's useful to know the specific string a function returns 😄

I would disagree. We should only use `Literal` if there is a clear value to it. In most cases however it just adds unnecessary complexity. Example, what if the return string changes? You'll also have to update the `Literal` typing as well.

`Literal` is most useful for `@overloads` IMO.

DanielNoord: I think for many of the methods that were being typed here the method could have also been a attribute, in which case I think it makes sense to do so as you would get similar information about the return value if it was an attribute.

But you are right, they shouldn't be used everywhere.

PR #1883, opened 2022-11-21
Inference context annotations

nickdrozd:
## Description

Add type annotations pertaining to inference contexts.

1. The file `context.py` is mostly typed.
2. Most if not all inference context args are annotated.
3. Two minor code errors exposed by the type checker are fixed.

## Type of Changes

## Related Issue

This fixes a code error. I don't know if it was associated with any known bug.

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -697,8 +697,6 @@ def infer_isinstance(callnode, context: InferenceContext | None = None):
     """Infer isinstance calls
 
     :param nodes.Call callnode: an isinstance call
-    :param InferenceContext context: context for call
-        (currently unused but is a common interface for inference)
     :rtype nodes.Const: Boolean Const value of isinstance call
 
     :raises UseInferenceDefault: If the node cannot be inferred
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -171,7 +171,7 @@ def register_builtin_transform(transform, builtin_name):
     an optional context.
     """
 
-    def _transform_wrapper(node, context=None):
+    def _transform_wrapper(node, context: InferenceContext | None = None):
         result = transform(node, context=context)
         if result:
             if not result.parent:
@@ -336,7 +336,7 @@ def _get_elts(arg, context):
     return items
 
 
-def infer_dict(node, context=None):
+def infer_dict(node, context: InferenceContext | None = None):
     """Try to infer a dict call to a Dict node.
 
     The function treats the following cases:
@@ -379,7 +379,7 @@ def infer_dict(node, context=None):
     return value
 
 
-def infer_super(node, context=None):
+def infer_super(node, context: InferenceContext | None = None):
     """Understand super calls.
 
     There are some restrictions for what can be understood:
@@ -458,7 +458,7 @@ def _infer_getattr_args(node, context):
     return obj, attr.value
 
 
-def infer_getattr(node, context=None):
+def infer_getattr(node, context: InferenceContext | None = None):
     """Understand getattr calls
 
     If one of the arguments is an Uninferable object, then the
@@ -486,7 +486,7 @@ def infer_getattr(node, context=None):
     raise UseInferenceDefault
 
 
-def infer_hasattr(node, context=None):
+def infer_hasattr(node, context: InferenceContext | None = None):
     """Understand hasattr calls
 
     This always guarantees three possible outcomes for calling
@@ -513,7 +513,7 @@ def infer_hasattr(node, context=None):
     return nodes.Const(True)
 
 
-def infer_callable(node, context=None):
+def infer_callable(node, context: InferenceContext | None = None):
     """Understand callable calls
 
     This follows Python's semantics, where an object
@@ -571,7 +571,7 @@ def infer_property(
     return prop_func
 
 
-def infer_bool(node, context=None):
+def infer_bool(node, context: InferenceContext | None = None):
     """Understand bool calls."""
     if len(node.args) > 1:
         # Invalid bool call.
@@ -594,7 +594,7 @@ def infer_bool(node, context=None):
     return nodes.Const(bool_value)
 
 
-def infer_type(node, context=None):
+def infer_type(node, context: InferenceContext | None = None):
     """Understand the one-argument form of *type*."""
     if len(node.args) != 1:
         raise UseInferenceDefault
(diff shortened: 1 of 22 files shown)


astroid/context.py
```diff
@@ -174,7 +178,7 @@ def copy_context(context: InferenceContext | None) -> InferenceContext:
     return InferenceContext()
 
 
-def bind_context_to_node(context, node):
+def bind_context_to_node(context: InferenceContext | None, node) -> InferenceContext:
```
DanielNoord, line 181: Could you remove the docstring typing here?

astroid/context.py
```diff
@@ -104,7 +108,7 @@ def inferred(self) -> _InferenceCache:
         """
         return _INFERENCE_CACHE
 
-    def push(self, node):
```
DanielNoord, line 107: Can you remove `rtype` here as well?

astroid/context.py
```diff
@@ -72,7 +76,7 @@ def __init__(self, path=None, nodes_inferred=None):
 
         e.g. the bound node of object.__new__(cls) is the object node
         """
-        self.extra_context = {}
+        self.extra_context: dict[NodeNG, InferenceContext] = {}
```
DanielNoord, line 79: Have you checked that this can't be `Instance` or its parent `Proxy`? In general, docstring typing is incorrect 😓

astroid/brain/brain_builtin_inference.py
```diff
@@ -693,7 +693,7 @@ def infer_issubclass(callnode, context=None):
     return nodes.Const(issubclass_bool)
 
 
-def infer_isinstance(callnode, context=None):
+def infer_isinstance(callnode, context: InferenceContext | None = None):
     """Infer isinstance calls
 
     :param nodes.Call callnode: an isinstance call
```
Pierre-Sassoulas, line 699: The docstring typing is duplicated and can be removed imo. (There are other instances of this)

astroid/context.py
```diff
@@ -42,7 +42,11 @@ class InferenceContext:
 
     max_inferred = 100
 
-    def __init__(self, path=None, nodes_inferred=None):
+    def __init__(
+        self,
+        path: set[tuple[NodeNG, str | None]] | None = None,
+        nodes_inferred: list[int] | None = None,
```
DanielNoord, line 48: Why is this a list 😅 Seems so illogical, oh well..

astroid/context.py
```diff
@@ -42,7 +42,11 @@ class InferenceContext:
...
+    ):
```
DanielNoord, line 49:
```suggestion
    ) -> None:
```

astroid/context.py
```diff
@@ -42,7 +42,11 @@ class InferenceContext:
 
     max_inferred = 100
 
-    def __init__(self, path=None, nodes_inferred=None):
+    def __init__(
+        self,
+        path: set[tuple[NodeNG, str | None]] | None = None,
```
DanielNoord, line 47: Where do you base this on?

DanielNoord (changes requested):
Only spotted this somewhere along the review: please review any docstring typing that is now explicit. It makes sure we don't have to keep two separate things up to date!

Pierre-Sassoulas (changes requested):
Good job finding those two bugs !

astroid/context.py
nickdrozd: Yes, I will cut this one.

astroid/context.py
nickdrozd: Isn't `__init__` exempt from return typing?

astroid/context.py
nickdrozd:
```
        :type: set(tuple(NodeNG, optional(str)))

        Path of visited nodes and their lookupname

        Currently this key is ``(node, context.lookupname)``
```

astroid/context.py
nickdrozd: Changed it to `Any` instead.

astroid/context.py
DanielNoord: It's very likely that this is indeed `SuccesfulInferenceResult` instead of `NodeNG`. Let's revert this and keep it without the typing for now. The PR is really useful anyway!

astroid/context.py
DanielNoord: Ah of course. Just force of habit for me, feel free to ignore and resolve!

astroid/context.py
```diff
@@ -64,24 +67,20 @@ def __init__(self, path=None, nodes_inferred=None):
         """
         self.callcontext: CallContext | None = None
         """The call arguments and keywords for the given context."""
-        self.boundnode = None
+        self.boundnode: NodeNG | None = None
```
DanielNoord, line 70: Again, are we sure that this is not more than `NodeNG`? I'd prefer to do such changes in a separate PR.

astroid/context.py
```diff
@@ -64,24 +67,20 @@ def __init__(self, path=None, nodes_inferred=None):
...
         """
-        :type: optional[NodeNG]
-
         The bound node of the given context
 
         e.g. the bound node of object.__new__(cls) is the object node
         """
-        self.extra_context = {}
+        self.extra_context: dict[Any, InferenceContext] = {}
```
DanielNoord, line 76: Let's not use `Any`. Since `mypy` isn't running anyway the typing also doesn't give us much so it might be better to just keep the docstring for now until we can be 100% sure we get this correctly.

DanielNoord (changes requested):

astroid/context.py
nickdrozd: I'd like to keep this one because for some reason Mypy specifically calls it out:

```
astroid/context.py:78: error: Need type annotation for "extra_context" (hint: "extra_context: Dict[<type>, <type>] = ...")  [var-annotated]
```

So when trying to do more typing, it adds noise.

astroid/context.py
nickdrozd: I changed `NodeNG` to `Any`. Even with that hole, this a pretty specific type to verify.

I have a feeling that this pattern is going to come up all over the place, where a docstring says it's `NodeNG`, but in reality it's `NodeNG | Proxy | Base | Etc`. Being able to verify the compound types without having to untangle all the node-like types would be easier.

DanielNoord: I don't really understand your last comment.

I agree it would be good to untangle where the docstring annotation is incorrect but I prefer to do so in separate PRs as checking that is so involved. If we save that change for a small follow-up PR we can focus on getting that right while we can now already merge the `context: InferenceContext | None` stuff.

nickdrozd:
@DanielNoord I took out the partial annotations. I'll put them into a separate PR.

DanielNoord (approved):

DanielNoord:
THanks a lot @nickdrozd :heart:

Pierre-Sassoulas (approved):
Nice 🎉 !

PR #1885, opened 2022-11-23
Add some type annotations to raw_building

nickdrozd:
## Description

Add some type annotations to raw_building. This isn't all of it, but it's a lot of it.

## Type of Changes


astroid/raw_building.py
```diff
@@ -177,6 +177,7 @@ def register_arguments(func: nodes.FunctionDef, args: list | None = None) -> Non
             func.set_local(func.args.vararg, func.args)
         if func.args.kwarg:
             func.set_local(func.args.kwarg, func.args)
+    assert args is not None
```
Pierre-Sassoulas, line 180: This is a little stressful :D I'd rather have a type ignore

Pierre-Sassoulas (changes requested):

astroid/raw_building.py
nickdrozd: How about adding this instead?

```python
    if args is None:
        return
```

nickdrozd: Or even better:

```python
    if args is None:
        if func.args.vararg:
            func.set_local(func.args.vararg, func.args)
        if func.args.kwarg:
            func.set_local(func.args.kwarg, func.args)
        args = func.args.args
        if args is None:
            return
```

astroid/raw_building.py
```diff
@@ -157,10 +157,10 @@ def build_function(
         argsnode.defaults[-1].parent = argsnode
     if args:
         register_arguments(func)
-    return func
+    return func  # type: ignore[no-any-return]
 
 
-def build_from_import(fromname, names):
+def build_from_import(fromname: str | None, names: list[str]) -> nodes.ImportFrom:
```
DanielNoord, line 163: This is dead code and only used in tests. Shall we deprecate it? Perhaps in a follow up PR?

Either way this should be:
`fromname: str`. It makes no sense to have an `ImportFrom` without a from I think?

astroid/raw_building.py
```diff
@@ -172,11 +172,13 @@ def register_arguments(func: nodes.FunctionDef, args: list | None = None) -> Non
     (i.e. def func(a, (b, c, d)): ...)
     """
     if args is None:
-        args = func.args.args
         if func.args.vararg:
             func.set_local(func.args.vararg, func.args)
         if func.args.kwarg:
             func.set_local(func.args.kwarg, func.args)
+        args = func.args.args
+        if args is None:
```
DanielNoord, line 180: See:
https://github.dev/nickdrozd/astroid/blob/0a9a0a17675124c2f21ef545704066785da12eb7/astroid/nodes/node_classes.py#L635

`args is None` is valid. Although we could exit early here I think it makes more sense to do a `if args` or `args = []`. Right now it seems as if we don't consider it valid if `args is None` and want to exit early while instead it is an exit early to avoid a `for ... in None`.

astroid/raw_building.py
```diff
@@ -324,7 +326,7 @@ def _build_from_function(
         object_build_function(node, member, name)
 
 
-def _safe_has_attribute(obj, member) -> bool:
+def _safe_has_attribute(obj, member: str) -> bool:
     try:
         return hasattr(obj, member)
     except Exception:  # pylint: disable=broad-except
```
DanielNoord, line 332: This seems like we could easily avoid the broad exception here.. But I'm overscoping :sweat_smile:

astroid/raw_building.py
```diff
@@ -559,6 +561,7 @@ def _astroid_bootstrapping() -> None:
         types.TracebackType,
     )
     for _type in builtin_types:
+        assert hasattr(_type, "__name__")
```
DanielNoord, line 564: What does `mypy` say here? `pyright` seems to understand this in online Github.

DanielNoord (changes requested):
Thanks again for doing this :smile:

astroid/nodes/node_classes.py
```diff
@@ -2550,6 +2550,8 @@ class EmptyNode(_base_nodes.NoChildrenNode):
 
     object = None
 
+    has_underlying_object: Callable[[EmptyNode], bool]
```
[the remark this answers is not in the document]
nickdrozd: I have no idea why it's defined there, or indeed why it's defined at all. Can it be deleted?

astroid/raw_building.py
nickdrozd: I'd be happy to delete it!

`ImportFrom.fromname` is currently specified as `str | None`. That could be changed, but of course that would be out of scope.

astroid/raw_building.py
nickdrozd: Is there any reason for this function to exist? Are there any objects for which `hasattr` could raise an error? My guess is that this is a holdover from Python 2 days.

astroid/raw_building.py
nickdrozd: `astroid/raw_building.py:565: error: "object" has no attribute "__name__"; maybe "__ne__" or "__new__"?  [attr-defined]`

nickdrozd: I have no idea what the type of `builtin_types` should be.

astroid/nodes/node_classes.py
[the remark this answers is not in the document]
DanielNoord: It's necessary for the assignment to the class in `raw_building`. What would actually be better is if we moved `_marker` into `const.py` and just define the method on this class itself and import `_marker` wherever it is needed.

astroid/raw_building.py
DanielNoord: @Pierre-Sassoulas Do we consider this public?

I think it is better to add a deprecation and then remove in `3.0`. We are working towards that anyway.

DanielNoord: On the topic of `ImportFrom.fromnam`: we're probably going to change that in `3.0` which would mean we would need to revisit this function. Let's already fix this here and use `str` only.

astroid/raw_building.py
DanielNoord: Yeah I don't think so. Let's remove, it is private anyway.

astroid/raw_building.py
DanielNoord:
```python
import types

builtin_types: tuple[type, ...] = (
    types.GetSetDescriptorType,
    types.GeneratorType,
    types.MemberDescriptorType,
    type(None),
    type(NotImplemented),
    types.FunctionType,
    types.MethodType,
    types.BuiltinFunctionType,
    types.ModuleType,
    types.TracebackType,
)

for i in builtin_types:
    print(i.__name__)

```

This works for `mypy`, so `tuple[type. ...]` I guess

nickdrozd:
```
astroid/raw_building.py:534: error: Incompatible types in assignment (expression has type "Tuple[Type[GetSetDescriptorType], Type[GeneratorType[Any, Any, Any]], Type[MemberDescriptorType], object, Type[_NotImplementedType], Type[FunctionType], Type[MethodType], Type[BuiltinFunctionType], Type[Module], Type[TracebackType]]", variable has type "Tuple[type, ...]")  [assignment]
```

:scream:

astroid/raw_building.py
nickdrozd: Well, initially `args` can be missing. If it is, we check the args of the function instead. But if those are missing too, then there isn't anything more to be done, so we can exit. At least that's my reading of the intention behind the logic. We could set `args = []` so that there's always something to iterate over, but that seems like artificially giving it something to do when there isn't anything to do.

astroid/raw_building.py
DanielNoord: Hmm, I'll have a look at this this weekend. Perhaps my `mypy` was out of date?

astroid/raw_building.py
DanielNoord: Okay, can we then add a comment explaining this? This is just being cautious for the moment somebody adds extra logic at the end of the function and then some nodes would never reach that.

astroid/const.py
```diff
@@ -33,3 +33,6 @@ class Context(enum.Enum):
 
 ASTROID_INSTALL_DIRECTORY = Path(__file__).parent
 BRAIN_MODULES_DIRECTORY = ASTROID_INSTALL_DIRECTORY / "brain"
+
+
+EMPTY_OBJECT_MARKER = object()
```
DanielNoord, line 38: Let's make this private like `marker` was: `_EMPTY_OBJECT_MARKER`.

DanielNoord (changes requested):
Last changes!

DanielNoord (approved):

Pierre-Sassoulas (approved):
Great work @nickdrozd, refactoring the method definition outside of the class definition is particularly important to make the code understandable and not easy to pull off 👍

PR #1887, opened 2022-11-27
Add some inference annotations

nickdrozd:
## Type of Changes

### DIFF
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -10,7 +10,7 @@ import functools
 import inspect
 import sys
 import warnings
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from typing import TypeVar
 
 import wrapt
@@ -89,7 +89,7 @@ class cachedproperty:
         return val
 
 
-def path_wrapper(func):
+def path_wrapper(func: Callable) -> Callable:
     """return the given infer function wrapped to handle the path
 
     Used to stop inference if the node has already been looked
@@ -97,7 +97,7 @@ def path_wrapper(func):
     """
 
     @functools.wraps(func)
-    def wrapped(node, context: InferenceContext | None = None, _func=func, **kwargs):
+    def wrapped(node, context: InferenceContext | None = None, _func=func, **kwargs) -> Iterator:
         """wrapper function handling context"""
         if context is None:
             context = InferenceContext()
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -97,7 +97,9 @@ def path_wrapper(func: Callable) -> Callable:
     """
 
     @functools.wraps(func)
-    def wrapped(node, context: InferenceContext | None = None, _func=func, **kwargs) -> Iterator:
+    def wrapped(
+        node, context: InferenceContext | None = None, _func=func, **kwargs
+    ) -> Iterator:
         """wrapper function handling context"""
         if context is None:
             context = InferenceContext()
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -10,7 +10,7 @@ import functools
 import inspect
 import sys
 import warnings
-from collections.abc import Callable, Iterator
+from collections.abc import Callable, Generator
 from typing import TypeVar
 
 import wrapt
@@ -99,7 +99,7 @@ def path_wrapper(func: Callable) -> Callable:
     @functools.wraps(func)
     def wrapped(
         node, context: InferenceContext | None = None, _func=func, **kwargs
-    ) -> Iterator:
+    ) -> Generator:
         """wrapper function handling context"""
         if context is None:
             context = InferenceContext()
--- a/astroid/decorators.py
+++ b/astroid/decorators.py
@@ -89,7 +89,7 @@ class cachedproperty:
         return val
 
 
-def path_wrapper(func: Callable) -> Callable:
+def path_wrapper(func):
     """return the given infer function wrapped to handle the path
 
     Used to stop inference if the node has already been looked
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -219,7 +219,7 @@ def is_supertype(type1, type2) -> bool:
     return _type_check(type1, type2)
 
 
-def class_instance_as_index(node) -> nodes.NodeNG | None:
+def class_instance_as_index(node) -> nodes.Const | None:
     """Get the value as an index for the given instance.
 
     If an instance provides an __index__ method, then it can
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -219,7 +219,7 @@ def is_supertype(type1, type2) -> bool:
     return _type_check(type1, type2)
 
 
-def class_instance_as_index(node):
+def class_instance_as_index(node) -> nodes.NodeNG | None:
     """Get the value as an index for the given instance.
 
     If an instance provides an __index__ method, then it can
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -219,7 +219,7 @@ def is_supertype(type1, type2) -> bool:
     return _type_check(type1, type2)
 
 
-def class_instance_as_index(node) -> nodes.Const | None:
+def class_instance_as_index(node: SuccessfulInferenceResult) -> nodes.Const | None:
     """Get the value as an index for the given instance.
 
     If an instance provides an __index__ method, then it can
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -80,8 +80,8 @@ def infer_end(
 
 # We add ignores to all assignments to methods
 # See https://github.com/python/mypy/issues/2427
-nodes.Module._infer = infer_end  # type: ignore[assignment]
-nodes.ClassDef._infer = infer_end  # type: ignore[assignment]
+nodes.Module._infer = infer_end
+nodes.ClassDef._infer = infer_end
 nodes.Lambda._infer = infer_end  # type: ignore[assignment]
 nodes.Const._infer = infer_end  # type: ignore[assignment]
 nodes.Slice._infer = infer_end  # type: ignore[assignment]
@@ -193,7 +193,7 @@ def _infer_map(
             if any(not elem for elem in (key, safe_value)):
                 raise InferenceError(node=node, context=context)
             # safe_value is SuccessfulInferenceResult as bool(Uninferable) == False
-            values = _update_with_replacement(values, {key: safe_value})  # type: ignore[dict-item]
+            values = _update_with_replacement(values, {key: safe_value})
     return values
 
 
@@ -216,7 +216,7 @@ def _higher_function_scope(node: nodes.NodeNG) -> nodes.FunctionDef | None:
     while current.parent and not isinstance(current.parent, nodes.FunctionDef):
         current = current.parent
     if current and current.parent:
-        return current.parent  # type: ignore[return-value]
+        return current.parent
     return None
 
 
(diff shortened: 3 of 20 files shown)


astroid/decorators.py
```diff
@@ -89,15 +89,17 @@ def __get__(self, inst, objtype=None):
         return val
 
 
-def path_wrapper(func):
+def path_wrapper(func: Callable) -> Callable:
```
DanielNoord, line 92: This should probably be a `Typevar` or even a `ParamSpec`.

astroid/helpers.py
```diff
@@ -219,7 +219,7 @@ def is_supertype(type1, type2) -> bool:
     return _type_check(type1, type2)
 
 
-def class_instance_as_index(node):
+def class_instance_as_index(node) -> nodes.NodeNG | None:
```
DanielNoord, line 222:
```suggestion
def class_instance_as_index(node) -> nodes.Const | None:
```

?

astroid/inference.py
```diff
@@ -80,14 +80,16 @@ def infer_end(
...
+nodes.ClassDef._infer = infer_end
 nodes.Lambda._infer = infer_end  # type: ignore[assignment]
 nodes.Const._infer = infer_end  # type: ignore[assignment]
 nodes.Slice._infer = infer_end  # type: ignore[assignment]
 
 
-def _infer_sequence_helper(node, context: InferenceContext | None = None):
+def _infer_sequence_helper(
+    node: _BaseContainerT, context: InferenceContext | None = None
+) -> list[nodes.NodeNG]:
```
DanielNoord, line 92: Is `node.elts` correctly typed?

astroid/inference.py
```diff
@@ -193,7 +195,7 @@ def _infer_map(
             if any(not elem for elem in (key, safe_value)):
                 raise InferenceError(node=node, context=context)
             # safe_value is SuccessfulInferenceResult as bool(Uninferable) == False
-            values = _update_with_replacement(values, {key: safe_value})  # type: ignore[dict-item]
+            values = _update_with_replacement(values, {key: safe_value})
```
DanielNoord, line 198: Have you tested this with ``mypy --strict``? I added these ignores based on strict mode so they might not be unused in that case.

astroid/inference.py
```diff
@@ -772,7 +776,7 @@ def _get_binop_flow(
     right_type: InferenceResult | None,
     context: InferenceContext,
     reverse_context: InferenceContext,
-):
+) -> list:
```
DanielNoord, line 779: I have PRs for all binary inference code paths after https://github.com/PyCQA/astroid/pull/1787. Let's fix those after that PR gets merged.

astroid/inference.py
```diff
@@ -1139,10 +1143,12 @@ def infer_empty_node(
 nodes.EmptyNode._infer = infer_empty_node  # type: ignore[assignment]
 
 
-def _populate_context_lookup(call, context):
+def _populate_context_lookup(
+    call: nodes.Call, context: InferenceContext | None
+) -> dict[nodes.NodeNG, InferenceContext]:
```
DanielNoord, line 1148: This feels like it should be `InferenceResult` or something.

astroid/interpreter/dunder_lookup.py
```diff
@@ -18,7 +18,7 @@
 from astroid.exceptions import AttributeInferenceError
 
 
-def _lookup_in_mro(node, name):
+def _lookup_in_mro(node, name) -> list:
```
DanielNoord, line 21: Can we already try narrowing this down?

astroid/nodes/_base_nodes.py
```diff
@@ -150,7 +150,7 @@ def do_import_module(self, modname: str | None = None) -> nodes.Module:
             use_cache=use_cache,
         )
 
-    def real_name(self, asname):
+    def real_name(self, asname) -> str:
```
DanielNoord, line 153:
```suggestion
    def real_name(self, asname: str) -> str:
```
?

astroid/protocols.py
```diff
@@ -349,7 +349,7 @@ def assend_assigned_stmts(
 nodes.AssignAttr.assigned_stmts = assend_assigned_stmts
 
 
-def _arguments_infer_argname(self, name, context):
+def _arguments_infer_argname(self, name: sttr, context: InferenceContext) -> Iterator:
```
DanielNoord, line 352: There are return statements here so should we use `Generator`?

DanielNoord (changes requested):
Initial pass!

astroid/inference.py
nickdrozd: `BaseContainer` has `self.elts: list[NodeNG] = []`

astroid/inference.py
nickdrozd: Yes, with version `mypy 0.991 (compiled: yes)` and `Python 3.8.10`. It warned that only some of the ignores are unused. I guess some have the nodes have the function types declared and some don't?

astroid/interpreter/dunder_lookup.py
nickdrozd: I'm not sure what it's supposed to be.

astroid/decorators.py
nickdrozd: Could you explain more? I haven't used those.

astroid/inference.py
nickdrozd: Which part?

astroid/decorators.py
DanielNoord: I think the PEP introducing `ParamSpec` is quite clear:
https://peps.python.org/pep-0612/

It explains its uses and those can be applied here.

astroid/inference.py
DanielNoord: Hmm that sounds weird. I need to investigate this but I'm fairly certain that it could be `Instance` as well.

astroid/inference.py
DanielNoord: The `nodes.NodeNG`

astroid/decorators.py
```diff
@@ -89,15 +89,17 @@ def __get__(self, inst, objtype=None):
...
 
     Used to stop inference if the node has already been looked
     at for a given `InferenceContext` to prevent infinite recursion
     """
 
     @functools.wraps(func)
-    def wrapped(node, context: InferenceContext | None = None, _func=func, **kwargs):
+    def wrapped(
+        node, context: InferenceContext | None = None, _func=func, **kwargs
+    ) -> Iterator:
```
DanielNoord, line 102: Let's use `Generator`, since this function has `return` statements.

astroid/helpers.py
```diff
@@ -219,7 +219,7 @@ def is_supertype(type1, type2) -> bool:
     return _type_check(type1, type2)
 
 
-def class_instance_as_index(node):
+def class_instance_as_index(node) -> nodes.Const | None:
```
DanielNoord, line 222:
```suggestion
def class_instance_as_index(node: bases.Instance) -> nodes.Const | None:
```

I did a quick check and this should work 😄

astroid/inference.py
DanielNoord: I checked and this is likely incorrect. In the construction of `Lists` in `DictModel.attr_values` we use the `keys` of a `Dict.items`. Those are `SuccessfulInferenceResults`.
So `elts` is indeed incorrectly typed.

astroid/inference.py
```diff
@@ -216,7 +218,7 @@ def _higher_function_scope(node: nodes.NodeNG) -> nodes.FunctionDef | None:
     while current.parent and not isinstance(current.parent, nodes.FunctionDef):
         current = current.parent
     if current and current.parent:
-        return current.parent  # type: ignore[return-value]
+        return current.parent
```
DanielNoord, line 221: This causes an additional message.

astroid/inference.py
```diff
@@ -743,7 +745,9 @@ def _bin_op(
     )
 
 
-def _get_binop_contexts(context, left, right):
+def _get_binop_contexts(
```
DanielNoord, line 748: Let's do this in my `BinOp` PRs. I'll prepare those somewhere this week.

astroid/inference.py
```diff
@@ -759,7 +763,7 @@ def _get_binop_contexts(context, left, right):
         yield new_context
 
 
-def _same_type(type1, type2):
+def _same_type(type1, type2) -> bool:
```
DanielNoord, line 766: This is interesting as adding annotations to `type1` will cause a type error... Hmm, we should probably investigate this.

astroid/protocols.py
```diff
@@ -349,7 +349,9 @@ def assend_assigned_stmts(
 nodes.AssignAttr.assigned_stmts = assend_assigned_stmts
 
 
-def _arguments_infer_argname(self, name, context):
+def _arguments_infer_argname(
+    self, name: str, context: InferenceContext
```
DanielNoord, line 353: `name` can be `None`, see code down below.

pyproject.toml
```diff
@@ -78,6 +78,8 @@ no_implicit_optional = true
 scripts_are_modules = true
 show_error_codes = true
 warn_redundant_casts = true
+allow_untyped_decorators = true
```
DanielNoord, line 81: Let's do this in another PR, if we really want to do this. But let's keep the discussion in that PR.

DanielNoord (changes requested):
Did another pass! Thanks for taking the time to do this, I know how difficult it is to get these right and PRs like this are really helpful to at least get the discussion and project going 😄

astroid/inference.py
nickdrozd: This function is used twice. In both cases, the inputs can be `InferenceResult | None`, but this function assumes that it will be `InferenceResult`.

nickdrozd: `Proxy` is involved too. Messy!

astroid/decorators.py
nickdrozd: That sounds out of scope for this PR. I can undo this change if you want.

astroid/inference.py
nickdrozd: Should this change be undone?

astroid/decorators.py
DanielNoord: Yeah, let's revert then

astroid/inference.py
DanielNoord: Can't you change it to `SuccessfulInferenceResult`?

astroid/inference.py
DanielNoord: 😅

astroid/inference.py
```diff
@@ -1139,10 +1141,12 @@ def infer_empty_node(
 nodes.EmptyNode._infer = infer_empty_node  # type: ignore[assignment]
 
 
-def _populate_context_lookup(call, context):
+def _populate_context_lookup(
+    call: nodes.Call, context: InferenceContext | None
+) -> dict[InferenceResult, InferenceContext]:
```
DanielNoord, line 1146: Have you actually checked that this is `InferenceResult` and not `SuccesfulInferenceResult`?

DanielNoord (changes requested):

Pierre-Sassoulas (approved):
Looks pretty good, there are a lot of fixes. I would be in favor of merging. We definitely need to activate mypy to be able to tell if something regressed without having to check locally by comparing the output before and after (I didn't do that so the approve is based on trust and the fact that actually verifying would take a lot of time.)

DanielNoord (approved):
I did verify this last time and all comments have been resolved! Thanks!

PR #1888, opened 2022-11-28
Misc type stuff

nickdrozd:
## Description

Just some various type annotations.

## Type of Changes

### DIFF
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import ast
 import sys
 import types
+from collections.abc import Callable
 from functools import partial
 from typing import NamedTuple
 
@@ -36,6 +37,7 @@ class ParserModule(NamedTuple):
     context_classes: dict[type[ast.expr_context], Context]
 
     def parse(self, string: str, type_comments: bool = True) -> ast.Module:
+        parse_func: Callable[[str], ast.Module]
         if self.module is _ast_py3:
             if PY38_PLUS:
                 parse_func = partial(self.module.parse, type_comments=type_comments)
--- a/astroid/brain/brain_uuid.py
+++ b/astroid/brain/brain_uuid.py
@@ -8,7 +8,7 @@ from astroid.nodes.node_classes import Const
 from astroid.nodes.scoped_nodes import ClassDef
 
 
-def _patch_uuid_class(node):
+def _patch_uuid_class(node) -> None:
     # The .int member is patched using __dict__
     node.locals["int"] = [Const(0, parent=node)]
 
--- a/astroid/brain/brain_uuid.py
+++ b/astroid/brain/brain_uuid.py
@@ -8,7 +8,7 @@ from astroid.nodes.node_classes import Const
 from astroid.nodes.scoped_nodes import ClassDef
 
 
-def _patch_uuid_class(node) -> None:
+def _patch_uuid_class(node: ClassDef) -> None:
     # The .int member is patched using __dict__
     node.locals["int"] = [Const(0, parent=node)]
 
--- a/astroid/brain/helpers.py
+++ b/astroid/brain/helpers.py
@@ -5,8 +5,8 @@
 from astroid.nodes.scoped_nodes import Module
 
 
-def register_module_extender(manager, module_name, get_extension_mod):
-    def transform(node):
+def register_module_extender(manager, module_name: str, get_extension_mod) -> None:
+    def transform(node) -> None:
         extension_module = get_extension_mod()
         for name, objs in extension_module.locals.items():
             node.locals[name] = objs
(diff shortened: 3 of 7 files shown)


Pierre-Sassoulas (approved):
👌

astroid/_ast.py
```diff
@@ -36,6 +37,7 @@ class ParserModule(NamedTuple):
     context_classes: dict[type[ast.expr_context], Context]
 
     def parse(self, string: str, type_comments: bool = True) -> ast.Module:
+        parse_func: Callable[[str], ast.Module]
```
DanielNoord, line 40: This isn't really correct with the `partial`, but let's use it for now. We're removing/revisiting most of this code anyway when we drop 3.7.

astroid/brain/brain_uuid.py
```diff
@@ -8,7 +8,7 @@
 from astroid.nodes.scoped_nodes import ClassDef
 
 
-def _patch_uuid_class(node):
+def _patch_uuid_class(node) -> None:
```
DanielNoord, line 11:
```suggestion
def _patch_uuid_class(node: nodes.ClassDef) -> None:
```

astroid/brain/helpers.py
```diff
@@ -5,8 +5,8 @@
 from astroid.nodes.scoped_nodes import Module
 
 
-def register_module_extender(manager, module_name, get_extension_mod):
-    def transform(node):
+def register_module_extender(manager, module_name: str, get_extension_mod) -> None:
+    def transform(node) -> None:
```
DanielNoord, line 9:
```suggestion
    def transform(node: nodes.Module) -> None:
```

astroid/brain/helpers.py
```diff
@@ -5,8 +5,8 @@
 from astroid.nodes.scoped_nodes import Module
 
 
-def register_module_extender(manager, module_name, get_extension_mod):
-    def transform(node):
+def register_module_extender(manager, module_name: str, get_extension_mod) -> None:
```
DanielNoord, line 8:
```suggestion
def register_module_extender(manager: AstroidManager, module_name: str, get_extension_mod: Callable[[], nodes.Module]) -> None:
```

astroid/manager.py
```diff
@@ -78,12 +78,12 @@ def __init__(self) -> None:
         self._transform = AstroidManager.brain["_transform"]
 
     @property
-    def register_transform(self):
+    def register_transform(self) -> Callable[[Any, Any, Any], None]:
```
DanielNoord, line 81: Let's not do this. We're actually losing information here as the current return type seen in IDEs will include the default `None` value for the third argument.

astroid/manager.py
```diff
@@ -78,12 +78,12 @@ def __init__(self) -> None:
...
         # This and unregister_transform below are exported for convenience
         return self._transform.register_transform
 
     @property
-    def unregister_transform(self):
+    def unregister_transform(self) -> Callable[[Any, Any, Any], None]:
```
DanielNoord, line 86: Same here.

astroid/util.py
```diff
@@ -22,7 +22,7 @@ def __get__(self, instance, owner=None):
     return DescriptorProxy(obj)
 
 
-def lazy_import(module_name):
+def lazy_import(module_name: str):
```
DanielNoord, line 25:
```suggestion
def lazy_import(module_name: str) -> Proxy:
```
?

astroid/util.py
```diff
@@ -37,7 +37,7 @@ def __repr__(self) -> str:
 
     __str__ = __repr__
 
-    def __getattribute__(self, name):
+    def __getattribute__(self, name: str):
```
DanielNoord, line 40:
```suggestion
    def __getattribute__(self, name: str) -> Any:
```

DanielNoord (changes requested):
Made some suggestions for some the types!

astroid/_ast.py
nickdrozd: What should it be instead? Isn't it a function that takes a string and gives back a module?

DanielNoord: Well `partial` is its one type I believe, but it should do for now!

astroid/brain/helpers.py
DanielNoord: Could you do this suggestion as well?

DanielNoord (changes requested):
One final suggestion!

Pierre-Sassoulas (approved):
👍

DanielNoord (approved):

DanielNoord:
@Pierre-Sassoulas Again a failing CI.. Any idea?

Pierre-Sassoulas:
This one we need to relaunch until it work, I guess it's github actions acting up.

PR #1889, opened 2022-11-28
Allow untyped decorators in mypy configuration

nickdrozd:
These are the settings I've been using for adding types. Obviously we're far from getting it to pass with strict settings, but strict makes it easier to see what is unannotated.

For me at least, adding types to decorators is totally unfeasible right now, so those should be allowed for the time being.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,8 @@ no_implicit_optional = true
 scripts_are_modules = true
 show_error_codes = true
 warn_redundant_casts = true
+allow_untyped_decorators = true
+strict = true
 
 [[tool.mypy.overrides]]
 # Importlib typeshed stubs do not include the private functions we use
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,6 @@ scripts_are_modules = true
 show_error_codes = true
 warn_redundant_casts = true
 allow_untyped_decorators = true
-strict = true
 
 [[tool.mypy.overrides]]
 # Importlib typeshed stubs do not include the private functions we use


DanielNoord:
Let's not do this. It makes it even harder to turn on `mypy` in CI and avoid people adding more issues to our long list of typing issues. The earlier we turn it on the better imo.

Pierre-Sassoulas (approved):

DanielNoord (changes requested):
Why would we do this? Are there any decorators that are not possible to type due to `mypy` being broken? Or is it just that it is hard to get them right with `ParamSpec`?

Pierre-Sassoulas:
I think activating mypy, even with forgiving option, is very important. We should work progressively toward the perfect typing and activating mypy early will help to not introduce new problems. (Then we can remove type ignore and forgiving options progressively).

DanielNoord:
Agreed! However, based on the comments on my review in the other PR I felt like this was at least partially motivated by the difficulty of typing decorators, not so much with issues in mypy itself.

Unless we find actual issues within mypy I think we should hold off on disabling this. As far as I can remember there aren't many issues about decorators preventing us from turning typing on anyway.

Pierre-Sassoulas:
> there aren't many issues about decorators preventing us from turning typing on anyway.

If we're trying to fix everything before activating mypy when are we going to activate mypy exactly ? If adding ``type: ignore`` everywhere was a reasonable option I would do that and activate mypy right now. I would be in favor of activating every possible forgiving option that remove some issue (not only decorator) and starting from there personally. It's a really cheap way to activate mypy faster.

DanielNoord:
That just hides crucial flaws that might prevent us from actually turning on `mypy` at some point anyway. I have gotten awfully close at giving up due to some patterns that are just (almost) impossible to type.

I think it would be good to get `astroid` in the same shape `pylint` was: no strict warnings turned on and no lenient warnings turned off. That should show the most important problems with our code in terms of "type-ability". This really doesn't provide us much benefit and I would happily help anybody getting a `ParamSpec` correct. (_In fact, with `ParamSpec` I think typing decorators has become some of easier things to type no that they have basically their own construct_).

Pierre-Sassoulas:
>  I have gotten awfully close at giving up due to some patterns that are just (almost) impossible to type.

Should we try to type those pattern or just give up and add type ignore where applicable forever or until we refactor it ? typing is great but even the best feature can have a bad ratio of cost vs usefulness.

DanielNoord:
I think we should. The main difficulty is not that they are decorators, which this setting is for, but rather that they are doing dynamic stuff that is not really allowed by `mypy`. Thus, if we really want to give up we should disable them specifically and give reasoning in a comment instead of ignoring just because they are decorators.

Pierre-Sassoulas:
All right, should we open an issue and merge this in the meantime ?

DanielNoord:
I don't think we should merge this, there is no immediate benefit and we can add ignores to decorators in specific PRs after individual review.
_See also the other typing PR that initially triggered this PR._

PR #1891, opened 2022-11-29
Mandatory fields

nickdrozd:

This is a fun one.

[This comment](https://github.com/PyCQA/astroid/pull/1885#discussion_r1030791292) got me to thinking about required fields. There are many places in the code where a field is assumed to be not none even though the field annotation gives it as optional. Mypy rightly complains about this.

There are three ways to deal with this: (1) not-none assertions (yikes), (2) instance checks (yuck), and (3) annotate the fields as mandatory. This PR attempts (3).

Running Mypy strict, this reduces warnings from **1534** to **1478**, which is pretty good. I'm assuming this also will improve performance slightly.

In many cases, making the fields mandatory means that they aren't set until `postinit`, and so there isn't anything node-specific to do in `__init__`. So the `__init__` method can be deleted altogether since it does nothing but delegate to `super`.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1670,47 +1670,17 @@ class Compare(NodeNG):
     [('<=', <Name.b l.1 at 0x7f23b2e9e2b0>), ('<=', <Name.c l.1 at 0x7f23b2e9e390>)]
     """
 
+    left: NodeNG
+    """The value at the left being applied to a comparison operator."""
+
+    ops: list[tuple[str, NodeNG]]
+    """The remainder of the operators and their relevant right hand value."""
+
     _astroid_fields = ("left", "ops")
-
-    def __init__(
-        self,
-        lineno: int | None = None,
-        col_offset: int | None = None,
-        parent: NodeNG | None = None,
-        *,
-        end_lineno: int | None = None,
-        end_col_offset: int | None = None,
-    ) -> None:
-        """
-        :param lineno: The line that this node appears on in the source code.
-
-        :param col_offset: The column that this node appears on in the
-            source code.
-
-        :param parent: The parent node in the syntax tree.
-
-        :param end_lineno: The last line this node appears on in the source code.
-
-        :param end_col_offset: The end column this node appears on in the
-            source code. Note: This is after the last symbol.
-        """
-        self.left: NodeNG | None = None
-        """The value at the left being applied to a comparison operator."""
-
-        self.ops: list[tuple[str, NodeNG]] = []
-        """The remainder of the operators and their relevant right hand value."""
-
-        super().__init__(
-            lineno=lineno,
-            col_offset=col_offset,
-            end_lineno=end_lineno,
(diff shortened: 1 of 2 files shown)


DanielNoord:
I think @cdce8p has some ideas about this, but at the same time his ideas might have changed now that we are planning for `3.0`. Let's discuss!

nickdrozd:
In non-strict mode, this change reduces Mypy errors from **364** to **313**. And that's just from the nodes that have been done so far.

Really, I think this is the only viable path forward. There are just too many type errors otherwise.

nickdrozd:
@cdce8p Any thoughts on this one?

cdce8p (changes requested):
The idea is good, however in it's current for this will be a massive breaking change. As you might have seen, there is already some logic in place to deprecate certain default values but all in all that needs more time more and more consideration before I would move forward with it.

This is certainly something we can do in some form for `3.0`, but it isn't a priority for me at the moment.

I'll be quite busy the next few weeks. Maybe I have some time after that towards late March  - April.

--
Sorry to disappoint you again here. Adding type annotations is valuable, however there are also other considerations at play here. It isn't particularly easiest to work with such old legacy code. Sometimes is better to be careful.

nickdrozd:
Maybe there are use cases I don't understand. Suppose this change were merged and released immediately. Who would notice? What exactly would break?

DanielNoord:
> Maybe there are use cases I don't understand. Suppose this change were merged and released immediately. Who would notice? What exactly would break?

There are dependants on `astroid`. Such as https://github.com/readthedocs/sphinx-autoapi.
We can't remove stuff without giving them time to react or prepare.

_That said, I would highly be in favour of just taking `astroid` private, but that has other issues assosciated with it._

Pierre-Sassoulas:
Last month a quarter of a million unique astroid downloads were not directly linked to pylint (according to pypystats: this is the number of astroid download last month minus the number of pylint download last month). This does not account for one of astroid's use case in pre-commit hook where it's analyzing code and is downloaded by pre-commit and not via pip. So breaking astroid's API can potentially break a sizable number of small packages and bother numerous maintainers and users. We need to deprecate with a proper waiting period before removing. 

Which is not to mean that we should suffer until the end of time because of the current state of the codebase and not touch anything. But we do need to be responsible about it and let other maintainers adapt in a reasonable time.

nickdrozd:
I understand that there are non-Pylint users of Astroid. But that isn't my question. My question is: which of those users would be impacted by this change? Does anyone know of any specific examples of code that would break as a result of the changes in this PR?

Or to put it another way, are there any test cases that could be added that would break in the presence of this change?

DanielNoord:
That's not really how dependencies and semantic versioning work.. We simply don't know how users use the tool, just as we don't know how people use `pylint`. We make assumptions and try not to change anything in the public API, without deprecations.

For `astroid` we haven't defined what is public and what isn't, so therefore everything kind of is... That's annoying, but something we sort of have to deal with.

PR #1892, opened 2022-11-29
Fix crash if numpy doesn't have version

tushar-deepsource:
## Type of Changes

## Related Issue

Resolves https://github.com/PyCQA/pylint/issues/7868

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,7 +24,7 @@ Release date: TBA
 
   Closes PyCQA/pylint#5225
 
-* Fix crash if ``numpy`` module doesn't have ``version`` attribute. 
+* Fix crash if ``numpy`` module doesn't have ``version`` attribute.
 
   Refs PyCQA/pylint#7868
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,10 @@ Release date: TBA
 
   Refs PyCQA/pylint#2567
 
+* Fix crash if `numpy` module doesn't have `version` attribute
+
+  Refs PyCQA/pylint#7868
+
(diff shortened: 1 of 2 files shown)


mbyrnepr2:
Thanks for the MR @tushar-deepsource.
I'm hesitant about the fix because, with a local `numpy.py` now getting imported, we run into `E1101: Module 'numpy' has no 'ndarray' member (no-member)` (using the [code sample](https://github.com/PyCQA/pylint/issues/7868#issue-1468175279) linked from the issue). Perhaps an alternative is to identify if there is a local `numpy.py` & try to ignore that one?

mbyrnepr2 (approved):

tushar-deepsource:
Is this waiting on something?

mbyrnepr2:
Hey @tushar-deepsource I think we can merge. I just realized there is no Change log entry. Can you add that, then I think we should be good to go.

Pierre-Sassoulas (approved):

DanielNoord (approved):

DanielNoord:
@tushar-deepsource Could you force-push the branch? Jobs seem to broken..

ChangeLog
```diff
@@ -24,6 +24,9 @@ Release date: TBA
 
   Closes PyCQA/pylint#5225
 
+* Fix crash if `numpy` module doesn't have `version` attribute
```
mbyrnepr2, line 27:
```suggestion
* Fix crash if ``numpy`` module doesn't have ``version`` attribute. 
```

mbyrnepr2: Just to be consistent with other logs

PR #1893, opened 2022-11-30
Use dict literals

nickdrozd:
## Description

Use dict literals as per new check https://github.com/PyCQA/pylint/issues/7690

## Type of Changes


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Shouldn't we also add the check ?

nickdrozd:
The warning is from `use-dict-literal`, so it's not a new check. But this particular circumstance is only triggered in the latest Pylint.

Pierre-Sassoulas:
Right, you're anticipating pylint 2.16.0 👌

PR #1910, opened 2022-12-21
Fix CI for pypy 3.8 and 3.9

Pierre-Sassoulas:
## Type of Changes

## Description

The CI is broken for pypy3.8 and 3.9 atm. Hoping for a cache issue. 🤞


Pierre-Sassoulas:
@cdce8p I'm pinging you because you implemented something similar originally. I don't think there's another way than to extend it right now, the issue seems to be in pypy's ast.

PR #1931, opened 2022-12-29
brain tip for numpy masked_invalid

clavedeluna:
## Type of Changes

## Description

This is my first time trying to update astroid's brain so tips are really appreciated. I'm trying to get the following 

```
import numpy as np

DATA = np.array([np.nan, 2, 3])
DATA = np.ma.masked_invalid(DATA)
print(DATA.mask)
assert isinstance(DATA, np.ma.MaskedArray)
```

To not emit `numpy_pylint.py:7:6: E1101: Instance of 'ndarray' has no 'mask' member (no-member)`
I got it to work with the code in this PR but I'm getting ` Redefinition of DATA type from .ndarray to numpy.ma.core.MaskedArray (redefined-variable-type)`. If I don't re-define `DATA` like this: then it works

```
DATA = np.ma.masked_invalid(np.array([np.nan, 2, 3]))
print(DATA.mask)
```

Closes https://github.com/PyCQA/pylint/issues/5715

### DIFF
--- a/astroid/brain/brain_numpy_ma.py
+++ b/astroid/brain/brain_numpy_ma.py
@@ -21,7 +21,7 @@ def numpy_ma_transform():
     import numpy.ma
     def masked_where(condition, a, copy=True):
         return numpy.ma.masked_array(a, mask=[])
-    
+
     def masked_invalid(a, copy=True):
         return numpy.ma.masked_array(a, mask=[])
     """
--- a/astroid/brain/brain_numpy_ma.py
+++ b/astroid/brain/brain_numpy_ma.py
@@ -21,6 +21,9 @@ def numpy_ma_transform():
     import numpy.ma
     def masked_where(condition, a, copy=True):
         return numpy.ma.masked_array(a, mask=[])
+    
+    def masked_invalid(a, copy=True):
+        return numpy.ma.masked_array(a, mask=[])
     """
     )
 
--- a/astroid/brain/brain_numpy_ma.py
+++ b/astroid/brain/brain_numpy_ma.py
@@ -11,7 +11,7 @@ from astroid.manager import AstroidManager
 
 def numpy_ma_transform():
     """
-    Infer the call of the masked_where function
+    Infer the call of various numpy.ma functions
 
     :param node: node to infer
     :param context: inference context
--- a/tests/unittest_brain_numpy_ma.py
+++ b/tests/unittest_brain_numpy_ma.py
@@ -13,6 +13,8 @@ except ImportError:
 
 from astroid import builder
 
+parametrize = pytest.mark.parametrize("alias_import", [True, False])
+
 
 @pytest.mark.skipif(HAS_NUMPY is False, reason="This test requires the numpy library.")
 class TestBrainNumpyMa:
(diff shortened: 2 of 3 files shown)


DanielNoord:
Tagging @mbyrnepr2 as they have experience with this brain

mbyrnepr2:
Thanks for the heads-up!

Looks great. The change is in right module ([source module of masked_invalid & masked_where](https://github.com/numpy/numpy/blob/main/numpy/ma/core.py#L2332)) & the return value looks reasonable as masked_invalid is a modified version of masked_where. Just [add a test](https://github.com/PyCQA/astroid/blob/main/tests/unittest_brain_numpy_ma.py) & I think its good to go! (you might need to parametrize the tests or generally spruce them up as you see fit).

astroid/brain/brain_numpy_ma.py
```diff
@@ -21,6 +21,9 @@ def numpy_ma_transform():
     import numpy.ma
     def masked_where(condition, a, copy=True):
         return numpy.ma.masked_array(a, mask=[])
+
+    def masked_invalid(a, copy=True):
```
mbyrnepr2, line 25: I know it is still work in progress but I noticed the docstring above is now out of date since it specifically mentions `masked_where`.

tests/unittest_brain_numpy_ma.py
```diff
@@ -13,41 +13,58 @@
...
         The "masked_where" node is an Attribute
         """
-        src = """
-        import numpy as np
+        import_str = (
+            "import numpy as np"
+            if alias_import
+            else "from numpy.ma import masked_where"
+        )
+        func_call = "np.ma.masked_where" if alias_import else "masked_where"
```
mbyrnepr2, line 42:
```suggestion
        func = "np.ma.masked_where" if alias_import else "masked_where"
```

Its just the function name so I think this is a bit clearer.

mbyrnepr2:
I like your test structure @clavedeluna but I was also thinking of something like this since there is plenty of similar logic in both tests, what do you think?
```python
    @pytest.mark.parametrize("alias_import", [True, False])
    @pytest.mark.parametrize("ma_function", ["masked_invalid", "masked_where"])
    def test_numpy_ma_returns_maskedarray(self, alias_import, ma_function):
        """
        Test that calls to numpy ma functions return a MaskedArray object.

        The `ma_function` node is an Attribute or a Name
        """
        import_str = (
            "import numpy as np"
            if alias_import
            else f"from numpy.ma import {ma_function}"
        )
        func = f"np.ma.{ma_function}" if alias_import else ma_function

        src = f"""
        {import_str}
        data = np.ndarray((1,2))
        {func}([1, 0, 0], data)
        """
        self._assert_maskedarray(src)
"""

mbyrnepr2:
[Changelog can be added here](https://github.com/PyCQA/astroid/blob/main/ChangeLog)

clavedeluna:
> I like your test structure @clavedeluna but I was also thinking of something like this since there is plenty of similar logic in both tests, what do you think?
> 
> ```python
>     @pytest.mark.parametrize("alias_import", [True, False])
>     @pytest.mark.parametrize("ma_function", ["masked_invalid", "masked_where"])
>     def test_numpy_ma_returns_maskedarray(self, alias_import, ma_function):
>         """
>         Test that calls to numpy ma functions return a MaskedArray object.
> 
>         The `ma_function` node is an Attribute or a Name
>         """
>         import_str = (
>             "import numpy as np"
>             if alias_import
>             else f"from numpy.ma import {ma_function}"
>         )
>         func = f"np.ma.{ma_function}" if alias_import else ma_function
> 
>         src = f"""
>         {import_str}
>         data = np.ndarray((1,2))
>         {func}([1, 0, 0], data)
>         """
>         self._assert_maskedarray(src)
> """
> ```

I actually thought of that but thought it would be harder to read the test. But ok will update

clavedeluna:
> [Changelog can be added here](https://github.com/PyCQA/astroid/blob/main/ChangeLog)

👍 👍  is there a plan to fail the pipeline if a changelog isn't updated? similar to pylint repo

mbyrnepr2:
> > [Changelog can be added here](https://github.com/PyCQA/astroid/blob/main/ChangeLog)
> 
> 👍 👍 is there a plan to fail the pipeline if a changelog isn't updated? similar to pylint repo

I'm not aware of such a plan (although I'm more out of touch with astroid than Pylint). Could be a good one to propose as a maintenance enhancement.

mbyrnepr2:
Thanks @clavedeluna 👍 you've made numpy users' lives easier.

PR #1932, opened 2022-12-31
Cache `ClassDef._metaclass_lookup_attribute`

jacobtylerwalls:
## Type of Changes

## Description
When linting `import returns.result` from the `returns` library, generates 93,315 cache hits to 443 misses, cutting the total linting time from 50s to 12s.

Closes PyCQA/pylint#4750


Pierre-Sassoulas (commented):
We might want to create an intermediary cached function without 'self' or we could ignore the flake8 warning. The perf increase looks amazing.

Pierre-Sassoulas (approved):
Amazing !

PR #1935, opened 2023-01-02
Use codecov and separate flags per OS

DanielNoord:
## Steps

## Description

Similar to `pylint`. However, this also uses the flag functionality, which makes it possible to report coverage per OS.
I (finally) got this to work with `pydocstringformatter` today and found out how to do this here as well.

One effect of this is that coverage will only be reported after all tests have run. We can consider making an additional job to only report linux coverage first, but that feels unnecessary. Let's see if this fits our needs for now.

## Type of Changes

### DIFF
--- a/requirements_test_min.txt
+++ b/requirements_test_min.txt
@@ -1,2 +1,4 @@
+coverage~=7.0
 pytest
+pytest-cov~=4.0
 typing-extensions>=3.10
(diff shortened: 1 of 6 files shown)


requirements_test_min.txt
```diff
@@ -1,2 +1,4 @@
+coverage~=7.0
 pytest
```
Pierre-Sassoulas, line 2: We would need to put pytest in ``requirements_test.txt`` then

Pierre-Sassoulas (approved):
👌

PR #1949, opened 2023-01-09
Initial pass with ``pydocstringformatter`` over tests

DanielNoord:
## Type of Changes

## Description

We're getting there 😄


tests/unittest_inference.py
```diff
@@ -1469,7 +1469,7 @@ def test(self):
             self.fail(f"expected to find an instance of Application in {inferred}")
 
     def test_list_inference(self) -> None:
-        """#20464"""
+        """#20464."""
```
Pierre-Sassoulas, line 1472:
```suggestion
```
Dead logilab link, better to remove.

PR #1956, opened 2023-01-09
[codecov] Use our own token to avoid being rate limited

Pierre-Sassoulas:
## Type of Changes

## Description

See https://github.com/codecov/codecov-action/issues/557\#issuecomment-1224970469

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -254,6 +254,7 @@ jobs:
           coverage xml -o coverage-linux.xml
       - uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true
           flags: linux
@@ -264,6 +265,7 @@ jobs:
           coverage xml -o coverage-windows.xml
       - uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true
           flags: windows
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
At least dependabot / pre-commit (?) and owner MR will work properly. This afternoon the dependabot opened maybe 5 merge request which contributed to the problem. (Also on astroid I do not have a fork, I work directly on the main repo, maybe because I already had contributor right the first time I worked on astroid).

DanielNoord:
Hmm, I have never ran into these issues on `pydocstringformatter` and `pylint` was also fine until today.

Having the coverage broken for any fork PR is really annoying, especially on `pylint`. I would be in favour of waiting with this and seeing if the rate limiting issues were just really high today or if there is a general trend.

Pierre-Sassoulas:
I agree that this is annoying, but this is a (partial) fix why should we wait ?

DanielNoord:
> I agree that this is annoying, but this is a (partial) fix why should we wait ?

Well, it isn't a real fix and breaks our actual workflow. Of the open PRs in both projects less than a quarter a non-fork.
With this change they won't be able to get coverage reports and pass CI.
Re-running manually is annoying, but seems like the better option for now.

Pierre-Sassoulas:
> With this change they won't be able to get coverage reports and pass CI. 

Huh ? Not sure about that. I think this revert to the default of using their (codecov) token ? Look here: https://github.com/PyCQA/pylint/pull/8044 it's a PR from my fork of pylint.

DanielNoord (approved):
Misinterpreted the comment and the docs. It is only for runs in the forked repository, not in the upstream one!

PR #1964, opened 2023-01-13
Do not require typing_extensions on Python 3.11

paper42:
19878a55e61ce8788db530240dba9570706a5aac added an unconditional dependency on typing_extensions to fix tests with Python 3.10 3.11.
This commit fixes this issue by only requiring typing_extensions on Python versions lower than 3.11 and using standard typing on Python 3.11

cc @Pierre-Sassoulas 

## Type of Changes

## Description

Refs #1944

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,8 @@ Release date: TBA
 
   Closes #1958
 
+* Remove unnecessary typing_extensions dependency on Python 3.11 and newer
+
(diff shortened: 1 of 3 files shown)


DanielNoord (approved):
If tests do indeed pass this is a nice fix! Thanks!

paper42:
The 3.10 test failure seems unrelated

PR #1970, opened 2023-01-17
Fix order of overwritten attributes in inherited dataclasses

DanielNoord:
## Type of Changes

## Description

Closes PyCQA/pylint#7881

Slowly going through all of these `dataclass` issues. Turns out some of the previous tests were actually incorrectly ordered as well.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,10 @@ Release date: TBA
 
   Closes #1958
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (changes requested):
LGTM apart from the nit.

Pierre-Sassoulas (approved):

PR #1973, opened 2023-01-20
Fix issues with `typing_extensions.TypeVar`

cdce8p:
## Description
Fix some issues with `typing_extensions.TypeVar` added in `4.4.0`.
Changes for pylint: https://github.com/PyCQA/pylint/pull/8089

While add it, clean up some more code. `typing_extensions` does include it's own copy of `NamedTuple` and `TypedDict` as well. _Although those changes might not actually be necessary, haven't seen a test failure for those yet. The existing tests pass even if the import is changed to `typing_extensions`. So didn't add any new ones._

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -40,7 +40,15 @@ ENUM_BASE_NAMES = {
     "enum.IntFlag",
 }
 ENUM_QNAME: Final[str] = "enum.Enum"
-TYPING_NAMEDTUPLE_BASENAMES: Final[set[str]] = {"NamedTuple", "typing.NamedTuple"}
+TYPING_NAMEDTUPLE_QUALIFIED: Final = {
+    "typing.NamedTuple",
+    "typing_extensions.NamedTuple",
+}
+TYPING_NAMEDTUPLE_BASENAMES: Final = {
+    "NamedTuple",
+    "typing.NamedTuple",
+    "typing_extensions.NamedTuple",
+}
 
 
 def _infer_first(node, context):
@@ -542,7 +550,7 @@ def infer_typing_namedtuple(
     except (InferenceError, StopIteration) as exc:
         raise UseInferenceDefault from exc
 
-    if func.qname() != "typing.NamedTuple":
+    if func.qname() not in TYPING_NAMEDTUPLE_QUALIFIED:
         raise UseInferenceDefault
 
     if len(node.args) != 2:
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -34,6 +34,7 @@ from astroid.nodes.node_classes import (
 from astroid.nodes.scoped_nodes import ClassDef, FunctionDef
 from astroid.util import Uninferable
 
+import sys
 if sys.version_info >= (3, 8):
     from typing import Final
 else:
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -35,12 +35,12 @@ from astroid.nodes.scoped_nodes import ClassDef, FunctionDef
 from astroid.util import Uninferable
 
 TYPING_TYPEVARS = {"TypeVar", "NewType"}
-TYPING_TYPEVARS_QUALIFIED = {
+TYPING_TYPEVARS_QUALIFIED: Final = {
     "typing.TypeVar",
     "typing.NewType",
     "typing_extensions.TypeVar",
 }
-TYPING_TYPEDDICT_QUALIFIED = {"typing.TypedDict", "typing_extensions.TypedDict"}
+TYPING_TYPEDDICT_QUALIFIED: Final = {"typing.TypedDict", "typing_extensions.TypedDict"}
 TYPING_TYPE_TEMPLATE = """
 class Meta(type):
     def __getitem__(self, item):
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -33,6 +33,10 @@ from astroid.nodes.node_classes import (
 )
 from astroid.nodes.scoped_nodes import ClassDef, FunctionDef
 from astroid.util import Uninferable
+if sys.version_info >= (3, 8):
+    from typing import Final
+else:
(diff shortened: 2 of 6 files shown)


requirements_test_min.txt
```diff
@@ -1,3 +1,4 @@
 coverage~=7.0
 pytest
 pytest-cov~=4.0
+typing-extensions>=4.4.0
```
Pierre-Sassoulas, line 4: Is this a test requirements or a runtime requirements ? Because we had an issue with this earlier, we want to know if astroid would fail with earlier version of typing_extensions see #1944. Maybe we need to test with the lower and higher version of ``typing_extensions `` and that would probably need to be handled with github actions and multiple jobs requiring both.

Pierre-Sassoulas (changes requested):

requirements_test_min.txt
cdce8p: It's a test requirement. `typing_extensions.TypeVar` was added in `4.4.0`.

> Maybe we need to test with the lower and higher version of `typing_extensions ` and that would probably need to be handled with github actions and multiple jobs requiring both.

Not sure it's worth it. This usually isn't an issue and the one time it is, we will see it fairly quickly.

There are alternatives
* We could bump the runtime requirement and make it unconditional.
* Or I could remove the test case I added here. It doesn't provide much value and I don't expect we'll introduce a regression anytime soon.

Pierre-Sassoulas: > We could bump the runtime requirement and make it unconditional.

What do you mean by make it unconditional ? ``typing_extensions`` is used by a lot of lib, I think we need to use the less constrained requirement possible.

cdce8p: > > We could bump the runtime requirement and make it unconditional.
> 
> What do you mean by make it unconditional ?

We don't install it for Python 3.11 currently. This here would need to be changed to
https://github.com/PyCQA/astroid/blob/c2519e52594fd5b971a0a1bf5094ec2fa307a7dd/pyproject.toml#L39

`"typing-extensions>=4.4.0"`

> `typing_extensions` is used by a lot of lib, I think we need to use the less constrained requirement possible.

I agree, however it shouldn't be an issue in this case. We would only require the latest version of `typing-extensions`.

Pierre-Sassoulas: Ok, make sense, let's do this then.

cdce8p: I tried something a bit different today with https://github.com/PyCQA/astroid/pull/1973/commits/b7c6b618a2d5305142ee661b442ccb785ee053a8. By moving the test case to a separate class and guarding it by `pytest.mark.skipif`, we can move the requirement to `requirements_test_brain.txt`. Those are only installed on linux and windows, not for PyPy. Maybe that is a bit better?

Let me know what you think. I can still move the requirement and update it in `pyproject.toml` like we talked about.

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -40,7 +40,15 @@
     "enum.IntFlag",
 }
 ENUM_QNAME: Final[str] = "enum.Enum"
-TYPING_NAMEDTUPLE_BASENAMES: Final[set[str]] = {"NamedTuple", "typing.NamedTuple"}
+TYPING_NAMEDTUPLE_QUALIFIED: Final = {
```
DanielNoord, line 43: Why did you remove the `set[str]`?

astroid/brain/brain_typing.py
```diff
@@ -34,9 +34,13 @@
 from astroid.nodes.scoped_nodes import ClassDef, FunctionDef
 from astroid.util import Uninferable
 
-TYPING_NAMEDTUPLE_BASENAMES = {"NamedTuple", "typing.NamedTuple"}
 TYPING_TYPEVARS = {"TypeVar", "NewType"}
-TYPING_TYPEVARS_QUALIFIED = {"typing.TypeVar", "typing.NewType"}
+TYPING_TYPEVARS_QUALIFIED = {
```
DanielNoord, line 38:
```suggestion
TYPING_TYPEVARS_QUALIFIED: Final = {
```

astroid/brain/brain_typing.py
```diff
@@ -34,9 +34,13 @@
...
+    "typing.TypeVar",
+    "typing.NewType",
+    "typing_extensions.TypeVar",
+}
+TYPING_TYPEDDICT_QUALIFIED = {"typing.TypedDict", "typing_extensions.TypedDict"}
```
DanielNoord, line 43:
```suggestion
TYPING_TYPEDDICT_QUALIFIED: Final = {"typing.TypedDict", "typing_extensions.TypedDict"}
```

astroid/brain/brain_namedtuple_enum.py
cdce8p: It's not necessary. Mypy can infer final types. Especially easy ones like this. It's a bit different with complex dictionary types though. Something like `dict[str, int | str | dict[str, Any]]`, there it can make a small performance difference.

DanielNoord (approved):

requirements_test_min.txt
Pierre-Sassoulas: Yeah it's actually nicer that way !

astroid/brain/brain_typing.py
```diff
@@ -34,9 +34,13 @@
 from astroid.nodes.scoped_nodes import ClassDef, FunctionDef
 from astroid.util import Uninferable
```
Pierre-Sassoulas, line 35:
```suggestion
from astroid.util import Uninferable
if sys.version_info >= (3, 8):
    from typing import Final
else:
    from typing_extensions import Final
```

Pierre-Sassoulas (approved):
I think there's a missing import.

astroid/brain/brain_typing.py
```diff
@@ -34,9 +34,18 @@
...
 
-TYPING_NAMEDTUPLE_BASENAMES = {"NamedTuple", "typing.NamedTuple"}
+if sys.version_info >= (3, 8):
```
Pierre-Sassoulas, line 37:
```suggestion
import sys
if sys.version_info >= (3, 8):
```

PR #1976, opened 2023-01-27
Add Lock to multiprocessing

clavedeluna:
## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/3313

### DIFF
--- a/astroid/brain/brain_multiprocessing.py
+++ b/astroid/brain/brain_multiprocessing.py
@@ -78,6 +78,7 @@ def _multiprocessing_managers_transform():
         Queue = JoinableQueue = queue.Queue
         Event = threading.Event
         RLock = threading.RLock
+        Lock = threading.Lock
         BoundedSemaphore = threading.BoundedSemaphore
         Condition = threading.Condition
         Barrier = threading.Barrier
--- a/astroid/brain/brain_multiprocessing.py
+++ b/astroid/brain/brain_multiprocessing.py
@@ -78,7 +78,7 @@ def _multiprocessing_managers_transform():
         Queue = JoinableQueue = queue.Queue
         Event = threading.Event
         RLock = threading.RLock
-        Lock = multiprocessing.synchronize.Lock()
+        Lock = threading.Lock
         BoundedSemaphore = threading.BoundedSemaphore
         Condition = threading.Condition
         Barrier = threading.Barrier
--- a/astroid/brain/brain_multiprocessing.py
+++ b/astroid/brain/brain_multiprocessing.py
@@ -78,7 +78,7 @@ def _multiprocessing_managers_transform():
         Queue = JoinableQueue = queue.Queue
         Event = threading.Event
         RLock = threading.RLock
-        Lock = threading.Lock
+        Lock = multiprocessing.synchronize.Lock()
         BoundedSemaphore = threading.BoundedSemaphore
         Condition = threading.Condition
         Barrier = threading.Barrier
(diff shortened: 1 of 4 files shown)


clavedeluna:
Hi maintainers, I'm attempting to fix https://github.com/PyCQA/pylint/issues/3313 . While this change no longer emits `testing.py:4: [E1101(no-member), ] Instance of 'SyncManager' has no 'Lock' member; maybe 'RLock'?` I'm now getting a ` Too many positional arguments for function call (too-many-function-args)` on that line. Trying to investigate and think it's likely due to [this line](https://github.com/python/cpython/blob/53d9cd95cd91f1a291a3923acb95e0e86942291a/Lib/threading.py#L108) . What else can I do here?

DanielNoord (approved):

DanielNoord (dismissed):
Can't you reference the private import instead?

astroid/brain/brain_multiprocessing.py
```diff
@@ -78,6 +78,7 @@ class SyncManager(object):
         Queue = JoinableQueue = queue.Queue
         Event = threading.Event
         RLock = threading.RLock
+        Lock = threading.Lock
```
Pierre-Sassoulas, line 81:
```suggestion
        Lock = multiprocessing.synchronize.Lock()
```
Would this work ? I'm basing this suggestion on https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock

clavedeluna: Yay this indeed works perfectly! Thank you!

Pierre-Sassoulas:
The ``tests/test_import_graph.py::test_checker_dep_graphs`` test break in pylint with this version of astroid but it seems to also break with the current ``main`` branch of astroid, so...

tests/unittest_brain.py
```diff
@@ -711,6 +711,7 @@ def test_multiprocessing_manager(self) -> None:
         joinable_queue = manager.JoinableQueue()
         event = manager.Event()
         rlock = manager.RLock()
+        lock = multiprocessing.synchronize.Lock()
```
cdce8p, line 714:
```suggestion
        lock = manager.Lock()
```

tests/unittest_brain.py
```diff
@@ -736,6 +737,10 @@ def test_multiprocessing_manager(self) -> None:
         rlock_name = "threading._RLock"
         self.assertEqual(rlock.qname(), rlock_name)
 
+        lock = next(module["lock"].infer())
+        lock_name = "multiprocessing.synchronize.Lock"
```
cdce8p, line 741:
```suggestion
        lock_name = "threading.lock"
```

cdce8p (changes requested):

clavedeluna:
Hi @cdce8p your review (one part outdated) is a bit confusing to me, I can't figure out if your suggestion of using `threading` is what we want or not, given we're focusing on `multiprocessing` module. Maybe if you add a commit with what you're suggesting I can test it and verify the original test case does work.

cdce8p:
> Maybe if you add a commit with what you're suggesting I can test it and verify the original test case does work.

Sorry for the confusion, that wasn't may intention. Pushed the commit. It's basically your original proposal but was a small change in `brain_threading.py` to make it work.

I think for the `SyncManager` it does make sense to use `Lock = threading.Lock` as this is closest to the ones already there.
https://github.com/python/cpython/blob/v3.11.1/Lib/multiprocessing/managers.py#L1214-L1215

clavedeluna:
@cdce8p ahh perfect! I got it now and it does indeed work 👏

cdce8p:
> @cdce8p ahh perfect! I got it now and it does indeed work 👏

Great! Could you add a short changelog entry for `2.13.4`. After that we can merge it.

cdce8p (approved):
Looks good! Thanks @clavedeluna 👍🏻

PR #1977, opened 2023-01-27
Add support for binary union types - Python 3.10

cdce8p:
## Description
Add basic inference support for runtime binary or unions added in Python 3.10.
Not completely sure this is 100% the right approach for it, but it works. We can always iterate on it later if we want / need to.

At runtime `int | None` is represented as an instance of `types.UnionType`.

Refs https://github.com/PyCQA/pylint/issues/8119

--
After this is merged, I think it would make sense to do a new release so we can include it in pylint `2.16.0`.

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -121,11 +121,12 @@ class Proxy:
         if proxied is None:
             # This is a hack to allow calling this __init__ during bootstrapping of
             # builtin classes and their docstrings.
-            # For Const and Generator nodes the _proxied attribute is set during bootstrapping
+            # For Const, Generator, and UnionType nodes the _proxied attribute
+            # is set during bootstrapping
             # as we first need to build the ClassDef that they can proxy.
             # Thus, if proxied is None self should be a Const or Generator
             # as that is the only way _proxied will be correctly set as a ClassDef.
-            assert isinstance(self, (nodes.Const, Generator))
+            assert isinstance(self, (nodes.Const, Generator, UnionType))
         else:
             self._proxied = proxied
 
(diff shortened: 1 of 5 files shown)


astroid/inference.py
```diff
@@ -814,6 +823,22 @@ def _get_binop_flow(
             _bin_op(left, binary_opnode, op, right, context),
             _bin_op(right, binary_opnode, op, left, reverse_context, reverse=True),
         ]
+
+    if (
+        PY310_PLUS
```
Pierre-Sassoulas, line 828: Isn't it py310 or future annotations used ?

Pierre-Sassoulas (commented):
I think this can go in 2.16.

Pierre-Sassoulas:
But shouldn't it be in astroid 2.14 ?

astroid/inference.py
cdce8p: I don't think so. AFAIK inference should model the runtime effects and for Python 3.9 `int | str` would throw a `TypeError`.

We do exclude handle `int | str` in typing contexts with `from __future__ import annotations` separately, in pylint.

cdce8p:
> But shouldn't it be in astroid 2.14 ?

It certainly could. Moved the changelog entry.
We would then only need to decide if we want to release `2.13.4` with the `typing_extensions.TypeVar` fix or just include it with `2.14`.

astroid/bases.py
```diff
@@ -669,3 +670,36 @@ def __repr__(self) -> str:
...
+
+class UnionType(BaseInstance):
+    """Special node representing new style typing unions.
+
+    Proxied class is set once for all in raw_building.
+    """
+
+    _proxied: nodes.ClassDef
+
+    def __init__(self, left, right, parent=None):
```
DanielNoord, line 683: Can we have typing here?

DanielNoord (changes requested):
One minor nit. Rest LGTM!

DanielNoord (approved):

Pierre-Sassoulas (changes requested):
We should probably cover ``callable``, ``pytype``, ``display_type``, ``bool_value`` and ``__str__`` if only to prevent regression later, right ?

cdce8p:
> We should probably cover `callable`, `pytype`, `display_type`, `bool_value` and `__str__` if only to prevent regression later, right ?

Added some more asserts. Although it might be that we need to go back and change some of them later. I don't know every detail of how the new class will be used by astroid and pylint.

cdce8p:
The isort issue will be fixed by #1989

Pierre-Sassoulas:
:thumbsup:

cdce8p:
> 👍

@Pierre-Sassoulas Think you actually need to approve it since you requested changes earlier. Or I could just bypass the branch protection in this case 😄

Pierre-Sassoulas (approved):
My bad 😄

PR #1979, opened 2023-01-28
Fix PyPy `ClassDef.fromlino` with decorators

cdce8p:
## Description
Followup to #1910
Never got around to reviewing it in time unfortunately.

The last PyPy release for 3.8 (v7.3.11) actually provided a fix for the ast lineno issue and is now compliant with the cpython implementation.

https://doc.pypy.org/en/latest/release-v7.3.11.html#python-3-8

/CC: @Pierre-Sassoulas

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,8 @@ Release date: TBA
 * Fix issues with ``typing_extensions.TypeVar``.
 
 
+* Fix ``ClassDef.fromlino`` for PyPy 3.8 (v7.3.11) if class is wrapped by a decorator.
+
 
 What's New in astroid 2.13.3?
 =============================
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -22,7 +22,7 @@ from typing import TYPE_CHECKING, ClassVar, NoReturn, TypeVar, overload
 from astroid import bases
 from astroid import decorators as decorators_mod
 from astroid import util
-from astroid.const import IS_PYPY, PY38, PY38_PLUS, PY39_PLUS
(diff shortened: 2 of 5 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2139,9 +2139,10 @@ def _newstyle_impl(self, context: InferenceContext | None = None):
     @cached_property
     def fromlineno(self) -> int | None:
         """The first line that this node appears on in the source code."""
-        if not PY38_PLUS or PY38 and IS_PYPY:
+        if not PY38_PLUS or IS_PYPY and PY38 and not PYPY_7_3_11_PLUS:
```
cdce8p, line 2142: Do we want to support multiple PyPy versions? I guess it's simple enough here, although I wouldn't add any more to CI.

ChangeLog
```diff
@@ -15,6 +15,8 @@ Release date: TBA
 * Fix issues with ``typing_extensions.TypeVar``.
 
 
+* Fix ``ClassDef.fromlino`` for PyPy 3.8 (v7.3.11) if class is wrapped by a decorator.
```
cdce8p, line 18: I can move this to `2.14.0` if we want to skip `2.13.4`.

Pierre-Sassoulas: Let's not skip, releasing is cheap (compared to fixing the milestones)

astroid/nodes/scoped_nodes/scoped_nodes.py
Pierre-Sassoulas: This feel like a lot of work but also added values. I guess on a case by case basis... ?

cdce8p: > This feel like a lot of work but also added values. I guess on a case by case basis... ?

Yeah, I was more referring to this specific case 😄

Pierre-Sassoulas (approved):

PR #1981, opened 2023-01-28
Fix wrong inference cache result for `typing.cast` calls inside functions

cdce8p:
## Description
The result for the `typing.cast` inference tip depend could depend on the arguments passed to the parent function.
Those aren't included in the cache key which could cause the first (invalid) inference result to be returned again.

This PR adds a new parameter to `inference_tip` -> `cached`, to be able to opt out of caching for specific inference tips.
In most cases this shouldn't be needed. Therefore the default is set to `True`.

https://github.com/PyCQA/astroid/blob/dfd88f5edc636df80c8cabd61a6b8d6bc8746ca9/astroid/inference_tip.py#L36-L38

```py
from typing import TypeVar, cast
T = TypeVar("T")
def ident(var: T) -> T:
    return cast(T, var)

ident(2)  #@
ident("Hello")  #@
```

Closes https://github.com/PyCQA/pylint/issues/8074

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,8 +14,11 @@ Release date: TBA
 
 * Fix issues with ``typing_extensions.TypeVar``.
 
-
 * Fix ``ClassDef.fromlino`` for PyPy 3.8 (v7.3.11) if class is wrapped by a decorator.
+
+* Fix wrong inference cache result for ``typing.cast`` calls inside functions.
+
+  Closes PyCQA/pylint#8074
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):

cdce8p:
Found a potentially better solution. Just need to do some performance testing first.

cdce8p:
Closing in favor of #1982

PR #1982, opened 2023-01-28
Preserve parent `CallContext` when inferring nested functions

cdce8p:
## Description
Alternative to #1981
At the moment `infer_call` creates a new `CallContext` for each `infer_call_result` call overwriting the current one. Result was that nested functions with dynamic values couldn't be inferred.

Instead of overwriting the `CallContext`, keep it in `parent_call_context` so it can be reused later.
With that nested functions can now be inferred. The custom inference tip for `typing.cast` is thus no longer necessary.

https://github.com/PyCQA/astroid/blob/dfd88f5edc636df80c8cabd61a6b8d6bc8746ca9/astroid/inference.py#L274-L278

```py
def f(x):
    return g(x)

def g(y):
    return y + 2

f(1)  #@
```

Closes https://github.com/PyCQA/pylint/issues/8074

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,8 +14,11 @@ Release date: TBA
 
 * Fix issues with ``typing_extensions.TypeVar``.
 
-
 * Fix ``ClassDef.fromlino`` for PyPy 3.8 (v7.3.11) if class is wrapped by a decorator.
+
+* Preserve parent CallContext when inferring nested functions.
+
+  Closes PyCQA/pylint#8074
 
 
 What's New in astroid 2.13.3?
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -28,7 +28,6 @@ from astroid.nodes.node_classes import (
     Const,
     JoinedStr,
     Name,
-    NodeNG,
     Subscript,
     Tuple,
 )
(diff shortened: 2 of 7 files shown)


Pierre-Sassoulas (approved):
Removing code and fixing bug at the same time ! You're the bane of bad productivity metrics, Marc ;)

cdce8p:
> Removing code and fixing bug at the same time ! You're the bane of bad productivity metrics, Marc ;)

Thanks 😄 Took me literally the whole day yesterday for such an easy fix. I find the constant recursion in astroid quite difficult to debug but maybe that's just me.

cdce8p:
@Pierre-Sassoulas What do you think about the `2.13.4` release? It could be a good time after the backport is merged. Shortly after that, we could do `2.14.0` with just #1977, it's basically ready. That way could update the astroid version in pylint for the next release.

DanielNoord:
Wonderful fix @cdce8p! Glad to see you opening some PRs again 😄

cdce8p:
> Glad to see you opening some PRs again 😄

Only for a short while, unfortunately. Will be quite busy the next month.

Pierre-Sassoulas:
I'm definitely going to release 2.13.4 tomorrow.

PR #1995, opened 2023-01-31
remove module builtins

clavedeluna:
## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/3327

Issue mentioned removing the assumptions that all modules have builtins solves the false positive, so testing it out to see if something breaks.

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -163,7 +163,6 @@ class ObjectModel:
 
 
 class ModuleModel(ObjectModel):
-
     @property
     def attr___path__(self):
         if not self._instance.package:
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -163,13 +163,6 @@ class ObjectModel:
 
 
 class ModuleModel(ObjectModel):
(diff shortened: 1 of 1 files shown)


clavedeluna:
I've ran the pylint test suite locally with python 3.8.9 and did receive this error / crash of the test run:

```
tests/test_functional.py::test_functional[regression_5408] Abort trap: 6
```

which isn't great. I also recognize my running on only one python version is probably insufficient. Maintainers, how do you usually handle this type of testing to make sure we don't merge astroid PRs that would break pylint? Is there a way to run the pylint repo actions by specifying to use a particular astroid branch? (something I've personally setup before for testing)

Pierre-Sassoulas:
I install astroid locally with an editable install and run the test suite for pylint with my current interpreter (then add the pylint-tested label on the astroid side)

clavedeluna:
@Pierre-Sassoulas that's exactly what I did!

Any tips on how to debug the crash for that one test?

Pierre-Sassoulas:
Sorry I'm not really familiar with the problem you're facing.

DanielNoord:
Closing as I don't think this is really feasible and actively worked on anymore.

PR #1998, opened 2023-01-31
Release 2.14.0

cdce8p:
_Added an additional changelog entry for #1978_


cdce8p:
@Pierre-Sassoulas The release are all done, label and maintenance branch created. Will update pylint later.

One other thing, I would suggest we remove the old `backport maintenance/2.xx.x` labels. In the end, they are just confusing and don't add much value. I've added a new label https://github.com/PyCQA/astroid/labels/backported which we can add to PRs once the backport is done. But aside from that, if the original PR does have a milestone attached, I think the backport label isn't necessary anymore.

Pierre-Sassoulas:
Make sense especially as we set the milestone too.

PR #1999, opened 2023-01-31
Numpy reshape shape to use *

### LINKED ISSUE #7883: False positive when using ndarray.reshape with separate arguments for each element of the shape
### Bug description

```
# pylint: disable=missing-docstring
x = np.zeros(12)
x.reshape(3, 2, 2, order='F')
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module src.a
a.py:4:0: E1121: Too many positional arguments for method call (too-many-function-args)
a.py:4:0: E1124: Argument 'order' passed by position and keyword in method call (redundant-keyword-arg)

```

### Expected behavior

no error reported, the syntax is correct as documented [here](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.reshape.html#numpy-ndarray-reshape)

### Pylint version

```shell
pylint 2.14.5
astroid 2.11.7
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:33) 
[Clang 13.0.1 ]
```

### OS / Environment

macOS 12.6

### Additional dependencies

numpy @ file:///private/var/folders/.../croots/recipe/numpy_and_numpy_base_1652801688716/work

clavedeluna:
## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/7883

Pylint issue demonstrates a FP. I think this is the right solution, I also tested with this code and made sure pylint no longer showed the messages.

```
import numpy as np
x = np.zeros(12)
x.reshape(3, 2, 2)
x.reshape((3, 2))
x.reshape(3, 2, 2, order='F')
x.reshape(3, 2, 2, order='F')
x.reshape((3, 2, 2), order='F')

a = np.arange(6).reshape((3, 2))

x.reshape(a, order='F')

np.reshape(a, (2, 3))

np.reshape(a, 6, order='F')
```

### DIFF
--- a/astroid/brain/brain_numpy_ndarray.py
+++ b/astroid/brain/brain_numpy_ndarray.py
@@ -121,7 +121,7 @@ def infer_numpy_ndarray(node, context: InferenceContext | None = None):
         def put(self, indices, values, mode='raise'): return None
         def ravel(self, order='C'): return np.ndarray([0, 0])
         def repeat(self, repeats, axis=None): return np.ndarray([0, 0])
-        def reshape(self, shape, order='C'): return np.ndarray([0, 0])
+        def reshape(self, *shape, order='C'): return np.ndarray([0, 0])
         def resize(self, new_shape, refcheck=True): return None
         def round(self, decimals=0, out=None): return np.ndarray([0, 0])
         def searchsorted(self, v, side='left', sorter=None): return np.ndarray([0, 0])
--- a/tests/brain/numpy/test_ndarray.py
+++ b/tests/brain/numpy/test_ndarray.py
@@ -150,6 +150,22 @@ class NumpyBrainNdarrayTest(unittest.TestCase):
                     msg=f"Illicit type for {attr_:s} ({inferred_values[-1].pytype()})",
                 )
 
+    def test_numpy_ndarray_reshape_accepts_variadic_args(self):
+        """Test that ndarray.reshape can be called with variadic positional args."""
+        code = """
+        import numpy as np
+        x = np.zeros(12)
+        x.reshape(3, 2, 2)  #@
+        x.reshape((3, 2))  #@
+        x.reshape(3, 2, 2, order='F')  #@
+        x.reshape((3, 2, 2), order='F')  #@
+        """
+        for node in builder.extract_node(code):
+            with self.subTest(node=node.as_string()):
+                inferred_values = list(node.infer())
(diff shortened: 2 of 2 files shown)


clavedeluna:
I'm not really sure a unit test is needed here but if it does please let me know and hit to where it should be.

astroid/brain/brain_numpy_ndarray.py
```diff
@@ -120,7 +120,7 @@ def ptp(self, axis=None, out=None): return np.ndarray([0, 0])
         def put(self, indices, values, mode='raise'): return None
         def ravel(self, order='C'): return np.ndarray([0, 0])
         def repeat(self, repeats, axis=None): return np.ndarray([0, 0])
-        def reshape(self, shape, order='C'): return np.ndarray([0, 0])
+        def reshape(self, *shape, order='C'): return np.ndarray([0, 0])
```
DanielNoord, line 123: https://github.com/numpy/numpy/blob/9d7dca4596d7aeb8f31a261592a33d3b38b93b5f/numpy/__init__.pyi#L1767-L1774

The actual code is a bit different.. I'm not sure how to express this as it feels like we might run into issues with the other form as well..

Pierre-Sassoulas (changes requested):
We install numpy during tests now, so it's possible to add automated tests for this (the example in the description at least).

astroid/brain/brain_numpy_ndarray.py
clavedeluna: not sure if I need to address this? not sure what the ask

Pierre-Sassoulas: I think Daniel's point is that there's multiple possible signature, and we only handle one here ?

Pierre-Sassoulas (commented):
Tests lgtm for what we have but I'm not sure it handles every possible combination from the overload

clavedeluna:
@Pierre-Sassoulas I'm going to be away and won't get back to this so feel free to update it and merge

astroid/brain/brain_numpy_ndarray.py
DanielNoord: Yes, we still need to handle this difference in this PR.

DanielNoord:
Closing for now, we can always reopen later :)

See also https://github.com/pylint-dev/astroid/pull/2710 for our poor `numpy` support.

PR #2005, opened 2023-02-01
Add some type annotations

nickdrozd:
## Type of Changes

## Description

Add more type annotations, and cut out some unnecessary docstring lines.

Running Mypy strict, this goes from 1479 errors to 1446.

### DIFF
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -213,7 +213,7 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
     def blockstart_tolineno(self):
         return self.lineno
 
-    def _elsed_block_range(self, lineno, orelse, last=None):
+    def _elsed_block_range(self, lineno, orelse, last=None) -> tuple[int, int]:
         """Handle block line numbers range for try/finally, for, if and while
         statements.
         """
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -11,7 +11,7 @@ import itertools
 import sys
 import typing
 import warnings
-from collections.abc import Generator, Iterable, Mapping
+from collections.abc import Generator, Iterable, Mapping, Iterator
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Optional, TypeVar, Union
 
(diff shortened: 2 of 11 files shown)


astroid/nodes/_base_nodes.py
```diff
@@ -213,7 +213,7 @@ class MultiLineWithElseBlockNode(MultiLineBlockNode):
     def blockstart_tolineno(self):
         return self.lineno
 
-    def _elsed_block_range(self, lineno, orelse, last=None):
+    def _elsed_block_range(self, lineno, orelse, last=None) -> tuple[int, int]:
```
cdce8p, line 216:
```suggestion
    def _elsed_block_range(self, lineno, orelse, last=None) -> tuple[int, int | None]:
```

Noticed a few times were you omitted the optional `None`. Please take a look at the changes again. Will comment on one or two, but quite likely I'll miss some.

astroid/nodes/node_classes.py
```diff
@@ -304,11 +304,10 @@ def postinit(self, elts: list[NodeNG]) -> None:
         self.elts = elts
 
     @classmethod
-    def from_elements(cls, elts=None):
+    def from_elements(cls, elts: list[NodeNG] = None):
```
cdce8p, line 307:
```suggestion
    def from_elements(cls, elts: list[NodeNG] | None = None):
```

astroid/nodes/node_classes.py
```diff
@@ -793,11 +788,7 @@ def arguments(self):
         return list(itertools.chain((self.posonlyargs or ()), self.args or ()))
 
     def format_args(self, *, skippable_names: set[str] | None = None) -> str:
-        """Get the arguments formatted as string.
-
-        :returns: The formatted arguments.
```
cdce8p, line 798: Please keep the docstring. Again, this is only one on many instances in this PR. Please take a look at it again. _It's fine to remove just the type from the docstring if it doesn't add any more value._

cdce8p (changes requested):
Thanks for doing the work, but tbh I don't really like this kind of changes.

It kind of feels like doing something without a clear objective. The changes aren't connected so it's much more difficult to check if they are correct. We can still improve the type annotations, but not like this. Just replacing all docstring annotations isn't helpful and can actually make typing it correctly later more difficult.

The approach I would suggest is to focus on a smaller area at a time. E.g. only `get_children` calls or line number annotations (`blockstart_tolineno`, `block_range`, ...).

It's probably easier to focus on pylint if you want to do something...

PR #2024, opened 2023-02-12
Fix '_infer_str_format_call' crash when the string it analyses are uninferable

Pierre-Sassoulas:
## Type of Changes

## Description

Closes PyCQA/pylint#8109

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,9 @@ What's New in astroid 2.14.2?
 =============================
 Release date: TBA
 
+* '_infer_str_format_call' won't crash anymore when the string it analyses are uninferable.
+
+  Closes PyCQA/pylint#8109
 
 
 What's New in astroid 2.14.1?
--- a/tests/brain/test_dataclasses.py
+++ b/tests/brain/test_dataclasses.py
@@ -1350,23 +1350,3 @@ def test_dataclass_with_properties() -> None:
     fourth_init: bases.UnboundMethod = next(fourth.infer())
     assert [a.name for a in fourth_init.args.args] == ["self", "other_attr", "attr"]
     assert [a.name for a in fourth_init.args.defaults] == ["Uninferable"]
-
-
-def test_dataclass_pylint_8109() -> None:
-    """https://github.com/PyCQA/pylint/issues/8109"""
-    function_def = astroid.extract_node(
-        """
-from dataclasses import dataclass
-
-@dataclass
-class Number:
-    amount: int | float
-    round: int = 2
-
-    def __str__(self): #@
-        number_format = "{:,.%sf}" % self.round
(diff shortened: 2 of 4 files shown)


ChangeLog
```diff
@@ -23,6 +23,9 @@ What's New in astroid 2.14.2?
 =============================
 Release date: TBA
 
+* '_infer_str_format_call' won't crash anymore when the string it analyses are uninferable.
```
DanielNoord, line 26: Let's just release `2.15` if we want to. There are fixes in there that are worth bringing out sooner than later imo.

tests/brain/test_dataclasses.py
```diff
@@ -1350,3 +1350,23 @@ def attr(self, value: int) -> None:
     fourth_init: bases.UnboundMethod = next(fourth.infer())
     assert [a.name for a in fourth_init.args.args] == ["self", "other_attr", "attr"]
     assert [a.name for a in fourth_init.args.defaults] == ["Uninferable"]
+
+
+def test_dataclass_pylint_8109() -> None:
```
DanielNoord, line 1355: Can you move this to the other tests for `format`? This is not really a `dataclass` issue right?

ChangeLog
Pierre-Sassoulas: There's the group exception for 3.11 that need to be done before releasing. I'm planning to try to do that next.

tests/brain/test_dataclasses.py
Pierre-Sassoulas: Yeah I don't think so but reproducing was not easy with only the format.

Pierre-Sassoulas: Well in fact it's (partly) a dataclass issue, the following do not raise anything with the old code:

```python
class Number:
    amount: int | float
    round: int = 2

    def __str__(self): #@
        number_format = "{:,.%sf}" % self.round
        return number_format.format(self.amount).rstrip("0").rstrip(".")
```

Pierre-Sassoulas: (I tried to reproduce by adding thing instead of the saner "removing things until it does not crash)

DanielNoord: Yeah, but it doesn't feel like this should be in the dataclasses test. As in, this is clearly related to our support of `format` in various scenarios instead of our support for `dataclasses`. If I read the `format` tests I would expect this test to be there as well.

Pierre-Sassoulas: [b1fd5d7](https://github.com/PyCQA/astroid/pull/2024/commits/b1fd5d71dd3aebd3110d7f67b7c3251ce136dfd8)

DanielNoord (approved):

PR #2028, opened 2023-02-12
Add Trystar nodes from python 3.11

### LINKED ISSUE #1516: TODO list for Python 3.11 support

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #1516

This is a work in progress, I never did that before. I'm creating the class on the model given in the other class although I instinctively really wanted to not use postinit and recursively rebuild nodes inside the class itself instead. Righ tnow it seems postinit do not populate the handlers property. Pushing so we can work collaboratively on it as this is one big piece of #1516

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4216,6 +4216,107 @@ class TryFinally(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
         yield from self.finalbody
 
 
+class TryStar(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
+    """Class representing an :class:`ast.TryStar` node."""
+
+    _astroid_fields = ("body", "handlers", "orelse", "finalbody")
+    _multi_line_block_fields = ("body", "handlers", "orelse", "finalbody")
+
+    def __init__(
+        self,
+        *,
+        lineno: int | None = None,
+        col_offset: int | None = None,
+        end_lineno: int | None = None,
+        end_col_offset: int | None = None,
+        parent: NodeNG | None = None,
+    ) -> None:
+        """
+        :param lineno: The line that this node appears on in the source code.
+        :param col_offset: The column that this node appears on in the
+            source code.
+        :param parent: The parent node in the syntax tree.
+        :param end_lineno: The last line this node appears on in the source code.
+        :param end_col_offset: The end column this node appears on in the
+            source code. Note: This is after the last symbol.
+        """
+        self.body: list[NodeNG] = []
+        """The contents of the block to catch exceptions from."""
+
+        self.handlers: list[ExceptHandler] = []
+        """The exception handlers."""
+
+        self.orelse: list[NodeNG] = []
+        """The contents of the ``else`` block."""
+
+        self.finalbody: list[NodeNG] = []
+        """The contents of the ``finally`` block."""
+
+        super().__init__(
+            lineno=lineno,
+            col_offset=col_offset,
+            end_lineno=end_lineno,
+            end_col_offset=end_col_offset,
+            parent=parent,
+        )
+
+    def postinit(
+        self,
+        *,
+        body: list[NodeNG] | None = None,
+        handlers: list[ExceptHandler] | None = None,
+        orelse: list[NodeNG] | None = None,
+        finalbody: list[NodeNG] | None = None,
+    ) -> None:
+        """Do some setup after initialisation.
+        :param body: The contents of the block to catch exceptions from.
+        :param handlers: The exception handlers.
+        :param orelse: The contents of the ``else`` block.
+        :param finalbody: The contents of the ``finally`` block.
+        """
+        if body:
+            self.body = body
+        if handlers:
+            self.handlers = handlers
+        if orelse:
+            self.orelse = orelse
+        if finalbody:
+            self.finalbody = finalbody
+
+    def _infer_name(self, frame, name):
+        return name
+
+    def block_range(self, lineno: int) -> tuple[int, int]:
+        """Get a range from a given line number to where this node ends."""
+        if lineno == self.fromlineno:
+            return lineno, lineno
+        if self.body and self.body[0].fromlineno <= lineno <= self.body[-1].tolineno:
+            # Inside try body - return from lineno till end of try body
+            return lineno, self.body[-1].tolineno
+        for exhandler in self.handlers:
+            if exhandler.type and lineno == exhandler.type.fromlineno:
+                return lineno, lineno
+            if exhandler.body[0].fromlineno <= lineno <= exhandler.body[-1].tolineno:
+                return lineno, exhandler.body[-1].tolineno
+        if self.orelse:
+            if self.orelse[0].fromlineno - 1 == lineno:
+                return lineno, lineno
+            if self.orelse[0].fromlineno <= lineno <= self.orelse[-1].tolineno:
+                return lineno, self.orelse[-1].tolineno
+        if self.finalbody:
+            if self.finalbody[0].fromlineno - 1 == lineno:
+                return lineno, lineno
+            if self.finalbody[0].fromlineno <= lineno <= self.finalbody[-1].tolineno:
+                return lineno, self.finalbody[-1].tolineno
+        return lineno, self.tolineno
+
+    def get_children(self):
+        yield from self.body
+        yield from self.handlers
+        yield from self.orelse
+        yield from self.finalbody
+
+
 class Tuple(BaseContainer):
     """Class representing an :class:`ast.Tuple` node.
 
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -1822,6 +1822,22 @@ class TreeRebuilder:
             return self.visit_tryexcept(node, parent)
         return None
 
+    def visit_trystar(self, node: ast.TryStar, parent: NodeNG) -> nodes.TryStar:
+        newnode = nodes.TryStar(
+            lineno=node.lineno,
+            col_offset=node.col_offset,
+            end_lineno=getattr(node, "end_lineno", None),
+            end_col_offset=getattr(node, "end_col_offset", None),
+            parent=parent,
+        )
+        newnode.postinit(
+            body=[self.visit(n, newnode) for n in node.body],
+            handlers=[self.visit(n, newnode) for n in node.handlers],
+            orelse=[self.visit(n, newnode) for n in node.orelse],
+            finalbody=[self.visit(n, newnode) for n in node.finalbody],
+        )
+        return newnode
+
     def visit_tuple(self, node: ast.Tuple, parent: NodeNG) -> nodes.Tuple:
         """Visit a Tuple node by returning a fresh instance of it."""
         context = self._get_context(node)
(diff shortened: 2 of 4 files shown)


DanielNoord:
What do you need help on with this PR?

Pierre-Sassoulas:
Right now I don't understand why the postinit is never called and handlers never populated, but I did not have enough time to retro-engineer things yet.

Pierre-Sassoulas:
There's anoother tasks for 3.11 though :)

DanielNoord:
> Right now I don't understand why the postinit is never called and handlers never populated, but I did not have enough time to retro-engineer things yet.

`postinit` needs to be called manually. It's just a convention, not something that is done automatically.

astroid/rebuilder.py
```diff
@@ -1822,6 +1822,15 @@ def visit_try(
...
         return None
 
+    def visit_trystar(self, node: ast.TryStar, parent: NodeNG) -> nodes.TryStar:
+        return nodes.TryStar(
+            lineno=node.lineno,
+            col_offset=node.col_offset,
+            end_lineno=getattr(node, "end_lineno", None),
+            end_col_offset=getattr(node, "end_col_offset", None),
+            parent=parent,
+        )
```
jacobtylerwalls, line 1832: Judging from `visit_try()` directly above this, we will need to do something very similar to populate the body, handlers, and finally.

jacobtylerwalls (approved):

Pierre-Sassoulas:
Thank you for fixing everything and unblocking the path to 2.17.0 @jacobtylerwalls :D I'm going to add more tests for inference and iterating on parents.

astroid/nodes/node_classes.py
```diff
@@ -4216,6 +4216,107 @@ def get_children(self):
...
+            self.body = body
+        if handlers:
+            self.handlers = handlers
+        if orelse:
+            self.orelse = orelse
+        if finalbody:
+            self.finalbody = finalbody
+
+    def _infer_name(self, frame, name):
+        return name
```
Pierre-Sassoulas, line 4287: Not sure how to cover this, TryStar are not supposed to be inferred. It's called only in ``infer_stmt``. What do you think @jacobtylerwalls ?

jacobtylerwalls: I think this isn't very material; I see that TryFinally doesn't even have an override. You'd have to create an inference context, give it a lookup name, and then assert that it's preserved after evaluating infer_stmt.

```python
>>> import astroid
>>> trystar = astroid.extract_node("""
... try:
...   1/0
... except* ValueError:
...   pass""")
>>> context = astroid.context.InferenceContext()
>>> context.lookupname = 'arbitraryName'
>>> iter = astroid.bases._infer_stmts([trystar], context)
>>> list(iter)
[Uninferable]
>>> context.lookupname
'arbitraryName'
```

jacobtylerwalls: Not a bad idea to add it, though.

Pierre-Sassoulas: I took what you suggested word for word.

Pierre-Sassoulas:
Can't test with the full test suite in pylint because of ``DeprecationWarning: is_standard_module() is deprecated. Use, is_stdlib_module() or module_in_path() instead`` but the crash is gone as expected. I say let's release astroid 2.15.0 and pylint 2.17.0 :)

astroid/nodes/node_classes.py
```diff
@@ -4216,6 +4216,107 @@ def get_children(self):
...
+class TryStar(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
+    """Class representing an :class:`ast.TryStar` node."""
+
+    _astroid_fields = ("body", "handlers", "orelse", "finalbody")
+    _multi_line_block_fields = ("body", "handlers", "orelse", "finalbody")
+
+    def __init__(
+        self,
+        *,
+        lineno: int | None = None,
```
DanielNoord, line 4228: I'm a little late to the party, but this feels incorrect. I don't think these should ever be `None` right?

astroid/nodes/node_classes.py
```diff
@@ -4216,6 +4216,107 @@ def get_children(self):
...
+        """
+        :param lineno: The line that this node appears on in the source code.
+        :param col_offset: The column that this node appears on in the
+            source code.
+        :param parent: The parent node in the syntax tree.
+        :param end_lineno: The last line this node appears on in the source code.
+        :param end_col_offset: The end column this node appears on in the
+            source code. Note: This is after the last symbol.
+        """
+        self.body: list[NodeNG] = []
```
DanielNoord, line 4243: This feel like it should be `SuccesfulInferenceResult` maybe?

astroid/nodes/node_classes.py
```diff
@@ -4216,6 +4216,107 @@ def get_children(self):
...
+    _multi_line_block_fields = ("body", "handlers", "orelse", "finalbody")
+
+    def __init__(
+        self,
+        *,
+        lineno: int | None = None,
+        col_offset: int | None = None,
+        end_lineno: int | None = None,
+        end_col_offset: int | None = None,
+        parent: NodeNG | None = None,
```
DanielNoord, line 4232: Shouldn't this be `LocalsDictNodeNG`?

DanielNoord (commented):
I'm late to the party, but I think the typing of the arguments is a bit too permissive. Would be good to narrow those down in a follow up so we don't make this part of the public API before releasing.

astroid/nodes/node_classes.py
```diff
@@ -4216,6 +4216,107 @@ def get_children(self):
...
+            col_offset=col_offset,
+            end_lineno=end_lineno,
+            end_col_offset=end_col_offset,
+            parent=parent,
+        )
+
+    def postinit(
+        self,
+        *,
+        body: list[NodeNG] | None = None,
```
DanielNoord, line 4266: All of these should never be None:
https://github.com/python/typeshed/blob/9472dc10b98570633e8777accf339dcec0179da4/stdlib/_ast.pyi#L178

jacobtylerwalls: But why would we bother to make `TryStar` the only exception? Wouldn't we just batch them all up with `with`, etc.?

DanielNoord: I don't really understand your comment, but I think it's better to be more restrictive where possible since we don't clearly define our public API here. There is just more options to deal with later if we ever want to refactor this.

I do like that these are keyword only!

DanielNoord: According to the typing `orelse` and `finalbody` should also only be `Statement` and not `NodeNG` I think, but that's another issue 😓

jacobtylerwalls: I'm just pointing out that the current typing in this PR matches the typing for many other nodes such as `TryExcept`, `With`, `If`, `For`, etc., and doesn't significantly increase technical debt if we're going to tighten those all in the future. (In fact, it would seem weird to me to consciously fix only _one_ at this time.)

DanielNoord: Hm, I think it would have been easier to tighten them now since we're "investigating" the node anyway and make a small comment about the typing being correct somewhere.
The issue is that right now we can't know for sure that any of the typing in `astroid` is correct. Marking `TryStart` as "reviewed and actually correct" would have been a good first step I think.

jacobtylerwalls: > @[DanielNoord](https://github.com/DanielNoord) DanielNoord [4 minutes ago](https://github.com/PyCQA/astroid/pull/2028#discussion_r1125745466)
According to the typing orelse and finalbody should also only be Statement and not NodeNG I think, but that's another issue 😓

Interesting. We should decide if we're moving forward with #1867 before delving into that, I would think. And honestly I'd still prefer fixing all affected nodes in one fell swoop. Better release story esp. if we have to tinker with constructors to not allow None, etc., which is more impactful breakage.

DanielNoord: Fair enough!

jacobtylerwalls:
> Can't test with the full test suite in pylint because of DeprecationWarning: is_standard_module() is deprecated. Use, is_stdlib_module() or module_in_path() instead

I'm doing that now to double-check things. It's as simple as:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 1c8a915ac..f3972f44e 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ test = "pytest"
 testpaths = ["tests"]
 python_files = ["*test_*.py"]
 addopts = "--strict-markers"
-filterwarnings = "error"
+filterwarnings = "ignore::DeprecationWarning"
 markers = [
     "primer_stdlib: Checks for crashes and errors when running pylint on stdlib",
     "primer_external_batch_one: Checks for crashes and errors when running pylint on external libs (batch one)",
```

AI0867:
Astroid incorrectly infers the type of the exceptions caught by TryStar ExceptHandler:
```
try:
    raise ExceptionGroup("group", [TypeError("type")])
except* TypeError as not_a_type_error:
    assert isinstance(not_a_type_error, ExceptionGroup)
```

In fact:
```
try:
    raise ExceptionGroup("group1", [TypeError("1"), ExceptionGroup("group2", [TypeError("2"), TypeError("3")]), TypeError("4")])
except* TypeError as not_a_type_error:
    assert isinstance(not_a_type_error.exceptions[1], ExceptionGroup)
```

In `except* ExceptionType as name`, name will contain a value of type `ExceptionGroup[UnionException]`, where `UnionException = Union[ExceptionGroup[UnionException], ExceptionType]`.

jacobtylerwalls:
> Astroid incorrectly infers the type of the exceptions caught by TryStar ExceptHandler:
> 
> ```
> 
> try:
> 
>     raise ExceptionGroup("group", [TypeError("type")])
> 
> except* TypeError as not_a_type_error:
> 
>     assert isinstance(not_a_type_error, ExceptionGroup)
> 
> ```
> 
> 
> 
> In fact:
> 
> ```
> 
> try:
> 
>     raise ExceptionGroup("group1", [TypeError("1"), ExceptionGroup("group2", [TypeError("2"), TypeError("3")]), TypeError("4")])
> 
> except* TypeError as not_a_type_error:
> 
>     assert isinstance(not_a_type_error.exceptions[1], ExceptionGroup)
> 
> ```
> 
> 
> 
> In `except* ExceptionType as name`, name will contain a value of type `ExceptionGroup[UnionException]`, where `UnionException = Union[ExceptionGroup[UnionException], ExceptionType]`.

Thanks, can you open an issue and state the current result in addition to the expected result?

PR #2029, opened 2023-02-12
Update ``FormattedValue.postinit`` and its brain

DanielNoord:
## Type of Changes

## Description

Refs https://github.com/PyCQA/astroid/issues/1516.

Decided to make this a breaking change as it just so much easier to update signatures when these are keyword only. I don't think a small breaking change at the top of the signature matters too much.

Using an `Enum` doesn't make a lot of sense as the `ast` module has it as an `int`. So we would need to of some casting to that `Enum` and that could be the source of nasty crashes,

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@ What's New in astroid 2.15.0?
 =============================
 Release date: TBA
 
+* ``Formattedvalue.postinit`` is now keyword only. This is to allow correct typing of the
+  ``Formattedvalue`` class.
+
+  Refs #1516
+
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
We should probably cover what's not currently covered ?

DanielNoord:
Shall we do that in a follow up? Functionally nothing changes here, other than `conversion` now being passed correctly in the brain were it was previously just dropped. This whole brain seems to be untested and I don't want to block this PR on me having time to look at the testing of the brain.

Pierre-Sassoulas (approved):

PR #2033, opened 2023-02-20
Update submodule_path after finding an editable install

noah-weingarden:
## Type of Changes

## Description

#1752 enabled `astroid` to find modules that use an editable install with import hooks, but it didn't enable finding *submodules* of a package that uses such an editable install. This PR updates `submodule_path` when necessary to fix that. All modules in the path after the first one should use `ImportlibFinder` as a result of this change. See [this comment](https://github.com/PyCQA/pylint/issues/7306#issuecomment-1435899922) for a more detailed description and a minimal reproducible example.

I didn't update the changelog because the bug this is fixing is tied to a different bugfix that hasn't been released yet, but let me know if I should add one.

Refs https://github.com/PyCQA/pylint/issues/7306


DanielNoord:
Thanks for this @noah-weingarden!

Do you think it will be possible to add a test for this?

noah-weingarden:
Hmm, you mentioned in https://github.com/PyCQA/astroid/pull/1752 that you don't want to introduce a dependency on setuptools, so that definitely makes it harder. Could we test it by mocking `sys.meta_path` to have an `_EditableFinder` class? I'm just worried that a test like that would be brittle if setuptools changes how this works on their end.

DanielNoord:
> Hmm, you mentioned in #1752 that you don't want to introduce a dependency on setuptools, so that definitely makes it harder. Could we test it by mocking `sys.meta_path` to have an `_EditableFinder` class? I'm just worried that a test like that would be brittle if setuptools changes how this works on their end.
> 
> To be honest though, I'm pretty confused in general about which aspects of the Python packaging world are stable. #1752 seems to depend on there being an `_EditableFinder` class in `sys.meta_path` which has a `find_spec` method, so if any of those things changes, `astroid` would probably need an update anyway, right? In that case, mocking `sys.meta_path` seems like it would make sense for now, but would fail to automatically reveal those hypothetical changes.

I have you are willing to put in the in the time to create the tests that mocks `meta_path` I think it would still be worthwhile. It is much easier to change such a test when `setuptools` changes their behaviour than being completely in the dark about what we expect `setuptools` to do.

noah-weingarden:
Alrighty, I made a test! If you remove the change I made to `astroid/interpreter/_import/spec.py`, the test will pass the first assertion and fail the second one because the latter imports a subpackage.

(Side-note: To get this to work locally, I had to make a separate import for `unittest.mock`. For some reason, just importing `unittest` as the test module currently does results in `AttributeError: module 'unittest' has no attribute 'mock'`. I didn't have time to look into why, but it might be some pytest magic?)

astroid/interpreter/_import/spec.py
```diff
@@ -453,8 +453,13 @@ def find_spec(modpath: list[str], path: Sequence[str] | None = None) -> ModuleSp
...
         )
         processed.append(modname)
-        if modpath and isinstance(finder, Finder):
-            submodule_path = finder.contribute_to_path(spec, processed)
+        if modpath:
+            if isinstance(finder, Finder):
+                submodule_path = finder.contribute_to_path(spec, processed)
+            # If modname is a package from an editable install, update submodule_path
+            # so that the next module in the path will be found inside of it using importlib.
+            elif finder.__name__ in {"_EditableFinder", "_EditableNamespaceFinder"}:
```
DanielNoord, line 461: I'm wondering if we should extract this logic into a class, but for now I think this is fine!

tests/testdata/python3/data/import_setuptools_pep660/__editable___example_0_1_0_finder.py
```diff
@@ -0,0 +1,68 @@
+import sys
```
DanielNoord, line 1: I guess this is copied from `setuptools`. Shall we provide a link in a module docstring?

astroid/interpreter/_import/spec.py
noah-weingarden: I agree. Do you think the logic will apply to all of the keys in _MetaPathFinderModuleTypes? (I'm not familiar with six). If so, an improvement for now would be to use that.

tests/testdata/python3/data/import_setuptools_pep660/__editable___example_0_1_0_finder.py
noah-weingarden: That's a good idea. They don't have seem to have any documentation on it though. Should we link to the file in setuptools that generates this code?

astroid/interpreter/_import/spec.py
DanielNoord: When I implemented this @jacobtylerwalls actually (rightly) stressed that we should probably be careful here: we rather want to have less support for installers we don't know than making assumptions that don't turn out to be true.

We can always add support for new ones if we encounter them.

The import system is apparently not that well defined..

tests/testdata/python3/data/import_setuptools_pep660/__editable___example_0_1_0_finder.py
noah-weingarden: https://github.com/pypa/setuptools/blob/main/setuptools/command/editable_wheel.py

DanielNoord: Yeah, preferably with a commit hash in the docstring as well.

Everything to make the `astroid` maintainer 10 years down the line what we were trying to achieve here :smile:

astroid/interpreter/_import/spec.py
noah-weingarden: Sounds good. I left the set alone, but moved it to be a top-level variable instead of treating it as a constant. Any new classes can be added to that set.

tests/testdata/python3/data/import_setuptools_pep660/__editable___example_0_1_0_finder.py
noah-weingarden: Done!

DanielNoord:
@noah-weingarden Could you rebase this?

jacobtylerwalls (approved):
Looks good!

PR #2050, opened 2023-03-12
Don't cache `decoratornames()`

jacobtylerwalls:
## Type of Changes

## Description
This method shouldn't be cached like a property, as it takes an argument.

See:
https://github.com/PyCQA/astroid/blob/91fd7b95fc943676f5aafb2d0ee881c4e167fa04/astroid/decorators.py#L32-L33


DanielNoord (approved):

jacobtylerwalls:
Any ideas about the regex brain failure @DanielNoord? Shall we just xfail it for now?

DanielNoord:
Perhaps? I don't know, it seems like `regex` has changed its implementation and we don't deal well with it. I don't really have time to investigate, but it should be fixable.

Issue #1780, opened 2022-09-13
Astroid is leaking memory


matusvalo:
Astroid is leaking memory when is run multiple times over same file using API

### Steps to reproduce

1. Pick larger file, e.g. file from pylint project: pylint/checkers/variables.py
2. Run following script:
```python
from pylint.lint import Run
from pylint.lint import pylinter
while True:
    Run(["pylint/checkers/variables.py"], exit=False)
    pylinter.MANAGER.clear_cache()
```

### Current behavior

After ~12 iteration memory goes over 1 GB and keeps growing
![image](https://user-images.githubusercontent.com/827060/189831617-615bd61d-7305-4171-9b79-5ad6d12cbfcc.png)

### Expected behavior

Allocated memory should not rise indefinitelly

### `python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"` output

2.13.0-dev0

matusvalo:
I did some investigation - I profiled memory using muppy:

```python
from pylint.lint import Run
from pylint.lint import pylinter
from pympler import muppy
from pympler import summary

Run(["pylint/checkers/variables.py"], exit=False)
pylinter.MANAGER.clear_cache()

all_objects = muppy.get_objects()
sum1 = summary.summarize(all_objects)
summary.print_(sum1)

for _ in range(15):
    Run(["pylint/checkers/variables.py"], exit=False)
    pylinter.MANAGER.clear_cache()

all_objects = muppy.get_objects()
sum1 = summary.summarize(all_objects)
summary.print_(sum1)
```

I got following result:

```
                                  types |   # objects |   total size
======================================= | =========== | ============
                                   dict |      221794 |     60.35 MB
                                   list |      212895 |     14.44 MB
                                    str |       67663 |      5.82 MB
                                    int |      185941 |      4.97 MB
                                  tuple |       52200 |      2.78 MB
        astroid.nodes.node_classes.Name |       39949 |      1.83 MB
       astroid.nodes.node_classes.Const |       36755 |      1.68 MB
                                   code |        8153 |      1.56 MB
                     tokenize.TokenInfo |       16652 |      1.40 MB
                                   type |        1311 |      1.18 MB
                                 method |       13220 |    826.25 KB
   astroid.nodes.node_classes.Attribute |       12100 |    567.19 KB
  astroid.nodes.node_classes.AssignName |       11793 |    552.80 KB
        astroid.nodes.node_classes.Call |        9234 |    432.84 KB
[...]

PR #2051, opened 2023-03-12
Change `_get_assign_nodes()` to `cached_property`

jacobtylerwalls:
## Type of Changes

## Description
`cached_property` does a much better job cleaning up after itself than we're doing.

Refs #1780


DanielNoord (commented):
Pehaps out of scope, but this seems to indicate that this returns `list[nodes.Assign]`? Perhaps nice to add as well?

Anyway, I feel like the name doesn't really do the property justice anymore. What about something like `_assign_nodes_in_scope`?

DanielNoord (approved):

PR #2057, opened 2023-03-13
Upgrade pre-commit configuration and move to ruff

Pierre-Sassoulas:
## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2926,7 +2926,7 @@ class ClassDef(
             if exc.args and exc.args[0] not in ("", None):
                 return exc.args[0]
             return None
-        return [first, *list(slots)]
+        return [first, *slots]
 
     # Cached, because inferring them all the time is expensive
     @decorators_mod.cached
@@ -3032,7 +3032,7 @@ class ClassDef(
                 ancestors = list(base.ancestors(context=context))
                 bases_mro.append(ancestors)
 
-        unmerged_mro = [[self], *bases_mro] + [inferred_bases]
+        unmerged_mro = [[self], *bases_mro, inferred_bases]
         unmerged_mro = list(clean_duplicates_mro(unmerged_mro, self, context))
         clean_typing_generic_mro(unmerged_mro)
         return _c3_merge(unmerged_mro, self, context)
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -19,9 +19,8 @@ from collections.abc import Generator, Iterator
 from functools import lru_cache
 from typing import TYPE_CHECKING, ClassVar, NoReturn, TypeVar, overload
 
-from astroid import bases
+from astroid import bases, util
 from astroid import decorators as decorators_mod
-from astroid import util
 from astroid.const import IS_PYPY, PY38, PY38_PLUS, PY39_PLUS, PYPY_7_3_11_PLUS
 from astroid.context import (
     CallContext,
@@ -2927,7 +2926,7 @@ class ClassDef(
             if exc.args and exc.args[0] not in ("", None):
                 return exc.args[0]
             return None
-        return [first] + list(slots)
+        return [first, *list(slots)]
 
     # Cached, because inferring them all the time is expensive
     @decorators_mod.cached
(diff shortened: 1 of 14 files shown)


doc/conf.py
```diff
@@ -62,7 +62,7 @@
 # built documents.
 #
 # The short X.Y version.
-from astroid.__pkginfo__ import __version__
+from astroid.__pkginfo__ import __version__  # noqa
```
DanielNoord, line 65: Why is this necessary?

pyproject.toml
```diff
@@ -65,14 +65,6 @@ python_files = ["*test_*.py"]
 testpaths = ["tests"]
 filterwarnings = "error"
```
DanielNoord, line 67: Is the `isort` config no longer neede?

requirements_test_pre_commit.txt
```diff
@@ -1,9 +1,5 @@
 -r requirements_test_min.txt
 black==23.1.0
```
DanielNoord, line 2: I'd be okay with removing this as well

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2927,7 +2926,7 @@ def _slots(self):
             if exc.args and exc.args[0] not in ("", None):
                 return exc.args[0]
             return None
-        return [first] + list(slots)
+        return [first, *list(slots)]
```
Pierre-Sassoulas, line 2929:
```suggestion
        return [first, *slots]
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -3033,7 +3032,7 @@ def _compute_mro(self, context: InferenceContext | None = None):
                 ancestors = list(base.ancestors(context=context))
                 bases_mro.append(ancestors)
 
-        unmerged_mro = [[self]] + bases_mro + [inferred_bases]
+        unmerged_mro = [[self], *bases_mro] + [inferred_bases]
```
Pierre-Sassoulas, line 3035:
```suggestion
        unmerged_mro = [[self], *bases_mro, inferred_bases]
```

pyproject.toml
Pierre-Sassoulas: I don't think ruff is using it, it's compatible with black by default.

requirements_test_pre_commit.txt
Pierre-Sassoulas: Removing black or removing the version ?

tests/test_regrtest.py
```diff
@@ -264,7 +264,7 @@ def test_unicode_in_docstring(self) -> None:
 
         class MyClass(object):
             def method(self):
-                "With unicode : {'’'} "
+                "With unicode : {'`'} "
```
Pierre-Sassoulas, line 267: This looks suspicious, having strange unicode characters is probably what we want to check.

pyproject.toml
DanielNoord: But do we get import sorting still then?

requirements_test_pre_commit.txt
DanielNoord: versioN!

pyproject.toml
Pierre-Sassoulas: We might not have intra group sorting with first / third parties, which is why isort is kept in pylint. I think in astroid we'll have "good enough sorting" because we don't have a lot of third parties -- and faster commits.

Pierre-Sassoulas: Nevermind, I just tested, ruff sort correctly first party and third party without any configuration.

Pierre-Sassoulas: We might want to use ruff instead of isort for wrong-import-order in pylint.

requirements_test_pre_commit.txt
```diff
@@ -1,9 +1,5 @@
...
-isort==5.12.0;python_version>='3.8'
-flake8==6.0.0;python_version>='3.8'
-flake8-typing-imports==1.14.0;python_version>='3.8'
-flake8-bugbear==23.2.13;python_version>='3.8'
-mypy==1.1.1
-pre-commit~=2.21
+black
+pylint
+mypy
+pre-commit
```
Pierre-Sassoulas, line 5:
```suggestion
ruff
pre-commit
```

DanielNoord (approved):

Issue #2060, opened 2023-03-15
Make astroid typable possibily with breaking change to permit most performance increase possibility


Pierre-Sassoulas:
astroid is very dynamic and very hard to type. A lot of effort went into typing it, and mypy is still not activated (#1287).  It prevents automated error detection in pylint and it prevent most performance possibility that always start with proper typing, be it mypyc, a partial rewrite in a rust crate (#2014) or cython.

I'm opening this discussion so we can discuss the breaking changes we need to do to make it easier to type astroid in the future.

nickdrozd:
Many node types have fields that are marked as optional when really they are required. These fields are assumed to be non-optional throughout the Astroid code, and that's a major source of type errors detected by Mypy.

For example, consider the `Assert` node. It has these fields:

```python
        self.test: NodeNG | None = None
        """The test that passes or fails the assertion."""

        self.fail: NodeNG | None = None  # can be None
        """The message shown when the assertion fails."""
```

An `assert` statement without a test is not valid syntax, so the `test` field needs to be required. On the other hand, the fail message really is optional, so the `fail` field really is optional.

A symptom of the problem is the frequent appearance of `# can be None` comments. These show that the existing type annotations are not reliable.

Now, the reason AFAICT that these fields are all marked as optional is that the type annotations are declared in `__init__` but the values for those fields are not available until `postinit` is called. There is an easy solution to this: declare the types outside of `__init__`. This has the happy side effect of making it possible to get rid of `__init__` definitions altogether for many nodes, since all they do besides declaring types is calling `super().__init__`.

nickdrozd:
https://github.com/PyCQA/astroid/pull/2061 is a concrete example of the kind of change I'm talking about.

AWhetter:
> Now, the reason AFAICT that these fields are all marked as optional is that the type annotations are declared in `__init__` but the values for those fields are not available until `postinit` is called.

This is exactly why. When I was first defining the types of everything all the way back when I added docstrings to everything, I had to make everything optional because of how `__init__` and `postinit` worked.

> There is an easy solution to this: declare the types outside of `__init__`. This has the happy side effect of making it possible to get rid of `__init__` definitions altogether for many nodes, since all they do besides declaring types is calling `super().__init__`.

We'd risk the potential for some of the attributes to be accidentally left unset by the `postinit` if we took this approach. If we think that's fine, then this seems like the easier solution.

[...]

PR #2061, opened 2023-03-16
Mandatory fields for Assign

nickdrozd:
## Type of Changes

## Description

This PR re-types the required fields of the `Assign` node so that they are non-optional, as described in https://github.com/PyCQA/astroid/issues/2060.

Two existing type errors are thereby fixed, as reported by Mypy. Here is one of them:

```python
    @cached_property
    def _assign_nodes_in_scope(self) -> list[nodes.Assign]:
        return [self, *self.value._assign_nodes_in_scope]
```

If `Assign.value` can be `None`, then this could fail. But if the field is required, there is no problem.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1202,62 +1202,25 @@ class Assign(_base_nodes.AssignTypeNode, _base_nodes.Statement):
     <Assign l.1 at 0x7effe1db8550>
     """
 
+    targets: list[NodeNG]
+    """What is being assigned to."""
+
+    value: NodeNG
+    """The value being assigned to the variables."""
+
+    type_annotation: NodeNG | None
+    """If present, this will contain the type annotation passed by a type comment"""
+
     _astroid_fields = ("targets", "value")
     _other_other_fields = ("type_annotation",)
 
-    def __init__(
-        self,
-        lineno: int | None = None,
-        col_offset: int | None = None,
-        parent: NodeNG | None = None,
-        *,
-        end_lineno: int | None = None,
-        end_col_offset: int | None = None,
-    ) -> None:
-        """
-        :param lineno: The line that this node appears on in the source code.
-
-        :param col_offset: The column that this node appears on in the
-            source code.
-
-        :param parent: The parent node in the syntax tree.
-
-        :param end_lineno: The last line this node appears on in the source code.
-
-        :param end_col_offset: The end column this node appears on in the
-            source code. Note: This is after the last symbol.
-        """
-        self.targets: list[NodeNG] = []
-        """What is being assigned to."""
-
-        self.value: NodeNG | None = None
-        """The value being assigned to the variables."""
-
-        self.type_annotation: NodeNG | None = None  # can be None
-        """If present, this will contain the type annotation passed by a type comment"""
-
-        super().__init__(
-            lineno=lineno,
-            col_offset=col_offset,
-            end_lineno=end_lineno,
-            end_col_offset=end_col_offset,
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
Shouldn't we change the constructor for ``_base_nodes.AssignTypeNode, _base_nodes.Statement`` too ?

DanielNoord:
@Pierre-Sassoulas Are we okay with merging breaking changes into `main`? I'd be in favour, but want to double check.

Pierre-Sassoulas:
Let's consider that if it's not breaking pylint then it's not a change that require a major version bump even if technically it's a breaking change. Especially if it's typing, as we never advertised astroid as typed.

nickdrozd:
> Shouldn't we change the constructor for `_base_nodes.AssignTypeNode, _base_nodes.Statement` too ?

It looks like those don't have their own constructors. They all just delegate to `NodeNG`.

Pierre-Sassoulas (commented):
I'd rather have the overridden constructor requiring 'targets' and remove the post init function altogether. But for that to work we would need to refactor the tree rebuilder. I suppose if we instantiate the leafs first and then recurse instead of what we're currently doing it's possible (?). What do you think @DanielNoord ?

DanielNoord:
> I'd rather have the overridden constructor requiring 'targets' and remove the post init function altogether. But for that to work we would need to refactor the tree rebuilder. I suppose if we instantiate the leafs first and then recurse instead of what we're currently doing it's possible (?). What do you think @DanielNoord ?

I mean, it's worth a shot but it would be a massive rewrite.

I would be okay with reviewing a complete rewrite and contributing but don't have the time to do to brunt of the work..

astroid/nodes/node_classes.py
```diff
@@ -1202,62 +1202,25 @@ class Assign(_base_nodes.AssignTypeNode, _base_nodes.Statement):
...
-            lineno=lineno,
-            col_offset=col_offset,
-            end_lineno=end_lineno,
-            end_col_offset=end_col_offset,
-            parent=parent,
-        )
+    _astroid_fields = ("targets", "value")
+    _other_other_fields = ("type_annotation",)
 
     def postinit(
```
DanielNoord, line 1217: https://github.com/pylint-dev/astroid/blob/345b91cd6d29318ac34352699ff75a0132ccda15/astroid/rebuilder.py#L808

Could you add asserts here that the typing we add here is correct?

And could you add a changelog entry? For the rest this LGTM!

Pierre-Sassoulas (approved):
I don't see myself doing a massive rewrite (when there are a lot of massive rewrite we could do) so let's merge, it's an upgrade in the current state of the code.

DanielNoord:
> I don't see myself doing a massive rewrite (when there are a lot of massive rewrite we could do) so let's merge, it's an upgrade in the current state of the code.

Agreed! But let's at least make sure that all the callsites of these functions, so everywhere we do `Assign(...` are in line with what we merge. Otherwise we will start wondering in 3 months time if we forget something and the new typing is incorrect as well :sweat_smile:

nickdrozd:
What about the rest of the node types? Many of the rest need a similar fix. (Some require more involved changes.) Should that be done here, or separate PRs? One PR per node type, or all at once?

DanielNoord:
Let's do separate PRs! I think I can check these fairly quickly but focusing on one node at a time makes it much easier. You can tag me for review (although please don't open 10 at once 😄 )

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #2062, opened 2023-03-17
Add typing to ``TransformVisitor``

DanielNoord:
## Type of Changes

## Description

I don't like it... But at least the functionality of `TransformVisitor` is now correctly exposed through its typing.

### DIFF
--- a/astroid/transforms.py
+++ b/astroid/transforms.py
@@ -4,13 +4,34 @@
 
 from __future__ import annotations
 
-import collections
-from typing import TYPE_CHECKING
+from collections import defaultdict
+from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


DanielNoord:
This does add typing to `TransformVisitor`, which is the main goal as it is sort of public. The rest of it is hidden behind custom logic.

Pierre-Sassoulas (commented):
Ok. Is there a performance impact using the intermediary transform dict instead of the default data structure ? I suppose there would be as it's an additional call in python instead of an optimized builtin data structure.

DanielNoord:
I have changed it to no longer use the intermediary dict. It just gave a lot of hassle with minimal gain. People should only ever need the `(un)register` methods anyway, and those are correctly typed and show expected behaviour.

This is ready for review again.

PR #2066, opened 2023-03-26
create `NodeNG.inferred_best()` and replace `inferred()[0]`

### LINKED ISSUE #2065: Improve `inferred` interface
### Steps to reproduce

1. Consider following code:
```pytohn
node = extract_node(
        """
if hasattr(typing, 'some_future_function'):
    future_function = typing.some_future_function
else:
    def future_function(arg: int) -> bool:
        return False
future_function
    """
)
inferred_any = node.inferred()[0]
assert inferred_any is not util.Uninferable
```

### Current behavior
This code fail, because it `inferred` function return inferred values in order of its occurrence in analyzed code.

### Expected behavior
A `NodeNG` interface should have a method to take only one best inferred value.

### Considered changes
1. If function `inferred` find any non `Uninferable` value, then remove all `Uninferable` from returning list:
    - probably many tests will fail,
    - maybe sometimes we want to know about uncertainty inferred value,
2. In function `inferred` before returning, we can sort values with most "valuable" values at the beginning:
    - some tests fail, because previously there wasn't such assumption about order,
3. Add another method `def inferred_best(self) -> InferenceResult` to `NodeNG`, where it returns single inferred value, possibly avoid Uninferable. 

I think the 3rd option is the best and I prepared some code, but before creating pull request I want to know others opinions.

### Rationale
It is important for other functionality (especially in PyCQA/pylint#7565 - where inferred value is considered as `Uninferable` despite having another better value).
`inferred()[0]` is a frequent code (about 60 occurrences in astroid and pylint), so it seems reasonable to replace it with `inferred_best()` - I tested such a change locally, and it seems ok.

5.16.0dev0

ostr00000:
## Type of Changes

## Description
Add `inferred_best` function to `NodeNG` interface.
Replace all code where `inferred()[0]` is used.

Closes #2065

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -623,8 +623,7 @@ class NodeNG:
         :rtype: InferenceResult
         """
         return sorted(
-            self.infer(),
-            key=lambda inf: isinstance(inf, util.UninferableBase)
+            self.infer(), key=lambda inf: isinstance(inf, util.UninferableBase)
         )[0]
 
     def instantiate_class(self):
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -612,6 +612,21 @@ class NodeNG:
         """
         return list(self.infer())
 
+    def inferred_best(self) -> InferenceResult:
+        """Return single inferred value, possibly avoid Uninferable.
+
+        Try to return value of type other than :class:`UninferableBase`.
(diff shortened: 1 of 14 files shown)


ostr00000:
I am closing this, since code with `inferred()[0]` should be refactored another way (`Ideally we would always check all results of infer() or use safe_infer to make sure we are using the "full" result and not [0] or [-1]`).
See #2065 for more details.

PR #2077, opened 2023-04-03
Fix constructors of ``AssignName``

DanielNoord:
## Type of Changes

## Description

Just going down the list. We might also just consider these to be breaking changes and just release `3.x` whenever we have done all nodes. Nobody is keeping us from releasing `4.x` soon after if we find additional issues.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,9 +11,9 @@ Release date: TBA
   Closes #1780
 
 * Improved signature of the ``__init__`` and ``__postinit__`` methods of the following nodes:
+  - ``nodes.AssignName``
   - ``nodes.Delete``
   - ``nodes.If``
-  - ``nodes.AssignName``
   - ``nodes.IfExp``
 
   These changes involve breaking changes to their API but should be considered bug fixes. We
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (dismissed):

Pierre-Sassoulas (approved):

ChangeLog
```diff
@@ -13,6 +13,7 @@ Release date: TBA
 * Improved signature of the ``__init__`` and ``__postinit__`` methods of the following nodes:
   - ``nodes.Delete``
   - ``nodes.If``
+  - ``nodes.AssignName``
```
DanielNoord, line 16:
```suggestion
  - ``nodes.AssignName``
  - ``nodes.Delete``
  - ``nodes.If``
```

Pierre-Sassoulas:
Let's not profusely sweet over this, 2.16 is okay, let's keep more radical changes to be considered as breaking changes. Something like breaking the API to get rid of cyclic-import for example.

ChangeLog
Pierre-Sassoulas: This was intentional, if all the addition are at the top we're going to get a conflict for each MR.

PR #2080, opened 2023-04-03
Don't add typing stubs for packages that we only use in tests

DanielNoord:
## Type of Changes

## Description

This just adds more dependencies while we don't even really use these packages. Let's just ignore the import errors and make our `mypy` config easier and more maintainable.


Pierre-Sassoulas (approved):
Fair enough

PR #2086, opened 2023-04-03
Fix various typing issues and the constructors of ``Keyword``

DanielNoord:
## Type of Changes

## Description

Best reviewed commit per commit. I saw some easy to fix issues in the `mypy` output so I jumped on them while also adding `Keyword`.


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1127,7 +1127,7 @@ def postinit(self, args: Arguments, body):
         self.args = args
         self.body = body
 
-    def pytype(self) -> Literal["bultins.instancemethod", "builtins.function"]:
+    def pytype(self) -> Literal["builtins.instancemethod", "builtins.function"]:
```
Pierre-Sassoulas, line 1130: 🤨 All right, we really need to activate mypy 😄

Pierre-Sassoulas (approved):

PR #2088, opened 2023-04-03
Fix some line numbering related methods

DanielNoord:
## Type of Changes

## Description

``mypy`` was screaming at me for how we used these. We heavily depend on these always returning an `int` so we should handle the `return 0` case wherever needed instead of having it optionally return `None`.
If you want to check if it is really `None` you should just use `node.lineno`.

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -491,15 +491,13 @@ class NodeNG:
                 parent = parent.parent
         return line or 0
 
-    def block_range(self, lineno):
+    def block_range(self, lineno: int) -> tuple[int, int]:
         """Get a range from the given line number to where this node ends.
 
         :param lineno: The line number to start the range at.
-        :type lineno: int
 
         :returns: The range of line numbers that this node belongs to,
             starting at the given line number.
-        :rtype: tuple(int, int or None)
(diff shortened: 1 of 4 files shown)


astroid/nodes/node_ng.py
```diff
@@ -446,15 +446,21 @@ def previous_sibling(self):
     # single node, and they rarely get looked at
 
     @cached_property
-    def fromlineno(self) -> int | None:
-        """The first line that this node appears on in the source code."""
+    def fromlineno(self) -> int:
+        """The first line that this node appears on in the source code.
```
Pierre-Sassoulas, line 450: Maybe the docstring only need to be made explicit once in the parent class ?

Pierre-Sassoulas (approved):
LGTM, with a possibility of nit.

astroid/nodes/node_ng.py
DanielNoord: I don't know, I don't think all IDEs follow `docstrings` successfully but I don't really mind. I'll let you decide.

Pierre-Sassoulas: The path of least resistance is merging as is, let's do that

PR #2094, opened 2023-04-04
Add return type to get_children

nickdrozd:
## Type of Changes

## Description

Add return annotations for `get_children` throughout. This reduces the number of warnings with strict Mypy, but it actually increases the number of non-strict warnings due to exposing other errors.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -11,7 +11,7 @@ import itertools
 import sys
 import typing
 import warnings
-from collections.abc import Generator, Iterable, Mapping
+from collections.abc import Generator, Iterable, Mapping, Iterator
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Optional, TypeVar, Union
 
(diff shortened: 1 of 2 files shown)


DanielNoord (changes requested):
This should be `SuccessfulInferenceResult`. See `get_children` of `nodes.Dict`.

DanielNoord (commented):
Oh and please remove the `rtype` comments for any signature you fix

nickdrozd:
Is there some way to get `Proxy` to "count" as a `NodeNG`? That would solve a quite a few existing type errors.

nickdrozd:
Run Mypy and search the output for `Item "NodeNG" of "Union[NodeNG, Proxy]" has no attribute`.

DanielNoord:
> Is there some way to get `Proxy` to "count" as a `NodeNG`? That would solve a quite a few existing type errors.

Yes, by using ``SuccessfulInferenceResult`` :smile:  Sadly there is no better way. `NodeNG` is simply incorrect.

nickdrozd:
I guess my question was partly rhetorical, since I am sure there must be a better way :smile: But I don't know what that better way is exactly.

A `Proxy` is just a stand-in for a real `NodeNG`, right? Or is it something else? The docstring is not very helpful: `A simple proxy object.`

But if it represents a real node, and can be called with all the same methods, then it ought to look like a real node to the type checker as well. `SuccessfulInferenceResult` appears to be a type alternative -- either this type or that type -- when really it's just a node or a node-alike, and those are practically the same thing.

At least that's my understanding of the situation.

PR #2095, opened 2023-04-04
Mandatory fields for While

nickdrozd:
## Type of Changes

## Description

Mandatory fields for `While`. Fixes one strict Mypy warning.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4011,12 +4011,12 @@ class While(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
     def postinit(
         self,
         test: NodeNG,
-        body: list[NodeNG] | None = None,
-        orelse: list[NodeNG] | None = None,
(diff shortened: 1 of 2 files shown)


astroid/nodes/node_classes.py
```diff
@@ -3999,64 +3999,24 @@ class While(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
...
-        )
+    orelse: list[NodeNG]
+    """The contents of the ``else`` block."""
 
     def postinit(
         self,
-        test: NodeNG | None = None,
+        test: NodeNG,
         body: list[NodeNG] | None = None,
         orelse: list[NodeNG] | None = None,
```
DanielNoord, line 4015: Please remove the defaults here as well. We are just making everything mandatory from now on as that is how `astroid` uses this internally and ideally how anybody else should also be using this.

DanielNoord (approved):

PR #2098, opened 2023-04-04
Check for call context in `PartialFunction.infer_call_result`

nickdrozd:
## Type of Changes

|

## Description

Check a context's call context. Fixes four Mypy errors.

### DIFF
--- a/astroid/objects.py
+++ b/astroid/objects.py
@@ -300,7 +300,9 @@ class PartialFunction(scoped_nodes.FunctionDef):
 
     def infer_call_result(self, caller=None, context: InferenceContext | None = None):
         if context:
-            assert context.callcontext, "CallContext should be set before inferring call result"
+            assert (
+                context.callcontext
+            ), "CallContext should be set before inferring call result"
             current_passed_keywords = {
                 keyword for (keyword, _) in context.callcontext.keywords
             }
--- a/astroid/objects.py
+++ b/astroid/objects.py
@@ -300,7 +300,7 @@ class PartialFunction(scoped_nodes.FunctionDef):
 
     def infer_call_result(self, caller=None, context: InferenceContext | None = None):
         if context:
-            assert context.callcontext
+            assert context.callcontext, "CallContext should be set before inferring call result"
             current_passed_keywords = {
                 keyword for (keyword, _) in context.callcontext.keywords
             }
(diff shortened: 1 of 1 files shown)


DanielNoord (commented):
Shouldn't this be an `assert`? We're inferring a `call` so `callcontext` should probably be set

DanielNoord:
I don't know, but it makes no sense to change the logic if we are inadvertently introducing bugs with it. This changes the behaviour of the function and feels a bit like moving [Chesterton's Fence](https://thoughtbot.com/blog/chestertons-fence).
I think a deepdive into `infer_call_result` would be really worthwhile but it should probably start from the top rather than doing individual changes in single definitions of it.

astroid/objects.py
```diff
@@ -300,6 +300,7 @@ def __init__(
 
     def infer_call_result(self, caller=None, context: InferenceContext | None = None):
         if context:
+            assert context.callcontext
```
DanielNoord, line 303:
```suggestion
            assert context.callcontext, "CallContext should be set before inferring call result"
```

DanielNoord (approved):

DanielNoord:
I have just cleaned up the signature of `infer_call_result`. However, if not had any insight into whether `callcontext` is always set.

DanielNoord:
I don't know. Asserting might very well be correct! Or we might want a subclass where the attribute has always been set?

jacobtylerwalls:
I think this change makes sense for now. Whether CallContext must be part of the context object depends on whether or not the call is to a property. Here, this PR is only altering `PartialFunction.infer_call_result`, in a clear case where the code would have crashed if the assumption CallContext was always set was violated.

Here is where the CallContext is usually set if you are going through `infer_call`:

https://github.com/pylint-dev/astroid/blob/92fe8294690a3f83eaadaa63c3ec5cbd5ca8c90a/astroid/inference.py#L291-L295

Let's just get the mypy errors fixed 👍

jacobtylerwalls (approved):

DanielNoord:
> Let's just get the mypy errors fixed +1

Although I heavily agree with this I'm also wary of adding technical debt by adding asserts and code that we don't fully understand. In this case it might be fine but I do think we should be re-evaluating on a case by case basis and not just add them everywhere.

jacobtylerwalls:
Agree! (I walked myself through the infer_call_result patterns before approving 📚 .)

PR #2099, opened 2023-04-04
Mandatory fields for Comprehension

nickdrozd:
## Type of Changes

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1524,6 +1524,18 @@ class Comprehension(NodeNG):
     'for x in some_values'
     """
 
+    target: NodeNG
+    """What is assigned to by the comprehension."""
+
+    iter: NodeNG
+    """What is iterated over by the comprehension."""
+
+    ifs: list[NodeNG]
+    """The contents of any if statements that filter the comprehension."""
+
+    is_async: bool
+    """Whether this is an asynchronous comprehension or not."""
+
     _astroid_fields = ("target", "iter", "ifs")
     _other_fields = ("is_async",)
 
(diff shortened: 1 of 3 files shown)


astroid/rebuilder.py
```diff
@@ -1009,7 +1009,7 @@ def visit_comprehension(
         self, node: ast.comprehension, parent: NodeNG
     ) -> nodes.Comprehension:
         """Visit a Comprehension node by returning a fresh instance of it."""
-        newnode = nodes.Comprehension(parent)
+        newnode = nodes.Comprehension(parent=parent)
```
DanielNoord, line 1012: Can you update this to also `provide` `lineno` etc? It's available on the `node`. No reason not to pass it.

astroid/nodes/node_classes.py
```diff
@@ -1524,58 +1524,34 @@ class Comprehension(NodeNG):
...
+    """The contents of any if statements that filter the comprehension."""
 
-        self.ifs: list[NodeNG] = []
-        """The contents of any if statements that filter the comprehension."""
+    is_async: bool
+    """Whether this is an asynchronous comprehension or not."""
 
-        self.is_async: bool | None = None
-        """Whether this is an asynchronous comprehension or not."""
+    _astroid_fields = ("target", "iter", "ifs")
```
DanielNoord, line 1539: Please keep these at the top. It makes it easier to `git blame` these lines to an actual change to them.

astroid/rebuilder.py
nickdrozd: Apparently it doesn't? https://docs.python.org/3/library/ast.html#abstract-grammar

DanielNoord: Oh that's weird. I shouldn't have trusted my IDE which said it should 😓 

I'm wondering: what do you think of still providing `None` as values? That way we can at least remove the default values from `NodeNG` at some point.

astroid/nodes/node_classes.py
```diff
@@ -1527,55 +1527,31 @@ class Comprehension(NodeNG):
...
+    ifs: list[NodeNG]
+    """The contents of any if statements that filter the comprehension."""
 
-        self.is_async: bool | None = None
-        """Whether this is an asynchronous comprehension or not."""
+    is_async: bool
+    """Whether this is an asynchronous comprehension or not."""
 
-        super().__init__(parent=parent)
+    optional_assign = True
```
DanielNoord, line 1542: Same comment as about `_astroid_fields`, no need to move this in this PR 😄

nickdrozd: Might as well keep it consistent.

astroid/rebuilder.py
nickdrozd: Sounds reasonable, although without a comment explaining why there are no line numbers I don't think it will be any clearer!

DanielNoord: Let's do that then!

DanielNoord (approved):

PR #2109, opened 2023-04-05
Clean up If fields

nickdrozd:
## Type of Changes

Refs #XXXX

Closes #XXXX

## Description

This brings the `If` constructor in line with the other nodes.

### DIFF
--- a/astroid/filter_statements.py
+++ b/astroid/filter_statements.py
@@ -150,14 +150,9 @@ def _filter_stmts(
                     _stmt_parents.append(stmt.parent)
                 # If the if statement is first-level and not within an orelse block
                 # we know that it will be evaluated
-                elif not if_parent.is_orelse:
+                else:
                     _stmts = [node]
                     _stmt_parents = [stmt.parent]
-                # Else we do not known enough about the control flow to be 100% certain
-                # and we append to possible statements
-                else:
-                    _stmts.append(node)
-                    _stmt_parents.append(stmt.parent)
             else:
                 _stmts = [node]
                 _stmt_parents = [stmt.parent]
--- a/astroid/filter_statements.py
+++ b/astroid/filter_statements.py
@@ -148,8 +148,7 @@ def _filter_stmts(
                     optional_assign = False
                     _stmts.append(node)
                     _stmt_parents.append(stmt.parent)
-                # If the if statement is first-level and not within an orelse block
-                # we know that it will be evaluated
+                # Else we assume that it will be evaluated
                 else:
                     _stmts = [node]
                     _stmt_parents = [stmt.parent]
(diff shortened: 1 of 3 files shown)


astroid/nodes/node_classes.py
```diff
@@ -2578,44 +2578,24 @@ class If(_base_nodes.MultiLineWithElseBlockNode, _base_nodes.Statement):
...
+    """The contents of the ``else`` block."""
 
     def postinit(self, test: NodeNG, body: list[NodeNG], orelse: list[NodeNG]) -> None:
         self.test = test
         self.body = body
         self.orelse = orelse
-        if isinstance(self.parent, If) and self in self.parent.orelse:
-            self.is_orelse = True
+
+    @cached_property
```
nickdrozd, line 2595: I don't like how the `is_orelse` field had a default value which was sometimes overwritten in `postinit`. So I changed it to have no default value and calculated the value in `postinit`. But then that raised some kind of circularity issue at build time. So I changed it to a `cached_property`, which is equivalent and I think ultimately more sensible.

DanielNoord: Is this condition ever True? I think I tested this and the condition never passed in our test suite. It might be redundant now.

nickdrozd: It's used in just one place that I'm aware of, in `_filter_stmts`:

```python
                # If the if statement is first-level and not within an orelse block
                # we know that it will be evaluated
                elif not if_parent.is_orelse:
                    _stmts = [node]
                    _stmt_parents = [stmt.parent]
```

nickdrozd: I added an exception to the else-block and it seems to never get reached:

```python
                # If the if statement is first-level and not within an orelse block
                # we know that it will be evaluated
                elif not if_parent.is_orelse:
                    _stmts = [node]
                    _stmt_parents = [stmt.parent]
                # Else we do not known enough about the control flow to be 100% certain
                # and we append to possible statements
                else:
                    raise Exception  # <-- never reached
                    _stmts.append(node)
                    _stmt_parents.append(stmt.parent)
```

So it seems the whole `is_orelse` field can be cut.

DanielNoord: Let's remove it then!

astroid/filter_statements.py
```diff
@@ -150,14 +150,9 @@ def _filter_stmts(
                     _stmt_parents.append(stmt.parent)
                 # If the if statement is first-level and not within an orelse block
                 # we know that it will be evaluated
-                elif not if_parent.is_orelse:
+                else:
```
DanielNoord, line 153: Could you update the comment here as well?

DanielNoord (commented):
Could you add this to the CHangelog and make a separate entry that this attribute has been removed as it serves no purpose and was never set correctly?

astroid/filter_statements.py
nickdrozd: I can delete it, but I don't have anything better to change it to.

DanielNoord (approved):

PR #2110, opened 2023-04-05
Assert parent is not None

nickdrozd:
## Type of Changes

## Description

Fixes one Mypy error.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -689,6 +689,7 @@ class Arguments(_base_nodes.AssignTypeNode):
         Can also return 0 if the line can not be determined.
         """
         lineno = super().fromlineno
+        assert self.parent is not None
         return max(lineno, self.parent.fromlineno or 0)
 
     @cached_property


DanielNoord (changes requested):
I think we should think about making this some sort of generic attribute of `NodeNG`. There are lots of nodes where we know that `parent` is not `None`. Not having to put asserts everywhere would be much better, but I haven't had the time to look into how to do this properly.

nickdrozd:
As far as assertions go, I think it's worth adding them if it helps getting Mypy running without errors.

DanielNoord:
> When is it possible for there to be no parent? Just for modules?

I think so, but also for nodes that are constructed out of thin air. We do that quite often.



> As far as assertions go, I think it's worth adding them if it helps getting Mypy running without errors.

I don't think so. Asserts change behaviour and can make `pylint` crash. We should only add them if we know they are true and can't find any other way to express what they are testing.

jacobtylerwalls:
> I think so, but also for nodes that are constructed out of thin air. We do that quite often.

The brains do this all over the place, see https://github.com/pylint-dev/astroid/issues/1490#issuecomment-1152619493.

PR #2111, opened 2023-04-06
Mandatory for Except types

nickdrozd:
## Type of Changes


DanielNoord (approved):
Thanks again for all this work @nickdrozd!

I know I can be quite strict in some of my reviews but that is mostly because `astroid` and `pylint` are so decoupled and I don't want to have to do a patch release of `astrodi` if we find we made a mistake.

PR #2112, opened 2023-04-06
Clean up Lambda fields

nickdrozd:
## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1049,6 +1049,12 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
     special_attributes = FunctionModel()
     """The names of special attributes that this function has."""
 
+    args: Arguments
+    """The arguments that the function takes."""
+
+    body: NodeNG
+    """The contents of the function body."""
+
     def implicit_parameters(self) -> Literal[0]:
         return 0
 
@@ -1065,42 +1071,15 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
 
     def __init__(
         self,
-        lineno=None,
-        col_offset=None,
-        parent=None,
+        lineno: int,
+        col_offset: int,
+        parent: NodeNG,
         *,
-        end_lineno=None,
-        end_col_offset=None,
+        end_lineno: int | None,
+        end_col_offset: int | None,
     ):
-        """
-        :param lineno: The line that this node appears on in the source code.
-        :type lineno: int or None
-
-        :param col_offset: The column that this node appears on in the
-            source code.
-        :type col_offset: int or None
-
-        :param parent: The parent node in the syntax tree.
-        :type parent: NodeNG or None
-
-        :param end_lineno: The last line this node appears on in the source code.
-        :type end_lineno: Optional[int]
-
-        :param end_col_offset: The end column this node appears on in the
-            source code. Note: This is after the last symbol.
-        :type end_col_offset: Optional[int]
-        """
         self.locals = {}
         """A map of the name of a local variable to the node defining it."""
-
-        self.args: Arguments
-        """The arguments that the function takes."""
-
-        self.body = []
-        """The contents of the function body.
-
-        :type: list(NodeNG)
-        """
 
         self.instance_attrs: dict[str, list[NodeNG]] = {}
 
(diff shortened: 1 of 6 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1049,6 +1049,24 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
     special_attributes = FunctionModel()
     """The names of special attributes that this function has."""
 
+    args: Arguments
+    """The arguments that the function takes."""
+
+    body: NodeNG
```
DanielNoord, line 1055: This is a change in behaviour and I think an inconsistency with `FunctionDef` which derives from this. `Lambda` seems to have a single node as a body whereas `FunctionDef` has a `list`. We might need to remove their coupling...

nickdrozd: I don't think this changes any behaviour, but yes, the `body` fields are inconsistent. Decoupling sounds like a good idea to me.

DanielNoord: Well, `mypy` will complain when we update the `FunctionDef` nodes right? So this change can't really be merged without doing the decoupling as well.

nickdrozd: That's just the tool telling us what we already know, which is that there are inconsistencies among the types. I don't think there's anything to be gained from leaving `body` unannotated in the two classes, especially since the typechecker warnings can be disabled.

DanielNoord: I think it better to leave them unannotated for now. Incorrect type annotations serve no purpose.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1213,22 +1213,17 @@ class FunctionDef(_base_nodes.MultiLineBlockNode, _base_nodes.Statement, Lambda)
...
-    """Whether this node indicates a function.
+    returns: NodeNG | None
 
-    For a :class:`FunctionDef` this is always ``True``.
+    decorators: node_classes.Decorators | None
+    """The decorators that are applied to this method or function."""
 
-    :type: bool
-    """
-    type_annotation = None
```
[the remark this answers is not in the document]
DanielNoord: Could you add a Changelog entry about this?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1350,12 +1300,12 @@ def __init__(
             frame = parent.frame(future=True)
             frame.set_local(name, self)
 
-    def postinit(
+    def postinit(  # type: ignore[override]
```
DanielNoord, line 1303: Why are we not also fixing up the `__init__`?

astroid/nodes/scoped_nodes/scoped_nodes.py
nickdrozd: Which one are you saying is incorrect? AFAICT, `Lambda.body` really is a single node, and `FunctionDef.body` really is a list.

astroid/nodes/scoped_nodes/scoped_nodes.py
nickdrozd: That will require a bit more work. `FunctionDef` isn't in great shape. It has deprecated args, lots unannotated fields, weird extra `__init__` args. Plus this `Lambda` inheritance problem.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Well, `Mypy` won't be able to understand `FunctionDef.body` or at least warn us about it. The typing makes no sense with both a single node and a list.

We should focus on the refactor first before changing these two I think.

nickdrozd: Nothing is being changed though. Those fields already have the values that they have, namely a single node for `Lambda.body` and a list of nodes for `FunctionDef.body`. That is is current situation on the ground, and these annotations reflect that.

Mypy doesn't like the inheritance, hence the `type: ignore[assignment]` comments. But other than that it can figure out what's going on. If something is changed to give `FunctionDef.body` something other than a list, Mypy will flag it, and likewise for giving `Lambda.body` a list. If anything, Mypy can't understand things as they are now, since it believes that the two `body` fields have the same type, which they don't.

DanielNoord:
I cleaned up this PR. @Pierre-Sassoulas would you mind reviewing it?

Pierre-Sassoulas (approved):
:thumbsup :

PR #2113, opened 2023-04-06
Clean up BaseContainer type fields

nickdrozd:
## Type of Changes

## Description

Refactor the `__init__` methods of `BaseContainer` and its descendants. Rather than have all the nodes deal with line number information, just do node-specific initialization and pass everything along to the next initializer. At the same time, centralize the `elts` initialization in `BaseContiner`, rather than having `List` and `Tuple` each do it on their own.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,12 +32,15 @@ Release date: TBA
   - ``nodes.If``
   - ``nodes.IfExp``
   - ``nodes.Keyword``
+  - ``nodes.List``
   - ``nodes.Name``
   - ``nodes.Raise``
   - ``nodes.Return
+  - ``nodes.Set
   - ``nodes.Slice``
   - ``nodes.Starred``
   - ``nodes.Subscript``
+  - ``nodes.Tuple``
   - ``nodes.UnaryOp``
   - ``nodes.While``
   - ``nodes.Yield``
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
I think this is a downgrade of the current code and don't really like the `args` and `kwargs`. Can we do `Set` or `Tuple` first and see if we can fix this while still using `keyword` arguments?

nickdrozd:
> I think this is a downgrade of the current code and don't really like the `args` and `kwargs`. Can we do `Set` or `Tuple` first and see if we can fix this while still using `keyword` arguments?

The thing that I don't like about the current system is that every node type has information about line numbers in its `__init__`. This leads to a proliferation of repeated text and makes it hard to figure out how nodes differ and hard to figure out which information is actually important and which is just repetitive noise.

Already `Set` doesn't have an `__init__` method, because it just passes along to `NodeNG`. `List` and `Tuple` can't do that because they have extra context arguments, but once those are set, they also just pass along to `NodeNG`. I don't see any reason for their `__init__` methods to contain any information about line numbers, since that information isn't used in the node.

DanielNoord:
@nickdrozd Do you plan on picking this up or should we close this PR?

PR #2115, opened 2023-04-11
Decouple ``FunctionDef`` and ``Lambda``

DanielNoord:
## Type of Changes

## Description

As discussed in https://github.com/pylint-dev/astroid/pull/2112 we really need to decouple this nodes. I think this warrants a 3.0 release for the next release. We don't need to do any of the other deprecations but this is just 100% necessary for `Mypy` but will also 100% break code of users.

Let me know what you think!

/CC @nickdrozd

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1244,7 +1244,12 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
         raise AttributeInferenceError(target=self, attribute=name)
 
 
-class FunctionDef(_base_nodes.MultiLineBlockNode, _base_nodes.Statement, Lambda):
+class FunctionDef(
+    _base_nodes.MultiLineBlockNode,
+    _base_nodes.FilterStmtsBaseNode,
+    _base_nodes.Statement,
+    LocalsDictNodeNG,
+):
     """Class representing an :class:`ast.FunctionDef`.
 
     >>> import astroid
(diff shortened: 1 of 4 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1454,6 +1479,62 @@ def extra_decorators(self) -> list[node_classes.Call]:
...
+
+    def display_type(self) -> str:
+        """A human readable type of this node.
+
+        :returns: The type of this node.
+        :rtype: str
+        """
+        if "method" in self.type:
+            return "Method"
+        return "Function"
```
Pierre-Sassoulas, line 1499: Should be trivial to cover (?)

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1454,6 +1479,62 @@ def extra_decorators(self) -> list[node_classes.Call]:
...
+        names += [elt.name for elt in self.args.kwonlyargs]
+        if self.args.kwarg:
+            names.append(self.args.kwarg)
+        return names
+
+    def getattr(
+        self, name: str, context: InferenceContext | None = None
+    ) -> list[NodeNG]:
+        if not name:
+            raise AttributeInferenceError(target=self, attribute=name, context=context)
```
Pierre-Sassoulas, line 1527: This line is not covered either.

Pierre-Sassoulas (changes requested):
Let's do the breaking change, but let's cover the new code if reasonable.

PR #2118, opened 2023-04-15
Fix infer_call_result() crash on methods called with_metaclass()

### LINKED ISSUE #1735: Code for supporting `with_metaclass` seems broken
So this code works as expected:
```python
node = astroid.extract_node('''
def foo():
    return 42
''')
print(list(node.infer_call_result(caller=node)))
```
```console
$ python a.py
[<Const.int l.3 at 0x105704070>]
```

But this breaks:
```python
node = astroid.extract_node('''
def with_metaclass(meta, *bases):
    return 42
''')
print(list(node.infer_call_result(caller=node)))
```

```console
$ python a.py
  File "/Users/tusharsadhwani/a.py", line 14, in <module>
    print(list(node.infer_call_result(node)))
  File "/Users/tusharsadhwani/venv/lib/python3.10/site-packages/astroid/nodes/scoped_nodes/scoped_nodes.py", line 1738, in infer_call_result
    metaclass = next(caller.args[0].infer(context), None)
TypeError: 'Arguments' object is not subscriptable
```
astroid seems to have special handling for functions called `with_metaclass`, which seems to be broken.

Am I using the method wrongly? I tried using it just as I saw it being used in the project itself.

jacobtylerwalls:
## Type of Changes

## Description

Closes #1735

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -61,6 +61,9 @@ What's New in astroid 2.15.3?
 =============================
 Release date: TBA
 
+* Fix ``infer_call_result()`` crash on methods called ``with_metaclass()``.
+
+  Closes #1735
 
 
 What's New in astroid 2.15.2?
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Nice !

Issue #2127, opened 2023-04-19
Resolving symlinks considered harmful


smurfix:
### Steps to reproduce

1. Get https://github.com/M-o-a-T/moat
2. git submodule update --init util micro
3. cd micro
4. pylint

### Current behavior

The problem which this archive exhibits is that `moat/micro/link.py` is a symlink that refers to `moat/micro/_embed/lib/moat/micro/link.py`. The directory `moat/micro` has an `_init__.py` file while `moat/micro/_embed/lib/moat/micro` does not.

Now, `modpath_from_file_with_callback`, in `astroid/mdutils.py`, calls astroid's `_get_relative_base_path` function which calls `realpath` and thus resolves this symlink. As a result, pylint's "is this a proper module" callback looks for the `__init__.py` file in the wrong place, decides that this is not a module, and refuses to process the import.

### Expected behavior

IMHO there's no good reason to call `realpath` on anything. Why not simply assume that the symlink is there for a good reason?

### `python -c "from astroid import __pkginfo__; print(__pkginfo__.version)"` output

2.14.2, but applies to master

Pierre-Sassoulas:
Did you search for the reason in the git history ? It feel like a Chesterton's fence (we should not remove it until we know why it's here in the first place)

smurfix:
Both absolute and real path lookups were added in cd76b495d

    modutils is using realpath at the moment which will resolve the symlink
    and don't allow the import as the file is detected outside of the
    available paths.

This change added two tests – which incidentally still work after removing that `realpath` call.

DanielNoord:
> Both absolute and real path lookups were added in [cd76b49](https://github.com/pylint-dev/astroid/commit/cd76b495d5e590138a90300d60cfcc5b4647cea5)
> 
> ```
> modutils is using realpath at the moment which will resolve the symlink
> and don't allow the import as the file is detected outside of the
> available paths.
> ```
> 
> This change added two tests – which incidentally still work after removing that `realpath` call.

As you can see from the diff we used `realpath` before it. I added a comment in your PR with the first introduction of `realpath`, which sadly doesn't include a test.

smurfix:
Ah, yes, sorry, the real reason `realpath` was added seems to be PyCQA/pylint#791

I'll ask there whether somebody can test this.

PR #2128, opened 2023-04-19
Don't resolve symlinks.

smurfix:
## Type of Changes

## Description

This patch removes code that resolves symlinks when resolving a module path.

This causes problems, as described in the linked-to bug, and IMHO since the Python interpreter doesn't resolve symlinks there's no good reason why a code analyzer should do so.

Closes: #2127.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,10 @@ What's New in astroid 2.16.0?
 =============================
 Release date: TBA
 
+* No longer resolve symlinks to modules.
+
+  Closes #2127
+
 * Remove ``@cached`` decorator (just use ``@cached_property`` from the stdlib).
 
   Closes #1780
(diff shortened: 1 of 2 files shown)


DanielNoord:
Could you add a test for this? This could very easily break in the future without a regression test.

See also
https://github.com/pylint-dev/astroid/pull/546
And
https://github.com/mariocj89/astroid/commit/cda9dc4ad0e2cc075553d133dd0fdab468d614d7

It seems there was a good reason to add symlink support, but since those commits also lack all tests we can't make a good judgement call right now.

jacobtylerwalls:
Closing per [comment](https://github.com/pylint-dev/pylint/issues/791#issuecomment-1521180769). Feel free to pursue this further and open a new PR. Thanks!

PR #2131, opened 2023-04-22
Fix some typing issues found while working on ``bases.py``

DanielNoord:
## Type of Changes

## Description

The import outside of top level is much better than having the `lazy_import` imo. We lose type annotations with it and it is a costly dependency of the package.


Pierre-Sassoulas (approved):
It also makes cyclic import explicit which is good !

Issue #2124, opened 2023-04-17
Reconsider caching removed in #1747


jacobtylerwalls:
### Steps to reproduce
When running pylint on one submodule of the pylint primer, `music21.stream.core`, the caching removed in #1747 causes 2,711 calls to `self.file_from_module_name(modname, context_file)` for the module name `music21`.

Look into whether we can reintroduce some sort of caching without recausing https://github.com/pylint-dev/pylint/issues/5151.

jacobtylerwalls:
Closing as duplicate of #2041.

PR #2135, opened 2023-04-23
Reduce file system access in `ast_from_file()`

jacobtylerwalls:
## Type of Changes

## Description
Found while looking into #2124.

I noticed that for projects such as `music21` (in the pylint primer), we are making tens of thousands of unnecessary calls to `get_source_file()` for a path such as `project.__init__.py`. Since that method accesses the filesystem with `os.path.exists()`, that can be a significant burden. On my M1 Mac running Python 3.11, music21 lints in 171 instead of 171.6 seconds now.

with cProfile on a script that runs `pylint music21`:
Before: 63,346 calls to `<frozen genericpath>:16(exists)`
After: 10,333 calls to `<frozen genericpath>:16(exists)`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,8 @@ Release date: TBA
 
   Closes #1780
 
+* Reduce file system access in ``ast_from_file()``.
+
 * ``nodes.FunctionDef`` no longer inherits from ``nodes.Lambda``.
   This is a breaking change but considered a bug fix as the nodes did not share the same
   API and were not interchangeable.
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
Amazing find ! Less important now that this is cached but maybe we can avoid calling os.path.exists altogether and ask for forgiveness instead (catch oserror) too ?

jacobtylerwalls:
> maybe we can avoid calling os.path.exists altogether and ask for forgiveness instead (catch oserror) too ?

I think that's slower:

```python
>>> timeit("""
... try:
...   with open('music21/__init__.py', 'r') as f:
...     pass
... except OSError:
...     pass
... """)
24.64154148600028
>>> timeit('os.path.exists("music21/__init__.py")', setup='import os')
5.797904410999763

Pierre-Sassoulas (approved):

PR #2136, opened 2023-04-23
Fixes

DanielNoord:
## Type of Changes

## Description

I bundled a couple of fixes as the commits themselves are quite easy to review.

We only have one ``lazy_descriptor`` remaining which is subject to a pretty difficult import conflict. After that we can remove `lazy_object_proxy` as a dependency.


Pierre-Sassoulas (approved):
I'm pretty sure removing the lazy imports is a startup boost for pylint. It would be nice to do a minimal benchmark at the end (so we can brag about it in the changelog).

PR #2138, opened 2023-04-23
Make all arguments to ``nodes.NodeNG.__init__`` required

DanielNoord:
## Type of Changes

## Description

Now that we plan on doing `3.0` instead I feel much more comfortable with this change. I have seen in `ast` that there are definite cases where we need all line information to be `None`. Either way, I think it is much better to make that explicit instead of relying on implicit defaults three way down in your inheritance pattern.

With this PR we make a lot of our ``__init__``'s more explicit. I am not in favour of the required code duplication but am very much in favour of making the `astroid` code more explicit and easy to understand.

### DIFF
--- a/astroid/brain/brain_fstrings.py
+++ b/astroid/brain/brain_fstrings.py
@@ -19,7 +19,13 @@ def _clone_node_with_lineno(
     cls = node.__class__
     other_fields = node._other_fields
     _astroid_fields = node._astroid_fields
-    init_params = {"lineno": lineno, "col_offset": node.col_offset, "parent": parent}
+    init_params = {
+        "lineno": lineno,
+        "col_offset": node.col_offset,
+        "parent": parent,
+        "end_lineno": node.end_lineno,
+        "end_col_offset": node.end_col_offset,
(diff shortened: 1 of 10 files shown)


astroid/brain/brain_fstrings.py
```diff
@@ -19,7 +19,13 @@ def _clone_node_with_lineno(
     cls = node.__class__
     other_fields = node._other_fields
     _astroid_fields = node._astroid_fields
-    init_params = {"lineno": lineno, "col_offset": node.col_offset, "parent": parent}
+    init_params = {
```
Pierre-Sassoulas, line 22: I think we should cover this line even if it wasn't covered before, it's easy to make mistakes with this kind of modification.

DanielNoord: I don't really have a good idea how to meaningfully test this. It would just be a `Am I indeed passing the arguments that I'm adding in this PR to this dict`. Doesn't seem that useful.

Pierre-Sassoulas (approved):

PR #2140, opened 2023-04-23
Remove cachedproperty decorator

cdce8p:
## Description
Followup to #2137
This PR completely removes the `cachedproperty` decorator. `functools.cached_property` is a full replacement.

Also update changelog: `2.16.0` -> `3.0.0`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,7 @@
 astroid's ChangeLog
 ===================
 
-What's New in astroid 2.16.0?
+What's New in astroid 3.0.0?
 =============================
 Release date: TBA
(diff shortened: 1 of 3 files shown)


DanielNoord (approved):

cdce8p:
@DanielNoord Is `deprecation` the right label? Maybe we should add one for `breaking-change`?

DanielNoord:
@cdce8p Not sure, I was looking in the labels to find one that fit. Feels like they could use a clean up anyway, perhaps by making organisation labels and re-using them across both repo's?

PR #2141, opened 2023-04-23
Remove unused constants

cdce8p:
## Description
Remove `PY38_PLUS` and previously deprecated `BUILTINS` constants.
Also remove `Load`, `Store`, and `Del` constants as those have already been aliases for `Context` enum.

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -39,9 +39,6 @@ if TYPE_CHECKING:
     from astroid.constraint import Constraint
 
 
-# TODO: check if needs special treatment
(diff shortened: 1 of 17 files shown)


astroid/bases.py
```diff
@@ -41,7 +41,6 @@
 
 # TODO: check if needs special treatment
 BOOL_SPECIAL_METHOD = "__bool__"
```
DanielNoord, line 43: This is perhaps out of scope, but I noticed today that we should probably hardcode this. It's only used once.

DanielNoord (approved):

PR #2143, opened 2023-04-24
Cleanup rebuilder after dropping py37

cdce8p:


Pierre-Sassoulas (approved):
This is SO satisfying !

PR #2156, opened 2023-04-26
Fix requirement files and package data

DanielNoord:
## Type of Changes

## Description

Closes https://github.com/pylint-dev/astroid/issues/2129

The first commit has been a pet peeve of mine for months. I really disliked how it was hard to see what you need to install when first opening the project. I hope this lay out is a bit more explicit.

The second commit adds the files as requested by @mtelka.
Tested with `pip install .` and the files ended up in the egg info. @mtelka could you test if this branch fits your needs?

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,6 +1,6 @@
 -r requirements_minimal.txt
 
-# Tools used during development, prefer installing these with pre-commit
+# Tools used during development, prefer running these with pre-commit
 black
 pre-commit
 pylint
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,0 +1,8 @@
+-r requirements_minimal.txt
+
+# Tools used during development, prefer installing these with pre-commit
+black
+pre-commit
+pylint
+mypy
+ruff
--- a/requirements_minimal.txt
+++ b/requirements_minimal.txt
@@ -1,0 +1,8 @@
+# Tools used when releasing
+contributors-txt>=0.7.4
+tbump~=6.9
+
+# Tools used to run tests
+coverage~=7.2
+pytest
+pytest-cov~=4.0
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,9 +35,8 @@ jobs:
         id: generate-python-key
         run: >-
           echo "key=base-venv-${{ env.CACHE_VERSION }}-${{
-            hashFiles('pyproject.toml', 'requirements_test.txt',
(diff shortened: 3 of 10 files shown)


DanielNoord:
Sorry for the spam guys. I'm walking to the coffee machine as we speak to hopefully prevent more commits :smile:

requirements_dev.txt
```diff
@@ -0,0 +1,8 @@
+-r requirements_minimal.txt
+
+# Tools used during development, prefer installing these with pre-commit
+black
+pre-commit
+pylint
```
Pierre-Sassoulas, line 6: pylint cannot be installed with pre-commt

requirements_minimal.txt
```diff
@@ -0,0 +1,8 @@
+# Tools used when releasing
+contributors-txt>=0.7.4
+tbump~=6.9.0
```
Pierre-Sassoulas, line 3:
```suggestion
tbump~=6.9
```

requirements_dev.txt
Pierre-Sassoulas: "prefer installing these with pre-commit", this is not possible for pylint as it's a system hook.

mtelka:
I reviewed changes and they looks reasonable, but I'm not sure how to test this properly without a sdist release.  Anyway, my workflow (packaging of Python projects for OpenIndiana) does not require perfect sdists because I can easily patch them (as I currently do with astroid 2.15.4), so if there will be something still not working as expected I'll report back.

Thank you!

hroncok:
This makes the files installed.

```
(venv) [tmp]$ pip install https://github.com/pylint-dev/astroid/archive/da246398cb42c1c0509ce572cfb93b6bc1776c1a.zip
...

(venv) [tmp]$ ls -1 venv/lib/python3.11/site-packages/
...
astroid
astroid-3.0.0a3.dev0.dist-info
...
requirements_dev.txt
requirements_full.txt
requirements_minimal.txt
...
tests
tox.ini
```

hroncok:
Don't `package-data` it, include it in the sdist only? Back in the day, that was done using `MANIFEST.in`. No idea if it's possible to do it via `pyproject.toml`.

DanielNoord:
@hroncok Sorry for taking so long to respond. This has been constantly on my radar, but I really have no idea how to fix this. If I `include` the files in `MANIFEST.in` and remove them from the `pyproject.toml` I still get them when doing `pip install .`. Is that expected? I'm completely lost here.

DanielNoord:
Hmm that did seem to work. No idea why it didn't on initial try locally. Thanks for the help!

PR #2157, opened 2023-04-30
Yield directly from `_explicit_inference` in `NodeNG.infer()`

jacobtylerwalls:
## Type of Changes

## Description
Given the number of calls to `infer()`, it's worth optimizing it. We shouldn't let incrementing a counter (`nodes_inferred`) interfere with yielding directly from a generator (or rechecking that things are not None).

## Benchmark
```python
from timeit import timeit
node = extract_node("""
if True:
    x = 1
else:
    x = 2
x #@""")
timeit(node.inferred, globals=globals())
```

On my M1 mac, Python 3.11.2:
**main** (aba2587):
3.59-3.62
**PR**:
3.57-3.59

Looks like a 1% improvement for this extremely simple benchmark; ideally we'd see more substantial results in more substantial cases.

Also profiled linting the astroid package, and although the entire time to lint the package fluctuated, the `tottime` spent in `infer()`(i.e. itself) reliably reduced from .559s to .499s.

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -149,11 +149,12 @@ class NodeNG:
         if self._explicit_inference is not None:
             # explicit_inference is not bound, give it self explicitly
             try:
-                # pylint: disable=not-callable
-                results = list(self._explicit_inference(self, context, **kwargs))
-                if context is not None:
-                    context.nodes_inferred += len(results)
-                yield from results
+                if context is None:
+                    yield from self._explicit_inference(self, context, **kwargs)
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
I love those perf improvments. It would be nice to be able to give a rough global perf increase when we release 3.0. The primer need to be run on the latest code from the primed libs, which do not permit to compare apple to apple if we bench on it. Maybe we could upgrade the primer only when we release a new pylint minor. I'm going to find or create a issue to discuss this.

PR #2159, opened 2023-04-30
Fix some issues found while working on ``context.py``

DanielNoord:
## Type of Changes

## Description

First commit is just random clean up.

Second commit is complete signature typing of `infer_argument`, which I found while working on `InferenceContext`. I know this will decrease coverage, but rather than fix calling this with `InferenceResult` (which we already do) I think it is nicer to make the expected behaviour explicit and just accept a drop in coverage. If we ever have `Mypy` running fully it will be much easier to update the signature and reject all other nodes.

Third commit was the original intent of this PR and cleans up the annotation of `InferenceContext`. Nice to have done I guess.


astroid/transforms.py
```diff
@@ -145,7 +145,7 @@ def unregister_transform(
         """Unregister the given transform."""
         self.transforms[node_class].remove((transform, predicate))  # type: ignore[index, arg-type]
 
-    def visit(self, module: nodes.Module) -> SuccessfulInferenceResult:
+    def visit(self, module: nodes.NodeNG) -> SuccessfulInferenceResult:
```
jacobtylerwalls, line 148: From poking at this I wonder if `nodes.Module | nodes.Unknown` is better (and also fix `AstroidManager.visit_transforms`, too)?

DanielNoord: There is brain that calls this with a node, although it might be Unknown.
I don't think there is really a reason not to accept the base node right?

jacobtylerwalls: Only reason would be if it's a symptom of a mistake, then we could throw a TypeError. It sort of does seem like a symptom of a mistake, but brains do all kinds of hacky things, so I dunno. If we're going to leave the typing this way, we should update the argument name.

DanielNoord: I think updating the argument name is better.

This seems like something an astroid plugin might want to do: call all transforms for visited class definitions for example. Think this is a valid pattern.

jacobtylerwalls: Sounds good!

Pierre-Sassoulas (dismissed):
I don't like the coverage loss especially since mypy is not even activated yet so it's a net loss atm. But I trust you about the typing, we're never going to have it in astroid otherwise.

jacobtylerwalls (approved):
Nice! (The argument name update is a nice to have.)

jacobtylerwalls (approved):

PR #2160, opened 2023-04-30
(Almost) finish typing ``bases.py``

DanielNoord:
## Type of Changes

## Description

Again quite a big PR but I think the commits make it clear what is the intent. First I had to clean up some code that I noticed while seeing what actually happens "in the wild" in the `bases.py` code.
Then I added typing to `bases.py` and finally I also fixed `safe_infer` so it is compatible with what (can) happen(s) in the wild.

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import collections
 import collections.abc
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterable, Iterator
 from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
 from astroid import nodes
@@ -66,7 +66,9 @@ POSSIBLE_PROPERTIES = {
 }
 
 
-def _is_property(meth, context: InferenceContext | None = None) -> bool:
+def _is_property(
+    meth: nodes.FunctionDef | UnboundMethod, context: InferenceContext | None = None
+) -> bool:
     from astroid import helpers  # pylint: disable=import-outside-toplevel
 
     decoratornames = meth.decoratornames(context=context)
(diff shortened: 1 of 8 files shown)


astroid/bases.py
```diff
@@ -461,19 +479,22 @@ def infer_call_result(
...
 
     def _infer_builtin_new(
         self,
-        caller: nodes.Call,
+        caller: SuccessfulInferenceResult | None,
         context: InferenceContext,
     ) -> collections.abc.Generator[
         nodes.Const | Instance | UninferableBase, None, None
     ]:
+        if not isinstance(caller, nodes.Call):
```
DanielNoord, line 496: As expected. Again, I think it makes more sense to bring this in line with what we see in the wild and build robust code (without tests) for it, than spending a lot of time to see why we are passing arguments to this function that are too broad.

astroid/bases.py
```diff
@@ -536,7 +563,7 @@ def _infer_type_new_call(self, caller, context):  # noqa: C901
             mcs = next(caller.args[0].infer(context=context))
         except StopIteration as e:
             raise InferenceError(context=context) from e
-        if mcs.__class__.__name__ != "ClassDef":
+        if not isinstance(mcs, nodes.ClassDef):
```
Pierre-Sassoulas, line 566: This is a behavior change for subclasses of nodes.ClassDef ?

Pierre-Sassoulas: There's more such examples after that but I did not comment on each one (with Const for example). Maybe sometime it's important because msc could actually be a subclass ? Or maybe it's a fix because subclasses were supposed to be included in this conditional ?

DanielNoord: Yeah, I think as long as tests pass this should be fine.

It very much feels like this was a relic of the `lazy_import` stuff where `nodes` was previously unavailable.
Using `name` is bad anyway as we would match against any object that is also called ``Name``

Pierre-Sassoulas (approved):

PR #2162, opened 2023-05-01
Fix urllib3 tests following the release of v2.0.0

Pierre-Sassoulas:
## Type of Changes

## Description

Found this while trying to run the full test suite after installing with ``pip3 install -r requirements_full.txt``. It seems astroid is never tested with all requirements installed in CI, not even on a single interpreter.

### DIFF
--- a/requirements_full.txt
+++ b/requirements_full.txt
@@ -9,5 +9,5 @@ python-dateutil
 PyQt6
 regex
 six
-urllib3
+urllib3>1,<2
 typing_extensions>=4.4.0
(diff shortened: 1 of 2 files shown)


DanielNoord:
I wonder if we should just remove this test? Does it make sense to test against an old package?

Pierre-Sassoulas:
Yeah, we're in a strange place where old astroid brain supporting old numpy (or whatever other lib) version that are probably not used much anymore are lingering in the code base. Feel like this would ideally deserve a lot more maintenance work if we want to do it right. (brain adapting to version of lib, and tests with various version of libs, removal of libs following a deprecation period, etc.). I agree with what you say about cleaning up the code and not testing old lib but by contrast to other libs in the codebase, the last urllib <2 release is still a very fresh release.

DanielNoord (approved):

PR #2158, opened 2023-04-30
Complete cache key for inference tip

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description
As identified by @kriek in #1828, part of the cache key (context) was missing.

Alternative to #1832, which removes the cache, this just adds the missing part of the key.

Closes #1828
Closes pylint-dev/pylint#7464
Closes pylint-dev/pylint#8074

Refs #1982 (reverted)

## Pylint test
The following diff in pylint will (almost) pass the test suite with this change. ~PR shortly.~ PR once I can debug the remaining failures with `deprecated_typing_alias`.

```diff
diff --git a/pylint/checkers/design_analysis.py b/pylint/checkers/design_analysis.py
index 701615d89..d605aed76 100644
--- a/pylint/checkers/design_analysis.py
+++ b/pylint/checkers/design_analysis.py
@@ -166,6 +166,8 @@ STDLIB_CLASSES_IGNORE_ANCESTOR = frozenset(
         "typing.AsyncContextManager",
         "typing.Hashable",
         "typing.Sized",
+        TYPING_NAMEDTUPLE,
+        TYPING_TYPEDDICT,
     )
 )
```

PR #2166, opened 2023-05-06
Improve typing of inference functions

jacobtylerwalls:
## Type of Changes

## Description
Follow-up to #2158.

Before, this was quite vague:
```python
InferFn = Callable[..., Any]
```

### DIFF
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -6,18 +6,21 @@
 
 from __future__ import annotations
 
+import sys
 from collections.abc import Callable, Iterator
+from typing import cast
 
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
 from astroid.nodes import NodeNG
-from astroid.typing import (
-    _P,
-    InferenceResult,
-    InferFn,
-    InferFnExplicit,
-    InferFnTransform,
-)
+from astroid.typing import InferenceResult, InferFn, InferFnExplicit, InferFnTransform
+
+if sys.version_info >= (3, 11):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
+_P = ParamSpec("_P")
 
 _cache: dict[
     tuple[InferFn, NodeNG, InferenceContext | None], list[InferenceResult]
@@ -39,8 +42,8 @@ def _inference_tip_cached(
     def inner(
         *args: _P.args, **kwargs: _P.kwargs
     ) -> Iterator[InferenceResult] | list[InferenceResult]:
-        node: NodeNG = args[0]
-        context: InferenceContext | None = args[1]
+        node = cast(NodeNG, args[0])
+        context = cast(InferenceContext | None, args[1])
 
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Iterator
+from typing import Any
 
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
@@ -31,7 +32,7 @@ def _inference_tip_cached(
     """Cache decorator used for inference tips."""
 
     def inner(
-        node: NodeNG, context: InferenceContext | None
+        node: NodeNG, context: InferenceContext | None, **kwargs: Any,
     ) -> Iterator[InferenceResult] | list[InferenceResult]:
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
@@ -47,7 +48,7 @@ def _inference_tip_cached(
             # with slightly different contexts while still passing the simple
             # test cases included with this commit.
             _CURRENTLY_INFERRING.add(partial_cache_key)
-            result = _cache[func, node, context] = list(func(node, context))
+            result = _cache[func, node, context] = list(func(node, context, **kwargs))
             # Remove recursion guard.
             _CURRENTLY_INFERRING.remove(partial_cache_key)
 
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -40,11 +40,8 @@ def _inference_tip_cached(
     def inner(
         *args: _P.args, **kwargs: _P.kwargs
     ) -> Iterator[InferenceResult] | list[InferenceResult]:
-        node = args[0]
-        context = args[1]
-        if TYPE_CHECKING:
-            assert isinstance(node, NodeNG)
-            assert context is None or isinstance(context, InferenceContext)
+        node: NodeNG = args[0]
+        context: InferenceContext | None = args[1]
 
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Iterator
-from typing import TYPE_CHECKING
 
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Callable, Iterator
-from typing import cast
+from typing import Optional, cast
 
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
@@ -43,7 +43,7 @@ def _inference_tip_cached(
         *args: _P.args, **kwargs: _P.kwargs
     ) -> Iterator[InferenceResult] | list[InferenceResult]:
         node = cast(NodeNG, args[0])
-        context = cast(InferenceContext | None, args[1])
+        context = cast(Optional[InferenceContext], args[1])
 
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -6,21 +6,12 @@
 
 from __future__ import annotations
 
-import sys
 from collections.abc import Callable, Iterator
-from typing import Optional, cast
 
 from astroid.context import InferenceContext
 from astroid.exceptions import InferenceOverwriteError, UseInferenceDefault
 from astroid.nodes import NodeNG
 from astroid.typing import InferenceResult, InferFn, InferFnExplicit, InferFnTransform
-
-if sys.version_info >= (3, 11):
-    from typing import ParamSpec
-else:
-    from typing_extensions import ParamSpec
-
-_P = ParamSpec("_P")
 
 _cache: dict[
     tuple[InferFn, NodeNG, InferenceContext | None], list[InferenceResult]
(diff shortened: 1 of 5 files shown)


astroid/inference_tip.py
```diff
@@ -6,20 +6,19 @@
...
-from astroid.typing import InferenceResult, InferFn
-
-if sys.version_info >= (3, 11):
-    from typing import ParamSpec
-else:
-    from typing_extensions import ParamSpec
-
-_P = ParamSpec("_P")
+from astroid.typing import (
+    _P,
```
jacobtylerwalls, line 16: Is there a better way to do this/name this?

DanielNoord: I think in general people tend to just redefine a `ParamSpec` per file, can't we do that here?

astroid/inference_tip.py
```diff
@@ -35,12 +34,18 @@ def clear_inference_tip_cache() -> None:
...
 
-    def inner(*args: _P.args, **kwargs: _P.kwargs) -> Iterator[InferenceResult]:
+    def inner(
+        *args: _P.args, **kwargs: _P.kwargs
+    ) -> Iterator[InferenceResult] | list[InferenceResult]:
         node = args[0]
         context = args[1]
+        if TYPE_CHECKING:
+            assert isinstance(node, NodeNG)
+            assert context is None or isinstance(context, InferenceContext)
```
DanielNoord, line 47:
```suggestion
        node: nodeNG = args[0]
        context: InferenceContext | None = args[1]
```

Think I would prefer this?

astroid/inference_tip.py
```diff
@@ -6,20 +6,19 @@
 
 from __future__ import annotations
 
-import sys
 from collections.abc import Callable, Iterator
+from typing import TYPE_CHECKING
```
jacobtylerwalls, line 10:
```suggestion
```

astroid/inference_tip.py
jacobtylerwalls: Actually, this doesn't work, gives:
```
astroid/inference_tip.py:44: error: Incompatible types in assignment (expression has type "object", variable has type "NodeNG")  [assignment]
astroid/inference_tip.py:45: error: Incompatible types in assignment (expression has type "object", variable has type "Optional[InferenceContext]")  [assignment]
```
I'll cast instead of asserting. I wanted to avoid the penalty in such a hot path of isinstance.

DanielNoord: Let's add a type ignore. If we improve the typing this might at some point then be flagged as unused.

jacobtylerwalls: [5fd854f](https://github.com/pylint-dev/astroid/pull/2166/commits/5fd854fff4439c4b322a706687d069e25e6a802e)

jacobtylerwalls: > If we improve the typing this might at some point then be flagged as unused.

How could the typing be better, unless we remove the dynamic *args? (An earlier version of this PR did that, but I wasn't sure that was alright to remove support for it.)

DanielNoord: I would say we can remove support, it's breaking changes season anyway.

jacobtylerwalls: ⛈️

jacobtylerwalls: 06f42bb

astroid/inference_tip.py
```diff
@@ -34,13 +26,13 @@ def clear_inference_tip_cache() -> None:
 
 
 def _inference_tip_cached(
-    func: Callable[_P, Iterator[InferenceResult]],
-) -> Callable[_P, Iterator[InferenceResult]]:
+    func: Callable[[NodeNG, InferenceContext | None], Iterator[InferenceResult]],
```
DanielNoord, line 29: I do think we have support for ``**kwargs``, which we should probably keep.

jacobtylerwalls: Curious why you think we should keep it. Can't find any examples. Trying to be extra-cautious about plugins?

DanielNoord: `ImportFrom._infer` uses an additional argument. I know, I hate it as well..

Pierre-Sassoulas: If ImportFrom is breaking the LSP  then we could design a breaking change where we don't, perhaps ?

jacobtylerwalls: I'm thinking of just moving the additional argument into **kwargs for now.

jacobtylerwalls: Already called by keyword:
https://github.com/pylint-dev/pylint/blob/8776ba0808f807a8915d96c0901526d7c648ec34/pylint/checkers/variables.py#L254

DanielNoord: Yeah but then `func` needs to be typed differently right?

jacobtylerwalls: I think I'd future that. There's not much value to adding it, it just raises the number of mypy errors, and it would cause retyping all the inference functions in `node_classes.py`.

DanielNoord: Hmm, most of the functions in `inference.py` already accept `**kwargs` like they should. Think it is better to reflect what we want the typing to be and work towards that than adding typing of which we know it is incorrect.

jacobtylerwalls: This diff
```diff
diff --git a/astroid/typing.py b/astroid/typing.py
index 85d7031d..80d585d9 100644
--- a/astroid/typing.py
+++ b/astroid/typing.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 from typing import (
     TYPE_CHECKING,
+    Any,
     Callable,
     Generator,
     Iterator,
@@ -75,9 +76,9 @@ InferBinaryOp = Callable[
     Generator[InferenceResult, None, None],
 ]
 
-InferFn = Callable[[_NodesT, Optional["InferenceContext"]], Iterator[InferenceResult]]
+InferFn = Callable[[_NodesT, Optional["InferenceContext"], Any], Iterator[InferenceResult]]
 InferFnExplicit = Callable[
-    [_NodesT, Optional["InferenceContext"]],
+    [_NodesT, Optional["InferenceContext"], Any],
     Union[Iterator[InferenceResult], List[InferenceResult]],
 ]
 InferFnTransform = Callable[[_NodesT, InferFn], _NodesT]
```
gives mypy errors like:
```
astroid/inference_tip.py:51: error: Too few arguments  [call-arg]
astroid/inference_tip.py:57: error: Incompatible return value type (got "Callable[[NodeNG, Optional[InferenceContext], KwArg(Any)], Union[Iterator[Union[NodeNG, UninferableBase, Proxy]], List[Union[NodeNG, UninferableBase, Proxy]]]]", expected "Callable[[Any, Optional[InferenceContext], Optional[Any]], Union[Iterator[Union[NodeNG, UninferableBase, Proxy]], List[Union[NodeNG, UninferableBase, Proxy]]]]")  [return-value]
```

To avoid it, I think I'd need a `Protocol`. I take your point we don't want wrong typing, but I just thought incremental improvements were still a good thing, since I don't have time to investigate Protocol right now.

DanielNoord: I can look into the `Protocol`. Want me to take over? You can focus on your PR for the `_infer` thing then 😄

jacobtylerwalls: Yes please!

DanielNoord:
@jacobtylerwalls think this is correct. Let me know what you think!

astroid/inference_tip.py
```diff
@@ -6,65 +6,67 @@
...
+    ) -> Generator[InferenceResult, None, None]:
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
             # If through recursion we end up trying to infer the same
             # func + node we raise here.
             raise UseInferenceDefault
         try:
-            return _cache[func, node, context]
+            yield from _cache[func, node, context]
+            return
```
jacobtylerwalls, line 50: Nice, this was really annoying before when it directly returned a list.

astroid/inference_tip.py
```diff
@@ -6,65 +6,67 @@
...
             # with slightly different contexts while still passing the simple
             # test cases included with this commit.
             _CURRENTLY_INFERRING.add(partial_cache_key)
-            result = _cache[func, node, context] = list(func(*args, **kwargs))
+            _cache[func, node, context] = list(func(node, context, **kwargs))
             # Remove recursion guard.
             _CURRENTLY_INFERRING.remove(partial_cache_key)
 
-        return iter(result)
+        yield from _cache[func, node, context]
```
jacobtylerwalls, line 62: I wonder if it still makes sense to save the `result` so it doesn't have to be looked up again. A very micro-optimization, but potentially worth it in such a hot path.

astroid/typing.py
```diff
@@ -67,3 +76,22 @@ class AstroidManagerBrain(TypedDict):
...
+
+
+class InferFn(Protocol, Generic[_SuccessfulInferenceResultT_contra]):
+    def __call__(
+        self,
+        node: _SuccessfulInferenceResultT_contra,
+        context: InferenceContext | None = None,
+        **kwargs: Any,
+    ) -> Generator[InferenceResult, None, None]:
+        ...
```
jacobtylerwalls, line 88: can we no-cover these?

jacobtylerwalls (commented):
New mypy messages as of b3b6c53a8024cf26e7b0fb2b663475174b63b686:
```
astroid/nodes/node_ng.py:147: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
astroid/nodes/node_ng.py:149: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
```

DanielNoord:
> New mypy messages as of b3b6c53a8024cf26e7b0fb2b663475174b63b686:
> 
> ```
> 
> astroid/nodes/node_ng.py:147: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
> 
> astroid/nodes/node_ng.py:149: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
> 
> ```

I'll change it to be Node again.

astroid/typing.py
DanielNoord: Yes!

astroid/inference_tip.py
DanielNoord: Ah, I had to remove the assignment previously because of type annotations but can probably add it again.
Will change!

DanielNoord:
> > New mypy messages as of [b3b6c53](https://github.com/pylint-dev/astroid/commit/b3b6c53a8024cf26e7b0fb2b663475174b63b686):
> > ```
> > 
> > astroid/nodes/node_ng.py:147: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
> > 
> > astroid/nodes/node_ng.py:149: error: Argument 1 to "__call__" of "InferFn" has incompatible type "NodeNG"; expected "Self"  [arg-type]
> > ```
> 
> I'll change it to be Node again.

Scherm­afbeelding 2023-05-07 om 23 21 22


This seems to be a `mypy` bug. `pyright` correctly infers `self` to be `Self` and doesn't complain. Would you be okay with a `type: ignore`?

jacobtylerwalls:
Works for me. The blame will point back to this discussion in case mypy improves before we turn it on.

DanielNoord:
> Works for me. The blame will point back to this discussion in case mypy improves before we turn it on.

Implemented!

jacobtylerwalls (commented):
LGTM, I can't click Approve as the OP. Happy to squash merge once you ✅

astroid/inference_tip.py
```diff
@@ -6,65 +6,67 @@
...
-    def inner(*args: _P.args, **kwargs: _P.kwargs) -> Iterator[InferenceResult]:
-        node = args[0]
-        context = args[1]
+    def inner(
+        node: _NodesT,
+        context: InferenceContext | None = None,
+        **kwargs: Any,
+    ) -> Generator[InferenceResult, None, None]:
         partial_cache_key = (func, node)
         if partial_cache_key in _CURRENTLY_INFERRING:
```
DanielNoord, line 44: Unrelated, but isn't this what the `path` wrapper does? Can we use that here?

DanielNoord (approved):

astroid/inference_tip.py
jacobtylerwalls: Perhaps, but at glance, it looks like that one is sensitive to specific InferenceContexts. Here, we're not ready to unleash recursive inference with every slightly different context.

jacobtylerwalls:
Thanks for the review and for taking over!

PR #2171, opened 2023-05-08
Move inference methods to nodes module(s)

### LINKED ISSUE #679: Remove monkey-patching of methods onto classes
In [this comment](https://github.com/PyCQA/astroid/issues/170#issuecomment-163118573), @PCManticore stated that there was a plan to remove monkey-patching of methods onto classes. inference.py and protocols.py seem to suggest that this hasn't happened.

Is this still a design goal?

It looks to me like this would be a fairly mechanical change. I am assuming that since it hasn't happened yet, I am missing something. Is it simply a "haven't had time" issue?

jacobtylerwalls:
## Type of Changes

## Description
Move `_infer_()` methods from `inference.py` to the nodes themselves (alternative to #2167).

## Benefits
- Avoid assigning methods to classes in inference.py and protocols.py (yuk!)
- Move LookupMixIn to _base_nodes as promised in the long-existing TODO comment
- We can now clearly see messages like:
```
astroid/nodes/node_classes.py:2349:4: W0221: Variadics removed in overriding 'Dict._infer' method (arguments-differ)
```

## Drawbacks
- There are some function-level imports created here that are symptoms of circular imports, but many of them are with the helpers module, and many of those are just for `safe_infer`, which could be extracted into its own helper module (it only needs UninferableBase). I don't think this jeopardizes the design.

Fixes #679


astroid/nodes/__init__.py
```diff
@@ -11,10 +11,6 @@
 """
 
 # Nodes not present in the builtin ast module:  DictUnpack, Unknown, and EvaluatedObject.
-
-# This is the only node we re-export from the private _base_nodes module. This
-# is because it was originally part of the public API and hasn't been deprecated.
-from astroid.nodes._base_nodes import Statement
```
jacobtylerwalls, line 17: removed!

astroid/nodes/node_classes.py
```diff
@@ -403,8 +433,6 @@ class AssignName(_base_nodes.NoChildrenNode, LookupMixIn, _base_nodes.ParentAssi
 
     _other_fields = ("name",)
 
-    infer_lhs: ClassVar[InferLHS[AssignName]]
-
```
jacobtylerwalls, line 407: yuk!

astroid/nodes/node_classes.py
```diff
@@ -1470,6 +1727,88 @@ def last_child(self):
         return self.ops[-1][1]
         # return self.left
 
+    # TODO: move to util?
```
jacobtylerwalls, line 1730: decision point

astroid/nodes/node_classes.py
```diff
@@ -1470,6 +1727,88 @@ def last_child(self):
...
+    @staticmethod
+    def _to_literal(node: SuccessfulInferenceResult) -> Any:
+        # Can raise SyntaxError or ValueError from ast.literal_eval
+        # Can raise AttributeError from node.as_string() as not all nodes have a visitor
+        # Is this the stupidest idea or the simplest idea?
```
jacobtylerwalls, line 1735: lol

astroid/bases.py
```diff
@@ -354,7 +353,16 @@ class Instance(BaseInstance):
     def __init__(self, proxied: nodes.ClassDef | None) -> None:
         super().__init__(proxied)
 
-    infer_binary_op: ClassVar[InferBinaryOp[Instance]]
+    @decorators.yes_if_nothing_inferred
+    def infer_binary_op(
+        self: Instance | nodes.ClassDef,
```
DanielNoord, line 358: This feels wrong, I don't think we should every type `self`. Why is this necessarry?

astroid/node_classes.py
```diff
@@ -48,7 +48,6 @@
     JoinedStr,
     Keyword,
     List,
-    LookupMixIn,
     Match,
```
DanielNoord, line 51: We should probably mention this in the release notes.

astroid/nodes/_base_nodes.py
```diff
@@ -10,15 +10,42 @@
...
+        [
+            InferenceResult,
+            Optional[InferenceResult],
+            Union[nodes.AugAssign, nodes.BinOp],
+            InferenceResult,
+            Optional[InferenceResult],
+            InferenceContext,
+            InferenceContext,
+        ],
+        "list[partial[Generator[InferenceResult, None, None]]]",
```
DanielNoord, line 47: Since this is under `TYPE_CHECKING` we should be able to do this without the `"`

astroid/nodes/_base_nodes.py
```diff
@@ -220,3 +247,524 @@ def _elsed_block_range(
...
+                    return
+
+                yield from results
+                return
+
+        yield util.Uninferable
+
+
+class AttributeNode(NodeNG):
+    expr: NodeNG
```
DanielNoord, line 672: This feels like this class should also define a `postinit` to couple this annotation to something actually setting it.

astroid/nodes/_base_nodes.py
```diff
@@ -220,3 +247,524 @@ def _elsed_block_range(
...
+        node: AssignedStmtsPossibleNode = None,
+        context: InferenceContext | None = None,
+        assign_path: list[int] | None = None,
+    ) -> Any:
+        """Returns the assigned statement (non inferred) according to the assignment type."""
+        return self.parent.assigned_stmts(node=self, context=context)
+
+
+class NameNode(LookupMixIn):
+    name: str
```
DanielNoord, line 741: See comment above.

astroid/nodes/_base_nodes.py
```diff
@@ -220,3 +247,524 @@ def _elsed_block_range(
...
+
+            if not stmts:
+                raise NameInferenceError(
+                    name=self.name, scope=self.scope(), context=context
+                )
+        context = copy_context(context)
+        context.lookupname = self.name
+        context.constraints[self.name] = get_constraints(self, frame)
+
+        return bases._infer_stmts(stmts, context, frame)
```
DanielNoord, line 770: I really like these new base classes, but don't really like how long this file is going to be and how many different types of bases nodes are in here. Not sure if this is fixable though.

astroid/nodes/node_classes.py
```diff
@@ -669,7 +718,9 @@ def postinit(
             type_comment_posonlyargs = []
         self.type_comment_posonlyargs = type_comment_posonlyargs
 
-    assigned_stmts: ClassVar[AssignedStmtsCall[Arguments]]
+    assigned_stmts: ClassVar[
+        AssignedStmtsCall[Arguments]
+    ] = protocols.arguments_assigned_stmts
```
DanielNoord, line 723: This is already much better. We could even remove the annotation here?

DanielNoord (commented):
Left some early review comments.

> ## Benefits
> ...

🚀  👍 

> ## Drawbacks
> * There are some function-level imports created here that are symptoms of circular imports, but many of them are with the helpers module, and many of those are just for `safe_infer`, which could be extracted into its own helper module (it only needs UninferableBase). I don't think this jeopardizes the design.

Agreed!

However, I think we are missing one major drawback (although not necessarily a drawback of this design).
We still have the issue of these classes being behemoths that (imo) try to do too much. With this change, and possibly the inclusion of `protocols.py` on the nodes themselves as well, there is not that much left besides the `astroid.nodes` package. I still don't know how I feel about this as it feels we are putting more logic on the `nodes` than is completely necessary.
There are also still reference to `AstroidManager()` or `MANAGER` which makes it impossible to decouple the global state used by `astroid` without starting the pass an actual `manager` object in these signatures.


> ## Todos:
> * [ ]  Get benchmarks

If we agree on this design I think we shouldn't be held back by performance degradation too much. We can more easily improve performance when we can actually work with the code.

DanielNoord (commented):
Left some early review comments.

> ## Benefits
> ...

🚀  👍 

> ## Drawbacks
> * There are some function-level imports created here that are symptoms of circular imports, but many of them are with the helpers module, and many of those are just for `safe_infer`, which could be extracted into its own helper module (it only needs UninferableBase). I don't think this jeopardizes the design.

Agreed!

However, I think we are missing one major drawback (although not necessarily a drawback of this design).
We still have the issue of these classes being behemoths that (imo) try to do too much. With this change, and possibly the inclusion of `protocols.py` on the nodes themselves as well, there is not that much left besides the `astroid.nodes` package. I still don't know how I feel about this as it feels we are putting more logic on the `nodes` than is completely necessary.
There are also still reference to `AstroidManager()` or `MANAGER` which makes it impossible to decouple the global state used by `astroid` without starting the pass an actual `manager` object in these signatures.


> ## Todos:
> * [ ]  Get benchmarks

If we agree on this design I think we shouldn't be held back by performance degradation too much. We can more easily improve performance when we can actually work with the code.

astroid/bases.py
jacobtylerwalls: merely copy paste. I'll clean it up to remove it.

astroid/nodes/_base_nodes.py
```diff
@@ -220,3 +247,524 @@ def _elsed_block_range(
...
+                values = tuple(i.value for i in inferred_positional)
+            else:
+                values = None
+        elif isinstance(other, nodes.Dict):
+            values: dict[Any, Any] = {}
+            for pair in other.items:
+                key = helpers.safe_infer(pair[0], context)
+                if not isinstance(key, nodes.Const):
+                    return (util.Uninferable,)
+                value = helpers.safe_infer(pair[1], context)
```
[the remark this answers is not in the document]
jacobtylerwalls: I agree, but this is a large enough PR that I think it would be beneficial to do it in a follow-up.

astroid/nodes/_base_nodes.py
jacobtylerwalls: It's more of a mixin. The actual concrete classes do have postinit methods. Should I document this or raise some sort of TypeError when attempting to instantiate?

astroid/nodes/_base_nodes.py
jacobtylerwalls: The concrete classes for this one lack a postinit, though!

jacobtylerwalls: fb05e0c65e443e0e438142821ba120680ac0dbff

astroid/nodes/_base_nodes.py
jacobtylerwalls: [c644b53](https://github.com/pylint-dev/astroid/pull/2171/commits/c644b53ea04af5e87cdf1ba536823dbe96b29775)

astroid/bases.py
jacobtylerwalls: [c644b53](https://github.com/pylint-dev/astroid/pull/2171/commits/c644b53ea04af5e87cdf1ba536823dbe96b29775)

astroid/node_classes.py
jacobtylerwalls: [b969095](https://github.com/pylint-dev/astroid/pull/2171/commits/b9690955e2006e3aef967edd5f352daae725e374)

astroid/nodes/node_classes.py
jacobtylerwalls: [c644b53](https://github.com/pylint-dev/astroid/pull/2171/commits/c644b53ea04af5e87cdf1ba536823dbe96b29775)

astroid/nodes/_base_nodes.py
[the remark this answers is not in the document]
jacobtylerwalls: Next PR for this ready [here](https://github.com/jacobtylerwalls/astroid/compare/inference-polymorphism...jacobtylerwalls:astroid:move-safe-infer?expand=1)

astroid/nodes/_base_nodes.py
DanielNoord: Perhaps make them `ABC`? However, I don't really like mixins in this way. We actually spend quite some time removing them from `pylint` because the way this mixins work don't allow us to guarantee that they function at runtime. We can't enforce that `expr` is being set.. Would it be feasible to make them have a `postinit` that needs to be called?

astroid/nodes/_base_nodes.py
DanielNoord: Do they need one? `name` gets set in the `init`?

astroid/helpers.py
```diff
@@ -324,3 +324,23 @@ def object_len(node, context: InferenceContext | None = None):
     raise AstroidTypeError(
         f"'{result_of_len}' object cannot be interpreted as an integer"
     )
+
+
+def _higher_function_scope(node: nodes.NodeNG) -> nodes.FunctionDef | None:
+    """Search for the first function which encloses the given
+    scope. This can be used for looking up in that function's
```
DanielNoord, line 331:
```suggestion
    scope.
    
    This can be used for looking up in that function's
```

ChangeLog
```diff
@@ -15,6 +15,11 @@ Release date: TBA
   Closes #1780
   Refs #2140
 
+* Remove the ``inference`` module. Node inference methods are now in the module
+  defining the node, rather than being associated to the node afterward.
+
+* Move ``LookupMixIn`` to ``astroid.nodes._base_nodes``.
```
DanielNoord, line 21:
```suggestion
* Move ``LookupMixIn`` to ``astroid.nodes._base_nodes`` and make it private.
```

astroid/nodes/node_classes.py
```diff
@@ -499,12 +538,21 @@ def __init__(
             parent=parent,
         )
 
+    def postinit(self, name: str) -> None:
```
DanielNoord, line 541: See other comment, I don't think this is necessary.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1469,6 +1485,44 @@ def is_generator(self) -> bool:
         """
         return bool(next(self._get_yield_nodes_skip_lambdas(), False))
 
+    def _infer(
+        self, context: InferenceContext | None = None, **kwargs: Any
+    ) -> Generator[objects.Property | FunctionDef, None, InferenceErrorInfo]:
+        from astroid import objects  # pylint: disable=import-outside-toplevel
```
DanielNoord, line 1491: I think this is a case of pretty bad cyclic imports. With this approach we can't avoid this I think which makes me doubtful about whether this is the right approach.

astroid/protocols.py
```diff
@@ -129,6 +107,8 @@ def const_infer_binary_op(
     context: InferenceContext,
     _: SuccessfulInferenceResult,
 ) -> Generator[ConstFactoryResult | util.UninferableBase, None, None]:
+    from astroid.nodes import const_factory  # pylint: disable=import-outside-toplevel
```
DanielNoord, line 110: On line 57 we already use this from `nodes` why did you change this?

DanielNoord (commented):
Still not sure how I feel about this. I think the use of `Mixins` and the import issues are not very nice, but you guys seem to agree that this is better so I'm also not sure how productive my opposition to this is 😄

astroid/nodes/_base_nodes.py
jacobtylerwalls: Whoops, you're totally right. Reverting.

astroid/protocols.py
jacobtylerwalls: Good catch. I shuffled this method out of this file and back into it, and missed this diff when I brought it back. Thank you!

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: Challenge accepted!

[Here is a branch](https://github.com/pylint-dev/astroid/compare/main...jacobtylerwalls:astroid:move-property-and-partial?expand=1) that removes these cyclic imports by moving `PartialFunction` and `Property` to `scoped_nodes`, which makes sense, since they inherit from `FunctionDef` anyway (what's the use of a Property being an object, anyway?)

The ease with which I could do this gives me hope that this is a _good_ direction that can enable easier refactors in iterative PRs. What do you think?

jacobtylerwalls: If we go with the `infer_object` approach, we're freezing the inference/objects/protocols ontologies in place, instead of actually questioning if everything is where it should be, which we should feel free to do during breaking changes season ☔

astroid/nodes/_base_nodes.py
jacobtylerwalls: Happy to do it, but just for my own understanding, why would a postinit help? When would it ever be called, and how does it help a static type checker? I'm sure there is a reason, I just haven't worked with `postinit` in astroid yet.

DanielNoord: In this case it should probably be an `init`. Sorry for the confusion!

jacobtylerwalls: If the mixins need to be robust against unintended instantiation, then they're more trouble then they're worth just for the DRY benefit. I'll just extract helper functions that can be shared.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: See my comment on the other pull request. I don't necessarily think we would be freezing anything in place.

I do think that we should merge your `PartialFunction` branch to `master` irrespective of the approach. Seems like a very sensible change!

astroid/helpers.py
```diff
@@ -324,3 +324,25 @@ def object_len(node, context: InferenceContext | None = None):
...
+    name fails.
+
+    :param node: A scope node.
+    :returns:
+        ``None``, if no parent function scope was found,
+        otherwise an instance of :class:`astroid.nodes.scoped_nodes.Function`,
+        which encloses the given node.
+    """
+    current = node
+    while current.parent and not isinstance(current.parent, nodes.FunctionDef):
```
DanielNoord, line 344: We could remove the `type: ignore` by doing a `return` within the `while` if `parent` is a `FunctionDef` I think.

DanielNoord (approved):
I have one nit, but I don't really mind merging this as is and seeing if there are smaller improvements later on. There is too much going on to make it perfect anyway.

While I still don't prefer this method it seems the discussion doesn't really take off and I don't want to block this any longer. I trust you fixed this and tested that `pylint` test suite still passes.

If Pierre still approves of this PR (the code, not the idea) then let's ship it 🚀

astroid/helpers.py
jacobtylerwalls: I removed the disable and no message was emitted ✨

jacobtylerwalls:
Sounds good @DanielNoord I really appreciate the care it took to reach this point, and of course I'm open to `infer_object()` down the road if we start showing tangible results with it. 👍 

I rebased and cleaned up the branch a bit.

Future PRs:
- Remove the mixins, they're not essential
- moving PartialFunction and Property to scoped_nodes
- move safe_infer to its own module
- maybe: remove monkey-patching from protocols.py like infer_binary_op

> I trust you fixed this and tested that pylint test suite still passes.

I don't know what I fixed other than the monkey-patching 😉 but yes the pylint suite passes. There's plenty to do in pylint, we should cut an astroid alpha ASAP and add 3.12 support and remove future=True everywhere.

astroid/nodes/_base_nodes.py
```diff
@@ -223,3 +250,526 @@ def _elsed_block_range(
...
+            [InferenceContext | None],
+            Generator[InferenceResult | util.BadOperationMessage, None, None],
+        ],
+        context: InferenceContext | None,
+        error: type[util.BadOperationMessage],
+    ) -> Generator[InferenceResult, None, None]:
+        for result in infer_callable(context):
+            if isinstance(result, error):
+                # For the sake of .infer(), we don't care about operation
+                # errors, which is the job of pylint. So return something
```
Pierre-Sassoulas, line 334:
```suggestion
                # errors, which is the job of a linter. So return something
```
Astroid Can be used without pylint :D

astroid/nodes/_base_nodes.py
Pierre-Sassoulas: I also greatly dislike mixin for the sake of being dry because having to initialise instance in a particular order and not being able to use typecheck (becauseof course you can't guarantee the instanciation particular order and a type checker will rightfully warn you) is order if magnitude worse than whatever design problem mixin are trying to solve.

astroid/nodes/_base_nodes.py
Pierre-Sassoulas: We could burst some files later on but we have other issues to deal with imo. (Especially cyclic imports that make this harder)

astroid/nodes/scoped_nodes/scoped_nodes.py
Pierre-Sassoulas: Maybe bursting the big node files is actually (one of) the way to solve circular imports?

Pierre-Sassoulas (approved):
LGTM, amazing refactor there's a lot of work here, and it certainely makes astroid more undrstandable

astroid/nodes/_base_nodes.py
jacobtylerwalls: Hah, good point! I'll catch this in the next PR.

jacobtylerwalls:
Thanks so much for the reviews. I'll merge this and then get started on the followup PRs for removing the mixins and some of the circular imports.

PR #2178, opened 2023-05-13
Improve performance of `looks_like_numpy_member()`

jacobtylerwalls:
## Type of Changes

## Description
This method is the hottest path in astroid by number of primitive calls(!) Every name and attribute goes through it, multiplied by the number of numpy method names we compare against.

We already know whether we have a `Name` or `Attribute` when registering the transform, so avoiding re-checking the type here saves 32% of the calls to `isinstance()` when linting the entire astroid package.

## Benchmark
for `pylint astroid` on my M1 mac running python 3.11.2, a savings of 0.4s out of 16.7s, or about 2.5% wall time:

**main**
```
 12602852    0.658    0.000    0.717    0.000 {built-in method builtins.isinstance}
  2056140    0.426    0.000    0.581    0.000 brain_numpy_utils.py:64(looks_like_numpy_member)
```

**PR**
```
  8491012    0.510    0.000    0.568    0.000 {built-in method builtins.isinstance}
  1534080    0.101    0.000    0.101    0.000 brain_numpy_utils.py:64(name_looks_like_numpy_member)
...
   522000    0.034    0.000    0.038    0.000 brain_numpy_utils.py:72(attribute_looks_like_numpy_member)
```

### DIFF
--- a/astroid/brain/brain_numpy_core_function_base.py
+++ b/astroid/brain/brain_numpy_core_function_base.py
@@ -6,7 +6,10 @@
 
 import functools
 
-from astroid.brain.brain_numpy_utils import infer_numpy_member, looks_like_numpy_member
+from astroid.brain.brain_numpy_utils import (
+    infer_numpy_member,
+    attribute_looks_like_numpy_member,
+)
 from astroid.inference_tip import inference_tip
 from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import Attribute
@@ -25,5 +28,5 @@ for func_name, func_src in METHODS_TO_BE_INFERRED.items():
     AstroidManager().register_transform(
         Attribute,
         inference_tip(inference_function),
-        functools.partial(looks_like_numpy_member, func_name),
+        functools.partial(attribute_looks_like_numpy_member, func_name),
     )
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
I'm wondering if we should treat numpy this way at all ? Could we create a pylint-numpy plugin that would be the only one doing the check ?

DanielNoord (commented):
Funny, I have local branch that fixes this exact issue.
Another improvement could be not to create a brain per method, but just pass a set of member names and see if the attribute name is in it. There is no need to have a separate brain for every possible member.

Do agree with Pierre though that ideally this should be moved out of the base astroid package.

Pierre-Sassoulas:
Not sure if you followed that @jacobtylerwalls but we created [a proof of concept "pylint-tensorflow" project](https://github.com/pylint-dev/pylint-tensorflow) to see if brains could be moved (tensorflow is heavy). Right now it's not in working condition, but if loading brains make astroid load slow, extending such plugins -- once we're clear about how they would work exactly -- would be within reach. Right now the idea would be to make user install the plugin if they use tensorflow, numpy, etc... Like they already do for django for example.

jacobtylerwalls:
> There is no need to have a separate brain for every possible member.

Oh, whoops, I'm unvolunteering for this part, as it's hairier than I anticipated. The method names correspond to specific function transforms. I think I'm going to take the W here and merge what's here.

Thanks for the reviews!

PR #2181, opened 2023-05-15
Add a bound to the inference tips cache (and fix regressions from #2158)

### LINKED ISSUE #2180: Inference tip cache is unbounded
See [failing pylint primer runs](https://github.com/pylint-dev/pylint/actions/runs/4973881710).

My mac shows 20% performance regression running `time pylint music21`.

bisected to #2158

#### initial ideas:
- the inference tip cache probably needs to be rewritten as an LRU cache.
- or, we need to hash equivalent InferenceContexts somehow (I [started to explore this](https://github.com/pylint-dev/astroid/pull/2158/commits/38955bf992cd0dda6b95b8d48f282a3f0b4a42bb) in #2158 but didn't ultimately pursue it)

jacobtylerwalls:
## Type of Changes

## Overview
Fix some performance regressions from #2158 (**1** and **2** below), and then improve on the status quo by adding a bound to the inference tips cache (**3**).

Closes #2180

## Description
**1**
Not every exit path was removing the recursion guard, so this "cache" meant to always return to empty after recursion concluded was not doing so: 6bdd92d1006ae73bd5747d7c3714e10de81f1818.

**2**
Add an `isEmpty()` method on `InferenceContext` so that `None` can be used as the portion of the cache key for any empty contexts instead of fresh, distinct `InferenceContext` instances: c807c03c48aea8e458a7db3df7f7b35e1cdffc6b.

**3**
Add a bound to the inference tips cache: 4757af207d96fa5ee50a453d452e35af083d49a4.

#### Benchmark for (3.)
Benchmark: `pylint home-assistant/core/homeassistant/helpers`
test code:
```diff
diff --git a/astroid/inference_tip.py b/astroid/inference_tip.py
index c0e121c5..7a484bf6 100644
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -19,6 +19,8 @@ from astroid.typing import (
     TransformFn,
 )
 
+hits = 0
+misses = 0
 _cache: OrderedDict[
     tuple[InferFn[Any], NodeNG, InferenceContext | None], list[InferenceResult]
 ] = OrderedDict()
@@ -52,6 +54,9 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
             context = None
         try:
             yield from _cache[func, node, context]
+            global hits
+            hits += 1
+            print("HITS: ", hits)
             return
         except KeyError:
             # Recursion guard with a partial cache key.
@@ -65,6 +70,9 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
                 result = _cache[func, node, context] = list(
                     func(node, context, **kwargs)
                 )
+                global misses
+                misses += 1
+                print("MISSES: ", misses)
             finally:
                 # Remove recursion guard.
                 _CURRENTLY_INFERRING.remove(partial_cache_key)
```
|maxsize|hits|misses|
|--------|----|-------|
|2|5279|4452|
|64|5946|3533|
|512|5979|3480|

Those are some pretty impressive results for a maxsize of 2! 😱 (I saw the same with several projects.)

For now, I decided to stick with 64.

### Future work
In a subsequent PR we could exploring going all the way down to a cache size of 1 (🙀), and then be able to throw away the ugly code we have here for managing dictionary entries, and just store `last_inference_tip`.

### DIFF
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+from collections import OrderedDict
 from collections.abc import Generator
 from typing import Any, TypeVar
 
@@ -18,9 +19,9 @@ from astroid.typing import (
     TransformFn,
 )
 
-_cache: dict[
+_cache: OrderedDict[
     tuple[InferFn[Any], NodeNG, InferenceContext | None], list[InferenceResult]
-] = {}
+] = OrderedDict()
 
 _CURRENTLY_INFERRING: set[tuple[InferFn[Any], NodeNG]] = set()
 
@@ -61,13 +62,18 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
             _CURRENTLY_INFERRING.add(partial_cache_key)
             try:
                 # May raise UseInferenceDefault
-                result = _cache[func, node, context] = list(func(node, context, **kwargs))
+                result = _cache[func, node, context] = list(
+                    func(node, context, **kwargs)
+                )
             finally:
                 # Remove recursion guard.
                 try:
                     _CURRENTLY_INFERRING.remove(partial_cache_key)
                 except KeyError:
                     pass  # Recursion may beat us to the punch.
+
+                if len(_cache) > 64:
+                    _cache.popitem(last=False)
 
         yield from result
 
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -75,7 +75,8 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
                 if len(_cache) > 64:
                     _cache.popitem(last=False)
 
-        yield from result
+        # https://github.com/pylint-dev/pylint/issues/8686
+        yield from result  # pylint: disable=used-before-assignment
 
     return inner
 
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -46,6 +46,9 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
             # func + node we raise here.
             _CURRENTLY_INFERRING.remove(partial_cache_key)
             raise UseInferenceDefault
+        if context is not None and context.is_empty():
+            # Fresh, empty contexts will defeat the cache.
+            context = None
         try:
             yield from _cache[func, node, context]
             return
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -44,6 +44,7 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
         if partial_cache_key in _CURRENTLY_INFERRING:
             # If through recursion we end up trying to infer the same
             # func + node we raise here.
+            _CURRENTLY_INFERRING.remove(partial_cache_key)
             raise UseInferenceDefault
         try:
             yield from _cache[func, node, context]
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Given the [transient failure](https://github.com/pylint-dev/astroid/actions/runs/4975856690/attempts/1) I just saw and re-ran to pass, we might want to restore the skip from c1e4c95b0d6e555a92484183b89be683dfbde458 (that's still described in the comment).

Pierre-Sassoulas (commented):
Look good ! I missed thé reason why we can't use lru cache after extracting a function ?

astroid/inference_tip.py
```diff
@@ -18,9 +19,9 @@
     TransformFn,
 )
 
-_cache: dict[
+_cache: OrderedDict[
```
DanielNoord, line 22: We are on `3.7+` right? `Dicts` are ordered by default?

jacobtylerwalls:
```python
>>> d = {}
>>> d.popitem(last=False)
TypeError: dict.popitem() takes no keyword arguments
```

jacobtylerwalls: The one thing `OrderedDict` is still good at :-)

jacobtylerwalls:
> Look good ! I missed thé reason why we can't use lru cache after extracting a function ?

Arbitrary **kwargs cannot be hashed in the cache key. So this only works if I omit them:

```diff
diff --git a/astroid/inference_tip.py b/astroid/inference_tip.py
index 9eda5b4f..6829d91b 100644
--- a/astroid/inference_tip.py
+++ b/astroid/inference_tip.py
@@ -6,8 +6,8 @@
 
 from __future__ import annotations
 
-from collections import OrderedDict
 from collections.abc import Generator
+from functools import lru_cache
 from typing import Any, TypeVar
 
 from astroid.context import InferenceContext
@@ -19,10 +19,6 @@ from astroid.typing import (
     TransformFn,
 )
 
-_cache: OrderedDict[
-    tuple[InferFn[Any], NodeNG, InferenceContext | None], list[InferenceResult]
-] = OrderedDict()
-
 _CURRENTLY_INFERRING: set[tuple[InferFn[Any], NodeNG]] = set()
 
 _NodesT = TypeVar("_NodesT", bound=NodeNG)
@@ -30,7 +26,14 @@ _NodesT = TypeVar("_NodesT", bound=NodeNG)
 
 def clear_inference_tip_cache() -> None:
     """Clear the inference tips cache."""
-    _cache.clear()
+    run_func.cache_clear()
+
+
+@lru_cache
+def run_func(func, node, context):
+    _CURRENTLY_INFERRING.add((func, node, context))
+    # Omit **kwargs! 😞
+    return list(func(node, context))
 
 
 def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
@@ -51,32 +54,13 @@ def _inference_tip_cached(func: InferFn[_NodesT]) -> InferFn[_NodesT]:
             # Fresh, empty contexts will defeat the cache.
             context = None
         try:
-            yield from _cache[func, node, context]
-            return
-        except KeyError:
-            # Recursion guard with a partial cache key.
-            # Using the full key causes a recursion error on PyPy.
-            # It's a pragmatic compromise to avoid so much recursive inference
-            # with slightly different contexts while still passing the simple
-            # test cases included with this commit.
-            _CURRENTLY_INFERRING.add(partial_cache_key)
+            yield from run_func(func, node, context)
+        finally:
+            # Remove recursion guard.
             try:
-                # May raise UseInferenceDefault
-                result = _cache[func, node, context] = list(
-                    func(node, context, **kwargs)
-                )
-            finally:
-                # Remove recursion guard.
-                try:
-                    _CURRENTLY_INFERRING.remove(partial_cache_key)
-                except KeyError:
-                    pass  # Recursion may beat us to the punch.
-
-                if len(_cache) > 64:
-                    _cache.popitem(last=False)
-
-        # https://github.com/pylint-dev/pylint/issues/8686
-        yield from result  # pylint: disable=used-before-assignment
+                _CURRENTLY_INFERRING.remove(partial_cache_key)
+            except KeyError:
+                pass  # Recursion may beat us to the punch.
 
     return inner
```

I guess it's a flaw in the status quo that we're leaving out **kwargs from the full cache key, but I think the next step should really be to replace this cache altogether with a reference to the last inference tip (like a cache with maxsize of 1) after doing some more serious benchmarking.

Pierre-Sassoulas (approved):
Ok thanks for the explanation. Having arbitrary kwargs is also a problem imo, albeit not easy to solve.

jacobtylerwalls:
Thanks for the review!

Pulling out from above since it was hiding behind the diff:
> I think the next step should really be to replace this cache altogether with a reference to the last inference tip (like a cache with maxsize of 1) after doing some more serious benchmarking.

But I think we should merge this now so we can unblock pylint alphas. I'll open a follow-up issue.

astroid/inference_tip.py
DanielNoord: TIL :smile:

PR #2182, opened 2023-05-15
Recognize stub ``pyi`` Python files.

mbyrnepr2:
Refs pylint-dev/pylint#4987

## Type of Changes

## Description
The result of this merge-request is to let Pylint analyse  `__init__.pyi` files.

Closes https://github.com/pylint-dev/pylint/issues/4987

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -44,10 +44,10 @@ logger = logging.getLogger(__name__)
 
 
 if sys.platform.startswith("win"):
-    PY_SOURCE_EXTS = ("py", "pyw")
+    PY_SOURCE_EXTS = ("py", "pyw", "pyi")
     PY_COMPILED_EXTS = ("dll", "pyd")
 else:
-    PY_SOURCE_EXTS = ("py",)
+    PY_SOURCE_EXTS = ("py", "pyi")
     PY_COMPILED_EXTS = ("so",)
 
 
@@ -476,7 +476,7 @@ def get_module_files(
             continue
         _handle_blacklist(blacklist, dirnames, filenames)
         # check for __init__.py
-        if not list_all and "__init__.py" not in filenames:
+        if not list_all and not {"__init__.py", "__init__.pyi"}.intersection(filenames):
             dirnames[:] = ()
             continue
         for filename in filenames:
@@ -499,6 +499,8 @@ def get_source_file(filename: str, include_no_ext: bool = False) -> str:
     """
     filename = os.path.abspath(_path_from_filename(filename))
     base, orig_ext = os.path.splitext(filename)
+    if orig_ext == ".pyi" and os.path.exists(f"{base}{orig_ext}"):
+        return f"{base}{orig_ext}"
     for ext in PY_SOURCE_EXTS:
         source_path = f"{base}.{ext}"
         if os.path.exists(source_path):
@@ -663,7 +665,7 @@ def _is_python_file(filename: str) -> bool:
 
     .pyc and .pyo are ignored
     """
-    return filename.endswith((".py", ".so", ".pyd", ".pyw"))
+    return filename.endswith((".py", ".pyi", ".so", ".pyd", ".pyw"))
 
 
 def _has_init(directory: str) -> str | None:
--- a/ChangeLog
+++ b/ChangeLog
@@ -156,6 +156,10 @@ Release date: TBA
 
   Refs pylint-dev/pylint#8554
 
+* Recognize stub ``pyi`` Python files.
+
+  Refs pylint-dev/pylint#4987
+
 
 What's New in astroid 2.15.4?
 =============================
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -163,7 +163,7 @@ class ImportlibFinder(Finder):
 
         for entry in submodule_path:
             package_directory = os.path.join(entry, modname)
-            for suffix in (".py", importlib.machinery.BYTECODE_SUFFIXES[0]):
+            for suffix in (".py", ".pyi", importlib.machinery.BYTECODE_SUFFIXES[0]):
                 package_file_name = "__init__" + suffix
                 file_path = os.path.join(package_directory, package_file_name)
                 if os.path.isfile(file_path):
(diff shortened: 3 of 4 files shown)


astroid/modutils.py
```diff
@@ -476,7 +476,7 @@ def get_module_files(
             continue
         _handle_blacklist(blacklist, dirnames, filenames)
         # check for __init__.py
-        if not list_all and "__init__.py" not in filenames:
+        if not list_all and not {"__init__.py", "__init__.pyi"}.intersection(filenames):
```
jacobtylerwalls, line 479: For performance you can just:
```suggestion
        if not list_all and {"__init__.py", "__init__.pyi"}.isdisjoint(filenames):
```

astroid/modutils.py
```diff
@@ -499,6 +499,8 @@ def get_source_file(filename: str, include_no_ext: bool = False) -> str:
     """
     filename = os.path.abspath(_path_from_filename(filename))
     base, orig_ext = os.path.splitext(filename)
+    if orig_ext == ".pyi" and os.path.exists(f"{base}{orig_ext}"):
+        return f"{base}{orig_ext}"
```
jacobtylerwalls, line 503: Happy to approve once this line is covered!

astroid/modutils.py
mbyrnepr2: Good one!

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Great !

sighingnow:
Only recognizing `__init__.pyi` is not sufficient for many cases, as `ImportlibFinder._SUFFIXES` doesn't contain `.pyi` so interface files like `a/b/c.pyi` won't take into effect.

Could we get such case fixed as well? 

/cc @Pierre-Sassoulas @jacobtylerwalls

mbyrnepr2:
Thank you @sighingnow. I think I would need a small working example to be sure I understand what the issue is. 
Let me see if I understood you correctly with the following example. Does this illustrate the issue from your point of view?:

```python
(venv310) $ tree not_working_ok/
not_working_ok/
└── b
    └── c.pyi

1 directory, 1 file
(venv310) $ tree working_ok/
working_ok/
└── b
    └── c.py

1 directory, 1 file
(venv310) $ cat working_ok/b/c.py 
x = 42
(venv310) $ diff working_ok/b/c.py not_working_ok/b/c.pyi 
(venv310) $ 
(venv310) $ pylint working_ok/
************* Module b.c
working_ok/b/c.py:1:0: C0114: Missing module docstring (missing-module-docstring)
working_ok/b/c.py:1:0: C0103: Constant name "x" doesn't conform to UPPER_CASE naming style (invalid-name)
working_ok/b/c.py:1:0: W0612: Unused variable 'x' (unused-variable)

-----------------------------------
Your code has been rated at 0.00/10

(venv310) $ pylint not_working_ok/
(venv310) $ <should be pylint messages here but there is not>
(venv310) $ 
(venv310) Marks-MacBook-Air-2:programming markbyrne$ pylint not_working_ok/b/c.pyi 
************* Module c
not_working_ok/b/c.pyi:1:0: C0114: Missing module docstring (missing-module-docstring)
not_working_ok/b/c.pyi:1:0: C0103: Constant name "x" doesn't conform to UPPER_CASE naming style (invalid-name)
not_working_ok/b/c.pyi:1:0: W0612: Unused variable 'x' (unused-variable)
```

Note that while looking into this I see that https://github.com/pylint-dev/pylint/blob/main/pylint/lint/pylinter.py#L603 will need to be updated to include "pyi" also; that will fix my example above although I'm not 100% sure if it matches your use-case.

mbyrnepr2:
Perhaps you could open a pull-request with your working hack @sighingnow because I've tried this also and I don't see it working unfortunately.
In the meantime I'll open a pull-request in Pylint because I think that will be necessary.

sighingnow:
Hi @mbyrnepr2,

I have opened the pull request:

I misunderstood the case above and would like to give a minimal example of my problem (sorry for the incorrect comment above).

I have (I use `head` to cat all files):

```
..../working_ok $ head -n 100 ./* 
==> ./c.py <==
def _init():
    import sys
    setattr(sys.modules[__name__], 'x', 42)
_init()

==> ./c.pyi <==
x: int = 42

==> ./d.py <==
from . import c
print('x from c: ', c.x)

==> ./__init__.py <==
```

From the case above you could see that `c` has defined `x` with an init function (protobuf works in a similar way) and we tell that `c` has `x` in the pyi file `c.pyi`. Without the hack in #2195, pylint says:

```bash
$ pylint d.py
************* Module working_ok.d
d.py:1:0: C0114: Missing module docstring (missing-module-docstring)
d.py:2:20: E1101: Module 'working_ok.c' has no 'x' member (no-member)

```

After the hack in #2195, the error disappears:

```bash
$ pylint d.py
************* Module working_ok.d
d.py:1:0: C0114: Missing module docstring (missing-module-docstring)

Your code has been rated at 5.00/10 (previous run: 0.00/10, +5.00)
```

Note the hack in #2195 only resolves the problem above, and is not enough to support other cases (not very familiar with the internal of pylint).

sighingnow:
Further tweaks the case above (adding `y` to `d.py` which doesn't exists in `c.pyi` as well), now pylint complaints as expected for `y`:

```bash
$ head -n 100 ./*
==> ./c.py <==
def _init():
    import sys
    setattr(sys.modules[__name__], 'x', 42)
_init()

==> ./c.pyi <==
x: int = 42

==> ./d.py <==
from . import c
print('x from c: ', c.x)
print('y from c: ', c.y)

==> ./__init__.py <==
```

mbyrnepr2:
So, in other words if a module `c.py` creates a name `x` dynamically (which Pylint typically won't be able to understand) then Pylint could check if the name exists in a `c.pyi` module and if it does, trust the `c.pyi` as the source of truth that the `x` name exists in `c.py` and hence there would be less `no-member` false positives. Correct me if I've misunderstood :) 

To be honest I didn't consider that dynamically assigned names would be in-scope for https://github.com/pylint-dev/pylint/issues/4987. My understanding is that the scope was more narrow than that - i.e. to make it possible for Pylint to be able to lint .pyi files (previously they were ignored or not picked-up by astroid when searching for modules).

@Pierre-Sassoulas @jacobtylerwalls what do you think?

sighingnow:
> to make it possible for Pylint to be able to lint .pyi files (previously they were ignored or not picked-up by astroid when searching for modules).

Agree that lint both `.py` and `.pyi` would be the correct solution. See also related issue: https://github.com/pylint-dev/pylint/issues/6281

PR #2185, opened 2023-05-18
Avoid duplicate inference results for `List[int]`

jacobtylerwalls:
## Type of Changes

## Description
Fix a regression in 3.0.0.a3 that became necessary once we fixed the cache key in 0740a0dd5e9cb48bb1a400aded498e4db1fcfca9.

If we re-run the inference function for inferring a slice like `List[int]`, we risk creating so many duplicates, especially if used in several files, that we hit the limit for max inference and end up introducing an `Uninferable`, which can cause some pylint checks to not fire.

This explains (and resolves!) the failures seen in https://github.com/pylint-dev/pylint/pull/8685#issuecomment-1546980683, which we were only seeing when running the test suite as a whole.

Skipping news because we're not writing changelogs specific to each alpha.

This should unblock https://github.com/pylint-dev/pylint/pull/8685.

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -316,6 +316,9 @@ def infer_typing_alias(
         # This is an issue in cases where the aliased class implements it,
         # but the typing alias isn't subscriptable. E.g., `typing.ByteString` for PY39+
         _forbid_class_getitem_access(class_def)
+
+    # Avoid re-instantiating this class every time it's seen
+    node._explicit_inference = lambda node, context: iter([class_def])
     return iter([class_def])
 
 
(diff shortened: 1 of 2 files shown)


astroid/brain/brain_typing.py
```diff
@@ -316,6 +316,9 @@ def infer_typing_alias(
         # This is an issue in cases where the aliased class implements it,
         # but the typing alias isn't subscriptable. E.g., `typing.ByteString` for PY39+
         _forbid_class_getitem_access(class_def)
+
+    # Avoid re-instantiating this class every time it's seen
+    node._explicit_inference = lambda node, context: iter([class_def])
```
DanielNoord, line 321: This feels like something we might need to do for a lot of brains.

We us a similar pattern for the `enum` brain. :sweat:

DanielNoord (approved):

astroid/brain/brain_typing.py
jacobtylerwalls: You're quite right. I traced the bulk of the primer noise to needing to do this in other places in the typing brain.

PR #2187, opened 2023-05-20
Optimize `_compute_mro()`

jacobtylerwalls:

## Description
- Optimize `clean_duplicates_mro()`
- Short-circuit in `_compute_mro()` for "builtins.object"

Performance improvement is more noticeable in projects making heavy use of inheritance. For the project `music21` from the pylint primer:

#### Benchmark
```
from pylint.lint import Run
Run(['music21'], exit=False)  # using project's own rcfile
```
**main**

```
452811674 function calls (383785563 primitive calls) in 161.566 seconds
...
427785/97707    0.934    0.000   10.187    0.000 scoped_nodes.py:2808(_compute_mro)
```
**PR**
```
450362344 function calls (381157337 primitive calls) in 157.018 seconds
...
427785/97707    0.692    0.000    8.160    0.000 scoped_nodes.py:2808(_compute_mro)
```

That's ~2s saved out of ~160s, or 1.25%.

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2806,6 +2806,9 @@ class ClassDef(
                 yield from baseobj.bases
 
     def _compute_mro(self, context: InferenceContext | None = None):
+        if self.qname() == "builtins.object":
+            return [self]
+
         inferred_bases = list(self._inferred_bases(context=context))
         bases_mro = []
         for base in inferred_bases:
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -133,24 +133,24 @@ def clean_typing_generic_mro(sequences: list[list[ClassDef]]) -> None:
     bases_mro.pop(position_in_inferred_bases)
 
 
-def clean_duplicates_mro(sequences, cls, context):
+def clean_duplicates_mro(
+    sequences: Iterable[Iterable[ClassDef]],
+    cls: ClassDef,
+    context: InferenceContext | None,
+) -> Iterable[Iterable[ClassDef]]:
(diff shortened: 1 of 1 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -133,24 +133,24 @@ def clean_typing_generic_mro(sequences: list[list[ClassDef]]) -> None:
     bases_mro.pop(position_in_inferred_bases)
 
 
-def clean_duplicates_mro(sequences, cls, context):
+def clean_duplicates_mro(
+    sequences: Iterable[Iterable[ClassDef]],
```
DanielNoord, line 137: https://github.com/pylint-dev/astroid/blob/feafcb83cf5b379cdb7613f48268a7525a591f40/astroid/nodes/scoped_nodes/scoped_nodes.py#L2806

This makes this `SuccesfulInferenceResult` although I wonder if `ClassDef.bases` could be narrowed down. I'd be okay with making this `ClassDef` for now and fixing later if necessary, but note that this will add one more `mypy` warning after `_inferred_bases` gets its signature typed.

DanielNoord (approved):

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: Yep. However, Pyright also says it could be `ClassDef` due to this:

https://github.com/pylint-dev/astroid/blob/feafcb83cf5b379cdb7613f48268a7525a591f40/astroid/nodes/scoped_nodes/scoped_nodes.py#L2803-L2804

Let's wait until we're ready to type this all together.

DanielNoord: Yeah of course, the `bases` line just broadens the type (perhaps/probably incorrectly)

PR #2198, opened 2023-06-06
Verify nodes' str() and repr() don't raise errors/warnings

### LINKED ISSUE #1881: Some `__str__` methods of nodes raise errors and warnings
### Steps to reproduce
```python
>>> fd = astroid.nodes.FunctionDef()
>>> str(fd)
  File "/Users/jacobwalls/astroid/astroid/nodes/node_ng.py", line 219, in __str__
    value = getattr(self, field)
            ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'FunctionDef' object has no attribute 'args'
```

Less severely, but still not great, launching python with warnings e.g. `python -Wall`:
```python
>>> cd = astroid.nodes.ClassDef()
>>> str(cd)
/Users/.../astroid/astroid/nodes/scoped_nodes/scoped_nodes.py:2041: DeprecationWarning: The 'ClassDef.doc' attribute is deprecated, use 'ClassDef.doc_node' instead.
```

***
We should add a unittest that generates the `str()` and `repr()` of all node types.

2.12.13

crazybolillo:
Calling str() or repr() on certain nodes fails either with errors or warnings. This commit adds unittests to find and reproduce this behavior with the endgoal of fixing it.

Closes #1881

## Type of Changes

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -210,7 +210,7 @@ class NodeNG:
             alignment = len(cname) + 1
         result = []
         for field in self._other_fields + self._astroid_fields:
-            value = getattr(self, field)
+            value = getattr(self, field, "Unknown")
             width = 80 - len(field) - alignment
             lines = pprint.pformat(value, indent=2, width=width).splitlines(True)
 
@@ -227,6 +227,11 @@ class NodeNG:
 
     def __repr__(self) -> str:
         rname = self.repr_name()
+        # The dependencies used to calculate fromlineno (if not cached) may not exist at the time
+        try:
+            lineno = self.fromlineno
+        except AttributeError:
+            lineno = 0
         if rname:
             string = "<%(cname)s.%(rname)s l.%(lineno)s at 0x%(id)x>"
         else:
@@ -234,7 +239,7 @@ class NodeNG:
         return string % {
             "cname": type(self).__name__,
             "rname": rname,
-            "lineno": self.fromlineno,
+            "lineno": lineno,
             "id": id(self),
         }
 
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -7,7 +7,9 @@
 from __future__ import annotations
 
 import copy
+import inspect
 import os
+import random
 import sys
 import textwrap
 import unittest
@@ -1880,3 +1882,35 @@ class TestPatternMatching:
         inferred = node.inferred()
         assert len(inferred) == 2
         assert [inf.value for inf in inferred] == [10, -1]
+
+
+@pytest.mark.parametrize(
+    "node",
+    [
+        node
+        for node in astroid.nodes.ALL_NODE_CLASSES
+        if node.__name__
+        not in ["_BaseContainer", "BaseContainer", "NodeNG", "const_factory"]
+    ],
+)
+@pytest.mark.filterwarnings("error")
+def test_str_repr_no_warnings(node):
+    parameters = inspect.signature(node.__init__).parameters
+
+    args = {}
+    for name, param_type in parameters.items():
+        if name == "self":
+            continue
+
+        if "int" in param_type.annotation:
+            args[name] = random.randint(0, 50)
+        elif "NodeNG" in param_type.annotation:
+            args[name] = nodes.Unknown()
+        elif "str" in param_type.annotation:
+            args[name] = ""
+        else:
+            args[name] = None
+
+    test_node = node(**args)
+    str(test_node)
+    repr(test_node)


crazybolillo:
I changed the test flavor to pytest and fixed all `str` and `repr` issues found in the tests. After all I had to add some more failsafes in other places.

crazybolillo:
Also something weird is going on with the PR, since it is trying to reapply a commit already found in the main branch (same hash and everything), I guess it has to do with me not having some sort of permission to update the PR or execute a certain flow.

Pierre-Sassoulas:
Could you rebase or merge main in your branch ?

crazybolillo:
> Could you rebase or merge main in your branch ?

It should be. My fork shows my branch being just one commit ahead and none behind. :open_mouth:

tests/test_nodes.py
```diff
@@ -1880,3 +1882,36 @@ def return_from_match(x):
...
+@pytest.mark.filterwarnings("error")
+def test_str_repr_no_warnings(node):
+    parameters = inspect.signature(node.__init__).parameters
+
+    args = {}
+    for name, param_type in parameters.items():
+        if name == "self":
+            continue
+
+        if "int" in param_type.annotation:
```
DanielNoord, line 1906: I like this trick, never thought of doing it like this!

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1350,7 +1350,7 @@ def fromlineno(self) -> int:
         # lineno is the line number of the first decorator, we want the def
         # statement lineno. Similar to 'ClassDef.fromlineno'
         lineno = self.lineno or 0
-        if self.decorators is not None:
+        if getattr(self, "decorators", None) is not None:
```
DanielNoord, line 1353: I'd prefer just setting `decorators` in the ``__init__``. Since it can be `None` anyway this doesn't interfere with the typing of the attribute.

astroid/nodes/node_ng.py
```diff
@@ -246,7 +246,7 @@ def accept(self, visitor):
     def get_children(self) -> Iterator[NodeNG]:
         """Get the child nodes below this node."""
         for field in self._astroid_fields:
-            attr = getattr(self, field)
+            attr = getattr(self, field, None)
```
DanielNoord, line 249: For which of the nodes is this necessary? `get_children` isn't only used in `__str__` so I'm worried people might depend on this returning an `AttributeError`.

astroid/nodes/node_ng.py
```diff
@@ -210,7 +210,7 @@ def __str__(self) -> str:
             alignment = len(cname) + 1
         result = []
         for field in self._other_fields + self._astroid_fields:
-            value = getattr(self, field)
+            value = getattr(self, field, "Unknown")
```
DanielNoord, line 213: :+1:

astroid/nodes/node_classes.py
```diff
@@ -859,33 +859,33 @@ def find_argname(self, argname, rec=False):
         return None, None
 
     def get_children(self):
```
DanielNoord, line 861: See other comment. If we could find a way where we only impact `__str__` and `__repr__` I would be very much in favour of that.

DanielNoord (changes requested):
Thanks for doing this! I really like the test you came up with!

crazybolillo:
Ah I see, makes sense to only touch `__str__` and `__repr__`.  Apparently the only node that had issues after patching `value = getattr(self, field)` was `MatchCase`.

The error is the following:
`AttributeError: 'MatchCase' object has no attribute 'pattern'`

I solved inside `__repr__` so nothing else is modified now.

Thanks for the review :handshake:

astroid/nodes/node_ng.py
```diff
@@ -227,14 +227,18 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         rname = self.repr_name()
+        try:
+            lineno = self.fromlineno
+        except AttributeError:
```
DanielNoord, line 232: Shall we add a comment like:
`# If the __postinit__ isn't called `fromlineno` could not exist`.

Although I must say, shouldn't `fromlineno` as property always exist?

crazybolillo: As far as my understanding goes, `fromlineno` does always exist. The problem is that it calls `fixed_source_line` and thats where the `AttributeError` is generated, specifically it seems to come from `get_children`.

I ran this on a different dev machine and got the error for a few other nodes besides `MatchCase` (weird). Here are some sample stack traces:

```
self = <[AttributeError("'Arguments' object has no attribute 'posonlyargs'") raised in repr()] Arguments object at 0x7f4c48347f90>

    def get_children(self):
>       yield from self.posonlyargs or ()
E       AttributeError: 'Arguments' object has no attribute 'posonlyargs'
```

```
@cached_property
def fromlineno(self) -> int:
....
>       if self.decorators is not None:
E       AttributeError: 'AsyncFunctionDef' object has no attribute 'decorators'
```

I guess the comment could be something like `The dependencies used to calculate fromlineno (if not cached) may not exist at the time`?

crazybolillo: Adding to my previous comment, in more general terms it seems `fromlineno` may execute certain function calls in an attempt to obtain the line number, and the resources used in said functions may not exist at the time of the read attempt.

DanielNoord: Yeah you suggestion for the comment seems fine! Thanks!

tests/test_nodes.py
```diff
@@ -1880,3 +1882,36 @@ def return_from_match(x):
...
+
+@pytest.mark.parametrize(
+    "node",
+    list(
+        filter(
+            lambda node: node.__name__
+            not in ["_BaseContainer", "BaseContainer", "NodeNG", "const_factory"],
+            astroid.nodes.ALL_NODE_CLASSES,
+        )
+    ),
```
jacobtylerwalls, line 1895: Won't block this merge over it, but stylistically I'd suggest a comprehension instead of listifying a filter.

```suggestion
    [
        node
        for node in astroid.nodes.ALL_NODE_CLASSES
        if node.__name__
        not in ["_BaseContainer", "BaseContainer", "NodeNG", "const_factory"]
    ],
```

jacobtylerwalls (commented):
Thanks for taking this on! Will approve and merge once the comment is added that's discussed above.

DanielNoord (approved):
Thanks for picking this up!

PR #2209, opened 2023-06-11
Fix interrupted `InferenceContext` call chains

jacobtylerwalls:
## Type of Changes

## Description

`ClassDef.getitem()` and `infer_argument()` both had interrupted call chains where the InferenceContext wasn't passed all the way through to `infer()`. This caused performance problems in packages such as `sqlalchemy` needing these features.

Closes pylint-dev/pylint#8150

Linting the example from pylint-dev/pylint#8150 now takes half as much time.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,11 @@ Release date: TBA
   Closes pylint-dev/pylint#7464
   Closes pylint-dev/pylint#8074
 
+* Fix interrupted ``InferenceContext`` call chains, thereby addressing performance
+  problems when linting ``sqlalchemy``.
+
+  Closes pylint-dev/pylint#8150
+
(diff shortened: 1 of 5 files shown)


DanielNoord (approved):
I'd love to talk about your strategy finding these fixes. Seems like you have a better strategy than I have 😄

jacobtylerwalls:
1. @mbyrnepr2 pointed me in the right direction to the enum brain
2. profile showed that is_enum_subclass() was the culprint, and the methods around mro() were taking about as long
3. wasted some effort trying to cache is_enum_subclass()
4. realized if caching wasn't working, then the InferenceContext was unnecessarily being refreshed
5. bunch of print statements to find out why the InferenceContexts were not the same
6. set a breakpoint and look at the call stack, find where they went missing! 😄

PR #2216, opened 2023-06-19
Infer function parameter types using annotations

ringohoffman:
## Type of Changes

## Description

Fixes https://github.com/pylint-dev/pylint/issues/4813
Fixes https://github.com/pylint-dev/pylint/issues/8781

Type annotations are not currently used for inferencing, which means that, for example, functions with parameter typed as a `logging.Logger` would not receive relevant pylint warnings such as `logging-fstring-interpolation`. 

To make use of these type annotations, we now modify `infer_assign` to add yield additional `Instance`s if an `AssignName` is being processed whose `parent` is an `Arguments` with annotations for the `AssignName`.

cc: @Pierre-Sassoulas

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@ What's New in astroid 3.0.0?
 =============================
 Release date: TBA
 
+* Use type annotations for type inference.
+
+  Closes pylint-dev/pylint#4813
+  Closes pylint-dev/pylint#8781
+
 * Remove support for Python 3.7.
 
   Refs #2137
--- a/astroid/inference.py
+++ b/astroid/inference.py
@@ -1168,10 +1168,30 @@ def infer_assign(
     assign node.
     """
     if isinstance(self.parent, nodes.AugAssign):
-        return self.parent.infer(context)
+        yield from self.parent.infer(context)
+        return
 
     stmts = list(self.assigned_stmts(context=context))
-    return bases._infer_stmts(stmts, context)
+    yield from bases._infer_stmts(stmts, context)
+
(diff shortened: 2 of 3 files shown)


DanielNoord:
The issue with this approach is that we have no way of knowing that the annotations are correct. Personally I wouldn't like `pylint` to assume all annotations are correct and think this should be handled with some sort of flag (`--consider-annotations-are-correct`).
Just turning this on by default doesn't seem like the best solution to me.

I'm wondering what other maintainers thinks of this.

ringohoffman:
> I think this should be handled with some sort of flag (`--consider-annotations-are-correct`).

I support this. If we could assume hinted return types were correct, I would think we could save a lot of time on inferencing, assuming we could skip manual inferencing.

> Personally I wouldn't like `pylint` to assume all annotations are correct

I understand that `astroid` was developed to infer types before type hints were introduced to Python. However, trusting type hints is what most type checkers do. Having to infer the return types even on standard library functions would seem like a major reason that `pylint` takes, for example, ~25 minutes to run on [`pytorch/torch`](https://github.com/pytorch/pytorch/tree/main/torch).

Pierre-Sassoulas:
We need to be careful here, using inference and not testing annotations is the differentiating factor that makes pylint find issues that other tools don't in partially typed code base.

On the other hand it would be nice to be able to detect that we're in a fully typed code base where we should trust the typing. This is not going to be trivial, unless we ask users to tell us via options. It seems possible to take typing in builtin into account without any drawbacks and without detecting anything or adding an option. (Because we know that typing are truthful and reliable in builtin)

PR #2217, opened 2023-06-19
Deprecate `rec` argument to `find_argname()`

jacobtylerwalls:
## Type of Changes

## Description
Follow-up to
3db2bdd6ebfc49db5ed7f379a7c8388849b97262.

(More unreachable code to remove, and avoid unnecessary isinstance tuple checks.)

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -10,6 +10,7 @@ import abc
 import itertools
 import sys
 import typing
+import warnings
 from collections.abc import Generator, Iterable, Iterator, Mapping
 from functools import cached_property, lru_cache
 from typing import (
@@ -506,6 +507,8 @@ class Name(_base_nodes.NoChildrenNode, LookupMixIn):
             yield from child_node._get_name_nodes()
 
 
+DEPRECATED_ARGUMENT_DEFAULT = object()
+
 class Arguments(_base_nodes.AssignTypeNode):
     """Class representing an :class:`ast.arguments` node.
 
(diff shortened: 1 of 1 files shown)


astroid/nodes/node_classes.py
```diff
@@ -836,26 +840,28 @@ def is_argument(self, name) -> bool:
...
         :type argname: str
 
-        :param rec: Whether or not to include arguments in unpacked tuples
-            in the search.
-        :type rec: bool
-
         :returns: The index and node for the argument.
         :rtype: tuple(str or None, AssignName or None)
         """
+        if rec is not DEPRECATED_ARGUMENT_DEFAULT:
```
Pierre-Sassoulas, line 857: That's pretty clever !

astroid/nodes/node_classes.py
```diff
@@ -836,26 +840,28 @@ def is_argument(self, name) -> bool:
...
+            warnings.warn(
+                "The rec argument will be removed in a future version.",
```
Pierre-Sassoulas, line 859:
```suggestion
                "The rec argument will be removed in astroid 4.0.",
```

Pierre-Sassoulas (changes requested):

DanielNoord (commented):
I would be okay with removing this in `3.x`. Any reason not to?

jacobtylerwalls:
No strong opinion here. I only just barely discovered it, so it didn't seem to hurt much to do one deprecation cycle.

jacobtylerwalls:
Oh, did you mean removing in astroid 3.1? Then we would leave the message as is. That's okay with me.

DanielNoord:
> Oh, did you mean removing in astroid 3.1? Then we would leave the message as is. That's okay with me.

I think we only remove in major releases, but if we mention `3.1` in the warning I think removing in `3.1` is fine as well!

Pierre-Sassoulas:
No that Jacob bothered with deprecating we might as well do the deprecation cycle but I think it's also ok to remove directly for this kind of niche API. (Very probable that no one use it even pylint might use it very little)

astroid/nodes/node_classes.py
jacobtylerwalls:
```suggestion
                "The rec argument will be removed in astroid 3.1.",
```

astroid/nodes/node_classes.py
jacobtylerwalls, line 857:
```suggestion
        if rec is not DEPRECATED_ARGUMENT_DEFAULT:  # pragma: no cover
```

Pierre-Sassoulas (approved):

PR #2223, opened 2023-06-23
Prefer stdlib modules over same-named modules on sys.path

jacobtylerwalls:
## Type of Changes

## Description
Prefer stdlib modules over same-named modules on sys.path

For example:
`import copy` now finds `copy` instead of `copy.py`.

This worked correctly before in at least some cases if there was a (more?) complete chain of `__init__.py` files from cwd all the way to the location of the `copy.py` module.

Closes pylint-dev/pylint#6535


ChangeLog
```diff
@@ -181,6 +181,11 @@ What's New in astroid 2.15.6?
 =============================
 Release date: TBA
 
+* Prefer standard library modules over same-named modules on sys.path. For example
```
jacobtylerwalls, line 184: Ah, Move this up to 3.0.

Pierre-Sassoulas (dismissed):

astroid/interpreter/_import/spec.py
```diff
@@ -157,6 +157,19 @@ def find_module(
                         location=getattr(spec.loader_state, "filename", None),
                         type=ModuleType.PY_FROZEN,
                     )
+                if (
```
DanielNoord, line 160: Could this clash with `namespace` modules?

jacobtylerwalls: Could you flesh this out a little more for me? (Sorry!)

DanielNoord:
```console
❯ tree .
.
├── argparse.py
└── test.py

1 directory, 2 files
```

```python
# test.py

import argparse

argparse.i_dont_exist()
```

```python
# argparse.py

def i_dont_exist():
    print("1")
```

```console
❯ python test.py
1
```

Won't this implementation default to `stdlib.argparse` instead of the local `argparse`?

jacobtylerwalls: Nope, set up this example and it's looking good 😎 

I added a call to argparse.ArgumentParser to make sure it's unavailable:

```
************* Module test
test.py:8:0: C0305: Trailing newlines (trailing-newlines)
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:6:0: E1101: Module 'argparse' has no 'ArgumentParser' member (no-member)

```

jacobtylerwalls: Oops, I didn't run the test case against the correct copy of astroid. More soon.

PR #2224, opened 2023-06-26
Optimize exception handling in import_module()

jacobtylerwalls:
## Type of Changes

## Description
Small performance boost. Don't repeat the same import operation for modules like `ntpath` that only exist on certain platforms (e.g. windows).

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -445,6 +445,10 @@ class Module(LocalsDictNodeNG):
             # skip here
             if relative_only:
                 raise
+            # Don't repeat the same operation, e.g. for missing modules
+            # like "_winapi" or "nt" on POSIX systems.
+            if modname == absmodname:
+                raise
         return AstroidManager().ast_from_module_name(modname)
 
     def relative_to_absolute_name(self, modname: str, level: int | None) -> str:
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
LGTM

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -445,6 +445,10 @@ def import_module(
             # skip here
             if relative_only:
                 raise
+            # Don't repeat the same operation, e.g. for missing modules
+            # like "_winapi" or "nt" on POSIX systems.
+            if modname == absmodname:
```
DanielNoord, line 450: Is there a difference between `absmodname` and `use_cache==False` and using `modname`? That would be a change in behaviour now.

jacobtylerwalls: This occurred to me also. Looks like we missed this in #1747 (we should have added use_cache=use_cache in the final line of the function). Are you okay with me making that in change in this PR or would you like a separate one?

DanielNoord: Let's do it in this one. This is fairly straightforward so should be fine to increase scope a little!

DanielNoord (approved):

jacobtylerwalls:
Thanks for the reviews!

PR #2214, opened 2023-06-18
Improve typing of builtins brain

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description
Improve typing of this brain module.

Resolves 14 mypy errors.

PR #2225, opened 2023-06-26
Further improve typing of builtins brain

jacobtylerwalls:
## Type of Changes

## Description
Resolves 12 mypy errors (including all for this module!)

Refs #2214


Pierre-Sassoulas (dismissed):

astroid/brain/brain_builtin_inference.py
```diff
@@ -212,7 +217,7 @@ def _transform_wrapper(node, context: InferenceContext | None = None):
 
     AstroidManager().register_transform(
         nodes.Call,
-        inference_tip(_transform_wrapper),
+        inference_tip(cast(InferFn, _transform_wrapper)),
```
DanielNoord, line 220: Are we not able to just change the signature instead of casting? Should be relatively straightforward?

jacobtylerwalls: Signatures are fine (well, other than two of them omitting **kwargs, but that didn't seem to be the issue), so I didn't find it straightforward (😅), but perhaps you can help!

```
astroid/brain/brain_builtin_inference.py:219: error: Argument 1 to "inference_tip" has incompatible type "Callable[[Any, InferenceContext | None], Any]"; expected "InferFn[<nothing>]"  [arg-type]
astroid/brain/brain_builtin_inference.py:1080: error: Argument 1 to "inference_tip" has incompatible type "Callable[[Any, InferenceContext | None], Any]"; expected "InferFn[<nothing>]"  [arg-type]
astroid/brain/brain_builtin_inference.py:1086: error: Argument 1 to "inference_tip" has incompatible type "Callable[[Call, InferenceContext | None, KwArg(Any)], Iterator[NodeNG | UninferableBase | Proxy]]"; expected "InferFn[<nothing>]"  [arg-type]
astroid/brain/brain_builtin_inference.py:1093: error: Argument 1 to "inference_tip" has incompatible type "Callable[[Call, InferenceContext | None, KwArg(Any)], Iterator[astroid.nodes.node_classes.List | astroid.nodes.node_classes.Set | astroid.nodes.node_classes.Tuple | astroid.nodes.node_classes.Dict | Const | EmptyNode | UninferableBase]]"; expected "InferFn[<nothing>]"  [arg-type]
```

I'm thinking it's a mypy issue, but I'm not certain which one.

jacobtylerwalls: Hold on, the **kwargs did fix two of them, looking into the next two...

jacobtylerwalls: Couldn't figure it out. There are so many open issues in mypy for unions and generics that I didn't narrow it down to one.

jacobtylerwalls: Added two disables in 
e223578

astroid/brain/brain_builtin_inference.py
```diff
@@ -127,6 +134,10 @@ def ljust(self, width, fillchar=None):
 """
 
 
+def _use_default() -> NoReturn:
```
jacobtylerwalls, line 137:
```suggestion
def _use_default() -> NoReturn:  # pragma: no cover
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -257,10 +270,12 @@ def _container_generic_transform(
     iterables: tuple[type[nodes.BaseContainer] | type[ContainerObjects], ...],
     build_elts: BuiltContainers,
 ) -> nodes.BaseContainer | None:
+    elts: Iterable
+
     if isinstance(arg, klass):
         return arg
     if isinstance(arg, iterables):
-        arg = cast(ContainerObjects, arg)
+        arg = cast(nodes.BaseContainer, arg)
```
DanielNoord, line 278: Both casts here are incorrect.

If `klass` is a specific subclass of `BaseContainer` (which it can be) then `arg` can still be any of the other subclasses of `BaseContainer`. I think we should remove the cast.

astroid/brain/brain_builtin_inference.py
```diff
@@ -257,10 +270,12 @@ def _container_generic_transform(
     iterables: tuple[type[nodes.BaseContainer] | type[ContainerObjects], ...],
     build_elts: BuiltContainers,
 ) -> nodes.BaseContainer | None:
+    elts: Iterable
```
DanielNoord, line 273: This contradicts line 301.

astroid/brain/brain_builtin_inference.py
```diff
@@ -399,6 +415,7 @@ def infer_dict(node: nodes.Call, context: InferenceContext | None = None) -> nod
     args = call.positional_arguments
     kwargs = list(call.keyword_arguments.items())
 
+    items: list[tuple[SuccessfulInferenceResult, SuccessfulInferenceResult]]
```
DanielNoord, line 418: Line 378 contradicts this, although that might be too broad.

astroid/brain/brain_builtin_inference.py
```diff
@@ -1075,11 +1094,13 @@ def _infer_str_format_call(
 
 AstroidManager().register_transform(
     nodes.Call,
-    inference_tip(_infer_copy_method),
+    inference_tip(_infer_copy_method),  # type: ignore[arg-type]
```
DanielNoord, line 1097: This can be avoided by using a `Generator` instead of an `Iterator` on the signature of this function.

Same goes for the other one, but that then creates other issues.. Can we yield from in that function? That would make it an iterator.

DanielNoord (changes requested):

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: What if I cast it to the type that `iterables` has. (Seems like here, I'm missing ContainerObjects).

DanielNoord: That shouldn't be necessary right? That's what the `isinstance` does?

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Line 301 is the `klass` argument? Having trouble following.

DanielNoord: Oops, sorry! New code line 300

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Unfortunately, no:
```python
reveal_type(arg)  # Type of arg: is NodeNG | Proxy
```

jacobtylerwalls:
```
astroid/brain/brain_builtin_inference.py:270: error: Item "NodeNG" of "NodeNG | Proxy" has no attribute "elts"  [union-attr]
astroid/brain/brain_builtin_inference.py:271: error: Item "NodeNG" of "NodeNG | Proxy" has no attribute "elts"  [union-attr]
astroid/brain/brain_builtin_inference.py:275: error: Item "NodeNG" of "NodeNG | Proxy" has no attribute "elts"  [union-attr]
```

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Got it now, thanks!

astroid/brain/brain_builtin_inference.py
DanielNoord: Hmm, that feels like a bug but `pyright` agrees with `mypy`. So I'm wondering if we're doing something wrong.. Let's do the cast to both types indeed, at least that is more correct than it was before.

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Nice catch! We should probably broaden it out. That means this is probably also too narrow:

https://github.com/pylint-dev/astroid/blob/d4f4452fe089f600bf9144ffdcd8e698816df3c2/astroid/nodes/node_classes.py#L294-L295

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Sorry, one more that I'm having trouble following :-)

Elsewhere we return iterators just fine, and they are typed this way:

https://github.com/pylint-dev/astroid/blob/e3ba1cadd4237d96d2e6c9a5341c7b652d389038/astroid/bases.py#L449-L453

So why would we change to Generator?

DanielNoord: Because `InferFn` expects a `Generator`. I think because we define it so explicitly we need to do so here as well.. I think we might also get away with changing `InferFn` to be an `Iterator` but since we don't actually run `mypy` that might create issues somewhere else that we don't know about..

jacobtylerwalls: Ho! Got it now. Okay. I'll audit the uses of `InferFn`. It's start took like like it should be `Iterator`.

> but since we don't actually run mypy that might create issues somewhere else that we don't know about..

That's why I'm trying to focus by file and make sure with each change the total number goes down and no new issues in the single file I'm focusing on (but yes... whackamole could ensue...)

DanielNoord: Yeah, that does make the PRs much easier to review.

What I have also noticed is that doing PRs per concept (such as `InferFn`) is also a good way of grouping things. It might make the big number bigger as the improved typing exposes more issues, but it makes review and reasoning about the issues easier as it is all related.

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Discussed below, resolving

DanielNoord (approved):

DanielNoord (approved):

jacobtylerwalls:
Thanks for the terrific review!

PR #2227, opened 2023-06-27
Small refactor after 3.12 support

DanielNoord:
## Type of Changes

## Description

Follow up to https://github.com/pylint-dev/astroid/pull/2219.

I didn't get round to reviewing, but had some last nits. This brings the nodes a little bit more in line with the other nodes (I also don't like it, but let's stay consistent).
More importantly, it also fixes an issue with the typing. `list[TypeVar, ParamSpec, ...]` is incorrect as per:
image
https://docs.python.org/3/library/typing.html#annotating-tuples

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -2705,16 +2705,19 @@ class ParamSpec(_base_nodes.AssignTypeNode):
     <ParamSpec l.1 at 0x7f23b2e4e198>
     """
 
+    _astroid_fields = ("name",)
+
+    name: AssignName
+
     def __init__(
         self,
         lineno: int,
         col_offset: int,
         parent: NodeNG,
         *,
-        end_lineno: int | None = None,
-        end_col_offset: int | None = None,
+        end_lineno: int | None,
+        end_col_offset: int | None,
     ) -> None:
-        self.name: AssignName | None
         super().__init__(
             lineno=lineno,
             col_offset=col_offset,
@@ -2723,7 +2726,7 @@ class ParamSpec(_base_nodes.AssignTypeNode):
             parent=parent,
         )
 
-    def postinit(self, *, name: AssignName | None) -> None:
+    def postinit(self, *, name: AssignName) -> None:
         self.name = name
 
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):

jacobtylerwalls:
Taking a look

astroid/nodes/node_classes.py
```diff
@@ -3375,8 +3379,8 @@ def __init__(
     def postinit(
         self,
         *,
-        name: AssignName | None,
-        type_params: list[TypeVar, ParamSpec, TypeVarTuple],
+        name: AssignName,
+        type_params: list[TypeVar | ParamSpec | TypeVarTuple],
```
jacobtylerwalls, line 3383: Ouch, thank you, I was just working too quickly.

astroid/nodes/node_classes.py
```diff
@@ -3350,20 +3353,21 @@ class TypeAlias(_base_nodes.AssignTypeNode):
     <TypeAlias l.1 at 0x7f23b2e4e198>
     """
 
-    _astroid_fields = ("type_params", "value")
+    _astroid_fields = ("name", "type_params", "value")
+
+    name: AssignName
+    type_params: list[TypeVar | ParamSpec | TypeVarTuple]
+    value: NodeNG
```
jacobtylerwalls, line 3360: I'm not totally sure why we're adopting this pattern instead of using the constructor.

astroid/nodes/node_classes.py
```diff
@@ -2705,16 +2705,19 @@ class ParamSpec(_base_nodes.AssignTypeNode):
     <ParamSpec l.1 at 0x7f23b2e4e198>
     """
 
+    _astroid_fields = ("name",)
```
jacobtylerwalls, line 2708: Thanks, I missed these because I went back and forth on whether this should be a `str` or an `AssignName` until I saw the references for the similar issue with `MatchStar` and friends pointing to https://github.com/python/cpython/issues/88160.

jacobtylerwalls (approved):

astroid/nodes/node_classes.py
DanielNoord: I don't like it as well, but I do think we should at least try and keep one style so that we can then refactor it to a desired style all at once?

jacobtylerwalls: No worries. Sometimes it seems like we prefer not to [mass-replace](https://github.com/pylint-dev/astroid/pull/2219#discussion_r1238936145), but not a big deal, let's keep moving. 👍

jacobtylerwalls (approved):
Let's ship it :-)

PR #2228, opened 2023-06-27
Avoid expensive list/tuple multiplication operations

jacobtylerwalls:
## Type of Changes

## Description

Closes pylint-dev/pylint#8748

Similar approach to #1610, instead of adding this guard to `safe_infer`, since it seems better to catch this as close to the source as possible instead of all of the places that call `infer()`.

### DIFF
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -167,7 +167,7 @@ def _multiply_seq_by_int(
     context: InferenceContext,
 ) -> _TupleListNodeT:
     node = self.__class__(parent=opnode)
-    if other.value > 1e8:
+    if isinstance(other.value, int) and other.value > 1e8:
         node.elts = [nodes.Const(NotImplemented)]
         return node
     filtered_elts = (
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -168,7 +168,7 @@ def _multiply_seq_by_int(
 ) -> _TupleListNodeT:
     node = self.__class__(parent=opnode)
     if isinstance(other.value, int) and other.value > 1e8:
-        node.elts = [nodes.Const(NotImplemented)]
+        node.elts = [util.Uninferable]
         return node
     filtered_elts = (
         helpers.safe_infer(elt, context) or util.Uninferable
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -230,7 +230,7 @@ def tl_infer_binary_op(
         as_index = helpers.class_instance_as_index(other)
         if not as_index:
             yield util.Uninferable
-        elif not isinstance(as_index.value, int):
+        elif not isinstance(as_index.value, int):  # pragma: no cover
             # already checked by class_instance_as_index() but faster than casting
             raise AssertionError("Please open a bug report.")
         else:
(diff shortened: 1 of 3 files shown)


astroid/protocols.py
```diff
@@ -167,6 +167,9 @@ def _multiply_seq_by_int(
     context: InferenceContext,
 ) -> _TupleListNodeT:
     node = self.__class__(parent=opnode)
+    if other.value > 1e8:
```
DanielNoord, line 170: Let's not pass a `Const` but an `int`:
```console
>>> "" >1e8
TypeError: '>' not supported between instances of 'str' and 'float'
>>>
```

There are two call sites, one is easy to fix. The other might need one more `isinstance`, but at least it won't crash and is type safe 😄

DanielNoord (changes requested):

astroid/protocols.py
```diff
@@ -167,6 +167,9 @@ def _multiply_seq_by_int(
     context: InferenceContext,
 ) -> _TupleListNodeT:
     node = self.__class__(parent=opnode)
+    if isinstance(other.value, int) and other.value > 1e8:
+        node.elts = [nodes.Const(NotImplemented)]
```
Pierre-Sassoulas, line 171: Shouldn't this be Uninferable instead of NotImplemented ?

astroid/protocols.py
DanielNoord: I meant changing `other` to `int`. The function name indicates that it should always be `int`.

astroid/protocols.py
jacobtylerwalls: I suppose that's better than implying there's one element in the list.

DanielNoord (approved):

astroid/protocols.py
```diff
@@ -221,14 +224,17 @@ def tl_infer_binary_op(
...
             yield not_implemented
             return
-        yield _multiply_seq_by_int(self, opnode, other, context)
+        yield _multiply_seq_by_int(self, opnode, other.value, context)
     elif isinstance(other, bases.Instance) and operator == "*":
         # Verify if the instance supports __index__.
         as_index = helpers.class_instance_as_index(other)
         if not as_index:
             yield util.Uninferable
+        elif not isinstance(as_index.value, int):
```
jacobtylerwalls, line 233:
```suggestion
        elif not isinstance(as_index.value, int):  # pragma: no cover
```

astroid/protocols.py
```diff
@@ -221,14 +224,17 @@ def tl_infer_binary_op(
...
             return
-        yield _multiply_seq_by_int(self, opnode, other, context)
+        yield _multiply_seq_by_int(self, opnode, other.value, context)
     elif isinstance(other, bases.Instance) and operator == "*":
         # Verify if the instance supports __index__.
         as_index = helpers.class_instance_as_index(other)
         if not as_index:
             yield util.Uninferable
+        elif not isinstance(as_index.value, int):  # pragma: no cover
+            # already checked by class_instance_as_index() but faster than casting
```
jacobtylerwalls, line 234: I'm not sure this is faster. I might change to this back to cast.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #2232, opened 2023-07-02
Move `safe_infer()` to util

jacobtylerwalls:
## Type of Changes

## Description
Avoid some circular imports by moving `safe_infer()` to `util`. Discussed at https://github.com/pylint-dev/astroid/pull/2171#discussion_r1187573283.

Also, [Remove shims for OperationError in exceptions module](https://github.com/pylint-dev/astroid/commit/1a770827277892641f3b62de2bcaac5b35f5c740)


DanielNoord (commented):
Haven't look at the code, but on a higher level: does it make sense to move this to `util/inference.py`?
That way we could define `safe_infer`, `infer_all` and some of the other inference utilities we need in `pylint` into one shared "API".

jacobtylerwalls:
I'm open to it, but I'd want to think carefully about it. Are you suggesting importing the astroid helpers and using those in pylint instead of the pylint helpers? The pylint safe_infer is a lot fancier. Could be a potential performance hit to port it to astroid and use it all the time there.

DanielNoord:
> I'm open to it, but I'd want to think carefully about it. Are you suggesting importing the astroid helpers and using those in pylint instead of the pylint helpers? The pylint safe_infer is a lot fancier. Could be a potential performance hit to port it to astroid and use it all the time there.

That's a good concern indeed. Let's keep it as is in this PR.

astroid/helpers.py
```diff
@@ -20,6 +20,10 @@
 from astroid.nodes import scoped_nodes
 from astroid.typing import InferenceResult
 
+# Don't remove the following safe_infer import without a deprecation notice.
```
DanielNoord, line 23: Do we want to do this now? What's the rationale for not doing it immediately?

jacobtylerwalls: f598696

DanielNoord (approved):

jacobtylerwalls:
Thanks for the review!

jacobtylerwalls:
I'll squash in the precommit fix. I undid it locally because I wasn't altering that code and figured I was on a newer version of ruff locally, but I guess not.

I'll add a test for the deprecation warning while at it

DanielNoord (approved):

jacobtylerwalls:
Missing line is just being moved from one location to another, so overriding the patch coverage check.

PR #2233, opened 2023-07-02
Fix ``typing`` issues in ``scoped_nodes.py``

DanielNoord:
## Type of Changes

## Description

Pretty all over the place, just zero'd in on the file and fixed what I could.
I think raising `ParentMissingError` is fine for now, we can always refactor later and we might actually get some useful reports from it.

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2369,7 +2369,7 @@ class ClassDef(  # pylint: disable=too-many-instance-attributes
         name: str,
         context: InferenceContext | None = None,
         class_context: bool = True,
-    ) -> list[SuccessfulInferenceResult]:
+    ) -> list[InferenceResult]:
         """Get an attribute from this class, using Python's attribute semantic.
 
         This method doesn't look in the :attr:`instance_attrs` dictionary
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -34,6 +34,7 @@ from astroid.exceptions import (
     InconsistentMroError,
     InferenceError,
     MroError,
+    ParentMissingError,
     StatementMissing,
     TooManyLevelsError,
 )
@@ -140,10 +141,10 @@ def clean_typing_generic_mro(sequences: list[list[ClassDef]]) -> None:
 
 
 def clean_duplicates_mro(
-    sequences: Iterable[Iterable[ClassDef]],
+    sequences: list[list[ClassDef]],
     cls: ClassDef,
     context: InferenceContext | None,
-) -> Iterable[Iterable[ClassDef]]:
+) -> list[list[ClassDef]]:
     for sequence in sequences:
         seen = set()
         for node in sequence:
@@ -241,7 +242,7 @@ class Module(LocalsDictNodeNG):
         self.pure_python = pure_python
         """Whether the ast was built from source."""
 
-        self.globals: dict[str, list[SuccessfulInferenceResult]]
+        self.globals: dict[str, list[InferenceResult]]
         """A map of the name of a global variable to the node defining the global."""
 
         self.locals = self.globals = {}
@@ -586,7 +587,7 @@ class Module(LocalsDictNodeNG):
 
     def _infer(
         self, context: InferenceContext | None = None, **kwargs: Any
-    ) -> Iterator[Module]:
+    ) -> Generator[Module, None, None]:
         yield self
 
 
@@ -887,7 +888,7 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
         :returns: 'method' if this is a method, 'function' otherwise.
         """
         if self.args.arguments and self.args.arguments[0].name == "self":
-            if isinstance(self.parent.scope(), ClassDef):
+            if self.parent and isinstance(self.parent.scope(), ClassDef):
                 return "method"
         return "function"
 
@@ -990,6 +991,8 @@ class Lambda(_base_nodes.FilterStmtsBaseNode, LocalsDictNodeNG):
         if (self.args.defaults and node in self.args.defaults) or (
             self.args.kw_defaults and node in self.args.kw_defaults
         ):
+            if not self.parent:
+                raise ParentMissingError(target=self)
             frame = self.parent.frame()
             # line offset to avoid that def func(f=func) resolve the default
             # value to the defined function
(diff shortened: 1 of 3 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2376,7 +2396,7 @@ def getattr(
             raise AttributeInferenceError(target=self, attribute=name, context=context)
 
         # don't modify the list in self.locals!
-        values: list[SuccessfulInferenceResult] = list(self.locals.get(name, []))
+        values: list[InferenceResult] = list(self.locals.get(name, []))
```
jacobtylerwalls, line 2399: Need to update the method's return annotation

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2883,7 +2903,7 @@ def _compute_mro(self, context: InferenceContext | None = None):
                 ancestors = list(base.ancestors(context=context))
                 bases_mro.append(ancestors)
 
-        unmerged_mro = [[self], *bases_mro, inferred_bases]
+        unmerged_mro: list[list[ClassDef]] = [[self], *bases_mro, inferred_bases]
```
jacobtylerwalls, line 2906: .bases is currently typed as list[SuccessfulInferenceResult]
```suggestion
        unmerged_mro: list[list[SuccessfulInferenceResult]] = [[self], *bases_mro, inferred_bases]
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2209,7 +2228,8 @@ def ancestors(
         if context is None:
             context = InferenceContext()
         if not self.bases and self.qname() != "builtins.object":
-            yield builtin_lookup("object")[1][0]
+            # This should always be a ClassDef (which we don't assert for)
+            yield builtin_lookup("object")[1][0]  # type: ignore[misc]
```
jacobtylerwalls, line 2232: I'd be okay with an assertion, but it's not a must-do either.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1886,7 +1901,9 @@ def __init__(
         self.is_dataclass: bool = False
         """Whether this class is a dataclass."""
 
-        self.type_params: list[nodes.TypeVar, nodes.ParamSpec, nodes.TypeVarTuple] = []
+        self.type_params: list[
+            nodes.TypeVar | nodes.ParamSpec | nodes.TypeVarTuple
+        ] = []
```
jacobtylerwalls, line 1906: ?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1932,7 +1949,8 @@ def postinit(
         *,
         position: Position | None = None,
         doc_node: Const | None = None,
-        type_params: list[nodes.TypeVar] | None = None,
+        type_params: list[nodes.TypeVar | nodes.ParamSpec | nodes.TypeVarTuple]
+        | None = None,
```
jacobtylerwalls, line 1953: Ah, thank you!

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -140,10 +141,10 @@ def clean_typing_generic_mro(sequences: list[list[ClassDef]]) -> None:
 
 
 def clean_duplicates_mro(
-    sequences: Iterable[Iterable[ClassDef]],
+    sequences: list[list[ClassDef]],
     cls: ClassDef,
     context: InferenceContext | None,
-) -> Iterable[Iterable[ClassDef]]:
+) -> list[list[ClassDef]]:
```
jacobtylerwalls, line 147: Why would we narrow these?

jacobtylerwalls (commented):
Thanks for marching onward with these!

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: `mypy` was very unhappy about assigning the return value to something that was previously a list. Then I tried to fix that but something else broken. I didn't want to spend more time, and since it is used only once I think it doesn't matter too much.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Scherm­afbeelding 2023-07-02 om 22 14 36

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: We use this so often I would be more inclined to create a `get_builtin_classdef` but for now I think we should just move on. There a probably much better solutions when we have a better codebase.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Then `clean_duplicates_mro` will rightfully complain. I think we should probably fix this function and add some assertions? But since a lot of these return `Any` now `mypy` is fine with it, for now 😄

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: Oh right, duh. I just literally couldn't see the change in the diff, I thought it was just a line break change. Sorry!

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: I was seeing a new mypy error

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: This was the one showing a new mypy error

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: Not a blocker, up to you!

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: Np

astroid/nodes/scoped_nodes/scoped_nodes.py
jacobtylerwalls: 237aaa2e1a5808c1ef0ed332d849a3ae7ea33334

jacobtylerwalls (approved):

PR #2234, opened 2023-07-04
Fix deprecation of `rec` argument to `find_argname()`

jacobtylerwalls:
## Type of Changes

## Description
Tested this in pylint, and saw I was incorrectly getting deprecation warnings. When using objects, they compare unidentical because astroid's copy of astroid is not pylint's copy of astroid. Using a string avoids this.

Follow-up to
698a376ff50a23958688d6cc4b9c8f7511a8c794.

Skip news.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -600,7 +600,7 @@ class Name(_base_nodes.LookupMixIn, _base_nodes.NoChildrenNode):
         return _infer_stmts(stmts, context, frame)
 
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Huh good to know, I thougth it was a neat trick and planned to use it.

DanielNoord (commented):
That's weird. We used to use the same trick in `pylint` and it seemed to work there..

jacobtylerwalls:
Maybe it wasn't tested in other consumers of pylint besides pylint itself?

PR #2231, opened 2023-07-02
Remove three mixin nodes from #2171

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description
Follow-up to 082774a. The DRY benefits of these three mixins (NameNode, AssignNode, AttributeNode) were so minimal that they can just be repeated.

The OperatorNode mixin is pretty huge, so I opted to leave it for now. (The nodes that inherit from it don't have as complicated a series of multiple mixin parents as the nodes changed in this PR, so it's not doing much harm.)

PR #2235, opened 2023-07-04
Restore `AssignAttr.infer_lhs()`

jacobtylerwalls:
## Type of Changes

## Description
Regression in 65df5e8efcc8dab41269be4252b22d21f634a2ae.

Here is the kind of "helper function" I was referring to when I said I would remove the mixins added in #2171. (Three of them removed in #2231.) One method went missing during that latter PR.

I got tricked into thinking I didn't need that method, but it was just because both the pylint and astroid unit tests didn't cover it. Only the pylint stdlib primer suite caught it.

Closes pylint-dev/pylint#8823

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1039,6 +1039,40 @@ def _format_args(
     return ", ".join(values)
 
 
+def _infer_attribute(
+    node: nodes.AssignAttr | nodes.Attribute,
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):

DanielNoord (commented):
I like this approach much more than adding even more `Mixins`, but probably personal preference. Thanks for fixing it like this!

PR #2236, opened 2023-07-04
Release branch 3.0.0a6

Pierre-Sassoulas:


jacobtylerwalls (changes requested):
Let's merge the last two PRs on the milestone, they're both unreleased bug fixes and we really need at least the crash fix.

jacobtylerwalls (approved):
Great! Everything will be correct when these are rebased, yeah?

jacobtylerwalls:
Btw, I think Marc was hoping that an entire astroid alpha would be devoted to just #1867, so that's why I created a new milestones for that PR only. There's some merge conflicts after the inference module was removed; I should be able to push some changes to his branch and review it.

PR #2242, opened 2023-07-08
Release branch 3.0.0a7

Pierre-Sassoulas:


jacobtylerwalls:
In https://github.com/pylint-dev/astroid/pull/1867#pullrequestreview-1520403279 I was hoping not to merge the Try node changes until we had 3.0.0a7 released, but now that the Try node changes are merged in, we have two large pylint PRs to do at the same time.

I'm okay with merging this as is, but in terms of doing the pylint upgrades, should we just keep the two PRs separate, and have the first one pin to a specific commit since we don't have another separate alpha for it?

Pierre-Sassoulas:
We could also cheat/rewrite the history or merge a branch install in pylint ?

jacobtylerwalls (approved):
> or merge a branch install in pylint

I'll just pin to the hotfix commit on the existing pylint upgrade PR and that pin will only be there for a day or two on main until we do the Try node update PR, and it won't be a problem.

Thanks!

PR #2243, opened 2023-07-08
[refactor] Remove variable for thing that never change in test_inference

Pierre-Sassoulas:
## Type of Changes

## Description

Small thing that annoy me: Having to go check a constant 5000 lines above the test I'm looking at and the constant will never ever change.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -58,8 +58,6 @@ def get_node_of_class(start_from: nodes.FunctionDef, klass: type) -> nodes.Attri
 
 builder = AstroidBuilder()
 
-EXC_MODULE = "builtins"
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (commented):
I like getting rid of BOOL_SPECIAL_METHOD because I think `__bool__` reads much better but am -0 on the other change. I generally like extracting a constant from a string that has to be used repeatedly, as it guards against typos.

Pierre-Sassoulas:
Wouldn't the test fail and show a diff of the two strings in case of typo ?

jacobtylerwalls:
Good point, I forgot these are unit tests.

> Having to go check a constant 5000 lines above

You gotta use mouse hovering! :D

jacobtylerwalls (approved):

PR #2244, opened 2023-07-08
[sys.argv] Make 'sys.argv' uninferable because it never is

Pierre-Sassoulas:
## Type of Changes

## Description

It's impossible to infer the value of sys.argv outside of static analysis where it's the linter own value that is irrelevant at run time. Even if we had one of the value during the program execution it can change for each execution of the program.

Refs https://github.com/pylint-dev/pylint/issues/7710 / https://github.com/pylint-dev/pylint/pull/9041

I'm not sure this is generic at all, the implementation is naive and based on the test case I created. I welcome any advice from astroid experts :)

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1062,7 +1062,12 @@ def _infer_attribute(
             if isinstance(owner, (ClassDef, Instance)):
                 frame = owner if isinstance(owner, ClassDef) else owner._proxied
                 context.constraints[node.attrname] = get_constraints(node, frame=frame)
-            yield from owner.igetattr(node.attrname, context)
+            if node.attrname == "argv" and owner.name == "sys":
+                # sys.argv will never be inferable during static analysis
+                # It's value would be the args passed to the linter itself
+                yield util.Uninferable
+            else:
+                yield from owner.igetattr(node.attrname, context)
         except (
             AttributeInferenceError,
             InferenceError,
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -7205,3 +7205,18 @@ class TestOldStyleStringFormatting:
         inferred = next(node.infer())
         assert isinstance(inferred, nodes.Const)
         assert inferred.value == "My name is Daniel, I'm 12.00"
+
+
+def test_sys_argv_uninferable() -> None:
+    """Regression test for https://github.com/pylint-dev/pylint/issues/7710."""
+    a: nodes.List = extract_node(
+        textwrap.dedent(
+            """
+    import sys
+
+    sys.argv"""
+        )
+    )
+    sys_argv_value = list(a._infer())
+    assert len(sys_argv_value) == 1
+    assert sys_argv_value[0] is Uninferable


Pierre-Sassoulas:
Testing in pylint is blocked until pylint is compatible with the latest astroid.

astroid/nodes/node_classes.py
```diff
@@ -1062,7 +1062,12 @@ def _infer_attribute(
             if isinstance(owner, (ClassDef, Instance)):
                 frame = owner if isinstance(owner, ClassDef) else owner._proxied
                 context.constraints[node.attrname] = get_constraints(node, frame=frame)
-            yield from owner.igetattr(node.attrname, context)
+            if node.attrname == "argv" and owner.name == "sys":
```
DanielNoord, line 1065: I'm not sure how often this happens, but this could conflict with implementations of `sys`. Perhaps we should just add a `sys` brain to handle determining whether it is `buitlin.sys`?

Pierre-Sassoulas:
Trying to release release 2.15.7, I feel like this could go in (or another approach can be designed at least).

astroid/nodes/node_classes.py
Pierre-Sassoulas: That's a false positive in pylint, and it happens relatively often (some duplicate opened for it) see https://github.com/canonical/apport/commit/b4e81f4d196a148259f2429aaae7a68f84d6dd24 for example. Maybe the brain approach is less costly ? But in general I think performance are less important than accuracy for a linter that is already slow.

jacobtylerwalls (approved):

mbyrnepr2 (approved):

astroid/nodes/node_classes.py
DanielNoord: Yeah I think we need a brain and use inference here.

Pierre-Sassoulas: How would you test the performance diff ?

jacobtylerwalls: @DanielNoord why do you think this is worth the inference penalty? You're imagining a custom script called "sys" that would lose the ability to infer an "argv"? I'm willing to let that be unhandled.

DanielNoord: Mostly because you can't control it as a user. If we want to prefer false negatives over false positives I think it is not nice that the user can't control this.

But if you guys think this is fine then go ahead!

Pierre-Sassoulas: I think ``sys.argv`` is actually ``Uninferable``. Often  ``Uninferable`` is 'soft' uninferable as in "our parsing does not permit to realize the variable MUST be none at this point because we do not have control flow so let's say we don't know", but I think this is an example of 'hard' ``Uninferable`` as in "we'll never know for sure what's in it, during static analysis".  How would the user control what is in ``sys.argv`` through a brain ? It seems to me that they face the same problem than us. If we want to give control to users over false negative vs false positive on ``Uninferable`` values, changes needs to be happening directly in pylint imo (but then every single astroid deficiency would create tons of FP).

DanielNoord: I agree that `stdlib.sys.argv` is `Uninferable`, but `user_module.sys.argv` isn't necessarily. That's why we would normally use inference: to see whether `sys` is builtin or from some user defined code.

Pierre-Sassoulas: Ha, right, so we should use qname then ? Although I'm not sure how often someone would happen to mimick sys.argv while also changing the meaning of sys.argv by putting something other than runtime args in it. (Or if this person deserves a linter 😛 )

jacobtylerwalls: I'm glad it was discussed, but yeah, I think we should stay focused on higher value cases and merge as is.

jacobtylerwalls: Decided to just test this out. I don't think we have a real problem. I added a "sys" module to astroid, and with this PR:
```py
>>> from astroid import extract_node
>>> n1 = extract_node("import astroid.sys; astroid.sys.argv")
>>> n1.inferred()[0]
<Const.int l.1 at 0x104c83f50>
>>> n2 = extract_node("import sys; sys.argv")
>>> n2.inferred()[0]
Uninferable
>>> n3 = extract_node("import astroid.sys as sys; sys.argv")
>>> n3.inferred()[0]
<Const.int l.1 at 0x104c83f50>
```

jacobtylerwalls:
Working on the backport

jacobtylerwalls:
Oof, sorry, I merged this without a changelog. Coming up...

Pierre-Sassoulas:
Thank you for fixing my PR @jacobtylerwalls 😅

PR #2257, opened 2023-07-19
Use the inference cache when no context is provided

jacobtylerwalls:
## Type of Changes

## Description
Previously, without an explicit `InferenceContext` instance, no checks were done against the global inference cache, and no entries were stored in it. This is a problem given how often the default `context=None` is used.

Refs #529. The third point of #529 muses about some sort of global inference cache, which was added later in cf6528cbc158097c4903f0cab68242ff14bb591b. This current PR just finishes the job. I'll leave remarks on #529 about the remaining possible improvements.

There will be one change needed in pylint where the checker depends on retroengineering a particular inference path being traveled twice, but I've already got a patch for it.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -51,6 +51,14 @@ Release date: TBA
   Closes pylint-dev/pylint#7464
   Closes pylint-dev/pylint#8074
 
+* Use the global inference cache when inferring, even without an explicit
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Sounds amazing !

PR #2261, opened 2023-07-22
fix default_value for kwonlyargs with no default

crazybolillo:
Refs #2213.

Requesting the default value for a kwonlyarg with no default would fail with an IndexError, it now correctly raises a NoDefault exception. This issue arised after changes in PR #2240.


jacobtylerwalls (approved):

jacobtylerwalls:
Thanks!

PR #2263, opened 2023-07-31
Exclude unassigned type-annotated class attributes from enum __members__ container

mbyrnepr2:
## Type of Changes

## Description

Exclude type-annotated class attributes, which have no assigned value, from the ``__members__`` container of an ``Enum`` class.

pylint-dev/pylint#7402

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -413,8 +413,6 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 else:
                     targets = stmt.targets
             elif isinstance(stmt, nodes.AnnAssign):
-                if stmt.value is None:
-                    continue
                 targets = [stmt.target]
             else:
                 continue
@@ -464,6 +462,8 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 for method in node.mymethods():
                     fake.locals[method.name] = [method]
                 new_targets.append(fake.instantiate_class())
+                if stmt.value == None:
+                    continue
                 dunder_members[local] = fake
             node.locals[local] = new_targets
 
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -413,6 +413,8 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 else:
                     targets = stmt.targets
             elif isinstance(stmt, nodes.AnnAssign):
+                if all(s is astroid.Uninferable for s in stmt.assigned_stmts()):
+                    continue
                 targets = [stmt.target]
             else:
                 continue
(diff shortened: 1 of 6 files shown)


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -413,6 +413,8 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                 else:
                     targets = stmt.targets
             elif isinstance(stmt, nodes.AnnAssign):
+                if all(s is astroid.Uninferable for s in stmt.assigned_stmts()):
```
jacobtylerwalls, line 416: Checking Uninferable might catch other truly uninferable assignment values, maybe?

```suggestion
                if stmt.value is None:
```

tests/brain/test_enum.py
```diff
@@ -493,3 +493,32 @@ def pear(self):
...
+            mass: int
+
+            def __init__(self, mass, radius):
+                self.mass = mass
+                self.radius = radius
+
+        Planet.MARS.value
+        """
+        )
+        enum_class.infer()
```
jacobtylerwalls, line 519:
```suggestion
```

tests/brain/test_enum.py
```diff
@@ -493,3 +493,32 @@ def pear(self):
...
+        enum_members = next(enum_class.igetattr("__members__"))
+        assert len(enum_members.items) == 2
+        mars, radius = enum_members.items
+        assert mars[1].name == "MARS"
+        assert radius[1].name == "radius"
```
jacobtylerwalls, line 524: This seems to go against the expected result described in the docstring? Or perhaps this is still WIP :-)

mbyrnepr2: Thanks for the quick feedback :D. Indeed the doc-string is out-dated since I added another scenario. Will remove it once I've figured out the 3.12 issue in the CI.

tests/brain/test_enum.py
mbyrnepr2: Thanks! I won't commit this directly because there is an unused variable to remove also.

astroid/brain/brain_namedtuple_enum.py
mbyrnepr2: This is the issue - we need to allow `red: int = "something"` but prevent `red: int` from entering the `__members__` container. Using `assigned_stmts()` let's me distinguish these two scenarios since `red: int` will have a value of `None` in the `stmt.value` approach.

jacobtylerwalls: Ah, if you could flesh out just a bit further that would help me understand. 👍 

> since red: int will have a value of None in the stmt.value approach

Isn't this exactly what we want? Tests pass with my suggestion and continue to distinguish the two cases.

jacobtylerwalls: Checking for a None value is the convention (I think) for elsewhere in pylint/astroid for checking annotations without values.

mbyrnepr2: Let me try it out! (I thought that I had already tried this during development and it wasn't working but perhaps it was something a bit different)

mbyrnepr2: Needed to move this logic further down the function so that something would be returned by the function - this manifested itself in a failing unittest testing an enum class with a single (annotated with no value) member.

mbyrnepr2: Funny enough the unittest only failed on 3.12

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Let's backport this :)

PR #2264, opened 2023-07-31
Update type annotations of PEP 695 nodes

### LINKED ISSUE #2253: Update location attributes of new Typing nodes
Now that https://github.com/python/cpython/issues/106145 is fixed we can make these attributes/arguments required again.
We added these in the last alphas so we can still add it without deprecation.

jacobtylerwalls:
## Type of Changes

## Description

Closes #2253

These attributes cannot be none in real-world situations, see https://github.com/python/cpython/issues/106145.

@DanielNoord they were already required; we're just changing the type annotations, yes? It's not enforced anywhere in the code.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -3381,8 +3381,8 @@ class ParamSpec(_base_nodes.AssignTypeNode):
         col_offset: int,
         parent: NodeNG,
         *,
-        end_lineno: int | None,
-        end_col_offset: int | None,
+        end_lineno: int,
+        end_col_offset: int,
     ) -> None:
         super().__init__(
             lineno=lineno,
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
Reading the cpython issue it's unclear to me if it was possible it was None in 3.10 and isn't anymore of if this was never possible that it's None. We still need to support older python interpreter if that's the case.

DanielNoord (approved):
Thanks! I had this on my todo list, but didn't get around to it before going on holiday.

_I'm trying to keep up with all the ongoing developments but it is quite hard. Feel free to tag me if I missed something in either project!_

DanielNoord:
> Reading the cpython issue it's unclear to me if it was possible it was None in 3.10 and isn't anymore of if this was never possible that it's None. We still need to support older python interpreter if that's the case.

This was only possible in beta's, so making this required is fine imo!

jacobtylerwalls:
These were new nodes in 3.12, so there's no real-world code that we need to support without these linenos.

Pierre-Sassoulas:
Damn, you're really keeping pylint bleeding edge, love that !

PR #2267, opened 2023-08-04
Register brains explicitly

DanielNoord:
## Type of Changes

## Description

This PR serves two purposes.

Firstly, it reduces the use of global state by explicitly passing an `AstroidManager` when registering brains instead of relying on the global singleton design. Related to this is the fact that we don't magically import any modules and that importing the brains modules _shouldn't_ have any side effects any more. In my opinion this explicitness is better although a bit verbose.
Secondly, by making the brains explicit we can start to think about a way to not register those we don't need. For example, in almost all of my projects I don't use any of the brains that are not for the `stdlib`. It would obviously be better not to register and call them. By making these imports explicit we can start excluding them based on certain flags based on during initialisation.

Let me know what you guys thinks of this! I tested this with `pytest` `main` and all seemed to be good.


astroid/brain/helpers.py
```diff
@@ -22,3 +22,112 @@ def transform(node: Module) -> None:
                     obj.parent = node
 
     manager.register_transform(Module, transform, lambda n: n.name == module_name)
+
+
+# pylint: disable-next=too-many-locals
+def register_all_brains(manager: AstroidManager) -> None:
+    from astroid.brain import (  # pylint: disable=import-outside-toplevel
```
jacobtylerwalls, line 29: This seems like a good candidate for `from astroid.brain import *`, but I have a feeling you will prefer the explicitness.

astroid/manager.py
```diff
@@ -456,10 +455,4 @@ def clear_cache(self) -> None:
         self.bootstrap()
 
         # Reload brain plugins. During initialisation this is done in astroid.__init__.py
```
jacobtylerwalls, line 457: Suggest changing astroid.__init__.py to astroid.manager

astroid/manager.py
```diff
@@ -456,10 +455,4 @@ def clear_cache(self) -> None:
...
-        for module in BRAIN_MODULES_DIRECTORY.iterdir():
-            if module.suffix == ".py":
-                module_spec = find_spec(f"astroid.brain.{module.stem}")
-                assert module_spec
-                module_object = module_from_spec(module_spec)
-                assert module_spec.loader
-                module_spec.loader.exec_module(module_object)
```
jacobtylerwalls, line 465: :rocket:

jacobtylerwalls (commented):
Great to be able to say good-by to iterdir!

jacobtylerwalls:
> Let me know what you guys thinks of this! I tested this with pytest main and all seemed to be good.

Did you test pylint?

astroid/brain/helpers.py
DanielNoord: Yeah guessed correctly 😄 

Do you mind keeping the explicitness (for now)?

DanielNoord:
> > Let me know what you guys thinks of this! I tested this with pytest main and all seemed to be good.
> 
> Did you test pylint?

That was a typo, yeah I did.

However, I'm now getting some issues that I feel might be due to the new `python -m pytest` changes. Could you maybe run this `pytest` for `pylint` with this branch of `astroid` and see if you get a green suite?

astroid/brain/helpers.py
jacobtylerwalls: Sure, that's fine. I'm still wondering if this should be in astroid.__init__ but I think there will have to be an import cycle to hide somewhere, and here is fine.

jacobtylerwalls:
Meaning without `python -m ` prepended? That's the fast resolution. If we want good ole `pytest` to work (with an editable install), we may need to add a `pylint/tests/__init__.py` and then alter some of the results that assert against module path names (tests.input instead of input, etc.)

jacobtylerwalls:
Sorry, I didn't mean to be throwing a wrench in people's setups, but we needed to do something. All the patching pylint does of sys.path is very surprising, and has prevented me from using pre-commit for over a year.

astroid/brain/helpers.py
jacobtylerwalls: Edited to say `astroid.brain.__init__`

DanielNoord: Oh, oops! Will fix tomorrow!

DanielNoord:
> Meaning without `python -m ` prepended? That's the fast resolution. If we want good ole `pytest` to work (with an editable install), we may need to add a `pylint/tests/__init__.py` and then alter some of the results that assert against module path names (tests.input instead of input, etc.)

Yeah. That is throwing errors for me on this branch that I can't explain. Perhaps you can? `pylint` runs fine so I think it is something off in the test suite.



> Sorry, I didn't mean to be throwing a wrench in people's setups, but we needed to do something. All the patching pylint does of sys.path is very surprising, and has prevented me from using pre-commit for over a year.

I think we're getting to a better place so that's good 😄 It did break my own `pre-commit` and `pytest` though so it is a bit hard to contribute to this process now 😅

jacobtylerwalls:
> Yeah. That is throwing errors for me on this branch that I can't explain. Perhaps you can? pylint runs fine so I think it is something off in the test suite.

Was it this?
```py
ImportError while loading conftest '/Users/jwalls/pylint/tests/conftest.py'.
tests/conftest.py:20: in <module>
    from pylint.testutils import MinimalTestReporter
E   ImportError: cannot import name 'MinimalTestReporter' from 'pylint.testutils' (unknown location)
```

That's actually "correct". Drumroll...

```py
cd ..
python3
>>> import pylint  # succeeds
>>> from pylint.testutils import MinimalTestReporter
ImportError: cannot import name 'MinimalTestReporter' from 'pylint.testutils' (unknown location)
```

Which is very similar to the reason I've been unable to use precommit since forever (apparently it's confessions day 😅 ):
```
jwalls@Jacobs-MacBook-Air-4 ~ % pylint
  File "/Library/Frameworks/Python.framework/Versions/3.12/bin/pylint", line 6, in <module>
    from pylint import run_pylint
ImportError: cannot import name 'run_pylint' from 'pylint' (unknown location)
```

***
Now, I've finally worked out why I was seeing the crashes reported [here](https://github.com/pylint-dev/pylint/issues/8854#issue-1805684684).

They go away with this:
```diff
diff --git a/pylint/__init__.py b/pylint/__init__.py
index cc0f609aa..27ef45d19 100644
--- a/pylint/__init__.py
+++ b/pylint/__init__.py
@@ -75,18 +75,19 @@ def modify_sys_path() -> None:
     stdlib or pylint's own modules.
     CPython issue: https://bugs.python.org/issue33053
 
-    - Remove the first entry. This will always be either "" or the working directory
+    - Remove the first entry if it is "" or ".".
     - Remove the working directory from the second and third entries
       if PYTHONPATH includes a ":" at the beginning or the end.
       https://github.com/pylint-dev/pylint/issues/3636
       Don't remove it if PYTHONPATH contains the cwd or '.' as the entry will
       only be added once.
-    - Don't remove the working directory from the rest. It will be included
-      if pylint is installed in an editable configuration (as the last item).
+    - Don't remove the explicit working directory from the rest. It will be
+      included explicitly if pylint is installed in an editable configuration,
+      but setuptools now makes no guarantee where it will appear in the path.
       https://github.com/pylint-dev/pylint/issues/4161
     """
     cwd = os.getcwd()
-    if sys.path[0] in ("", ".", cwd):
+    if sys.path[0] in ("", "."):
         sys.path.pop(0)
     env_pythonpath = os.environ.get("PYTHONPATH", "")
     if env_pythonpath.startswith(":") and env_pythonpath not in (f":{cwd}", ":."):
```

jacobtylerwalls:
I looked into it further. I'll revert the most recent change to pyproject.toml, which is to say, I'll restore the simulation of `editable_mode=compat`. It's too disruptive to ask everybody to add the repo root to their path when running tests. We can leave the workaround in place until we're ready to fix all the files. Sorry for hijacking your PR.

DanielNoord:
@jacobtylerwalls No I'm getting issues with `pytest -k test_issue_template_on_fatal_errors`. Do you get those as well? I don't really understand why that tests fails..

astroid/brain/helpers.py
Pierre-Sassoulas: This is going to need a refactor when we introduce conditional brain loading later (which is the point of the change, right ?). At this point we're either going to need to organize the brain under an enum of library (thinking ``{AstroidBrainLibrary.NUMPY: ["brain_numpy", "brain_numpy_core"]}``, and they're going to be explicitly sorted but we need to think about external brain registering if there is a pylint-numpy plugin with an external brain, or we'll filter them one way or another (thinking of something like "brain function contain the string 'numpy'" or "brain function match the regex *numpy*") and external brain will get auto-loaded. Probably better to argue about the final design than the tamporary one.

Pierre-Sassoulas (approved):
Great ! Not being forced to load all brains is going to be such a time saver on all run (especially the one on a single file without any dependencies). and astroid 3.0 is the moment to do this kind of changes.

astroid/brain/helpers.py
DanielNoord: Yeah, or we just move them into plugins? :smile: 

I don't know, removing this implicitness is an improvement either way.

jacobtylerwalls (approved):
Sorry, didn't realize this was waiting on me. All good, pylint suite passes (modulo existing errors I see all the time that predate this PR).

PR #2268, opened 2023-08-04
Allow ``AsStringVisitor`` to visit ``objects.PartialFunction``

DanielNoord:
## Type of Changes

## Description

Closes https://github.com/pylint-dev/pylint/issues/8881

Also added some annotations. Let's not back port as we are nearing the release of the version.

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING
 from astroid import nodes
 
 if TYPE_CHECKING:
+    from astroid import objects
     from astroid.nodes import Const
     from astroid.nodes.node_classes import (
         Match,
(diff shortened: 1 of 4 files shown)


astroid/nodes/as_string.py
```diff
@@ -441,6 +442,10 @@ def visit_pass(self, node) -> str:
         """return an astroid.Pass node as string"""
         return "pass"
 
+    def visit_partialfunction(self, node: objects.PartialFunction) -> str:
+        """Return an objects.PartialFunction as string."""
+        return self.visit_functiondef(node)
```
jacobtylerwalls, line 447: This is better than crashing, but it's not a great representation as it's identical to the regular function. Oh well.

jacobtylerwalls (approved):
Thanks!

PR #2269, opened 2023-08-05
Optimize root by not calling ``root()`` itself

DanielNoord:
## Type of Changes

## Description

We tend to call `root()` quite often. When linting `astroid` on `main` we spent 0.63% of our execution time in the `root()` function according to:
```python -m cProfile -o output.pstats -s cumtime -m pylint astroid && gprof2dot -f pstats output.pstats | dot -Tsvg -o out.svg```:
out2

On this branch the function doesn't even show up in the svg because it doesn't meet the default threshold.
out

Scherm­afbeelding 2023-08-05 om 14 14 35
As can be seen from this crop, we mostly call this function recursively so it no longer showing up in the profile after removing that makes sense.

### DIFF
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -332,9 +332,12 @@ class NodeNG:
 
         :returns: The root node.
         """
-        if self.parent:
-            return self.parent.root()
-        return self  # type: ignore[return-value] # Only 'Module' does not have a parent node.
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Yeah the recursion was 'elegant' but it's also a lot of call for non reason.

PR #2270, opened 2023-08-05
Remove unnecessary if statement

DanielNoord:
## Type of Changes

## Description

The todo comment seems to be true. I tested this with `pylint` and there doesn't seem to be a need for this check. Since we remove a call to the somewhat expensive `frame()` method it is nice to not have to do this.


Pierre-Sassoulas (approved):
If it doesn't change anything in pylint's primer then let's merge.

PR #2271, opened 2023-08-05
Optimize ``qname()`` by catching an exception instead

DanielNoord:
## Type of Changes

## Description

out2
On `main` when linting `astroid` we 1,78% of the runtime is caused by `qname`, which makes `403460` calls to `isinstance` which accounts for 0,13% of runtime.

out
This removes all of those calls, bringing down the number to 1,62%.

This was added in https://github.com/pylint-dev/astroid/pull/2130 to make the code type safe, but catching the exception serves the same purpose without paying the performance cost.


Pierre-Sassoulas (approved):
Nice optimisation !

PR #2272, opened 2023-08-05
Modernise the documentation and bring it in line with ``pylint``

DanielNoord:
## Type of Changes

## Description

This PR changes the theme of our documentation to be `furo` to have both projects use the same (and more modern) theme.
This prompted me to clean up the configuration files as they were a mess and it was difficult to see what was `astroid` specific and what autogenerated.

I also changed it so that every node now has its own page instead of producing one large unnavigable list. This works much better on `furo`.

Lastly, we no longer use the dead snakes project for links to `ast` objects, but instead link to `CPython`. I point to `/dev` as `/3` does not have `TypeAlias` and other objects yet.

_At some point we should perhaps reconsider if we even want to autogenerate our API like this. It creates a lot of link issues in Sphinx._


Pierre-Sassoulas (approved):
What a cleanup !

PR #2277, opened 2023-08-07
Infer user-defined enum classes by checking if the class is a subtype of ``enum.Enum``

mbyrnepr2:
## Type of Changes

## Description

Infer user-defined enum classes by checking if the class is a subtype of ``enum.Enum``.

Closes pylint-dev/pylint#8897


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -644,14 +636,7 @@ def _get_namedtuple_fields(node: nodes.Call) -> str:
 
 def _is_enum_subclass(cls: astroid.ClassDef) -> bool:
     """Return whether cls is a subclass of an Enum."""
-    try:
-        return any(
-            klass.name in ENUM_BASE_NAMES
-            and getattr(klass.root(), "name", None) == "enum"
-            for klass in cls.mro()
```
mbyrnepr2, line 651: `cls.mro()` wasn't returning the `enum` ancestors in the case of the example on the original issue; hence the class never made became inferred.

Pierre-Sassoulas (commented):
On top of the fix, it seems like a more sensible approach indeed, good job. Did we have tests for each hard coded subclasses from the removed constants ? (I want to make sure that intEnum are still handled correctly)

mbyrnepr2:
> On top of the fix, it seems like a more sensible approach indeed, good job. Did we have tests for each hard coded subclasses from the removed constants ? (I want to make sure that intEnum are still handled correctly)

Yes for [IntEnum](https://github.com/pylint-dev/astroid/blob/main/tests/brain/test_enum.py#L112).
No for `IntFlag` - there's tests for it in [Pylint](https://github.com/pylint-dev/pylint/blob/main/tests/functional/i/implicit/implicit_flag_alias.py#L2) but makes sense to have one here in astroid with the others.

mbyrnepr2:
This change to using `is_subtype_of` has opened the door to another issue as can be seen in the failing unittest. The failing test expects the class to be subscriptable and these classes currently don't have `__getitem__` methods. Not sure how straightforward that fix will be. I hope to dig into it soon.

jacobtylerwalls:
Sorry if this test will be difficult to rewrite! Maybe an extra inference call in the setup?

mbyrnepr2:
@jacobtylerwalls I've tried a few things to modify the test but no luck; when the program enters the enum brain module and into the `is_subtype_of` function we are getting a StopIteration (if I recall correctly) and the context path doesn't behave as expected. I've also had little time lately 😬.

astroid/brain/brain_namedtuple_enum.py
jacobtylerwalls: Looks like a slight performance improvement, too, based on the small sample size of profiling astroid.

jacobtylerwalls (approved):

PR #2281, opened 2023-08-21
Add `max_inferable_values` property to AstroidManager

### LINKED ISSUE #2280: `max_inferable_values` is only tunable on the class
### Steps to reproduce

Set `MANAGER.max_inferable_values` to any arbitrary number:
```py
astroid.MANAGER.max_inferable_values = 1_000
```

### Current behavior
the `limit` defined in `NodeNG.infer` will still be the default limit (100).
It comes from the fact that `limit` is defined by `AstroidManager.max_inferable_values` (the class) and not the `MANAGER` instance, like it was in `2.0.0`:
```py
class NodeNG:
    ...
    def infer(...):
        ...
        limit = AstroidManager.max_inferable_values  # 100
        ...

```

### Expected behavior
[Since `2.0.0`](https://pylint.readthedocs.io/projects/astroid/en/latest/changelog.html#what-s-new-in-astroid-2-0) the limit should be tunable:

> _"Limit the maximum amount of interable result in an NodeNG.infer() call to 100 by default for performance issues with variables with large amounts of possible values. The max inferable value can be tuned by setting the max_inferable_values flag on astroid.MANAGER."_

I saw no update in the changelog regarding this, so the limit should be `1_000` in my case, not `100`.

3.0.0a9
(I also tried with 2.15.6)

jacobtylerwalls:
## Type of Changes

## Description
Closes #2280

Exposes a more usable interface. Users should be able to interact with an instance (e.g. MANAGER) without having to drill down to the class.

### DIFF
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -408,9 +408,11 @@ class BorgAstroidManagerTC(unittest.TestCase):
     def test_max_inferable_values(self) -> None:
         mgr = manager.AstroidManager()
         original_limit = mgr.max_inferable_values
+
         def reset_limit():
             nonlocal original_limit
             manager.AstroidManager().max_inferable_values = original_limit
+
         self.addCleanup(reset_limit)
 
         mgr.max_inferable_values = 4


ChangeLog
```diff
@@ -326,6 +326,11 @@ Release date: 2023-03-06
 
   Refs #1780
 
+* ``max_inferable_values`` can now be set on ``AstroidManager`` instances, e.g. ``astroid.MANAGER``
+  besides just the ``AstroidManager`` class itself.
+
+  Closes # 2280
```
jacobtylerwalls, line 332:
```suggestion
  Closes #2280
```

DanielNoord (approved):

Pierre-Sassoulas (approved):
👌

PR #2289, opened 2023-09-07
Exclude the ``_ignore_`` attribute from the ``__members__`` container.

mbyrnepr2:
Closes pylint-dev/pylint#9015

## Type of Changes

## Description
Exclude the ``_ignore_`` attribute from the ``__members__`` container.

Closes [#9015](https://github.com/pylint-dev/pylint/issues/9015)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,8 @@ Release date: TBA
 * Exclude class attributes from the ``__members__`` container of an ``Enum`` class when they are
   ``nodes.AnnAssign`` nodes with no assigned value.
 
+* Fix a false positive for ``no-member`` when accessing the ``_name_`` and ``_value_`` sunders of an Enum member value.
+
   Refs pylint-dev/pylint#7402
 
 * Remove ``@cached`` and ``@cachedproperty`` decorator (just use ``@cached_property`` from the stdlib).
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,8 +22,6 @@ Release date: TBA
 * Exclude class attributes from the ``__members__`` container of an ``Enum`` class when they are
   ``nodes.AnnAssign`` nodes with no assigned value.
 
-* Fix a false positive for ``no-member`` when accessing the ``_name_`` and ``_value_`` sunders of an Enum member value.
-
   Refs pylint-dev/pylint#7402
 
 * Remove ``@cached`` and ``@cachedproperty`` decorator (just use ``@cached_property`` from the stdlib).
--- a/ChangeLog
+++ b/ChangeLog
@@ -224,7 +224,7 @@ Release date: TBA
   Closes pylint-dev/pylint#8802
 
 
-* Exclude the ``_ignore_`` attribute from the ``__members__`` container.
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
Look like we might need to also exclude ``_missing_``?  https://docs.python.org/3/library/enum.html

mbyrnepr2:
`_missing_` is a class method so Pylint is already happy with that naming. What do you think @Pierre-Sassoulas ?

Pierre-Sassoulas (commented):
😄 Hmm, and what about ``_name_`` for example ? Just stumbled upon https://docs.python.org/3/library/enum.html#supported-sunder-names (and learnt about sunder vs dunder, nice) it doesn't seem to be all functions in this case.

Pierre-Sassoulas:
Thank you for investigating ! If I understand correctly ``_value_`` is somewhat similar and ``_names_`` is not ? (It seems ``ValueError: _names_ are reserved for future Enum use`` should raise an error message in pylint, but it's very specific to enums, I don't know if we should do it and what message to raise.) If that's the case we could do ``_ignore_`` and ``_value_`` together and ``_names_`` later ? If not let's just do ``_ignore_`` to not overcomplicate things.

mbyrnepr2:
Ah my example made things confusing I think :)
The Python output `_names_ are reserved for future Enum use` is referring to sunder names in general (`_names_`).
Also `_value_` and `_name_` do produce the `invalid-name`  so I'll see about handling that here.


`ValueError: _names_ are reserved for future Enum use` could be a separate issue I think.

ChangeLog
```diff
@@ -221,6 +223,11 @@ Release date: TBA
 
   Closes pylint-dev/pylint#8802
 
+
+* Exclude the ``_ignore_`` attribute from the ``__members__`` container.
+
+  Closes pylint-dev/pylint#9015
```
mbyrnepr2, line 229:
```suggestion
* Fix false positives for ``no-member`` and ``invalid-name`` when using the ``_name_``, ``_value_`` and ``_ignore_`` sunders in Enums.

  Closes pylint-dev/pylint#9015
```

ChangeLog
```diff
@@ -22,6 +22,8 @@ Release date: TBA
 * Exclude class attributes from the ``__members__`` container of an ``Enum`` class when they are
   ``nodes.AnnAssign`` nodes with no assigned value.
 
+* Fix a false positive for ``no-member`` when accessing the ``_name_`` and ``_value_`` sunders of an Enum member value.
+
```
mbyrnepr2, line 26:
```suggestion
```

Pierre-Sassoulas (approved):
Great !

Pierre-Sassoulas:
Maybe we shouldn't close the pylint issue until astroid is released upgraded and we could clean-up the history and use two commits instead of a squash ? Or we can squash, I'll let you decide :)

Pierre-Sassoulas:
We actually need to squash for the backport job to work automatically

jacobtylerwalls:
Looks like this wasn't backported, so I'll move all the labels and such here and on the pylint issue. Life goes on. :D

PR #2313, opened 2023-09-30
Make manager required in AstroidBuilder/InspectBuilder

### LINKED ISSUE #251: Refactor circular dependency in builder, manager, and raw_building
Originally reported by: **BitBucket: [ceridwenv](http://bitbucket.org/ceridwenv), GitHub: @ceridwen**

---

Constructing Instance objects for classes imported from another module in raw_building requires accessing the functions for building ASTs from files in manager, which create a circular dependency because builder imports raw_building, manager imports both others, and raw_building imports manager.  Note that there are two preexisting circular dependencies that manager avoids with lazy imports, but the imports weren't sufficiently lazy to deal with the new dependency I introduced so I had to increase the laziness using lazy_object_proxy.

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/251

jacobtylerwalls:
## Type of Changes

## Description
Closes #251

Remove delayed imports in AstroidManager by requiring AstroidManager to be provided to AstroidBuilder/InspectBuilder.

A couple new delayed imports became necessary, but they're in helpers like `parse()` and `_astroid_bootstrapping()`. We could address those by:
- Requiring AstroidManager for parse() calls also.
- Moving `_astroid_bootstrapping()` to manager.py.

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -18,11 +18,14 @@ import types
 import warnings
 from collections.abc import Iterable
 from contextlib import redirect_stderr, redirect_stdout
-from typing import Any, Union
+from typing import TYPE_CHECKING, Any, Union
 
 from astroid import bases, nodes
 from astroid.const import _EMPTY_OBJECT_MARKER, IS_PYPY
 from astroid.nodes import node_classes
+
+if TYPE_CHECKING:
+    from astroid.manager import AstroidManager
 
 logger = logging.getLogger(__name__)
 
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -22,7 +22,6 @@ from typing import Any, Union
 
 from astroid import bases, nodes
 from astroid.const import _EMPTY_OBJECT_MARKER, IS_PYPY
-from astroid.manager import AstroidManager
 from astroid.nodes import node_classes
 
 logger = logging.getLogger(__name__)
@@ -37,8 +36,6 @@ _FunctionTypes = Union[
     types.ClassMethodDescriptorType,
 ]
 
-# the keys of CONST_CLS eg python builtin types
-_CONSTANTS = tuple(node_classes.CONST_CLS)
 TYPE_NONE = type(None)
 TYPE_NOTIMPLEMENTED = type(NotImplemented)
 TYPE_ELLIPSIS = type(...)
@@ -424,8 +421,8 @@ class InspectBuilder:
 
     bootstrapped: bool = False
 
-    def __init__(self, manager_instance: AstroidManager | None = None) -> None:
-        self._manager = manager_instance or AstroidManager()
+    def __init__(self, manager_instance) -> None:
+        self._manager = manager_instance
         self._done: dict[types.ModuleType | type, nodes.Module | nodes.ClassDef] = {}
         self._module: types.ModuleType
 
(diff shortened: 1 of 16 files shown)


Pierre-Sassoulas (commented):
This one bring a lot of joy. Were you able to use ``nodes.X`` instead of ``X`` thanks to the fixed circular imports ?

jacobtylerwalls:
I didn't have a choice, because the second import statement [from nodes import ...] was failing because nodes was now partially installed at that point. I think before, it was an accident that nodes was fully initialized already.

DanielNoord (commented):
Although really in favour, this do introduce more global state for `AstroidManager()`, see also https://github.com/pylint-dev/astroid/issues/2048.

As long as `AstroidManager` is a singleton we'll keep accessing shared memory in multiprocessing scenarios. Ideally, to me, we would change it in `4.x` so that the manager needs to be instantiated.

jacobtylerwalls:
> Although really in favour, this do introduce more global state for AstroidManager(), see also https://github.com/pylint-dev/astroid/issues/2048.

Getting back up to speed on this to see if I can get it in for 4.0. @DanielNoord can you remind me what you meant by "more global state"? I'm not sure I follow.

DanielNoord:
From my testing with multiprocessing on of the issues is that the cache of all nodes is stored on the manager, which means that we have a lot of dependency on this one shared object. Since it is global, and not passed or instantiated in `pylint`, it is quite hard to extract that dependency and (for example) give every process its own cache.

I'm fine with this PR, but I do think at some point we need to look into making the manager something that is instantiated by the process that wants to do inference.

jacobtylerwalls:
Thanks.

> As long as AstroidManager is a singleton we'll keep accessing shared memory in multiprocessing scenarios. Ideally, to me, we would change it in 4.x so that the manager needs to be instantiated.

Makes sense. For another day :-)

astroid/raw_building.py
```diff
@@ -424,8 +421,8 @@ class InspectBuilder:
 
     bootstrapped: bool = False
 
-    def __init__(self, manager_instance: AstroidManager | None = None) -> None:
-        self._manager = manager_instance or AstroidManager()
+    def __init__(self, manager_instance) -> None:
```
Pierre-Sassoulas, line 424:
```suggestion
    def __init__(self, manager_instance: AstroidManager) -> None:
```
Unleess there's a reason to not type I don't understand?

astroid/rebuilder.py
```diff
@@ -1855,7 +1929,9 @@ def visit_matchas(self, node: ast.MatchAs, parent: NodeNG) -> nodes.MatchAs:
             )
             return newnode
 
-        def visit_matchor(self, node: ast.MatchOr, parent: NodeNG) -> nodes.MatchOr:
+        def visit_matchor(
+            self, node: ast.MatchOr, parent: nodes.NodeNG
+        ) -> nodes.MatchOr:
             newnode = nodes.MatchOr(
                 lineno=node.lineno,
                 col_offset=node.col_offset,
```
Pierre-Sassoulas, line 1937: This file would deserve to be a package, github refuse to display the diff, like it's a ``package-lock.json``. (Just a passing observation, I'm not suggesting we immediately do something about it)

Pierre-Sassoulas (commented):
LGTM, just a nit.

astroid/raw_building.py
```diff
@@ -22,7 +22,6 @@
 
 from astroid import bases, nodes
 from astroid.const import _EMPTY_OBJECT_MARKER, IS_PYPY
-from astroid.manager import AstroidManager
 from astroid.nodes import node_classes
```
Pierre-Sassoulas, line 25:
```suggestion
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from astroid.manager import AstroidManager
```

Pierre-Sassoulas (approved):

DanielNoord (approved):

PR #2318, opened 2023-10-14
Remove `restore_path()` decorator

jacobtylerwalls:
## Type of Changes

## Description
Reverts 0d4f73d.
Refs pylint-dev/pylint#9138: reduces number of unraisable `ValueError` exceptions in the python 3.12 pylint primer from 3 to 1. (see [result](https://github.com/jacobtylerwalls/pylint/actions/runs/6520082681)) (EDIT: turns out this "result" is a little indeterminate, though...)

There's been churn around this decorator before, but I'm in favor of removing some indirection.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,10 @@ What's New in astroid 3.0.1?
 ============================
 Release date: TBA
 
+* Prevent unraisable ``ValueError`` exceptions in generator cleanup by removing
+  ``context.restore_path()`` decorator.
+
+  Refs pylint-dev/pylint#9138
 
 
 What's New in astroid 3.0.0?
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Some tests to adjust on Python <=3.9, but otherwise [primer result looks reasonable](https://github.com/jacobtylerwalls/pylint/pull/25#issuecomment-1763193942). Still have one unraisable exception though.

Pierre-Sassoulas (approved):

jacobtylerwalls:
Thanks for the review. I'm going to put this one on hold for now; I might want to go in a different direction if I'm able to narrow down why the pylint primer is so noisy.

PR #2319, opened 2023-10-14
Remove infrastructure around deprecating default args

jacobtylerwalls:
## Type of Changes

## Description
Following the deprecation period, we can adjust the behavior as promised and then remove all this cruft.


Pierre-Sassoulas (commented):
Should we keep the decorator to maybe reuse it next deprecation cycle ?

jacobtylerwalls:
Yeah, I wondered, and I'm happy to do it if you prefer. It just seems like a bulky implementation that that's not in scope for this project. It caused a 15% performance regression before it was guarded under a module-level check. I just felt like lower-tech solutions like this are a better fit IMO:
https://github.com/pylint-dev/astroid/blob/b8eec68bd77138ef33f6223170ab922366c9c55a/astroid/nodes/node_classes.py#L941

Pierre-Sassoulas (approved):

PR #2328, opened 2023-10-25
Disable test_crypt_brain on Python >= 3.13

### LINKED ISSUE #2327: astroid depends on 'crypt', which is removed in Python 3.13
Still present in git and 3.0.1.
https://bugzilla.redhat.com/show_bug.cgi?id=2246054

According to https://docs.python.org/3.13/whatsnew/3.13.html:

PEP 594: Remove the crypt module and its private _crypt extension, deprecated in Python 3.11.
The hashlib module is a potential replacement for certain use cases.
Otherwise, the following PyPI projects can be used:
- bcrypt: Modern password hashing for your software and your servers.
- passlib: Comprehensive password hashing framework supporting over 30 schemes.
- argon2-cffi: The secure Argon2 password hashing algorithm.
- legacycrypt: Wrapper to the POSIX crypt library call and associated functionality.
(Contributed by Victor Stinner in gh-104773.)

limburgher:
Resolves #2327 but leaves two failures remaining:

FAILED tests/test_inference.py::test_dataclasses_subscript_inference_recursion_error_39
FAILED tests/brain/test_brain.py::TypingBrain::test_typing_annotated_subscriptable

### DIFF
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -1976,7 +1976,7 @@ def test_oserror_model() -> None:
     assert strerror.value == ""
 
 
-@pytest.mark.skipif(PY313_PLUS, reason="Python >= 3.13")
+@pytest.mark.skipif(PY313_PLUS, reason="Python >= 3.13 no longer has a crypt module")
 def test_crypt_brain() -> None:
     module = MANAGER.ast_from_module_name("crypt")
     dynamic_attrs = [
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -1975,6 +1975,7 @@ def test_oserror_model() -> None:
     assert isinstance(strerror, astroid.Const)
     assert strerror.value == ""
 
+
 @pytest.mark.skipif(PY313_PLUS, reason="Python >= 3.13")
 def test_crypt_brain() -> None:
     module = MANAGER.ast_from_module_name("crypt")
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
Would you mind adding a constant like:
https://github.com/pylint-dev/astroid/blob/d3e8a9f29892f0fb5a880ec60d669c51e5aaad37/astroid/const.py#L12

And using that?

limburgher:
> Would you mind adding a constant like:
> 
> https://github.com/pylint-dev/astroid/blob/d3e8a9f29892f0fb5a880ec60d669c51e5aaad37/astroid/const.py#L12
> 
> And using that?

Will do, update incoming.

tests/brain/test_brain.py
```diff
@@ -1976,6 +1976,7 @@ def test_oserror_model() -> None:
     assert strerror.value == ""
 
 
+@pytest.mark.skipif(PY313_PLUS, reason="Python >= 3.13")
```
DanielNoord, line 1979:
```suggestion
@pytest.mark.skipif(PY313_PLUS, reason="Python >= 3.13 no longer has a crypt module")
```

DanielNoord (approved):
Thanks! We'll definitely need to fix the other tests as well, but I think it's more likely that will happen when `3.13` is a little bit more stable.

limburgher:
Thank you! I may take a look at those if I have time. I may not. :)

DanielNoord:
Thanks @limburgher, always appreciated when downstream packagers bring fixes to upstream!

PR #2340, opened 2023-12-10
Avoid duplicate inference results for `Tuple[Optional[int], ...]`

jacobtylerwalls:
## Type of Changes

## Description
Follow-up to #2185. Should help reduce noise in the pylint primer.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,10 +18,10 @@ What's New in astroid 3.0.2?
 Release date: TBA
 
 * Avoid duplicate inference results for some uses of ``typing.X`` constructs like
-  ``Tuple[Optional[int], ...]``. This was causing pylint to occassionally omit
+  ``Tuple[Optional[int], ...]``. This was causing pylint to occasionally omit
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Blocked by a pylint patch release

DanielNoord (approved):

jacobtylerwalls:
@Pierre-Sassoulas I'd like to unblock this PR so we can get an astroid patch release out and start getting better pylint primer results. We need a pylint patch release with the isort changes to do it, or we could just temporarily pin pylint to main in astroid CI. I'm leaning toward the former, since it's been a couple months since a pylint patch release. WDYT?

This would mean doing things in the reverse order we generally do them (first the pylint patch, then the astroid patch), but then we can upgrade astroid in pylint main without releasing.

Pierre-Sassoulas:
I was trying to prepare a pylint patch release "soon" ™️, let's release then. (What is in the current milestone are not blockers, only nice to have.)

PR #2343, opened 2023-12-12
Bump astroid to 3.0.2, update changelog

jacobtylerwalls:
What's New in astroid 3.0.2?
============================
Release date: 2023-12-12

* Avoid duplicate inference results for some uses of ``typing.X`` constructs like
  ``Tuple[Optional[int], ...]``. This was causing pylint to occasionally omit
  messages like ``deprecated-typing-alias``.
  Closes pylint-dev/pylint#9220

### DIFF
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -2,5 +2,5 @@
 # For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
 
-__version__ = "3.1.0-dev0"
+__version__ = "3.0.2"
 version = __version__
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,9 +9,15 @@ Release date: TBA
 
 
 
-What's New in astroid 3.0.2?
+What's New in astroid 3.0.3?
 ============================
 Release date: TBA
+
+
+
+What's New in astroid 3.0.2?
+============================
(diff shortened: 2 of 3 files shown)


astroid/__pkginfo__.py
```diff
@@ -2,5 +2,5 @@
 # For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
 
-__version__ = "3.1.0-dev0"
+__version__ = "3.0.2"
```
jacobtylerwalls, line 5: Don't worry; the tagged 3.0.1 went out with this correctly as 3.0.1. This was a mistake where instead of amending the merge commit to main, I added a commit to the maintenance branch as discussed [here](https://github.com/pylint-dev/astroid/pull/2323#issuecomment-1763576235)

Pierre-Sassoulas (approved):

astroid/__pkginfo__.py
Pierre-Sassoulas: Shouldn't this be two commits one with each version then we rebase ? Approving because you probably also thougth of something that work :)

jacobtylerwalls: The mistake already happened. There shouldn't have been `"3.1.0-dev0"` on this branch at all, so this squash merge will fix a bug. You might be thinking of going from 3.0.2 -> 3.1.0-dev0 on the main branch.

PR #2344, opened 2023-12-12
Update release procedure

jacobtylerwalls:
Add some wisdom ✨

### DIFF
--- a/doc/release.md
+++ b/doc/release.md
@@ -91,10 +91,10 @@ is the version under development on `main`.)
   example `v2.3.6`). This merge is required so `pre-commit autoupdate` works for pylint.
 - Fix version conflicts properly, meaning preserve the version numbers of the form
   `X.Y.0-devZ` (For example: `2.4.0-dev6`).
-- Open a merge request against main. Ensure a merge commit is used, because pre-commit need
-  the patch release tag to be in the main branch history to consider the patch release as the latest 
-  version and this won't be the case with rebase or squash. You can
-  defend against trigger-happy future selves by enabling auto-merge with the merge
+- Open a merge request against main. Ensure a merge commit is used, because pre-commit
+  need the patch release tag to be in the main branch history to consider the patch
+  release as the latest version and this won't be the case with rebase or squash. You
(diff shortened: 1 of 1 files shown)


doc/release.md
```diff
@@ -65,23 +65,40 @@ branch:
...
-  `2.4.0-dev6`) before pushing on the main branch
+- Freeze the main branch.
+- Branch `post-X.Y-1.Z` from `main`.
+- `git merge maintenance/X.Y-1.x`: this should have the changelog for `X.Y-1.Z+1` (For
+  example `v2.3.6`). This merge is required so `pre-commit autoupdate` works for pylint.
+- Fix version conflicts properly, meaning preserve the version numbers of the form
+  `X.Y.0-devZ` (For example: `2.4.0-dev6`).
+- Open a merge request against main. Ensure a merge commit is used, because our tooling
+  depends on distinct commits: one that was tagged with the patch release, and another
+  that restores the correct version number (the merge commit you made manually). You can
```
Pierre-Sassoulas, line 96:
```suggestion
- Open a merge request against main. Ensure a merge commit is used, because pre-commit need
  the patch release tag to be in the main branch history to consider the patch release as the latest 
  version and this won't be the case with rebase or squash. Our tooling also depends on distinct 
  commits: one that was tagged with the patch release, and another
  that restores the correct version number (the merge commit you made manually). You can
```

> Our tooling also depends on distinct  commits: one that was tagged with the patch release, and another  that restores the correct version number (the merge commit you made manually).
  
  Regarding that : Is it true ? The tbump script, right ?

Pierre-Sassoulas:
Thank you for taking the time to update this :) !

doc/release.md
Pierre-Sassoulas:
```suggestion
- Open a merge request against main. Ensure a merge commit is used, because pre-commit need
  the patch release tag to be in the main branch history to consider the patch release as the latest 
  version and this won't be the case with rebase or squash. You can
```

Pierre-Sassoulas (dismissed):

Pierre-Sassoulas (approved):

PR #2346, opened 2023-12-12
Merge maintenance following 3.0.2 release

jacobtylerwalls:


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Just thought about the reason we had to merge back initially, it's also so the changelog make sense (but we do not modify the changelog like that anymore in pylint).

PR #2348, opened 2023-12-14
Set parent when inferring __name__ as __main__

jacobtylerwalls:
Follow-up to a110905. Fixes pylint failures with `StatementMissing` because there was no parent.


jacobtylerwalls:
We may just want to fix these in pylint another way.

jacobtylerwalls:
Yeah, this doesn't actually address the failure. We can just fix it in pylint in `_loopvar_name`.

PR #2366, opened 2024-01-28
Fix annotations for statement method

cdce8p:
## Description
`Statement` isn't exported form `astroid.nodes` (anymore ?). All other annotations were updated to use `_base_nodes.Statement`. It seems these two were missed.

--
_Side note_
Any particular reason the `base_nodes` module should be considered private, i.e. prefixed with `_`? Probably missed that discussion.

### DIFF
--- a/astroid/filter_statements.py
+++ b/astroid/filter_statements.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
 
 def _get_filtered_node_statements(
     base_node: nodes.NodeNG, stmt_nodes: list[nodes.NodeNG]
-) -> list[tuple[nodes.NodeNG, nodes.Statement]]:
+) -> list[tuple[nodes.NodeNG, _base_nodes.Statement]]:
     statements = [(node, node.statement()) for node in stmt_nodes]
     # Next we check if we have ExceptHandlers that are parent
     # of the underlying variable, in which case the last one survives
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Daniël proposed it [here](https://github.com/pylint-dev/astroid/issues/1633#issue-1275395455):
> mixins.py isn't included in the __init__.py of astroid and don't provide much use to anybody without actually using the nodes themselves. So I think we can consider this private API.

jacobtylerwalls (approved):

cdce8p:
> Daniël proposed it [here](https://github.com/pylint-dev/astroid/issues/1633#issue-1275395455):
> 
> > mixins.py isn't included in the **init**.py of astroid and don't provide much use to anybody without actually using the nodes themselves. So I think we can consider this private API.

Hmm 🤔 Just interesting that we need to import `_base_nodes` even in `pylint`. _The `nodes.Statement` issue exists there too._ So it's probably not all that private...

Pierre-Sassoulas:
I agree with @cdce8p if we use the import in pylint it should be public (even if it's only for typing I guess).

DanielNoord:
I think my main rationale was that it is an implementation detail and ideally should be exported as a `Protocol`. Making it private gave a bit more flexibility while we were still refactoring the code to become fully typed/type-safe.

However, I burned out quite a bit on trying to fully type `astroid` as it just seems impossible. I would be okay with making it public, but don't think plugin writers should really need it. It seems private to our "core" library. Not sure how we should indicate it as being so.

PR #2375, opened 2024-02-04
Prefer `.pyi` stubs

jacobtylerwalls:
## Type of Changes

## Description
Closes pylint-dev/pylint#9185

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 3.2.0?
 ============================
 Release date: TBA
 
+* ``.pyi`` stub files are now preferred over ``.py`` files when resolving imports.
+
+  Closes pylint-dev/#9185
+
 * ``igetattr()`` returns the last same-named function in a class (instead of
   the first). This avoids false positives in pylint with ``@overload``.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,7 +7,7 @@ What's New in astroid 3.2.0?
 ============================
 Release date: TBA
 
-* ``.pyi`` stub files are now preferred over ``.py`` files when resolving imports.
+* ``.pyi`` stub files are now preferred over ``.py`` files when resolving imports, (except for numpy).
 
   Closes pylint-dev/#9185
 
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -162,7 +162,7 @@ class ImportlibFinder(Finder):
 
         # We're looping on pyi first because if a pyi exists there's probably a reason
         # (i.e. the code is hard or impossible to parse), so we take pyi into account
-        # But we're not quite ready to do this for numpy
+        # But we're not quite ready to do this for numpy, see https://github.com/pylint-dev/astroid/pull/2375
         suffixes = (".pyi", ".py", importlib.machinery.BYTECODE_SUFFIXES[0])
         numpy_suffixes = (".py", ".pyi", importlib.machinery.BYTECODE_SUFFIXES[0])
         for entry in submodule_path:
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -162,6 +162,8 @@ class ImportlibFinder(Finder):
 
         for entry in submodule_path:
             package_directory = os.path.join(entry, modname)
+            # We're looping on pyi first because if a pyi exists there's probably a reason
+            # (i.e. the code is hard or impossible to parse), so we take pyi into account
             for suffix in (".pyi", ".py", importlib.machinery.BYTECODE_SUFFIXES[0]):
                 package_file_name = "__init__" + suffix
                 file_path = os.path.join(package_directory, package_file_name)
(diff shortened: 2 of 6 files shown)


tests/brain/test_attr.py
```diff
@@ -90,7 +90,7 @@ class Eggs:
     def test_attrs_transform(self) -> None:
         """Test brain for decorators of the 'attrs' package.
 
-        Package added support for 'attrs' a long side 'attr' in v21.3.0.
+        Package added support for 'attrs' alongside 'attr' in v21.3.0.
```
Pierre-Sassoulas, line 93: 😄 !

tests/brain/test_attr.py
```diff
@@ -153,36 +153,12 @@ class Eggs:
...
-        o = BarFoo(d=1)
-        o.d['answer'] = 42
-
-        @my_mutable
-        class FooFoo:
-            d = attrs.field(default=attrs.Factory(dict))
-
-        p = FooFoo(d=1)
-        p.d['answer'] = 42
         """
```
Pierre-Sassoulas, line 156: Duplicate of line 154 ?

astroid/interpreter/_import/spec.py
```diff
@@ -162,7 +162,7 @@ def find_module(
 
         for entry in submodule_path:
             package_directory = os.path.join(entry, modname)
-            for suffix in (".py", ".pyi", importlib.machinery.BYTECODE_SUFFIXES[0]):
+            for suffix in (".pyi", ".py", importlib.machinery.BYTECODE_SUFFIXES[0]):
```
Pierre-Sassoulas, line 165:
```suggestion
            # We're looping on pyi first because if a pyi exists there's probably a reason
            # (i.e. the code is hard or impossible to parse), so we take pyi into account
            for suffix in (".pyi", ".py", importlib.machinery.BYTECODE_SUFFIXES[0]):
```

Pierre-Sassoulas (commented):
Looks like a clever way of doing thing. Regarding the error I have an intuition that our (now outdated/wrong) brain / tests and the actual content of the numpy pyi are conflicting which would be a good thing, but I don't have the time to test that hypothesis right now.

tests/brain/test_attr.py
jacobtylerwalls: No, it was tests newly failing. I'm hoping the new situation is fine.

jacobtylerwalls:
> but I don't have the time to test that hypothesis right now.

Neither do I. It's not straightfoward. A soft launch of this without numpy seems wise. Pushed an ugly workaround.

Pierre-Sassoulas:
Yeah it would work 😄 But if we're right and our tests are actually outdated (high possibility seeing they are very old and numpy is under heavy development), we're going to prevent numpy from having this new feature. Tbf I did not intend to check everything just check one test to confirm or not. Trying to find the motivation to download numpy and start reading atm. (Also we have a ton of brain for numpy, I those brains could become a pylint-numpy plugin so we don't have to block astroid for lib specific behavior) .

jacobtylerwalls:
@Pierre-Sassoulas do you think we can get this into astroid 3.2 as is?

jacobtylerwalls:
[pylint primer result](https://github.com/jacobtylerwalls/pylint/pull/33#issuecomment-2094310272)

Pierre-Sassoulas:
Yeah we can release as is and not make astroid worse than it is at the moment. I keep this on my mind but handling the flow of issues and then, at the end of a big pile of issues, finding the motivation to open numpy's code and compare it to the astroid's numpy brains and unit tests ..is proving difficult. Might need to create an issue for the numpy / pyi / outdated tests checks in the end.

Pierre-Sassoulas:
Or a pylint numpy plugin.

jacobtylerwalls:
I hear you. Plus, what if we get negative user-feedback on this first push -- don't want to sink time into a big numpy yak shave that turns out to be wrong-track!

ChangeLog
```diff
@@ -7,6 +7,10 @@ What's New in astroid 3.2.0?
 ============================
 Release date: TBA
 
+* ``.pyi`` stub files are now preferred over ``.py`` files when resolving imports.
```
Pierre-Sassoulas, line 10:
```suggestion
* ``.pyi`` stub files are now preferred over ``.py`` files when resolving imports, (except for numpy).
```
Might avoid unnecessary troubleshoot for numpy user ?

astroid/interpreter/_import/spec.py
```diff
@@ -160,9 +160,14 @@ def find_module(
                 pass
             submodule_path = sys.path
 
+        # We're looping on pyi first because if a pyi exists there's probably a reason
+        # (i.e. the code is hard or impossible to parse), so we take pyi into account
+        # But we're not quite ready to do this for numpy
```
Pierre-Sassoulas, line 165:
```suggestion
        # But we're not quite ready to do this for numpy, see https://github.com/pylint-dev/astroid/pull/2375
```

Pierre-Sassoulas (commented):
Some nits :)

Pierre-Sassoulas (approved):

mbyrnepr2:
> Or a pylint numpy plugin.

Should that be standalone and a dependency for pylint-ml?

Pierre-Sassoulas:
Yeah that sounds good but we'd need a proof of concept first. Probably on a smaller lib with less users affected if we drop the brain from astroid/pylint by default to test the water.

PR #2380, opened 2024-02-09
Include modname in AST warnings

cdce8p:
## Type of Changes

## Description
Python 3.12 change the DeprecationWarning for `invalid escape sequence` to a `SyntaxWarning`. These are now printed when the module is imported / the ast tree build. An example

```
# currently
<unknown>:158: SyntaxWarning: invalid escape sequence '\.'

# with this change
plexapi.base:158: SyntaxWarning: invalid escape sequence '\.'
```

Usually it's an actual filename for warnings and not just the modname. However, it should be good enough for now.

Long term it might make sense to adjust the warnings filter for astroid to make all `SyntaxWarnings` -> `SyntaxErrors` which would get re-raised. Thus it would be possible to add a `pylint: disable` (but only because the whole import isn't analyzed at all / skipped after the error).

For know this will at least help debug where the warning is coming from. Users can still choose to ignore them
```bash
python -Wignore:invalid\ escape\ sequence:SyntaxWarning -m pylint ...
```

### DIFF
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -22,7 +22,11 @@ class ParserModule(NamedTuple):
     bin_op_classes: dict[type[ast.operator], str]
     context_classes: dict[type[ast.expr_context], Context]
 
-    def parse(self, string: str, type_comments: bool = True) -> ast.Module:
+    def parse(
+        self, string: str, type_comments: bool = True, filename: str | None = None
+    ) -> ast.Module:
+        if filename:
+            return ast.parse(string, filename=filename, type_comments=type_comments)
         return ast.parse(string, type_comments=type_comments)
 
 
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -173,7 +173,9 @@ class AstroidBuilder(raw_building.InspectBuilder):
     ) -> tuple[nodes.Module, rebuilder.TreeRebuilder]:
         """Build tree node from data and add some informations."""
         try:
-            node, parser_module = _parse_string(data, type_comments=True)
+            node, parser_module = _parse_string(
+                data, type_comments=True, modname=modname
+            )
         except (TypeError, ValueError, SyntaxError) as exc:
             raise AstroidSyntaxError(
                 "Parsing Python code failed:\n{error}",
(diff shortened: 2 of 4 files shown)


tests/test_builder.py
```diff
@@ -416,6 +416,18 @@ def test_data_build_invalid_x_escape(self) -> None:
         with self.assertRaises(AstroidSyntaxError):
             self.builder.string_build('"\\x1"')
 
+    def test_data_build_error_filename(self) -> None:
+        """Check that error filename is set to modname if given."""
+        with pytest.raises(AstroidSyntaxError, match="invalid escape sequence") as ctx:
+            self.builder.string_build("'\\d+\\.\\d+'")
+        assert isinstance(ctx.value.error, SyntaxError)
+        assert ctx.value.error.filename == "<unknown>"
```
cdce8p, line 424: Not exactly sure why the `SyntaxError` is raised for all Python version. Additionally this should only emit a `SyntaxWarning` at the moment, not a `SyntaxError` (which is reraised as `AstroidSyntaxError`).

Might be some pytest magic though..

astroid/builder.py
```diff
@@ -466,11 +468,13 @@ def _extract_single_node(code: str, module_name: str = "") -> nodes.NodeNG:
 
 
 def _parse_string(
-    data: str, type_comments: bool = True
+    data: str, type_comments: bool = True, modname: str | None = None
```
DanielNoord, line 471:
```suggestion
    data: str, type_comments: bool = True, modname: str | None
```

Why not make it required?

astroid/_ast.py
```diff
@@ -22,7 +22,11 @@ class ParserModule(NamedTuple):
     bin_op_classes: dict[type[ast.operator], str]
     context_classes: dict[type[ast.expr_context], Context]
 
-    def parse(self, string: str, type_comments: bool = True) -> ast.Module:
+    def parse(
+        self, string: str, type_comments: bool = True, filename: str | None = None
```
DanielNoord, line 26:
```suggestion
        self, string: str, type_comments: bool = True, filename: str | None
```

astroid/builder.py
cdce8p: Non-default arguments can't follow ones with default.

DanielNoord: My bad!

DanielNoord (approved):
Thanks!

PR #2385, opened 2024-02-18
Log RecursionErrors out as warnings during node transformation

jacobtylerwalls:
## Type of Changes

## Description
Closes pylint-dev/pylint#8842

This only addresses recursion errors during node transformation. We should do something similar for the inference system, but suggesting to do that in pylint, not here.


tests/test_transforms.py
```diff
@@ -258,3 +261,19 @@ def transform_class(cls):
...
+
+    def test_transform_aborted_if_recursion_limited(self):
+        def transform_call(node: Call) -> Const:
+            return node
+
+        self.transformer.register_transform(
+            nodes.Call, transform_call, _looks_like_dataclass_field_call
+        )
+        original_limit = sys.getrecursionlimit()
+        sys.setrecursionlimit(1000)
```
jacobtylerwalls, line 273: Notice that pytest does use a higher limit, as @DanielNoord pointed out before. But I'd rather delegate this to users at first; we can always tinker later.

jacobtylerwalls: pylance doesn't seem to use a higher limit.

tests/test_transforms.py
```diff
@@ -258,3 +262,24 @@ def transform_class(cls):
...
+
+        if IS_PYPY:
+            original_limit = 1000  # pypy doesn't expose this
+            sys.setrecursionlimit(600)
+        else:
+            original_limit = sys.getrecursionlimit()
+            sys.setrecursionlimit(1000)  # use the default
+
+        try:
+            with self.assertWarns(UserWarning):
```
DanielNoord, line 282: Might want to assert on the message as well to make sure we're not catching something else?

DanielNoord (commented):
Nice idea! One optional comment

tests/test_transforms.py
```diff
@@ -258,3 +262,24 @@ def transform_class(cls):
...
+        self.transformer.register_transform(
+            nodes.Call, transform_call, _looks_like_dataclass_field_call
+        )
+
+        if IS_PYPY:
+            original_limit = 1000  # pypy doesn't expose this
+            sys.setrecursionlimit(600)
+        else:
+            original_limit = sys.getrecursionlimit()
+            sys.setrecursionlimit(1000)  # use the default
```
Pierre-Sassoulas, line 279:
```suggestion
        # pypy doesn't expose its recursion limit
        original_limit = 1000 if IS_PYPY else sys.getrecursionlimit()
        sys.setrecursionlimit(10)  # really low value to be able to test easily
```

Then we can use a smaller test fixture.

jacobtylerwalls: 10 is too low, it will cause other unrelated recursion errrors.

jacobtylerwalls: but pypy seems like it does expose its recursion limit; i probably read outdated information while debugging. will update.

tests/test_transforms.py
```diff
@@ -258,3 +262,22 @@ def transform_class(cls):
...
+            nodes.Call, transform_call, _looks_like_dataclass_field_call
+        )
+
+        original_limit = sys.getrecursionlimit()
+        sys.setrecursionlimit(500 if IS_PYPY else 1000)
+
+        try:
+            with self.assertWarns(
+                UserWarning, msg="try: --init-hook='import sys; sys.setrecursionlimit"
+            ):
```
cdce8p, line 280: I would use something like `with pytest.warns(UserWarning, match="try: --init-hook='import sys; sys.setrecursionlimit")` here.

jacobtylerwalls: Thanks, I often forget that this project uses pytest ;)

Pierre-Sassoulas (approved):
LGTM, maybe there's a middle ground between 1000 and 10 for the tests ? 😄 I'll let you decide

cdce8p (approved):

Pierre-Sassoulas (dismissed):
The warning level make so much sense that I wonder why we didn't do that sooner. 🎉

astroid/nodes/as_string.py
```diff
@@ -363,7 +363,10 @@ def visit_generatorexp(self, node) -> str:
 
     def visit_attribute(self, node) -> str:
         """return an astroid.Getattr node as string"""
-        left = self._precedence_parens(node, node.expr)
+        try:
+            left = self._precedence_parens(node, node.expr)
+        except RecursionError:
+            left = f"({node.expr.accept(self)})"
```
cdce8p, line 369: Seems like these lines aren't covered by the test case. Is it actually necessary to catch `RecursionError` here? Doing so in `_visit_generic` seems to be enough.

jacobtylerwalls: Well, I tried, but the test crashes `pytest`. Any ideas?
```diff
diff --git a/tests/test_nodes.py b/tests/test_nodes.py
index 6ea25fd8..713591e4 100644
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -29,7 +29,7 @@ from astroid import (
     transforms,
     util,
 )
-from astroid.const import PY310_PLUS, PY312_PLUS, Context
+from astroid.const import IS_PYPY, PY310_PLUS, PY312_PLUS, Context
 from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidBuildingError,
@@ -49,6 +49,7 @@ from astroid.nodes.node_classes import (
 from astroid.nodes.scoped_nodes import ClassDef, FunctionDef, GeneratorExp, Module
 
 from . import resources
+from tests.testdata.python3.recursion_error import LONG_CHAINED_METHOD_CALL
 
 abuilder = builder.AstroidBuilder()
 
@@ -279,6 +280,21 @@ everything = f""" " \' \r \t \\ {{ }} {'x' + x!r:a} {["'"]!s:{a}}"""
         assert nodes.Unknown().as_string() == "Unknown.Unknown()"
         assert nodes.Unknown(lineno=1, col_offset=0).as_string() == "Unknown.Unknown()"
 
+    @staticmethod
+    def test_recursion_error_trapped() -> None:
+        original_limit = sys.getrecursionlimit()
+        sys.setrecursionlimit(500 if IS_PYPY else 1000)
+        
+        try:
+            with pytest.warns(UserWarning, match="unable to transform"):
+                ast = abuilder.string_build(LONG_CHAINED_METHOD_CALL)
+
+            attribute = ast.body[1].value.func
+            assert attribute.as_string()
+
+        finally:
+            sys.setrecursionlimit(original_limit)
+
 
 @pytest.mark.skipif(not PY312_PLUS, reason="Uses 3.12 type param nodes")
 class AsStringTypeParamNodes(unittest.TestCase):
```

cdce8p: I guess that the except clause itself also triggers a new `RecursionError`.
```py
f"({node.expr.accept(self)})"
```

Just as POC, try:
```py
f"({node.expr})"
```

jacobtylerwalls: I wondered about that, but pylint lints the file just fine.

jacobtylerwalls: With your change, the output begins:

```py
(Call(func=<Attribute.add l.5 at 0x1075eaf30>,
args=[<Const.str l.10 at 0x1075eb680>],
keywords=[<Keyword l.10 at 0x1075eb6e0>])).add('name', value='value').add('name', value='value')...
```

instead of 
```py
b.builder('name').add('name', value='value').add('name', value='value')...
```

tests/test_nodes.py
```diff
@@ -279,6 +280,15 @@ def test_as_string_unknown() -> None:
...
         assert nodes.Unknown(lineno=1, col_offset=0).as_string() == "Unknown.Unknown()"
 
+    @staticmethod
+    def test_recursion_error_trapped() -> None:
+        with pytest.warns(UserWarning, match="unable to transform"):
+            ast = abuilder.string_build(LONG_CHAINED_METHOD_CALL)
+
+        attribute = ast.body[1].value.func
+        with pytest.raises(UserWarning):
+            attribute.as_string()
```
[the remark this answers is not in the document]
jacobtylerwalls: absolutely, but it seems pytest is doing something magical if the using same recursion limit is not a problem in the interactive interpreter but is a problem inside pytest?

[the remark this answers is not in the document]
cdce8p: Yeah, pytest and warnings capture is just magic sometimes 😅
Since the test pass an it works in pylint I would be fine with merging this as is (even with the one uncovered line).

cdce8p (approved):

PR #2386, opened 2024-02-19
Suppress SyntaxWarnings when parsing modules

cdce8p:
## Description
Following the conclusion in https://github.com/pylint-dev/pylint/issues/9322, ignore `SyntaxWarnings` when parsing modules with `ast.parse`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -27,6 +27,10 @@ Release date: TBA
 * Include modname in AST warnings. Useful for ``invalid escape sequence`` warnings
   with Python 3.12.
 
+* Suppress ``SyntaxWarnings`` when parsing modules.
+
+  Closes pylint-dev/pylint#9322
+
 
 
 What's New in astroid 3.0.3?
--- a/ChangeLog
+++ b/ChangeLog
@@ -27,7 +27,7 @@ Release date: TBA
 * Include modname in AST warnings. Useful for ``invalid escape sequence`` warnings
   with Python 3.12.
 
-* Suppress ``SyntaxWarnings`` when parsing modules.
+* Suppress ``SyntaxWarning`` for invalid escape sequences on Python 3.12 when parsing modules.
 
   Closes pylint-dev/pylint#9322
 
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
What do you think about ignoring SyntaxWarning only if SyntaxWarning is missing from the warning filters registry, as it is by default? That would allow users to exercise control over the warnings.

cdce8p:
> What do you think about ignoring SyntaxWarning only if SyntaxWarning is missing from the warning filters registry, as it is by default? That would allow users to exercise control over the warnings.

Not sure it's worth the effort. For Home Assistant the solution should be start linting all requirements as well.
What I don't like about it currently is that modifying the warnings filter for each file is quite a performance hit. Maybe it's better to do that globally?

ChangeLog
```diff
@@ -27,6 +27,10 @@ Release date: TBA
 * Include modname in AST warnings. Useful for ``invalid escape sequence`` warnings
   with Python 3.12.
 
+* Suppress ``SyntaxWarnings`` when parsing modules.
```
jacobtylerwalls, line 30:
```suggestion
* Suppress ``SyntaxWarning`` for invalid escape sequences on Python 3.12 when parsing modules.
```

jacobtylerwalls (approved):
Great, thank you!

Next step would be to update pylint-dev/pylint@2034874 as part of the astroid 3.1 upgrade.

@Pierre-Sassoulas do you want to move forward on cutting astroid 3.1? I think leaving the numpy workaround in #2375 is alright for now.

cdce8p:
> @Pierre-Sassoulas do you want to move forward on cutting astroid 3.1?

I think a new release would make sense. This, together with the fix for the `ValueError`, is one of the last blockers before I can upgrade Home Assistant to use 3.12 for the pylint check.

Pierre-Sassoulas:
Sure we were planning on releasing with Jacob. I was trying to find the time to check numpy pyi consistency against our astroid numpy's brain. (I don't remember the issue and I'm on mobile). It Can probably be dealt with later.

PR #2397, opened 2024-03-15
drop test units which requires python 2.x

kloczek:
Already it is not possible to build this module with python 2.x (`requires-python = ">=3.8.0"` from pyproject.toml) so keeping those units in the code does not make to much sense.

## Type of Changes

## Description

Refs #XXXX

Closes #XXXX

### DIFF
--- a/tests/test_scoped_nodes.py
+++ b/tests/test_scoped_nodes.py
@@ -47,6 +47,7 @@ from astroid.nodes.scoped_nodes.scoped_nodes import _is_metaclass
 
 from . import resources
 
+
 def _test_dict_interface(
     self: Any,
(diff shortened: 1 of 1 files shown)


kloczek:
BTW using `grep -ri 'python 2'` is possible to find more places where is potentially clean some code or comments.

jacobtylerwalls:
Astroid can still analyze python 2 code, however, and that's what's being tested here.

jacobtylerwalls:
There's still a difference between python 2 code and python 3 code that uses six to maintain compatibility (perhaps forever). Six does magic to the classes that can confuse astroid if we don't keep this functionality.

PR #2424, opened 2024-05-06
Add python 3.13 to the continuous integration

Pierre-Sassoulas:
## Type of Changes

## Description
Add compatibility with Python 3.13.

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -508,6 +508,8 @@ def _looks_like_dataclasses(node: nodes.Module) -> bool:
 
 
 def _resolve_private_replace_to_public(node: nodes.Module) -> None:
+    """In python/cpython@6f3c138, a _replace() method was extracted from
+    replace(), and this indirection made replace() uninferable."""
     if "_replace" in node.locals:
         node.locals["replace"] = node.locals["_replace"]
 
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -19,7 +19,7 @@ from typing import Literal, Union
 
 from astroid import bases, context, nodes
 from astroid.builder import parse
-from astroid.const import PY310_PLUS
+from astroid.const import PY310_PLUS, PY313_PLUS
 from astroid.exceptions import AstroidSyntaxError, InferenceError, UseInferenceDefault
 from astroid.inference_tip import inference_tip
 from astroid.manager import AstroidManager
@@ -503,6 +503,15 @@ def _looks_like_dataclass_field_call(
     return inferred.name == FIELD_NAME and inferred.root().name in DATACLASS_MODULES
 
 
+def _looks_like_dataclasses(node: nodes.Module) -> bool:
+    return node.qname() == "dataclasses"
+
+
+def _resolve_private_replace_to_public(node: nodes.Module) -> None:
+    if "_replace" in node.locals:
+        node.locals["replace"] = node.locals["_replace"]
+
+
 def _get_field_default(field_call: nodes.Call) -> _FieldDefaultReturn:
     """Return a the default value of a field call, and the corresponding keyword
     argument name.
@@ -608,6 +617,13 @@ def _infer_instance_from_annotation(
 
 
 def register(manager: AstroidManager) -> None:
+    if PY313_PLUS:
+        manager.register_transform(
+            nodes.Module,
+            _resolve_private_replace_to_public,
+            _looks_like_dataclasses,
+        )
+
     manager.register_transform(
         nodes.ClassDef, dataclass_transform, is_decorated_with_dataclass
     )
--- a/astroid/brain/brain_pathlib.py
+++ b/astroid/brain/brain_pathlib.py
@@ -8,6 +8,7 @@ from collections.abc import Iterator
 
 from astroid import bases, context, inference_tip, nodes
 from astroid.builder import _extract_single_node
+from astroid.const import PY313_PLUS
 from astroid.exceptions import InferenceError, UseInferenceDefault
 from astroid.manager import AstroidManager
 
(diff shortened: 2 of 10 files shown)


requirements_full.txt
```diff
@@ -8,6 +8,7 @@ numpy>=1.17.0; python_version<"3.12"
 python-dateutil
 PyQt6
 regex
+setuptools
```
Pierre-Sassoulas, line 11: Oof, is that avoidable ? I don't see what changed in this MR to warrant that, si it means we already were depending on setuptools ?

jacobtylerwalls: Exactly. Uninstall setuptools, run `pytest -k namespace`, see:
```py
tests/test_manager.py:124: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    import warnings
    
        warnings.simplefilter("ignore", DeprecationWarning)
>       __import__("pkg_resources").declare_namespace(__name__)
E       AttributeError: module 'pkg_resources' has no attribute 'declare_namespace'

tests/testdata/python3/data/path_pkg_resources_1/package/__init__.py:5: AttributeError
===================================== short test summary info ======================================
FAILED tests/test_manager.py::AstroidManagerTest::test_identify_old_namespace_package_protocol - AttributeError: module 'pkg_resources' has no attribute 'declare_namespace'
========================== 1 failed, 10 passed, 1675 deselected in 1.18s ===========================
```

We could do this in a separate MR if you like, but I figured a separate commit was enough.

Pierre-Sassoulas: Isn't this a test dependencie and not a runtime dependencie ?

jacobtylerwalls: I think that's what this file is.

tests/test_inference.py
```diff
@@ -4453,8 +4453,7 @@ def test_getitem_of_class_raised_type_error(self) -> None:
         # and reraise it as a TypeError in Class.getitem
         node = extract_node(
             """
-        def test():
-            yield
+        def test(): ...
```
DanielNoord, line 4456: This changes the function to no longer be a generator right? Why do we want that? https://github.com/pylint-dev/astroid/pull/2424/commits/8a9e944c3acb30592bb8d7754330f107f16eb890 doesn't explain this

tests/brain/test_pathlib.py
```diff
@@ -23,7 +23,10 @@ def test_inference_parents() -> None:
     inferred = name_node.inferred()
     assert len(inferred) == 1
     assert isinstance(inferred[0], bases.Instance)
-    assert inferred[0].qname() == "pathlib._PathParents"
+    if PY313_PLUS:
+        assert inferred[0].qname() == "builtins.tuple"
+    else:
+        assert inferred[0].qname() == "pathlib._PathParents"
```
DanielNoord, line 29: Do we depend on this in `pylint`? I guess CI will tell us

astroid/brain/brain_pathlib.py
```diff
@@ -27,10 +28,11 @@ def _looks_like_parents_subscript(node: nodes.Subscript) -> bool:
         value = next(node.value.infer())
     except (InferenceError, StopIteration):
         return False
+    parents = "builtins.tuple" if PY313_PLUS else "pathlib._PathParents"
```
DanielNoord, line 31: I'm wondering if this is too broad. Did we run `pylint` CI with this?

astroid/brain/brain_dataclasses.py
```diff
@@ -503,6 +503,15 @@ def _looks_like_dataclass_field_call(
     return inferred.name == FIELD_NAME and inferred.root().name in DATACLASS_MODULES
 
 
+def _looks_like_dataclasses(node: nodes.Module) -> bool:
+    return node.qname() == "dataclasses"
+
+
+def _resolve_private_replace_to_public(node: nodes.Module) -> None:
+    if "_replace" in node.locals:
+        node.locals["replace"] = node.locals["_replace"]
```
DanielNoord, line 512: Could you explain this change? Doesn't that make us incompatible with people adding their own `replace` method?

jacobtylerwalls: In python/cpython@6f3c138dfa868b32d3288898923bbfa388f2fa5d, a `_replace()` method was [extracted](https://github.com/python/cpython/blob/6f3c138dfa868b32d3288898923bbfa388f2fa5d/Lib/dataclasses.py#L1552) from `replace()`. Astroid became unable to infer the result of `replace()` due to this added indirection.

> Doesn't that make us incompatible with people adding their own replace method?

I'm not sure, but if so, isn't this also a problem with most/all brain transformations?

tests/test_inference.py
jacobtylerwalls: Good question. Apparently, on 3.13, generators now have `__class_getitem__` (at least according to astroid), defeating the purpose of this test. From the git blame, it's not clear to me why this would need to be a generator in the first place. It seems to just need to be something that raises an `AttributeInferenceError` when you getitem(4) on it.

astroid/brain/brain_dataclasses.py
DanielNoord: Ah, make sense! Could you add a comment with a link to that commit here?

DanielNoord (approved):

astroid/brain/brain_pathlib.py
jacobtylerwalls: Need to look closer, but at a glance nothing seems related: https://github.com/pylint-dev/pylint/actions/runs/9840792687/job/27165860010?pr=9776

jacobtylerwalls:
Thanks for the reviews!

PR #2431, opened 2024-05-12
Add the ``__annotations__`` attribute to the ``ClassDef`` object model.

mbyrnepr2:
## Type of Changes

## Description
Add the ``__annotations__`` attribute to the ``ClassDef`` object model.
Pylint now does not emit a ``no-member`` error when accessing ``__annotations__`` in the following cases:

**case 1.**
```python
class Test:
    print(__annotations__)
```

**case 2.**
```python
from typing import TypedDict

OtherTypedDict = TypedDict('OtherTypedDict', {'a': int, 'b': str})
print(OtherTypedDict.__annotations__)
```

**case 1** is similar to the behaviour of some of the other special attributes:
```python
class Test:
    print(__module__)
    print(__qualname__)
    print(__annotations__)
```
It turns out that this fix also fixed **case 2**.
However, it's still unclear why case 2 was only a false positive for Python 3.9 and older 😬 

Closes pylint-dev/pylint#7126

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -493,7 +493,7 @@ class ClassModel(ObjectModel):
         super().__init__()
 
     @property
-    def attr___annotations__(self) -> Unkown:
+    def attr___annotations__(self) -> node_classes.Unkown:
         return node_classes.Unknown()
 
     @property
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -493,6 +493,10 @@ class ClassModel(ObjectModel):
         super().__init__()
 
     @property
+    def attr___annotations__(self):
+        return node_classes.Unknown()
+
+    @property
     def attr___module__(self):
         return node_classes.Const(self._instance.root().qname())
 
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -493,7 +493,7 @@ class ClassModel(ObjectModel):
         super().__init__()
 
     @property
-    def attr___annotations__(self):
+    def attr___annotations__(self) -> Unkown:
         return node_classes.Unknown()
 
     @property
--- a/tests/test_object_model.py
+++ b/tests/test_object_model.py
@@ -857,3 +857,47 @@ def test_lru_cache(parentheses) -> None:
     assert wrapped.name == "foo"
     cache_info = next(ast_nodes[2].infer())
     assert isinstance(cache_info, astroid.Instance)
+
+
+def test_class_annotations() -> None:
+    """Test that the `__annotations__` attribute is avaiable in the class scope"""
+    annotations, klass_attribute = builder.extract_node(
+        """
+        class Test:
+            __annotations__  #@
+        Test.__annotations__  #@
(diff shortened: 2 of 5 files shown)


DanielNoord (approved):
I'm fine with the change, but should this behind a version flag?

See https://docs.python.org/3/howto/annotations.html. It seems it was only added to all objects in 3.10+?

mbyrnepr2:
> I'm fine with the change, but should this behind a version flag?
> 
> See https://docs.python.org/3/howto/annotations.html. It seems it was only added to all objects in 3.10+?

I see what you mean @DanielNoord great point. On Python 3.8, for example, we can demonstrate your point (this corresponds to **case 2** in the description above):
```python
>>> class Fruit:
...     pass
... 
>>> Fruit().__annotations__
AttributeError: 'Fruit' object has no attribute '__annotations__'
```

Now, let's look at **case 1** on Python 3.8:
```python
>>> class Fruit:
...     print(__annotations__)
... 
{}
>>> 
```
Without the fix, Pylint _does_ emit the error here. At this moment I don't know if it's better to make a distinction or keep it as-is but I'm open to ideas.

DanielNoord:
I think a distinction would be better as that better represents the actual versions right?

mbyrnepr2:
I would need to investigate a bit further @DanielNoord but my thinking is that we need the attribute anyway to satisfy both cases; so is the distinction necessary? Possibly but I need to see how that could work

astroid/interpreter/objectmodel.py
```diff
@@ -492,6 +492,10 @@ def __init__(self):
 
         super().__init__()
 
+    @property
+    def attr___annotations__(self):
```
DanielNoord, line 496:
```suggestion
    def attr___annotations__(self) -> Unkown:
```

Might as well.

DanielNoord (dismissed):
Thinking about this some more, we shouldn't optimize for old versions. Most people will be running their linter with any of the newer versions anyway. Let's merge! (with one last nit :))

DanielNoord (approved):

jacobtylerwalls:
> However, it's still unclear why case 2 was only a false positive for Python 3.9 and older 😬

At a glance, I expect that's because on 3.9 and below, the base class `__annotations__` [was returned](https://docs.python.org/3/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older).

tests/test_object_model.py
```diff
@@ -857,3 +857,47 @@ def foo():
...
+    assert isinstance(annotations, nodes.Name)
+    # The `__annotations__` attribute is `Uninferable`:
+    assert next(annotations.infer()) is astroid.Uninferable
+
+    # Test that we can access the class annotations:
+    assert isinstance(klass_attribute, nodes.Attribute)
+
+
+def test_class_annotations_typed_dict() -> None:
+    """Test that we can access class annotations on various TypedDicts"""
```
jacobtylerwalls, line 881: This is not important at the present moment, but elsewhere I've been encourage to avoid boilerplate like "test that" and "we", and I think it's a helpful edit ;)

jacobtylerwalls (approved):
Thanks for this!

PR #2433, opened 2024-05-12
Improve inference for generic classes (PEP 695)

cdce8p:
## Type of Changes

## Description

Generic classes with PEP 695 should be considered subscriptable, i.e. have a `__class_getitem__` method.
```py
class Foo[T]: ...
```

Refs https://github.com/pylint-dev/pylint/issues/9406
_Will close the issue after adding some more tests in pylint._

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -205,17 +205,9 @@ def infer_typing_generic_class_pep695(
     node: ClassDef, ctx: context.InferenceContext | None = None
 ) -> Iterator[ClassDef]:
     """Add __class_getitem__ for generic classes. Python 3.12+."""
-    try:
-        value = next(node.infer())
-    except (InferenceError, StopIteration) as exc:
-        raise UseInferenceDefault from exc
-
-    if not (isinstance(value, ClassDef) and value.type_params):
-        raise UseInferenceDefault
-
     func_to_add = _extract_single_node(CLASS_GETITEM_TEMPLATE)
-    value.locals["__class_getitem__"] = [func_to_add]
-    return iter([value])
+    node.locals["__class_getitem__"] = [func_to_add]
+    return iter([node])
 
 
 def _looks_like_typedDict(  # pylint: disable=invalid-name
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -196,6 +196,28 @@ def infer_typing_attr(
     return node.infer(context=ctx)
 
 
+def _looks_like_generic_class_pep695(node: ClassDef) -> bool:
+    """Check if class is uses type parameter. Python 3.12+."""
+    return len(node.type_params) > 0
+
+
+def infer_typing_generic_class_pep695(
+    node: ClassDef, ctx: context.InferenceContext | None = None
+) -> Iterator[ClassDef]:
+    """Add __class_getitem__ for generic classes. Python 3.12+."""
+    try:
+        value = next(node.infer())
+    except (InferenceError, StopIteration) as exc:
+        raise UseInferenceDefault from exc
+
+    if not (isinstance(value, ClassDef) and value.type_params):
+        raise UseInferenceDefault
+
+    func_to_add = _extract_single_node(CLASS_GETITEM_TEMPLATE)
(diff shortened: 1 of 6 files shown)


astroid/brain/brain_typing.py
```diff
@@ -196,6 +196,28 @@ def infer_typing_attr(
     return node.infer(context=ctx)
 
 
+def _looks_like_generic_class_pep695(node: ClassDef) -> bool:
+    """Check if class is uses type parameter. Python 3.12+."""
```
Pierre-Sassoulas, line 200:
```suggestion
    """Check if class is using type parameter. Python 3.12+."""
```

Pierre-Sassoulas (approved):
LGTM

astroid/brain/brain_typing.py
```diff
@@ -196,6 +196,28 @@ def infer_typing_attr(
...
+    return len(node.type_params) > 0
+
+
+def infer_typing_generic_class_pep695(
+    node: ClassDef, ctx: context.InferenceContext | None = None
+) -> Iterator[ClassDef]:
+    """Add __class_getitem__ for generic classes. Python 3.12+."""
+    try:
+        value = next(node.infer())
```
jacobtylerwalls, line 209: Thanks for tackling this. One question for my own information: why do we need to use inference? Can we just check `node.type_params`?

cdce8p: Tbh I'm not quite sure either. All other methods do this as well. My best guess is that with the `brain_**` we overwrite the default inference. However, since we still want that, we need to call `node.infer()` ourselves.

astroid/protocols.py
```diff
@@ -924,8 +924,7 @@ def generic_type_assigned_stmts(
...
     assign_path: None = None,
 ) -> Generator[nodes.NodeNG, None, None]:
-    """Return empty generator (return -> raises StopIteration) so inferred value
-    is Uninferable.
+    """Hack. Return any Node so inference doesn't fail
+    when evaluating __class_getitem__. Revert if it's causing issues.
     """
-    return
-    yield
+    yield nodes.Const(None)
```
cdce8p, line 930: This really isn't the "correct" solution but it works as the result is ignored anyway when inferring the `__class_getitem__` call. It just can't raise an Exception - that would lead to `Uninferable`.

The typing brain solves that by inferring the `T = TypeVar(...)` calls as new classes which is better but probably also not fully right.

_If it was actually uses, I believe we would've seen issue reports for it by now._
https://github.com/pylint-dev/astroid/blob/16da308bb3af0a13a7db88765de354355497981a/astroid/brain/brain_typing.py#L139-L144
https://github.com/pylint-dev/astroid/blob/16da308bb3af0a13a7db88765de354355497981a/astroid/brain/brain_typing.py#L47-L58

astroid/brain/brain_typing.py
jacobtylerwalls: Interestingly, if I add `assert node is value`, it passes.

Can we use this opportunity to explore avoiding inference here?

jacobtylerwalls (changes requested):
Open to different ideas here, but I think this would be a nice opportunity to verify things still work if we don't use inference.

jacobtylerwalls:
Should we guard this under if node.type_params: ?

cdce8p:
> Should we guard this under if node.type_params: ?

That is done by the predicate. No need to duplicate it I think
```py
def _looks_like_generic_class_pep695(node: ClassDef) -> bool:
    """Check if class is using type parameter. Python 3.12+."""
    return len(node.type_params) > 0
```

cdce8p:
> Can we use this opportunity to explore avoiding inference here?

Investigated that a bit more. With the previous setup, we would end up calling the inference_tip twice as the `infer` call was on `node` itself. _All other function is `brain_typing` infer an attribute like `func` or `args[0]`._ As the second call happened during the first inference pass, it would just get filtered out by the inference cache. So it's fine to remove it here. Also tested it with Home Assistant were everything continues to work as expected.

https://github.com/pylint-dev/astroid/blob/e43e045179fe4df7ba2aed96ad5ef180232f39cc/astroid/inference_tip.py#L45-L49

jacobtylerwalls (approved):

PR #2453, opened 2024-06-24
Visit the transforms before handling delayed assattr nodes

### LINKED ISSUE #2190: brain_gi uses wrong Gtk version for analysis
### Steps to reproduce
This only happens on systems where both Gtk 3 and Gtk 4 are available.

```python
import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk

cell = Gtk.CellRendererText()
#cell.props.xalign = 1.0

Gtk.Builder().connect_signals({})
```
When you run pylint on this code, it works as expected; however, when you uncomment the `cell.props.xalign` line, it breaks with 
```test.py:8:0: E1101: Instance of 'Builder' has no 'connect_signals' member (no-member)```
(`connect_signals` is only available in Gtk3, not in Gtk4)

### Current behavior
```test.py:8:0: E1101: Instance of 'Builder' has no 'connect_signals' member (no-member)```

### Expected behavior
**E1101** is not there (as `Builder` does have `connect_signals` member in Gtk 3.0).

### More info

I did a little bit more digging and found out that in the second (failing) case, following exception is raised in `_register_require_version`:
```Namespace Gtk is already loaded with version 4.0```

Debug prints in each function in brain_gi.py further showed that in the first case, `require_version` is handled first as expected. In the second case, `_register_require_version` is called after everything else has been processed, which is wrong (newer Gtk was already loaded at this point).

I found this in astroid 2.12.5 and pylint 2.15.2 with Python 3.7 and 3.11, but I also tried the latest and greatest (astroid 2.15.5 & pylint 2.17.4), and the same issue is still reproducible.

This issue was initially reported here:
https://github.com/pylint-dev/pylint/issues/6352

### LINKED ISSUE #6352: Pylint can't find some PyGObject classes' members
### Bug description

Hello,

I have a simple class inheriting from PyGObject's `Gtk.Window` class. Pylint cannot `Gtk.Window`'s `add` and `show_all` members, yet the code runs fine:

```Python
"""
The main window of the application.
"""

from typing import Tuple

import gi
gi.require_version("Gtk", "3.0")
# pylint: disable=wrong-import-position
from gi.repository import Gtk
# pylint: enable=wrong-import-position

class MainWindow(Gtk.Window):
    """
    The main window of the application.
    """

    WINDOW_SIZE: Tuple[int, int] = (500, 250)
    MAIN_PADDING_PX: int = 10
    FRAMES_VERTICAL_SPACE_PX: int = 10

    INPUT_FILE_FRAME_LABEL: str = "Input file"
    INPUT_FILE_BUTTON_LABEL: str = "Choose input file"

    def __init__(self, title: str):
        super().__init__(title=title)
        self.set_size_request(*self.WINDOW_SIZE)

        # --- Add widgets --- #

        # Main container
        main_box: Gtk.Box = Gtk.Box.new(orientation=Gtk.Orientation.VERTICAL,
                                        spacing=self.FRAMES_VERTICAL_SPACE_PX)
        self.add(main_box)
        MainWindow.__pad_on_all_sides(main_box, self.MAIN_PADDING_PX)

        # Input file picker
        input_file_frame: Gtk.Frame = Gtk.Frame.new(label=self.INPUT_FILE_FRAME_LABEL)
        main_box.pack_start(child=input_file_frame, expand=True, fill=True, padding=0)

        hbox: Gtk.Box = Gtk.Box.new(orientation=Gtk.Orientation.HORIZONTAL)
        input_file_frame.add(hbox)

        self.input_file_button = Gtk.Button.new(label=self.INPUT_FILE_BUTTON_LABEL)

        # --- Connect and show --- #
        self.connect("destroy", Gtk.main_quit)
        self.show_all()

    @staticmethod
    def __pad_on_all_sides(widget: Gtk.Widget, padding: int) -> None:
        widget.set_margin_bottom(padding)
        widget.set_margin_top(padding)
        widget.set_margin_left(padding)
        widget.set_margin_right(padding)

```

Interestingly, if I remove the following line, pylint stops complaining. It looks like adding a member to the MainWindow class which inherits from Gtk.Window throws off pylint.

```Python
self.input_file_button = Gtk.Button.new(label=self.INPUT_FILE_BUTTON_LABEL)
```

### Configuration

```ini
I am using pylint's default configuration.
```

### Command used

```shell
pylint main_window.py
```

### Pylint output

```shell
************* Module main_window
main_window.py:34:8: E1101: Instance of 'MainWindow' has no 'add' member (no-member)
main_window.py:48:8: E1101: Instance of 'MainWindow' has no 'show_all' member (no-member)

Your code has been rated at 6.43/10 (previous run: 6.43/10, +0.00)
```

### Expected behavior

I would expect pylint to find these two members, as they clearly exist since the code runs fine.

### Pylint version

```shell
pylint 2.13.5
astroid 2.11.2
Python 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
```

### OS / Environment

EndeavourOS KDE version
5.17.3-arch1-1 kernel
All packages up-to-date

### Additional dependencies

astroid==2.11.2
black==22.3.0
click==8.1.2
dill==0.3.4
isort==5.10.1
lazy-object-proxy==1.7.1
mccabe==0.7.0
mypy==0.942
mypy-extensions==0.4.3
pathspec==0.9.0
platformdirs==2.5.1
pycairo==1.21.0
PyGObject==3.42.0
pylint==2.13.5
tomli==2.0.1
typing_extensions==4.1.1
wrapt==1.14.0

marmarek:
## Type of Changes

## Description

When importing Gtk, it looks like this:

    import gi
    gi.require_version('Gtk', '3.0')
    from gi.repository import Gtk

It is vital that gi.require_version() is made before related 'from gi.repository import ...'. The brain_gi tries to do that using transforms. And it works unless Gtk is imported as part of delayed assattr handling.

Fix this by handling transforms earlier.

Closes https://github.com/pylint-dev/astroid/issues/2190
Closes https://github.com/pylint-dev/pylint/issues/6352

### DIFF
--- a/astroid/brain/brain_gi.py
+++ b/astroid/brain/brain_gi.py
@@ -245,6 +245,6 @@ def _register_require_version(node):
 
 def register(manager: AstroidManager) -> None:
     manager.register_failed_import_hook(_import_gi_module)
-    manager.register_transform(
+    manager.register_early_transform(
         nodes.Call, _register_require_version, _looks_like_require_version
     )
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -102,6 +102,7 @@ class AstroidBuilder(raw_building.InspectBuilder):
             if self._apply_transforms:
                 # We have to handle transformation by ourselves since the
                 # rebuilder isn't called for builtin nodes
+                node = self._manager.visit_early_transforms(node)
                 node = self._manager.visit_transforms(node)
         assert isinstance(node, nodes.Module)
         return node
@@ -164,6 +165,10 @@ class AstroidBuilder(raw_building.InspectBuilder):
                 for symbol, _ in from_node.names:
                     module.future_imports.add(symbol)
             self.add_from_names_to_locals(from_node)
+        # Visit the transforms
+        if self._apply_transforms:
+            module = self._manager.visit_early_transforms(module)
+
         # handle delayed assattr nodes
         for delayed in builder._delayed_assattr:
             self.delayed_assattr(delayed)
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -62,6 +62,7 @@ class AstroidManager:
         "extension_package_whitelist": set(),
         "module_denylist": set(),
         "_transform": TransformVisitor(),
+        "_early_transform": TransformVisitor(),
         "prefer_stubs": False,
     }
 
@@ -75,6 +76,7 @@ class AstroidManager:
         ]
         self.module_denylist = AstroidManager.brain["module_denylist"]
         self._transform = AstroidManager.brain["_transform"]
+        self._early_transform = AstroidManager.brain["_early_transform"]
         self.prefer_stubs = AstroidManager.brain["prefer_stubs"]
 
     @property
@@ -111,6 +113,15 @@ class AstroidManager:
         return self._transform.unregister_transform
 
     @property
+    def register_early_transform(self):
+        # This and unregister_early_transform below are exported for convenience
+        return self._early_transform.register_transform
+
+    @property
+    def unregister_early_transform(self):
+        return self._early_transform.unregister_transform
+
+    @property
     def builtins_module(self) -> nodes.Module:
         return self.astroid_cache["builtins"]
 
@@ -125,6 +136,10 @@ class AstroidManager:
     def visit_transforms(self, node: nodes.NodeNG) -> InferenceResult:
         """Visit the transforms and apply them to the given *node*."""
         return self._transform.visit(node)
+
+    def visit_early_transforms(self, node: nodes.NodeNG) -> InferenceResult:
+        """Visit the early transforms and apply them to the given *node*."""
+        return self._early_transform.visit(node)
 
     def ast_from_file(
         self,
@@ -466,6 +481,9 @@ class AstroidManager:
         self.astroid_cache.clear()
         # NB: not a new TransformVisitor()
         AstroidManager.brain["_transform"].transforms = collections.defaultdict(list)
+        AstroidManager.brain["_early_transform"].transforms = collections.defaultdict(
+            list
+        )
 
         for lru_cache in (
             LookupMixIn.lookup,
--- a/astroid/test_utils.py
+++ b/astroid/test_utils.py
@@ -73,6 +73,7 @@ def brainless_manager():
     m.astroid_cache = {}
     m._mod_file_cache = {}
     m._transform = transforms.TransformVisitor()
+    m._early_transform = transforms.TransformVisitor()
     m.extension_package_whitelist = set()
     m.module_denylist = set()
     return m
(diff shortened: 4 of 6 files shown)


marmarek:
My change made one test fail now:
```
____________________ TypingBrain.test_typing_generic_slots _____________________

self = <tests.brain.test_brain.TypingBrain testMethod=test_typing_generic_slots>

    def test_typing_generic_slots(self):
        """Test slots for Generic subclass."""
        node = builder.extract_node(
            """
        T = TypeVar('T')
        class A(Generic[T]):
            __slots__ = ['value']
            def __init__(self, value):
                self.value = value
        """
        )
        inferred = next(node.infer())
        slots = inferred.slots()
>       assert len(slots) == 1
E       TypeError: object of type 'NoneType' has no len()
```

I don't fully understand why, yet, but I think one option to make my change less invasive is to add "early transforms" and leave the original transforms in the old place. And then make brain_gi use the early transforms for the gi.require_version thing. Is that okay?

Pierre-Sassoulas:
Thank you for the MR. I agree with the proposed early transform, it feels like a big API decision but applying the transform twice so it's possible to apply transform after delayed attr creation is a lot worse. Let's hear another maintainer opinion before committing to it, it's not like I'm a brain expert myself.

DanielNoord:
Have you also tried running the `pylint` testsuite with this version of `astroid`? I'm wondering if this is a fix we can indeed make, the failing test doesn't make me too hopeful.

marmarek:
When running pylint test suite on plain main branch I got 12 failures already, but my change introduces 13th: `test_functional[assigning_non_slot_4509]` (`assigning-non-slot` message is not raised).
The failure here is also related to slots, so it's probably the same thing.

I'll change my patch to not affect other brains.

marmarek:
While at it, I've added also some basic test for brain_gi.

marmarek:
> tests\brain\test_gi.py s                                                 [ 20%]

So, the test I added doesn't run in CI (which also explain minimal test coverage of brain_gi.py). Is it worth installing "gi" python module + Gtk3 (and maybe Gtk4 too?) in there to make the test run? Or is it enough for it to run when called manually locally (assuming those dependencies are installed)?

DanielNoord:
Yeah I don't think we want to add those dependencies to the CI.

I do wonder if this warrants the API change. I can't really comment on that part (and I think most maintainers are a bit afraid of making such a decision). @Pierre-Sassoulas @jacobtylerwalls Do you have any opinions here?

jacobtylerwalls:
> I do wonder if this warrants the API change.

Yeah, I'm reluctant to add an early transforms API for this. We should consider alternatives:
- in _register_require_version(): manipulate sys.modules or use importlib.reload to reimport gi if necessary
- remove brain_gi in favor of a new plugin to be homed at pylint-dev/pylint-gi

DanielNoord:
A new plugin is probably what we would want considering the maintainer time we have for both `astroid` and `pylint`. Closing this for now, sorry!

PR #2455, opened 2024-07-01
Make the "test_helpers.py" test file runnable by itself

oprypin:
## Type of Changes

## Description

Currently, although running `pytest` passes, **running `pytest tests/test_helpers.py`** does not, it **gives the following error**:

    ...
      File "/tmp/astroid/tests/test_helpers.py", line 20, in setUp
        self.builtins = astroid_manager.astroid_cache[builtins_name]
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
    KeyError: 'builtins'

This started being the case since commit 33d0e8445f5d17aeae94d65918436ddb8affcc72, previously the file was runnable by itself.

This is a test-only fix, changelog non-applicable.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -8,6 +8,7 @@ import unittest
 import pytest
 
 from astroid import builder, helpers, manager, nodes, raw_building, util
+from astroid.builder import AstroidBuilder
 from astroid.const import IS_PYPY
 from astroid.exceptions import _NonDeducibleTypeHierarchy
 from astroid.nodes.scoped_nodes import ClassDef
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
I'm not sure we want to do that, we never launch a test file directly (``pytest -k`` can help with that), and there's going to be a perf cost each time we run the test and ensure bootstrap for nothing.

oprypin:
With `pytest -k` the test also fails.

And will you ever want to run tests in random order or in parallel?

Pierre-Sassoulas (approved):
Nice catch then :smile:

PR #2458, opened 2024-07-07
Fix AssertionError when inferring a property consisting of a partial function

jacobtylerwalls:
## Type of Changes

## Description
Closes pylint-dev/pylint#9214

Thanks Martin Belanger for the report and Bryce Guinta for the test case.

@brycepg would you like to review?

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2518,6 +2518,10 @@ class ClassDef(  # pylint: disable=too-many-instance-attributes
                 elif isinstance(inferred, objects.Property):
                     function = inferred.function
                     if not class_context:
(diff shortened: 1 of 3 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2518,6 +2518,10 @@ def igetattr(
                 elif isinstance(inferred, objects.Property):
                     function = inferred.function
                     if not class_context:
+                        if not context.callcontext:
```
jacobtylerwalls, line 2521: We may have the identical problem just below this, but I'm hesitant to just add it without a test case in hand. One thing at a time :-)

DanielNoord (approved):

brycepg:
Nice! deceptively straightforward :)

PR #2459, opened 2024-07-08
Implement inference for `JoinedStr` and `FormattedValue`

ericvergnaud:
## Type of Changes

## Description

`node.inferred()` would return `Uninferable` if inference involved a `JoinedStr`.
`JoinedStr` involves `FormattedValue` so this is addressed for both classes in the same PR


astroid/nodes/node_classes.py
```diff
@@ -4667,6 +4667,34 @@ def get_children(self):
...
+    ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
+        if self.format_spec is None:
+            yield from self.value.infer(context, **kwargs)
+            return
+        uninferable_already_generated = False
+        for format_spec in self.format_spec.infer(context, **kwargs):
+            if not isinstance(format_spec, Const):
+                if not uninferable_already_generated:
+                    yield util.Uninferable
+                    uninferable_already_generated = True
```
DanielNoord, line 4681: Why no do a `return` here and drop the constant?

astroid/nodes/node_classes.py
```diff
@@ -4728,6 +4756,30 @@ def postinit(self, values: list[NodeNG] | None = None) -> None:
...
+    ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
+        if len(nodes) == 1:
+            yield from nodes[0]._infer(context, **kwargs)
+            return
+        for prefix in nodes[0]._infer(context, **kwargs):
+            for suffix in cls._infer_from_values(nodes[1:], context, **kwargs):
+                result = ""
+                for node in (prefix, suffix):
+                    if node is util.Uninferable:
+                        result += "{Uninferable}"
```
DanielNoord, line 4776: It feels like we should exit early here and not continue to product "half" matching values

astroid/nodes/node_classes.py
ericvergnaud: There are a number of scenarios (such as loops) where `inferred` will return not one but multiple values. We don't want to drop the properly inferred values just because one (or more) of them can't be inferred.

astroid/nodes/node_classes.py
ericvergnaud: That's philosophical, depending on the strategy for dealing with partially inferred collections of values

jacobtylerwalls: It would help to make a decision about this if we could see a test case. At the moment, I do not think we should manufacture a constant that would never exist at runtime with the substring "Uninferable" in it.

astroid/nodes/node_classes.py
```diff
@@ -4728,6 +4756,30 @@ def postinit(self, values: list[NodeNG] | None = None) -> None:
...
+                    elif isinstance(node, Const):
+                        result += str(node.value)
+                    else:
+                        result += node.as_string()
```
jacobtylerwalls, line 4780: Also missing coverage.

tests/test_inference.py
```diff
@@ -652,6 +653,34 @@ def process_line(word_pos):
...
+            result = f"result: {value:{width}.{precision}}!"
+            """
+        ast = parse(code, __name__)
+        node = ast["result"]
+        inferred = node.inferred()
+        self.assertEqual(len(inferred), 1)
+        value_node = inferred[0]
+        self.assertIsInstance(value_node, Const)
+        self.assertEqual(value_node.value, "result:      12.35!")
+
```
jacobtylerwalls, line 683: Please also update the assertion in test_no_crash_on_const_reconstruction.

astroid/nodes/node_classes.py
jacobtylerwalls: There's "coincidental" coverage of this line in `test_typing_extensions_types` based on the current implementation of typing, but it would be better to have explicit coverage.

jacobtylerwalls (changes requested):
Thanks for contributing this feature!

tests/test_inference.py
jacobtylerwalls: Actually, that will be moot after https://github.com/pylint-dev/astroid/pull/2484.

astroid/nodes/node_classes.py
DanielNoord: Yes, you are completely right.

astroid/nodes/node_classes.py
DanielNoord: I agree with Jacob. I think `Uninferable` makes sense to those that have worked on `astroid` and `pylint`, but this will also be shown to users who might have no idea what it means. I would be in favour of exiting early and not creating partially correct values.

astroid/nodes/node_classes.py
```diff
@@ -4728,6 +4756,30 @@ def postinit(self, values: list[NodeNG] | None = None) -> None:
...
         yield from self.values
 
+    def _infer(
+        self, context: InferenceContext | None = None, **kwargs: Any
+    ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
+        yield from self._infer_from_values(self.values)
+
+    @classmethod
+    def _infer_from_values(
+        cls, nodes: list[NodeNG], context: InferenceContext | None = None, **kwargs: Any
```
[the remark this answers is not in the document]
ericvergnaud: True, but doesn't that change span beyond the scope of this PR ?

astroid/nodes/node_classes.py
```diff
@@ -4728,6 +4756,30 @@ def postinit(self, values: list[NodeNG] | None = None) -> None:
     def get_children(self):
         yield from self.values
 
+    def _infer(
+        self, context: InferenceContext | None = None, **kwargs: Any
+    ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
+        yield from self._infer_from_values(self.values)
```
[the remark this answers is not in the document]
ericvergnaud: Done

astroid/nodes/node_classes.py
ericvergnaud: This can never happen, line is now removed

astroid/nodes/node_classes.py
ericvergnaud: Done

tests/test_inference.py
ericvergnaud: Done anyway

astroid/nodes/node_classes.py
ericvergnaud: It is useful to know which fragment(s) of a JoinedStr cause inference failure, but I agree that i's misleading to return half-baked strings.

astroid/nodes/node_classes.py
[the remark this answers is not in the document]
jacobtylerwalls: Good point. And thanks!

ChangeLog
```diff
@@ -37,6 +37,10 @@ What's New in astroid 3.2.3?
 ============================
 Release date: 2024-07-11
 
+
+* Implement inference for JoinedStr and FormattedValue
```
jacobtylerwalls, line 41: Please move this to the 3.3.0 milestone above.

ericvergnaud: Done

jacobtylerwalls (dismissed):

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Great work !

DanielNoord (approved):

DanielNoord:
Oh, I just saw the failing tests. I do think we need to address it.

Pierre-Sassoulas:
It's failing on main too, we've merged multiple MR with 3.13 failing in the last week. (Due to a New beta of python 3.13). I think we could (should?) fix it before astroid 3.3.0 release.

ericvergnaud:
> Oh, I just saw the failing tests. I do think we need to address it.

TBH I'm 100% sure that the failure is unrelated to this PR. The test simply calls `as_string` which is unchanged. Smells like an unrelated bug with 3.13-dev.

jacobtylerwalls:
See #2478

PR #2468, opened 2024-07-12
Avoid reporting unary/binary op type errors for ambiguous inference

### LINKED ISSUE #2467: Pylint checks against incorrect type with properties that have a getter and setter
### Steps to reproduce

1. Install pylint 3.2.5 with astroid 3.2.3
2. Set up a Python script with the following code:
    ```python
   class Point:
       def __init__(self, x: float, y: float):
           self._x = x
           self._y = y

       @property
       def x(self) -> float:
           return self._x

       @x.setter
       def x(self, x: float):
           self._x = x

       @property
       def y(self) -> float:
           return self._y

       @y.setter
       def y(self, y: float):
           self._y = y

   test = Point(1.2, 3.4)
   value_neg = -test.x
   ```
4. Run pylint 

### Current behavior

The following error is emitted: E1130: bad operand type for unary -: Point (invalid-unary-operand-type)

### Expected behavior

pylint reports no errors

3.2.3

jacobtylerwalls:
## Type of Changes

## Description
Closes #2467

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,6 +18,10 @@ What's New in astroid 3.2.4?
 ============================
 Release date: TBA
 
+* Avoid reporting unary/binary op type errors when inference is ambiguous.
+
+  Closes #2467
+
 
 What's New in astroid 3.2.3?
 ============================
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Fixes/avoids a regression in pylint, not technically an astroid regression. Inference improved in #2458, exposing the fact that these utils don't really handle ambiguous results very cleanly.

DanielNoord (approved):

PR #2486, opened 2024-07-28
Improve class util typing

jacobtylerwalls:
## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1775,7 +1775,7 @@ def _class_type(
     klass: ClassDef,
     ancestors: set[str] | None = None,
     context: InferenceContext | None = None,
-):
+) -> Literal["class", "exception", "metaclass"]:
     """return a ClassDef node type to differ metaclass and exception
     from 'regular' classes
     """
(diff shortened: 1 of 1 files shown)


astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1790,7 +1798,7 @@ def _class_type(klass, ancestors=None, context: InferenceContext | None = None):
         for base in klass.ancestors(recurs=False):
             name = _class_type(base, ancestors)
             if name != "class":
-                if name == "metaclass" and not _is_metaclass(klass):
+                if name == "metaclass" and klass._type != "metaclass":
```
jacobtylerwalls, line 1801: This was checked and then set above on L1785

Pierre-Sassoulas (approved):
Nice catch ! Just saw the 'if class name ends with exception' bit around L1787 and it seems fragile, but also like we could add 'or error'

DanielNoord (approved):

jacobtylerwalls:
> Just saw the 'if class name ends with exception' bit around L1787 and it seems fragile, but also like we could add 'or error'

I'm assuming it's because ValueError etc eventually inherit from Exception and BaseException.

PR #2491, opened 2024-08-04
Fix release tests

jacobtylerwalls:
See https://github.com/actions/setup-python/issues/866#issuecomment-2154968862

[Passing run](https://github.com/pylint-dev/astroid/actions/runs/10236734136/job/28319161167)

### DIFF
--- a/.github/workflows/release-tests.yml
+++ b/.github/workflows/release-tests.yml
@@ -20,7 +20,11 @@ jobs:
         with:
           # virtualenv 15.1.0 cannot be installed on Python 3.10+
           python-version: 3.9
+        env:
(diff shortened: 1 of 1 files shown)


DanielNoord (approved):
Do you understand why were are experiencing this while we don't use Python 3.5? Changes make sense to me and the successful run is good, but it seems weird we need this for a version that is not yet EOL?

jacobtylerwalls:
I wondered that also. I don't know for a fact, but it might have something to do with [this behavior change](https://pip.pypa.io/en/stable/topics/https-certificates/#using-system-certificate-stores) before Python 3.10.

PR #2507, opened 2024-08-12
Remove unnecessary isinstance calls from transform predicates

correctmost:
## Type of Changes

## Description
This PR reduces the number of isinstance calls when linting yt-dlp by ~390k.

### Before

```
 13449993    2.005    0.000    2.015    0.000 {built-in method builtins.isinstance}
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `pylint --recursive=y .` | 32.888 ± 0.092 | 32.777 | 32.968 | 1.00 |

### After

```
 13058540    1.977    0.000    1.989    0.000 {built-in method builtins.isinstance}
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `pylint --recursive=y .` | 32.867 ± 0.154 | 32.667 | 33.020 | 1.00 |

### DIFF
--- a/astroid/brain/brain_numpy_ndarray.py
+++ b/astroid/brain/brain_numpy_ndarray.py
@@ -151,8 +151,8 @@ def infer_numpy_ndarray(node, context: InferenceContext | None = None):
     return node.infer(context=context)
 
 
-def _looks_like_numpy_ndarray(node) -> bool:
-    return isinstance(node, Attribute) and node.attrname == "ndarray"
+def _looks_like_numpy_ndarray(node: Attribute) -> bool:
+    return node.attrname == "ndarray"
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
This could be okay, but the isinstance call could be here to prevent a crash that was reported in pylint too. Did you check the git log to understand why the check was here ? (Also I'd be more comfortable is we were testing that branch using the primer before merging even if of course it's not going to catch everything.)

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #2509, opened 2024-08-12
Make Finder.find_module static to facilitate future caching

correctmost:
## Type of Changes

## Description

I would like to cache the results of `find_module` soon.  To make that possible, the function parameters will need to be hashable.  My plan is to split up the changes to reduce the risk of regressions and to make them easier to review.

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -86,9 +86,9 @@ class Finder:
     def __init__(self, path: Sequence[str] | None = None) -> None:
         self._path = path or sys.path
 
+    @staticmethod
     @abc.abstractmethod
     def find_module(
-        self,
         modname: str,
         module_parts: Sequence[str],
         processed: list[str],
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
LGTM, but let's hear from another maintainer.

DanielNoord (approved):

astroid/interpreter/_import/spec.py
```diff
@@ -378,9 +378,7 @@ def _find_spec_with_path(
 ) -> tuple[Finder | _MetaPathFinder, ModuleSpec]:
     for finder in _SPEC_FINDERS:
         finder_instance = finder(search_path)
```
jacobtylerwalls, line 380: Micro optimization, but seems like we can now move finder_instance below the continue.

jacobtylerwalls: Ah, no need then. It didn't occur to me there would be a side effect. Fine as is.

jacobtylerwalls (approved):

PR #2529, opened 2024-08-27
Use a tuple for processed parameter to facilitate future caching

correctmost:
## Type of Changes

## Description
I would like to cache the results of `find_module` soon. To make that possible, the function parameters will need to be hashable. My plan is to split up the changes to reduce the risk of regressions and to make them easier to review.

This is a follow-up to #2509 and #2516.

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -91,7 +91,7 @@ class Finder:
     def find_module(
         modname: str,
         module_parts: tuple[str],
-        processed: list[str],
+        processed: tuple[str],
         submodule_path: Sequence[str] | None,
     ) -> ModuleSpec | None:
         """Find the given module.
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
'Processed' is changing at every loop, I suppose it's never going to be the same twice ? Is it really a good candidate for caching ? (Something is appended to it right after the newly added cast to tuple)

astroid/interpreter/_import/spec.py
```diff
@@ -457,7 +457,7 @@ def _find_spec(module_path: tuple, path: tuple) -> ModuleSpec:
     while modpath:
         modname = modpath.pop(0)
         finder, spec = _find_spec_with_path(
-            _path, modname, module_parts, processed, submodule_path or path
+            _path, modname, module_parts, tuple(processed), submodule_path or path
         )
         processed.append(modname)
```
Pierre-Sassoulas, line 462: Here

Pierre-Sassoulas (approved):
Thank you for digging into this, You're doing an amazing work with all the performance improvements and detailed benchmark results. I wonder why we're calling find_module so much with identical values in the first place. But I have too much on my plate to investigate right now.

Pierre-Sassoulas:
Yes definitely. My unverified intuition here is that if we make import 'canonical' and create a mapping between the canonical import and the result of find module we could avoid searching for module and only translate an import to its canonical form instead. Caching find_module directly is close enough.

PR #2530, opened 2024-08-28
Enable mypy checking for astroid/interpreter/_import/

correctmost:
This commit also removes unnecessary tuple -> list conversions in `_find_spec`.

## Type of Changes

## Description

Automatic mypy checking would've helped catch some type issues in my recent PRs.  Hopefully more files and directories can be added in the future.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,14 +58,10 @@ repos:
     rev: v1.11.2
     hooks:
       - id: mypy
-        name: mypy
-        entry: mypy
         language: python
-        types: [python]
-        args: ["--config-file=pyproject.toml"]
+        pass_filenames: false
         require_serial: true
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -58,14 +58,10 @@ repos:
...
       - id: mypy
-        name: mypy
-        entry: mypy
         language: python
-        types: [python]
-        args: []
+        pass_filenames: false
         require_serial: true
         additional_dependencies: ["types-typed-ast"]
-        exclude: tests/testdata| # exclude everything, we're not ready
```
Pierre-Sassoulas, line 68: ~~We should still exclude testdata right?~~ nvm, saw that everything happens in the pyproject.toml now.

DanielNoord (approved):
Awesome!

PR #2531, opened 2024-08-28
Fix to order of module discovery

akamat10:
## Type of Changes

## Description
* Addresses the issues with module path discovery identified on https://github.com/pylint-dev/pylint/issues/9883
* Updated tests to reflect current python behavior when calling a script. Details of the expected behavior described [here](https://stackoverflow.com/questions/33743880/what-does-from-future-import-absolute-import-actually-do)


akamat10:
One thing this fix needs to consider is how the order of module discovery changes based on whether a user calls:

`python3.12 pkg/script.py`

vs

`python3.12 -m pkg.script`

The difference in behavior is described [here](https://stackoverflow.com/questions/33743880/what-does-from-future-import-absolute-import-actually-do) in a stackoverflow post.

The test fixes to `test_nonregr_absolute_import` case will now match the former case but perhaps some thought needs to be given to the latter case and have some test cases around it.

Pierre-Sassoulas:
Sorry for not reviewing this sooner, this is the kind of change that are hard to assess/test and it's frightening 😅

akamat10:
No need to be sorry. I should reevaluate this given it has been a while since I have looked at this one.

DanielNoord:
Closing for now, we can always reopen later :)

PR #2532, opened 2024-08-30
fix the documentation and the error message of modpath_from_file

temyurchenko:
The doc should clarify that the search will always include sys.path.

The error message should reflect the actual paths used for checking.

## Type of Changes


Pierre-Sassoulas (approved):
Thank you 👌

Issue #2517, opened 2024-08-26
KeyError in _set_proxied with getset_descriptor class


correctmost:
### Steps to reproduce

The following code triggers a `KeyError` in astroid:

```python
dict.fromkeys.__doc__.c=p
```

This bug was discovered by OSS-Fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62324 (report not public yet)

### Current behavior
```
Traceback (most recent call last):
  File "pylint/pylint/utils/ast_walker.py", line 91, in walk
    callback(astroid)
  File "pylint/pylint/checkers/classes/class_checker.py", line 1739, in visit_assignattr
    self._check_in_slots(node)
  File "pylint/pylint/checkers/classes/class_checker.py", line 1774, in _check_in_slots
    inferred = safe_infer(node.expr)
               ^^^^^^^^^^^^^^^^^^^^^
  File "pylint/pylint/checkers/utils.py", line 1372, in safe_infer
    inferred_types.add(_get_python_type_of_node(value))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pylint/pylint/checkers/utils.py", line 1339, in _get_python_type_of_node
    return pytype()
           ^^^^^^^^
  File "astroid/astroid/nodes/node_classes.py", line 2159, in pytype
    return self._proxied.qname()
           ^^^^^^^^^^^^^
  File "astroid/astroid/raw_building.py", line 596, in _set_proxied
    return _CONST_PROXY[const.value.__class__]
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
KeyError: <class 'getset_descriptor'>
```

### Expected behavior
No crash

### Version
d8dbc462243a65d72153ace2e7ac87daa9ffc1cf

Issue #1490, opened 2022-03-26
Audit the correctness of adding parent.name to child's locals in `ClassDef.__init__()` and `FunctionDef.__init__()`


jacobtylerwalls:
Apparently, bugs can be caused by providing the `parent` argument to `ClassDef.__init__()`, because the name of that node (the ancestor) will be added to the locals of the child node, and this may not be intended (it wasn't intended in the case of subclassing a `namedtuple`. We need to audit whether it is happening anywhere else and if that behavior is correct.

See discussion at https://github.com/PyCQA/astroid/pull/1489#issuecomment-1079788680.

jacobtylerwalls:
Caused another false positive at https://github.com/PyCQA/pylint/issues/6895

jacobtylerwalls:
I guess next step is to see what fails / depends on the current behavior.

DanielNoord:
> I guess next step is to see what fails / depends on the current behavior.

Commenting out the relevant lines crashes already in the bootstrapping of the `str` `ClassDef`. So, it seems like this this is necessary for 90% of the cases.

I wonder if we can come up with some sort of rule for the cases where it shouldn't be added. Is there some sort of commonality between the currently identified cases?

DanielNoord:
> I think it might only be a "gotcha" when creating fake nodes in the brain. When creating a fake ClassDef from scratch in the namespace brain, there's no logical parent, since we don't have the argparse module -- we could probably get away with not setting a parent at all.

We have deprecated `parent == None` for any node other than `nodes.Module`. I think this is beneficial in `pylint` as it removes a lot of `isinstance` checks.

jacobtylerwalls:
Same problem with inferring calls to `property()`, which creates a Property (inherited from FunctionDef.

This elaborate workaround is not observed when inferring a property:
https://github.com/pylint-dev/astroid/blob/ee2d4bd07f4d2b9646ba8904ad6337d474c66fa6/astroid/inference.py#L1246-L1267

jacobtylerwalls:
Fixed in #2588, can remove any related workarounds later as Refs #2588...

PR #2536, opened 2024-09-03
Clarify on non-Module roots

temyurchenko:
The PR fixes a missing parent on the "__main__" node and weakens the typing of "NodeNG.root" to better reflect the reality

I don't know what entry to fill in in the ChangeLog.

## Type of Changes

## Description

Refs #120, partially.

### DIFF
--- a/astroid/brain/brain_argparse.py
+++ b/astroid/brain/brain_argparse.py
@@ -21,7 +21,7 @@ def infer_namespace(node, context: InferenceContext | None = None):
         "Namespace",
         lineno=node.lineno,
         col_offset=node.col_offset,
-        parent=AstroidManager().adhoc_module,  # this class is not real
+        parent=AstroidManager().synthetic_root,  # this class is not real
         end_lineno=node.end_lineno,
         end_col_offset=node.end_col_offset,
     )
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -165,7 +165,7 @@ def _extend_builtins(class_transforms):
 
 def on_bootstrap():
     """Called by astroid_bootstrapping()."""
-    AstroidManager().cache_module(build_module("__astroid_adhoc"))
+    AstroidManager().cache_module(build_module("__astroid_synthetic"))
 
     _extend_builtins(
         {
@@ -653,7 +653,7 @@ def infer_property(
         # node.frame. It's somewhere in the builtins module, but we are special
         # casing it for each "property()" call, so we are making up the
         # definition on the spot, ad-hoc.
-        parent=AstroidManager().adhoc_module,
+        parent=AstroidManager().synthetic_root,
     )
     prop_func.postinit(
         body=[],
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -194,7 +194,7 @@ def infer_named_tuple(
     """Specific inference function for namedtuple Call node."""
     tuple_base: nodes.Name = _extract_single_node("tuple")
     class_node, name, attributes = infer_func_form(
-        node, tuple_base, parent=AstroidManager().adhoc_module, context=context
+        node, tuple_base, parent=AstroidManager().synthetic_root, context=context
     )
 
     call_site = arguments.CallSite.from_call(node, context=context)
@@ -360,7 +360,7 @@ def infer_enum(
     class_node = infer_func_form(
         node,
         enum_meta,
-        parent=AstroidManager().adhoc_module,
+        parent=AstroidManager().synthetic_root,
         context=context,
         enum=True,
     )[0]
--- a/astroid/manager.py
+++ b/astroid/manager.py
@@ -116,8 +116,8 @@ class AstroidManager:
         return self.astroid_cache["builtins"]
 
     @property
-    def adhoc_module(self) -> nodes.Module:
-        return self.astroid_cache["__astroid_adhoc"]
+    def synthetic_root(self) -> nodes.Module:
+        return self.astroid_cache["__astroid_synthetic"]
 
     @property
     def prefer_stubs(self) -> bool:
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -21,7 +21,7 @@ from typing import (
     overload,
 )
 
-from astroid import util
+from astroid import nodes, util
 from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidError,
@@ -43,7 +43,6 @@ else:
 
 
 if TYPE_CHECKING:
-    from astroid import nodes
     from astroid.nodes import _base_nodes
 
 
@@ -332,11 +331,13 @@ class NodeNG:
         :returns: The root node.
         """
         if not (parent := self.parent):
-            return self  # type: ignore[return-value] # Only 'Module' does not have a parent node.
+            assert isinstance(self, nodes.Module)
+            return self
 
         while parent.parent:
             parent = parent.parent
-        return parent  # type: ignore[return-value] # Only 'Module' does not have a parent node.
+        assert isinstance(parent, nodes.Module)
+        return parent
 
     def child_sequence(self, child):
         """Search for the sequence that contains this child.
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -44,7 +44,6 @@ from astroid.nodes import (
     Arguments,
     Const,
     NodeNG,
-    Unknown,
     _base_nodes,
     const_factory,
     node_classes,
@@ -2037,7 +2036,7 @@ class ClassDef(
             col_offset=0,
             end_lineno=0,
             end_col_offset=0,
-            parent=Unknown(),
+            parent=caller.parent,
         )
 
         # Get the bases of the class.
@@ -2071,7 +2070,6 @@ class ClassDef(
                 if isinstance(attr, node_classes.Const) and isinstance(attr.value, str):
                     result.locals[attr.value] = [value]
 
-        result.parent = caller.parent
         return result
 
     def infer_call_result(
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -1602,7 +1602,7 @@ class FunctionDef(
                     col_offset=0,
                     end_lineno=0,
                     end_col_offset=0,
-                    parent=AstroidManager().adhoc_module,
+                    parent=AstroidManager().synthetic_root,
                 )
                 new_class.hide = True
                 new_class.postinit(
--- a/astroid/objects.py
+++ b/astroid/objects.py
@@ -279,7 +279,7 @@ class PartialFunction(scoped_nodes.FunctionDef):
             name,
             lineno=lineno,
             col_offset=col_offset,
-            parent=AstroidManager().adhoc_module,
+            parent=AstroidManager().synthetic_root,
             end_col_offset=0,
             end_lineno=0,
         )
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -670,9 +670,8 @@ def _astroid_bootstrapping() -> None:
             col_offset=0,
             end_lineno=0,
             end_col_offset=0,
-            parent=nodes.Unknown(),
+            parent=astroid_builtin,
         )
-        _UnionTypeType.parent = astroid_builtin
         union_type_doc_node = (
             nodes.Const(value=types.UnionType.__doc__)
             if types.UnionType.__doc__
@@ -707,9 +706,8 @@ def _astroid_bootstrapping() -> None:
                 col_offset=0,
                 end_lineno=0,
                 end_col_offset=0,
-                parent=nodes.Unknown(),
+                parent=astroid_builtin,
             )
-            klass.parent = astroid_builtin
             doc = _type.__doc__ if isinstance(_type.__doc__, str) else None
             klass.postinit(
                 bases=[],
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -37,6 +37,7 @@ from astroid.exceptions import (
     ParentMissingError,
     StatementMissing,
 )
+from astroid.manager import AstroidManager
 from astroid.nodes.node_classes import (
     AssignAttr,
     AssignName,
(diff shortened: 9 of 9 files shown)


temyurchenko:
The reason the CI is failing is that at the class construction stage, it has "Unknown" for a parent. On the previous revision, the result was just that all builtin (or not just?) classes have "Unknown" as their `locals['__module__']`.
I'm thinking of a better way to approach it in this version.

DanielNoord:
I think this will cause a lot of type errors in `pylint` as we assume `root()` always returns... Can we get away by asserting?

temyurchenko:
> I think this will cause a lot of type errors in `pylint` as we assume `root()` always returns... Can we get away by asserting?

I'm currently in the process of eliminating non-module roots. Without that, the assert is going to fail a loot of times.

temyurchenko:
> I think this will cause a lot of type errors in `pylint` as we assume `root()` always returns... Can we get away by asserting?

What is the difference between `Uninferable` and `InferenceError`? Which one should be use and in what cases? Is there some documentation on it? Or a specific person to ask?

DanielNoord:
> > I think this will cause a lot of type errors in `pylint` as we assume `root()` always returns... Can we get away by asserting?
> 
> I'm currently in the process of eliminating non-module roots. Without that, the assert is going to fail a loot of times.

I don't understand what you mean with this?

> > I think this will cause a lot of type errors in `pylint` as we assume `root()` always returns... Can we get away by asserting?
> 
> What is the difference between `Uninferable` and `InferenceError`? Which one should be used and in what cases? Is there some documentation on it? Or a specific person to ask?

It depends on the situation. Ideally `InferenceError` isn't exposed to the users of `astroid` such as `pylint` so that they only need to handle `Uninferable`.

temyurchenko:
> I don't understand what you mean with this?

The `root()` currently returns non-Module nodes, often. So, if we just assert that the return is a Module, that assert is going to fail often.

> It depends on the situation. Ideally `InferenceError` isn't exposed to the users of `astroid` such as `pylint` so that they only need to handle `Uninferable`.

What is the logic for preferring one or the other internally? I often see lines like:

```
            if not name.startswith("__") and self.has_dynamic_getattr(context):
                # class handle some dynamic attributes, return a Uninferable object
                yield util.Uninferable
            else:
                raise InferenceError(
                    str(error), target=self, attribute=name, context=context
                ) from error
```
(this one is from `scoped_nodes.ClassDef.igetattr`)

temyurchenko:
> > I don't understand what you mean with this?
> 
> The `root()` currently returns non-Module nodes, often. So, if we just assert that the return is a Module, that assert is going to fail often.
> 
> > It depends on the situation. Ideally `InferenceError` isn't exposed to the users of `astroid` such as `pylint` so that they only need to handle `Uninferable`.
> 
> What is the logic for preferring one or the other internally? I often see lines like:
> 
> ```
>             if not name.startswith("__") and self.has_dynamic_getattr(context):
>                 # class handle some dynamic attributes, return a Uninferable object
>                 yield util.Uninferable
>             else:
>                 raise InferenceError(
>                     str(error), target=self, attribute=name, context=context
>                 ) from error
> ```
> 
> (this one is from `scoped_nodes.ClassDef.igetattr`)



> What is the logic for preferring one or the other internally? I often see lines like:

I see, raising the InferenceError doesn't let adding new inference results, while just yielding uninferrable does allow it.

temyurchenko:
Now, this should also close #1490. 

I tried my best to give complete descriptions to the commits, because they are lengthy. I also changed the annotation of `root()` back to just `Module` (with an added assert). 

I'm happy to elaborate more on changes, since there are quite a few of them.

As a side note, I still feel that it's not a very honest type signature, given how many nodes we are creating out-of-tree and thus without a parent (see `const_factory`). Furthermore, there is no enforcement of the `parent is not None` invariant anywhere in the codebase, so the accidents are bound to creep in.

temyurchenko:
Also fixes #2517

astroid/bases.py
```diff
@@ -5,6 +5,8 @@
 """This module contains base classes and functions for the nodes and some
 inference utils.
 """
+# isort: off
```
DanielNoord, line 8: Please remove this line

astroid/bases.py
```diff
@@ -273,27 +275,26 @@ def igetattr(
         try:
             context.lookupname = name
             # XXX frame should be self._proxied, or not ?
-            get_attr = self.getattr(name, context, lookupclass=False)
-            yield from _infer_stmts(
-                self._wrap_attr(get_attr, context), context, frame=self
-            )
+            attrs = self.getattr(name, context, lookupclass=False)
```
DanielNoord, line 278: Could you split out these changes into a separate commit/PR? I'd like to see the effect of these changes on the tests by itself.

astroid/brain/brain_argparse.py
```diff
@@ -20,13 +20,10 @@ def infer_namespace(node, context: InferenceContext | None = None):
         "Namespace",
         lineno=node.lineno,
         col_offset=node.col_offset,
-        parent=nodes.Unknown(),
+        parent=AstroidManager().adhoc_module,  # this class is not real
```
DanielNoord, line 23: Not sure about the name of `adhoc`. But I also don't have better suggestions

astroid/brain/brain_builtin_inference.py
```diff
@@ -4,6 +4,8 @@
 
 """Astroid hooks for various builtins."""
 
+# isort: off
```
DanielNoord, line 7: Please remove this line

astroid/brain/brain_builtin_inference.py
```diff
@@ -641,12 +646,13 @@ def infer_property(
 
     prop_func = objects.Property(
         function=inferred,
-        name=inferred.name,
+        name="<property>",
```
DanielNoord, line 649: Why this change? Doesn't that lose value?

astroid/brain/brain_builtin_inference.py
```diff
@@ -641,12 +646,13 @@ def infer_property(
...
         lineno=node.lineno,
         col_offset=node.col_offset,
+        # ↓ semantically, the definition of this property isn't within
+        # node.frame (or anywhere else, really)
+        parent=AstroidManager().adhoc_module,
```
DanielNoord, line 654: I don't understand this change? Why doesn't the property have the parent of the function being decorated as its parent?

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -146,15 +148,11 @@ def infer_func_form(
...
         end_col_offset=node.end_col_offset,
-        parent=nodes.Unknown(),
+        parent=parent,
     )
-    # A typical ClassDef automatically adds its name to the parent scope,
-    # but doing so causes problems, so defer setting parent until after init
-    # see: https://github.com/pylint-dev/pylint/issues/5982
-    class_node.parent = node.parent
     class_node.postinit(
         # set base class=tuple
```
DanielNoord, line 154: I think L156 can be remove?

astroid/interpreter/objectmodel.py
```diff
@@ -21,6 +21,8 @@
 mechanism.
 """
 
+# isort: off
```
DanielNoord, line 24: Not going to comment on all, but please remove these lines

DanielNoord (commented):
Awesome changes! But also pretty hard to review.

Maintainer time on `astroid` is sparse and nobody has actively worked on its internals for about a year.

I'd be open to spending some time of the next days/weeks to get this merged but I an only approve smaller bite sized PRs. If you're op for that let's go down that route!

For example, the first commit could be a separate PR to declutter this PR.
The second commit does a lot of things at once (as indicated by the commit message). I think a lot of changes in there could be separated into separate PRs, also to show their effects on the tests more clearly. That helps with reviewing them.

temyurchenko:
> I'd be open to spending some time of the next days/weeks to get this merged but I an only approve smaller bite sized PRs. If you're op for that let's go down that route!

For sure.

> For example, the first commit could be a separate PR to declutter this PR.

Will do.

> The second commit does a lot of things at once (as indicated by the commit message). I think a lot of changes in there could be separated into separate PRs, also to show their effects on the tests more clearly. That helps with reviewing them.

Will do.

Regarding isort. I've grepped the project and found a few similar isort comment. Based on that I concluded that the project uses isort to sort imports. However, when I ran isort on it, it resulted in quite a few changes, thus the "isort: off" comments. Could you please tell me if I should use isort? If so, is there a special config for the project? I wasn't able to find one.

astroid/brain/brain_argparse.py
temyurchenko: I can come up with other options. Please, tell what you are unsure about it, to guide the thought process for new names.

astroid/brain/brain_builtin_inference.py
temyurchenko: Right, I talk about it in the commit message:

- fixing construction of in-place properties (`bar =
   property(getter)`). They just create a nameless object, not the one
   with the name of the getter. Thus, the name was changed to
   "<property>". Furthermore, the definition of that property is not
   attached to any scope, as it's again nameless.

This change is for "inplace", nameless, properties, not the ones created with a decorator. 

What does it mean for it to lose the value? It's still accessible through an `Assign` field, for example.

astroid/brain/brain_builtin_inference.py
temyurchenko: See the comment above, this is about «inplace» properties, not the «decorator» properties.

DanielNoord:
https://github.com/pylint-dev/astroid/blob/7954bac2f4e84e3f0b1096be246b238a5651c76e/pyproject.toml#L146

We use `ruff` to do our import sorting now. Any comments you found are probably outdated (and could be removed in a separate PR)

astroid/brain/brain_builtin_inference.py
DanielNoord: Ah, makes sense. I understand the name change a little better now (although I would probably do `f"<property> {inferred.name}"` or something, so we don't lose that context.

I don't really understand the `parent` changes though.

```python
class A:
    temperature = property(get_temperature, set_temperature)
```

For `temperature` I would still expect `A` to be the `parent` no? Not the `adhoc` module

astroid/brain/brain_argparse.py
DanielNoord: @Pierre-Sassoulas Any suggestions here?

astroid/brain/brain_builtin_inference.py
temyurchenko: For `temperature` it is. However, temperature is just a name in the AssignName. It is assigned a call to the constructor of actual `property`. The actual `property` class is defined somewhere in the builtins module. However, since we are treating `property` specially, we are creating our own definition of it, separate for each use. Kind of like an Instance. And that special definition isn't defined anywhere. We've just made it up on the spot, ad-hoc. 

Treating it as if it was defined in `A` is furthermore incorrect, since it would cause it to be in `A`s body, or searchable from `A`.

temyurchenko:
> I think a lot of changes in there could be separated into separate PRs, also to show their effects on the tests more clearly.

Is it possible to have something like an merge train on github? That is, for one PR to depend on some previous PR? It's possible to emulate with merging into branches, but then you'd have to do reviews on my local fork, since I can't create branches in this repository.

DanielNoord:
> > I think a lot of changes in there could be separated into separate PRs, also to show their effects on the tests more clearly.
> 
> 
> 
> Is it possible to have something like an merge train on github? That is, for one PR to depend on some previous PR? It's possible to emulate with merging into branches, but then you'd have to do reviews on my local fork, since I can't create branches in this repository.
> 
> 
> 
> If not, the simplest way would probably be do just breakup commits within this PR. Would you be open to that?

Sadly it doesn't. 

I'm okay with splitting up in this PR but I think that increases the risk of me (and other maintainers) finding it hard to approve of the full PR and merge it.

Separate PRs make it much easier to spend 15-20 minutes during lunch to review and merge the smaller commits. From experience I know although it involves more rebase work in the end it significantly increases the chances of `astroid` PRs landing on `main`.

temyurchenko:
> I'm okay with splitting up in this PR but I think that increases the risk of me (and other maintainers) finding it hard to approve of the full PR and merge it.

Sure, I'll do my best to achieve that.

temyurchenko:
I've made the first batch of PRs that cover a big part of this one. One thing is, #2557 is better reviewed the last.

Pierre-Sassoulas:
The secret to merge PR with volunteer maintainers has been told. I have multiple 5mn free windows on mobile during the days. I'll hit that merge button when my rust code is compiling, when the pasta's water is heating, when I'm waiting for the train, or when a butterfly distracted me and I checked the notification on my phone. I have some 10mn/20mn windows during the week. But one hour of uninterrupted concentration time on PC ? I have to *plan* ahead for that, it's not going to happens (not in a reasonable time at least)

temyurchenko:
> The secret to merge PR with volunteer maintainers has been told. I have multiple 5mn free windows on mobile during the days. I'll hit that merge button when my rust code is compiling, when the pasta's water is heating, when I'm waiting for the train, or when a butterfly distracted me and I checked the notification on my phone. I have some 10mn/20mn windows during the week. But one hour of uninterrupted concentration time on PC ? I have to _plan_ ahead for that, it's not going to happens (not in a reasonable time at least)

I did my best to split off the PR into smaller ones, as you can see in the comment above. They currently don't cover the whole of this PR, but I will finish the job some time later.

astroid/brain/brain_builtin_inference.py
DanielNoord: I understand now, thanks!

DanielNoord:
I think I went through all PRs and merged the first one that should solve most rebase issues. Let me know when you rebased the others or need me to do another review!

DanielNoord:
To give an update: changes coming a long nicely. @temyurchenko has been very helpful in rebasing and responding to comments. 

https://github.com/pylint-dev/astroid/pull/2563 is on my radar, but after a long week at work I don't really have the capacity to have a good look at it. I hope to do so this weekend or on Monday.

DanielNoord (dismissed):
Awesome! Great job working with us to get this merged. Working with you was a joy!

_I would be okay with adding a Changelog for this as I think you work should be recognized!_

astroid/brain/brain_argparse.py
Pierre-Sassoulas: Maybe ``synthetic_root_node``, ``placeholder_root_node``,  ``placeholder_parent_node``,  ``placeholder_module_node`` ?  wasn't sure what adhoc meant exactly before checking the definition to be honest.

Pierre-Sassoulas (changes requested):
Thank you for your work on this @temyurchenko ! There's some warning in the CI job that need to be fixed before merging and we could rename "adhoc_module" if it makes sense, but LGTM.

temyurchenko:
> Awesome! Great job working with us to get this merged. Working with you was a joy!
> 
> _I would be okay with adding a Changelog for this as I think you work should be recognized!_

Thank you, same for you and @Pierre-Sassoulas! 

I will do one final pass over the changes here, fix the commit message and fix the the linting errors. Not sure, if the changelog is needed, since this is mostly an internal change.

temyurchenko:
> we could rename "adhoc_module" if it makes sense

Okay, I'll go with "synthetic_root".

Pierre-Sassoulas (approved):

PR #2538, opened 2024-09-04
Remove unused _path member from Finder

correctmost:
## Type of Changes

## Description

`_path` appears to be unused in Pylint and astroid.

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -83,8 +83,8 @@ class ModuleSpec(NamedTuple):
 class Finder:
     """A finder is a class which knows how to find a particular module."""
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Feel like there's a design issue there, either the parameter is useless and should not be in the API or it is and it should be used. Maybe the smallest amount of change would be to start using ``_path`` in child class after initialization ? I.e.:

```python
class ZipFinder(Finder):
    """Finder that knows how to find a module inside zip files."""
    def __init__(self, path: Sequence[str]) -> None:
        super().__init__(path)
        for entry_path in self._path:
            # ...
```           
A little ugly, I know, but I have no idea if external code use this Finder API and if we can change it like this without breaking user code (There's approx 500k download of astroid each months that are not linked to a pylint download.)

Issue #2014, opened 2023-02-07
[discussion] Use a rust crate to make astroid faster, or part of astroid faster


Pierre-Sassoulas:
I was reading [this article](https://www.peterbaumgartner.com/blog/wrapping-a-rust-crate-in-a-python-package/) lately. With pydantic, ruff and other python package using rust to provide faster python using binding offered by https://github.com/PyO3/pyo3 . I've been wondering if we should explore the possibility to do that too.

Potential contributor to astroid are scare, so we rejected doing cython earlier (In https://github.com/PyCQA/astroid/pull/606#issuecomment-498015223 and elsewhere with hippo91 but I can't find the link right now). 

However:
- rust is well loved and has been the most loved language for a few year on stackoverflow. So we might find contributors after all ?
- ``ruff`` performance proves it's possible as a linter and also that it's a massive boost (10x or more) and a game changer
- astroid's code is convoluted, maybe a way to avoid painfully typing it and maintaining it ? Daniel wanted to make it internal, well, porting to rust is a way to do whatever we like ? 😄 

Problem I foresee:
- I don't know how astroid works, it's not a given that we're going to architecture a new package better ?
- We rely on the ast module right now, I don't know if we can rely on the ast from rust. The fact that ruff do not handle pattern matching seems to indicate that no we can't.

Maybe it's possible to use rust on the bottleneck only, i.e. on inference.

tushar-deepsource:
A few notes:
- `ruff` uses RustPython, and RustPython's AST doesn't yet parse match statements. However you can instead use `libcst`'s parser, which is heavier but supports match statements. (Although, maintenance of libcst has slowed down recently and I've found a few bugs in it myself.)
- You can definitely write an astroid tree on top of those two parser's AST.

The major problem I see is that if you move away from Python's `ast` module, you gain a massive burden of maintaining whatever ast library that you use, be it RustPython or something else, everytime a user finds a bug or Python has a syntax changing update.

DanielNoord:
I think the main issue with contributing to `astroid` is the state of the codebase. It's just very unclear how everything works.
Any attempt to fix that is met with "breaking changes" concerns. I would love to spend some time just breaking full API compatibility within `astroid` and make sure all typing of the `nodes` is correct. That would already help so much, but to do so we would need to make a conscious decision to break backwards compatibility with a `3.0` cut off.

Pierre-Sassoulas:
I think it's a given that breaking the API will need to be done:
[...]

PR #2541, opened 2024-09-04
Fix additional mypy errors and expand CI checks to more files

correctmost:
## Type of Changes

## Description
Clean-up of some of the simpler mypy errors.

The import changes are to fix errors like this:
```
astroid/brain/brain_namedtuple_enum.py:16: error: Module "astroid" does not explicitly export attribute "inference_tip"  [attr-defined]
```


Pierre-Sassoulas (approved):
Wow, amazing that you were able to activate so much files at once and use ``nodes.X`` type import style without triggering a cyclic import. This is great work, and very important work on the path to make astroid really faster in the future 👏. (refs #2014)

I'm going to wait for a review by Daniel, Marc or Jacob that are more knowledgeable in astroid typing than myself but this is 🔥

pyproject.toml
```diff
@@ -64,7 +64,40 @@ filterwarnings = "error"
 [tool.mypy]
 python_version = "3.9"
 files = [
+    "astroid/_ast.py",
+    "astroid/_backport_stdlib_names.py",
+    "astroid/astroid_manager.py",
+    "astroid/brain/brain_crypt.py",
```
DanielNoord, line 70: Can we use `globs` here? And just enable for all of `brain`?

astroid/brain/brain_hypothesis.py
```diff
@@ -36,12 +36,12 @@ def is_decorated_with_st_composite(node) -> bool:
...
 
 
-def remove_draw_parameter_from_composite_strategy(node):
+def remove_draw_parameter_from_composite_strategy(node: FunctionDef) -> FunctionDef:
     """Given that the FunctionDef is decorated with @st.composite, remove the
     first argument (`draw`) - it's always supplied by Hypothesis so we don't
     need to emit the no-value-for-parameter lint.
     """
-    del node.args.args[0]
+    del node.args.args[0]  # type: ignore[union-attr]
```
DanielNoord, line 44: Shouldn't we do asserts here? Instead of ignoring?

astroid/brain/brain_hypothesis.py
```diff
@@ -36,11 +38,14 @@ def is_decorated_with_st_composite(node) -> bool:
     return False
 
 
-def remove_draw_parameter_from_composite_strategy(node):
+def remove_draw_parameter_from_composite_strategy(node: FunctionDef) -> FunctionDef:
     """Given that the FunctionDef is decorated with @st.composite, remove the
     first argument (`draw`) - it's always supplied by Hypothesis so we don't
     need to emit the no-value-for-parameter lint.
     """
+    if TYPE_CHECKING:
```
DanielNoord, line 46: I think it is better to actually do this assert always. `mypy` is showing us we have unsafe code, we should make it safe by asserting our assumption so we can see where they don't hold and handle it accordingly.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #2554, opened 2024-09-09
Fix partial function parent

temyurchenko:
A part of getting rid of non-Module roots, see https://github.com/pylint-dev/astroid/pull/2536


DanielNoord (approved):
If you could rebase this I can merge it!

PR #2555, opened 2024-09-09
Fix namedtuple parent

temyurchenko:
A part of getting rid of non-Module roots, see https://github.com/pylint-dev/astroid/pull/2536

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -74,7 +74,9 @@ def _find_func_form_arguments(node, context):
 
 def infer_func_form(
     node: nodes.Call,
-    base_type: list[nodes.NodeNG],
+    base_type: nodes.NodeNG,
+    *,
+    parent: nodes.NodeNG,
     context: InferenceContext | None = None,
     enum: bool = False,
 ) -> tuple[nodes.ClassDef, str, list[str]]:
@@ -147,15 +149,10 @@ def infer_func_form(
         col_offset=node.col_offset,
         end_lineno=node.end_lineno,
         end_col_offset=node.end_col_offset,
-        parent=nodes.Unknown(),
+        parent=parent,
     )
-    # A typical ClassDef automatically adds its name to the parent scope,
-    # but doing so causes problems, so defer setting parent until after init
-    # see: https://github.com/pylint-dev/pylint/issues/5982
-    class_node.parent = node.parent
     class_node.postinit(
-        # set base class=tuple
-        bases=base_type,
+        bases=[base_type],
         body=[],
         decorators=None,
     )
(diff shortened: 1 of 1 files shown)


astroid/brain/brain_namedtuple_enum.py
```diff
@@ -195,25 +193,18 @@ def infer_named_tuple(
...
-        nodes.Name(
-            name="tuple",
-            parent=node.root(),
-            lineno=0,
-            col_offset=0,
-            end_lineno=None,
-            end_col_offset=None,
-        )
-    ]
+    tuple_base = util.safe_infer(_extract_single_node("tuple"))
```
DanielNoord, line 196: Why are you inferring here? Shouldn't this just be the `nodes.Name` for `tuple`? Similar to the comment you make about `EnumMeta` preferably being the `Name` reference?

temyurchenko: My motivation for using the actual definition instead of a Name was to avoid the situation where the actual tuple is shadowed, and thus the Name refers to something different. 

However, looking at it my current implementation doesn't achieve that. A better way to do it would be something like `tuple_base = builtin_module["tuple"]`. If you think that my worry above is weakly-founded, I can revert to a `Name`

temyurchenko: (I talk about it in some more detail in the commit message)

temyurchenko: Hmm, actually it appears that `tuple_base: nodes.Name = _extract_single_node("tuple")` works pretty well. It's not going to be shadowed, because it belongs to an empty module. Should we go with that?

DanielNoord (approved):
Although I think I'd like @Pierre-Sassoulas to also have a look here

temyurchenko:
(removed outdated comment, "#set base class=tuple")

DanielNoord (approved):

Pierre-Sassoulas (commented):
I feel like those kind of change would benefit from being tested in pylint using the primers. This is not done automatically it would be bad to have surprises after we release a version of astroid. Do you know how to do that @temyurchenko ? (pip install a branch of astroid in editable mode then launch pylint's tests then pylint's primer)

temyurchenko:
> I feel like those kind of change would benefit from being tested in pylint using the primers. This is not done automatically it would be bad to have surprises after we release a version of astroid. Do you know how to do that @temyurchenko ? (pip install a branch of astroid in editable mode then launch pylint's tests then pylint's primer)

@Pierre-Sassoulas, the pylint tests have a fail due to a bug unrelated to this PR (I've submitted a fix here: https://github.com/pylint-dev/pylint/pull/9966). The primer succeeds. So, this one should be safe to merge.

temyurchenko:
Is there anything else to be done with this PR or is it okay to merge, @Pierre-Sassoulas?

Pierre-Sassoulas (approved):
Sorry this went under my radar. Great job !

PR #2556, opened 2024-09-09
Fix doc  

temyurchenko:

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -2061,7 +2061,16 @@ class Const(_base_nodes.NoChildrenNode, Instance):
         :param end_col_offset: The end column this node appears on in the
             source code. Note: This is after the last symbol.
         """
-        self.value: Any = value
+        if getattr(value, "__name__", None) == "__doc__":
+            warnings.warn(  # pragma: no cover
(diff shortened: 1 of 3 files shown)


astroid/nodes/node_classes.py
```diff
@@ -2061,7 +2061,14 @@ def __init__(
         :param end_col_offset: The end column this node appears on in the
             source code. Note: This is after the last symbol.
         """
-        self.value: Any = value
+        if getattr(value, "__name__", None) == "__doc__":
+            raise AssertionError(  # pragma: no cover
```
DanielNoord, line 2065: Is this still needed now that you have fixed them?

temyurchenko: It was tricky to debug, so I thought I might leave this as a help for future situations, as it's trivial to introduce this error. If you think it's unnecessary, I'll delete it

DanielNoord: I think if we want debugging we should make it a warning instead. It would be very annoying for users to have `pylint` crash just because we introduced a small oopsie somewhere. Before we release both `astroid` and `pylint` we're often a couple weeks further so I think the `AssertionError` is too much here.

DanielNoord (approved):

PR #2557, opened 2024-09-09
Require build class parent

temyurchenko:

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -178,6 +178,13 @@ def function_to_method(n, klass):
     return n
 
 
+def _attach_to_parent(node: NodeNG, name: str, parent: NodeNG):
+    frame = parent.frame()
+    frame.set_local(name, node)
+    if frame is parent:
+        frame._append_node(node)
+
+
 class Module(LocalsDictNodeNG):
     """Class representing an :class:`ast.Module` node.
 
@@ -1935,7 +1942,7 @@ class ClassDef(  # pylint: disable=too-many-instance-attributes
             parent=parent,
         )
         if parent and not isinstance(parent, Unknown):
-            parent.frame().set_local(name, self)
+            _attach_to_parent(self, name, parent)
 
         for local_name, node in self.implicit_locals():
             self.add_local_node(node, local_name)
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -1704,8 +1704,7 @@ def test_infer_dict_from_keys() -> None:
     """
     )
     for node in bad_nodes:
-        with pytest.raises(InferenceError):
-            next(node.infer())
+        assert isinstance(next(node.infer()), util.UninferableBase)
 
     # Test uninferable values
     good_nodes = astroid.extract_node(
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -1704,7 +1704,9 @@ def test_infer_dict_from_keys() -> None:
     """
     )
     for node in bad_nodes:
-        assert isinstance(next(node.infer()), util.UninferableBase)
+        with pytest.raises(InferenceError):
+            if isinstance(next(node.infer()), util.UninferableBase):
+                raise InferenceError
 
     # Test uninferable values
     good_nodes = astroid.extract_node(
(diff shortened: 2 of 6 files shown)


tests/brain/test_brain.py
```diff
@@ -1705,7 +1705,8 @@ def test_infer_dict_from_keys() -> None:
     )
     for node in bad_nodes:
         with pytest.raises(InferenceError):
-            next(node.infer())
+            if isinstance(next(node.infer()), util.UninferableBase):
```
DanielNoord, line 1708: Why is this necessary?

astroid/nodes/node_classes.py
```diff
@@ -2061,7 +2061,14 @@ def __init__(
         :param end_col_offset: The end column this node appears on in the
             source code. Note: This is after the last symbol.
         """
-        self.value: Any = value
+        if getattr(value, "__name__", None) == "__doc__":
+            raise AssertionError(  # pragma: no cover
```
DanielNoord, line 2065: I don't think this belongs in this PR right?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -178,6 +178,13 @@ def function_to_method(n, klass):
     return n
 
 
+def attach_to_parent(node: NodeNG, name: str, parent: NodeNG):
```
DanielNoord, line 181:
```suggestion
def _attach_to_parent(node: NodeNG, name: str, parent: NodeNG):
```

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -178,6 +178,13 @@ def function_to_method(n, klass):
...
+    frame = parent.frame()
+    frame.set_local(name, node)
+    if frame is parent:
+        frame._append_node(node)
```
DanielNoord, line 185: This feels like a crucial change that we don't have a test for. Should we add one?

DanielNoord (commented):
Awesome!

astroid/nodes/node_classes.py
temyurchenko: The first three commits in this PR will be merged in the previous PRs. Unfortunately, they are a necessity. The issue that these commits fix are uncovered by the last commit, so without them there are going to be errors.

tests/brain/test_brain.py
temyurchenko: For this one, I don't know exactly, but my intuition was the following. Before, when we had `Unknown` for a parent, when inference encountered Unknown, it would raise (`InferenceError`). Now, there is no `Unknown`, but the calls still aren't correct. So, we just fail to infer and yield `Uninferable`. 

However, both responses kind of seem to make sense, so I opted to allow either of the responses (`InferenceError` or `Uninferable`) to pass the tests.

DanielNoord: Can we be certain that it is either of the two? So, just assert that it is always `Uniferable` or always the exception? If so, I'd prefer just testing for that

temyurchenko: Yes, right now it's always `Uninferable`. I'll change to test specifically for that.

astroid/nodes/scoped_nodes/scoped_nodes.py
temyurchenko: So, usually the classes get added to the body of the parent in the positinit constructor. There is only one exception to my knowledge: `raw_building._add_dunder_class`. It was added solely to support some `pyqt` inference and it is creating (and attaching) adhoc classes for the purposes of being later detected by the transform process. 

That situation is already tested by the three qt tests (which fail without this change). In other situations it's useless, afaics.

Overall, the process described above seems very perverse, the "dunder" class shouldn't be recreated within a function each time. It is its own class that is defined outside. We should rather consult the type annotation in the transform process. However, it is a question for another PR.

astroid/nodes/scoped_nodes/scoped_nodes.py
temyurchenko: fixed

tests/brain/test_brain.py
temyurchenko: fixed

temyurchenko: Ah, one test fails under `pypy` with `InferenceError`. I'd rather go with the initial `pytest.raises` clause then

DanielNoord (approved):

PR #2562, opened 2024-09-11
enforce a non-None parent in build_function

temyurchenko:

We also remove `add_local_node` to avoid redundancy. Instead we do the
   attachment to the parent scope in the constructor of `FunctionDef`.

We append a node to the body of the frame when it is also the
   parent. If it's not a parent, then the node should belong to the
   "body" of the parent if it existed. An example is a definition
   within an "if", where the parent is the If node, but the frame is
   the whole module.

## Type of Changes

## Description

Refs #XXXX

Closes #XXXX

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -136,6 +136,7 @@ def build_class(
 
 def build_function(
     name: str,
+    parent: nodes.NodeNG,
     args: list[str] | None = None,
     posonlyargs: list[str] | None = None,
     defaults: list[Any] | None = None,
@@ -149,7 +150,7 @@ def build_function(
         name,
         lineno=0,
         col_offset=0,
-        parent=node_classes.Unknown(),
+        parent=parent,
         end_col_offset=0,
         end_lineno=0,
     )
@@ -321,6 +322,7 @@ def object_build_function(
 
     return build_function(
         getattr(member, "__name__", "<no-name>"),
+        node,
         args,
         posonlyargs,
         defaults,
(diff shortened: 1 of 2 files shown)


temyurchenko:
Review #2564 first, please

DanielNoord (commented):
This needs to be rebased and given a new name to the PR, but the `localname` change LGTM.

For posterity the `localname` stuff was introduced in https://github.com/pylint-dev/astroid/commit/4cdfc82457a0962c2dcb1ebc2a1129d4e4863170

If I understand it correctly, when doing the initial `object_build` we try to access `getattr(obj, name).__name__` However, this can fail so the code was changed to also have a reference to `name` itself as a fallback.
The changes in this PR still allow this pattern so they look good. Please rebase so I can merge it.

DanielNoord:
Never mind my previous comment. The second commit in this PR was merged in https://github.com/pylint-dev/astroid/pull/2557.

The first commit was correctly split out to https://github.com/pylint-dev/astroid/pull/2564 which I'll approve and merge shortly with the same comment.

@temyurchenko Is a commit missing from this PR? Or should it be closed/

temyurchenko:
> Never mind my previous comment. The second commit in this PR was merged in #2557.

Hmm, no, that commit was about `build_object`. This one is about `build_function`. The changes are similar but applied to different symbols.

DanielNoord (approved):
Last commit in this PR (after rebase of course) LGTM!

DanielNoord:
Could you rebase this so I can merge it? :)

temyurchenko:
> Could you rebase this so I can merge it? :)

Done!

PR #2563, opened 2024-09-11
use the ClassDef attribute inference process in Instances

temyurchenko:

Instances often have «special_attributes». ClassDef does a lot of
   transformations of attibutes during inference, so let's reuse them
   in Instance.

We are fixing inference of the "special" attributes of
   Instances. Before these changes, the FunctionDef attributes
   wouldn't get wrapped into a BoundMethod. This was facilitated by
   extracting the logic of inferring attributes into
   'FunctionDef._infer_attrs' and reusing it in BaseInstance.

  This issue isn't visible right now, because the special attributes
   are simply not found due not being attached to the parent (the
   instance). Which in turn is caused by not providing the actual
   parent in the constructor. Since we are on a quest to always
  provide a parent, this change is necessary.

### DIFF
--- a/tests/test_scoped_nodes.py
+++ b/tests/test_scoped_nodes.py
@@ -1295,7 +1295,7 @@ class ClassNodeTest(ModuleLoader, unittest.TestCase):
             self.assertIsInstance(inferred[0], BoundMethod)
         inferred = list(Instance(cls).igetattr("m4"))
         self.assertEqual(len(inferred), 1)
-        self.assertIsInstance(inferred[0], nodes.FunctionDef)
+        self.assertIsInstance(inferred[0], BoundMethod)
 
     def test_getattr_from_grandpa(self) -> None:
         data = """
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -273,10 +273,9 @@ class BaseInstance(Proxy):
         try:
             context.lookupname = name
             # XXX frame should be self._proxied, or not ?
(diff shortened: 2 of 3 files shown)


temyurchenko:
Note that the biggest source of changes is just moving a bunch of code into a separate function

tests/test_scoped_nodes.py
```diff
@@ -1295,7 +1295,7 @@ def func(arg1, arg2):
             self.assertIsInstance(inferred[0], BoundMethod)
         inferred = list(Instance(cls).igetattr("m4"))
         self.assertEqual(len(inferred), 1)
-        self.assertIsInstance(inferred[0], nodes.FunctionDef)
+        self.assertIsInstance(inferred[0], BoundMethod)
```
DanielNoord, line 1298: I think this change is incorrect, and therefore shows something is up with this PR.

To reproduce:
```python
class Clazz(object): ...


def func(arg1, arg2):
    "function that will be used as a method"
    return arg1.value + arg2


Clazz.m3 = func
inst = Clazz()
inst.m4 = func

print(Clazz.m3)
print(inst.m3)
print(inst.m4)
```


```shell
<function func at 0x109582340>
<bound method func of <__main__.Clazz object at 0x1095ab9b0>>
<function func at 0x109582340>
```

This should remain a `FunctionDef` and not a `BoundMethod`.

DanielNoord (changes requested):

temyurchenko:
Closing in favour of #2584

PR #2564, opened 2024-09-11
always use the local name when building definitions

temyurchenko:

Sometimes a class accesses a member by a different name than
   "__name__" of that member: in pypy3 `list.__mul__.__name__ ==
   "__rmul__"`.

As a result, in the example above we weren't able to find
   "list.__mul__", because it was recorded only as "list.__rmul__".

### DIFF
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -266,7 +266,9 @@ def object_build_class(
 ) -> nodes.ClassDef:
     """create astroid for a living class object"""
     basenames = [base.__name__ for base in member.__bases__]
-    return _base_class_object_build(node, member, basenames, localname=localname)
+    # N.B. ↓ why do we need to override the localname?
+    name = getattr(member, "__name__", localname)
+    return _base_class_object_build(node, member, basenames, name)
 
 
 def _get_args_info_from_callable(
@@ -316,7 +318,7 @@ def object_build_function(
     ) = _get_args_info_from_callable(member)
 
     func = build_function(
-        getattr(member, "__name__", None) or localname,
+        localname,
         args,
         posonlyargs,
         defaults,
@@ -342,9 +344,7 @@ def object_build_methoddescriptor(
 ) -> None:
     """create astroid for a living method descriptor object"""
     # FIXME get arguments ?
-    func = build_function(
-        getattr(member, "__name__", None) or localname, doc=member.__doc__
-    )
+    func = build_function(localname, doc=member.__doc__)
     node.add_local_node(func, localname)
     _add_dunder_class(func, node, member)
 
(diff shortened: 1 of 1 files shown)


DanielNoord (changes requested):
For posterity the localname stuff was introduced in https://github.com/pylint-dev/astroid/commit/4cdfc82457a0962c2dcb1ebc2a1129d4e4863170

If I understand it correctly, when doing the initial object_build we try to access getattr(obj, name).__name__ However, this can fail so the code was changed to also have a reference to name itself as a fallback.

Now that I'm re-reviewing this. Could you add a link to the commit in the `NB` comment?

temyurchenko:
> Now that I'm re-reviewing this. Could you add a link to the commit in the `NB` comment?

Hmm, the commit above answers the question «why have local name» with «because sometimes `.__name__` is missing». However, my question is «why possibly we might want to use `.__name__` over of local name».

temyurchenko:
> If I understand it correctly, when doing the initial object_build we try to access getattr(obj, name).**name** However, this can fail so the code was changed to also have a reference to name itself as a fallback.

It doesn't exactly fail, it's just that in pypy they do stuff like list.__mul__ = list.__rmul__. And now even though list.__mul__ is accessed via __mul__ localname, it's actually the __rmul__ object with the corresponding .__name__.

DanielNoord:
So I found the original bug report:
https://www.mail-archive.com/python-projects@lists.logilab.org/msg00009.html

I think the fix as in this PR is still correct: we should always use `__name__` as that is the name of the object we are building. `localname` should only be used a fallback if `__name__` isn't set due to a c-extension not behaving like regular Python code.

What do you think @temyurchenko?

temyurchenko:
> So I found the original bug report: https://www.mail-archive.com/python-projects@lists.logilab.org/msg00009.html
> 
> I think the fix as in this PR is still correct: we should always use `__name__` as that is the name of the object we are building. `localname` should only be used a fallback if `__name__` isn't set due to a c-extension not behaving like regular Python code.
> 
> What do you think @temyurchenko?
> 
> Edit: Upon reading the code again I think the change to `object_build_methoddescriptor` is actually incorrect. We should still prefer `__name__` there

Looking at the linked report, the issue was the following: symbols from C modules don't have `__name__`, thus we fail if we try to use `__name__`. 

It's still not clear to me, why we should prefer `__name__` over `localname`. I showed an example above where preferring `__name__` over `localname` leads to a bug and is just incorrect logically. If the class chooses to use a symbol under a different name (i.e. the `localname`), that's how we should look it up. Otherwise, the symbol isn't going to be found.

Why should we prefer `__name__`?

temyurchenko:
Tested this one with pylint tests and the primer, no regressions.

temyurchenko:
> Why should we prefer `__name__`?

I've digged and I found an answer for why we have to prefer `__name__` for classes. The classes are cached in `builder._done`. So, when we build a class once with a `localname`, all the other builds reuse that class with the same `localname`. Which is incorrect.

There are ways to deal with that. I feel like the clearest one is to separate attachment to the locals of a parent from the constructor. Since any object can be attached to the parent by a variety of different names that have no relation to the object itself, it would be clearer to not make it the object's responsibility.

So, I am creating a different PR meant to supersede this one. In that PR, I split out adding symbols to the `locals` of a parent from the constructor to the outside. So, closing this one in favour of #2588.

PR #2578, opened 2024-09-23
fix issue when inferring single-node or non-const JoinedStr

ericvergnaud:
## Type of Changes

## Description

Supports inference of single-node and non-const JoinedStr

Fixes #[9947](https://github.com/pylint-dev/pylint/issues/9947)


jacobtylerwalls (approved):
Thanks for the quick fix!

PR #2583, opened 2024-09-24
change the type annotation error heuristic

temyurchenko:
The previous one depended on the message from "typed_ast", which is
   not used anymore.

Instead, we check if there is a "# type:" substring in the source line
   of the exception. This can yield some false positives, but probably
   rarely.

## Description

Partially addresses https://github.com/pylint-dev/pylint/issues/3757

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import ast
 import os
+import re
 import textwrap
 import types
 import warnings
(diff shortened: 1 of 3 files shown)


DanielNoord (approved):

temyurchenko:
> Thank you for the dive into the internal, much appreciated. Do you think it's possible to start testing this @temyurchenko ? I expected the coverage to go up but it wasn't tested before which explain that the dead code using a dead library is still there.

I'll try to add a test

temyurchenko:
> Thank you for the dive into the internal, much appreciated. Do you think it's possible to start testing this @temyurchenko ? I expected the coverage to go up but it wasn't tested before which explain that the dead code using a dead library is still there.

I've added a test and unblocked another one. It seems to have fixed https://github.com/pylint-dev/pylint/issues/3757 and possibly https://github.com/pylint-dev/pylint/issues/3556.

Pierre-Sassoulas (approved):
Amazing !!

PR #2585, opened 2024-09-25
Move typing imports and some definitions behind guard

nickdrozd:
## Type of Changes

## Description

Taking https://github.com/pylint-dev/pylint/pull/9964 for a spin. Idea here is that imports that are only used for typechecking should only be imported for typechecking. This will clarify module dependency structure and possibly also reduce startup time and memory use. Pylint / Astroid is slow and complicated, so this sort of change will feed two birds with one scone.

Mostly this changes imports. There are also some type definitions that are used only for typechecking. The new checker doesn't look at those (yet), so they have not been systematically moved under the type guard. But ultimately they should be.

### DIFF
--- a/astroid/arguments.py
+++ b/astroid/arguments.py
@@ -4,12 +4,16 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from astroid import nodes
 from astroid.bases import Instance
 from astroid.context import CallContext, InferenceContext
 from astroid.exceptions import InferenceError, NoDefault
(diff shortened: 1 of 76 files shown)


Pierre-Sassoulas:
Lot of work on this one, nice 🔥 I'm going to benchmark this when I find the time

cdce8p (changes requested):
https://github.com/pylint-dev/pylint/pull/9970#pullrequestreview-2329514138

PR #2588, opened 2024-09-27
control setting nodes as local outside of the constructor

temyurchenko:
1. The main reason is that a node might be assigned to its parent via an «alias»:

Sometimes a class accesses a member by a different name than
   "__name__" of that member: in pypy3 `list.__mul__.__name__ ==
   "__rmul__"`.

As a result, in the example above we weren't able to find
   "list.__mul__", because it was recorded only as "list.__rmul__".

2. Sometimes we want to have a parent semantically, but we don't want to add it to the list of locals. For example, when inferring properties we are creating ad-hoc properties. We wouldn't want to add another symbol to the locals every time we do an inference. (actually, there's a very good question as to why we are doing those ad-hoc properties but that's out of scope)

No pylint regressions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,10 @@ What's New in astroid 3.3.5?
 ============================
 Release date: TBA
 
+* Control setting local nodes outside of the supposed local's constructor.
+
+  Closes pylint-dev/astroid/issues/1490
+
 * Fix regression with f-string inference.
 
   Closes pylint-dev/pylint#9947
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -178,13 +178,6 @@ def function_to_method(n, klass):
     return n
 
 
-def _attach_to_parent(node: NodeNG, name: str, parent: NodeNG):
-    frame = parent.frame()
-    frame.set_local(name, node)
-    if frame is parent:
-        frame._append_node(node)
-
-
 class Module(LocalsDictNodeNG):
     """Class representing an :class:`ast.Module` node.
 
@@ -1173,9 +1166,6 @@ class FunctionDef(
             end_col_offset=end_col_offset,
             parent=parent,
         )
-        if parent and not isinstance(parent, Unknown):
-            frame = parent.frame()
-            frame.set_local(name, self)
 
     def postinit(
         self,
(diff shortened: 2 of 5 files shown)


temyurchenko:
Unfortunately, this PR turned out to be a bit big. A big part of changes is renaming `name` to `alias` in `InspectBuilder.object_build`. I felt that `name` was quite confusing as to what relation it had to the object being built. In fact, it is just an alias for an object within its parent. The object itself might have a completely different, very own name.

temyurchenko:
Supersedes #2564, quoting my comment from there for clarity:

> I've digged and I found an answer for why we have to prefer __name__ for classes. The classes are cached in builder._done. So, when we build a class once with a localname, all the other builds reuse that class with the same localname. Which is incorrect.

> There are ways to deal with that. I feel like the clearest one is to separate attachment to the locals of a parent from the constructor. Since any object can be attached to the parent by a variety of different names that have no relation to the object itself, it would be clearer to not make it the object's responsibility.

> So, I am creating a different PR meant to supersede this one. In that PR, I split out adding symbols to the locals of a parent from the constructor to the outside. So, closing this one in favour of https://github.com/pylint-dev/astroid/pull/2588.

DanielNoord (dismissed):
Changes make sense to me!

I have asked Jacob for review as he has looked into this pattern before and might know of edge cases that I don't know about.
Awesome work on the investigation!

jacobtylerwalls:
Thanks for this! Does this close #1490? If so, can you add a changelog?

> No pylint regressions.

Thank you for checking the pylint test suite. I think these are the kinds of PRs that would also benefit from a pylint primer run. I've toyed with adding a github action to astroid to run the pylint primer action. I need to just finish it up and get it in. Would you mind waiting for me to do that? I promise I won't hold this up for weeks.

temyurchenko:
> Thank you for checking the pylint test suite. I think these are the kinds of PRs that would also benefit from a pylint primer

I've ran the primer as well. 

> I need to just finish it up and get it in. Would you mind waiting for me to do that? I promise I won't hold this up for weeks.

There are a few other PRs dependent on this one (#2562, and, eventually #2536), but if it doesn't take too long, I'm happy to wait.

jacobtylerwalls:
Oh that's great news that you ran the primer. No need to wait then.

temyurchenko:
> > Oh that's great news that you ran the primer. No need to wait then.
> 
> Let me give it just one more try, to be safe

Yep, all good

temyurchenko:
> Thanks for this! Does this close #1490? If so, can you add a changelog?

The wording of #1490 is confusing. It says that the problem is that the parent is added to the locals of the child, but the problem is that the child is added to the locals of the parent. Looking at #1489, indeed it is the latter that is the problem.

I initially intended for https://github.com/pylint-dev/astroid/pull/2536#issuecomment-2334879039 to close #1490, but this one can probably also be considered to be closing this. 

What should changelog look like? Can I just add the following to the `./ChangeLog` file?

```
* Control setting local nodes outside of the supposed local's constructor.

  Closes pylint-dev/astroid/issues/1490
```

Pierre-Sassoulas:
Yes, only this syntax is used for pylint issues, for issues from astroid it can be simplified to:

```
* Control setting local nodes outside of the supposed local's constructor.

  Closes #1490
```

temyurchenko:
> Yes, only this syntax is used for pylint issues, for issues from astroid it can be simplified to:
> 
> ```
> * Control setting local nodes outside of the supposed local's constructor.
> 
>   Closes #1490
> ```

Thanks, I'll note this for the future.

Pierre-Sassoulas (approved):
I'm drowning in astroid myself, so take this approve for what it's worth but I think this change is pretty good and going into the right direction design wise.

jacobtylerwalls (approved):
Wonderful, I can remove the workarounds that stemmed from #1490 in another PR, and many thanks for noticing that the wording was wrong 👍 .

PR #2589, opened 2024-09-30
Change order of search path to fix inconsistency between pylint and astroid.

akamat10:
## Type of Changes

## Description

While testing https://github.com/pylint-dev/pylint/issues/9967, I noticed that the qualified package names are incorrect when linted files are outside of the source-roots specified. This PR changes the priority of the search path which then correctly picks up the right qualified module name. It makes sense to give higher priority to the additional search paths over `sys.path`. This is in line with priority order of [expand_modules](https://github.com/pylint-dev/pylint/blob/c0ecd70e6ff9e92ee686671c7e0f152ab643f25e/pylint/lint/expand_modules.py#L92) function in pylint. Hopefully the maintainers agree that this inconsistency needs to be fixed.

Refs https://github.com/pylint-dev/pylint/issues/9955

### DIFF
--- a/astroid/util.py
+++ b/astroid/util.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import contextlib
 import sys
 import warnings
+from collections.abc import Iterator, Sequence
 from typing import TYPE_CHECKING, Any, Final, Literal
 
 from astroid.exceptions import InferenceError
--- a/astroid/util.py
+++ b/astroid/util.py
@@ -5,6 +5,8 @@
 
 from __future__ import annotations
 
+import contextlib
+import sys
 import warnings
 from typing import TYPE_CHECKING, Any, Final, Literal
 
@@ -157,3 +159,24 @@ def safe_infer(
         return None  # there is some kind of ambiguity
     except StopIteration:
         return value
+
+def _augment_sys_path(additional_paths: Sequence[str]) -> list[str]:
+    original = list(sys.path)
+    changes = []
+    seen = set()
+    for additional_path in additional_paths:
+        if additional_path not in seen:
+            changes.append(additional_path)
+            seen.add(additional_path)
+
+    sys.path[:] = changes + sys.path
+    return original
+
+@contextlib.contextmanager
+def augmented_sys_path(additional_paths: Sequence[str]) -> Iterator[None]:
+    """Augment 'sys.path' by adding entries from additional_paths."""
+    original = _augment_sys_path(additional_paths)
+    try:
+        yield
+    finally:
+        sys.path[:] = original
--- a/astroid/util.py
+++ b/astroid/util.py
@@ -160,6 +160,7 @@ def safe_infer(
     except StopIteration:
         return value
 
+
 def _augment_sys_path(additional_paths: Sequence[str]) -> list[str]:
     original = list(sys.path)
     changes = []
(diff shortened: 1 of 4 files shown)


akamat10:
Also, both `expand_module` in `pylint` and `modpath_from_file_with_callback` in `astroid` are taking sys.path for module discovery with `expand_module` calling `modpath_from_file_with_callback`. It may be makes sense to simplify such that it is only happening in one of these functions

Pierre-Sassoulas:
Yeah let's make pylint rely on astroid's fonction or at least comment on the similarity if we want to reduce coupling.

akamat10:
Great! Let me add a test with this PR before this is merged.

tests/test_modutils.py
```diff
@@ -175,6 +175,41 @@ def test_import_symlink_with_source_outside_of_path(self) -> None:
...
+                    modutils.modpath_from_file(module_file), [sub_dirname, mod_name]
+                )
+
+                # With sub directory as additional directory, return relative to
+                # sub directory
+                self.assertEqual(
+                    modutils.modpath_from_file(f"{sub_dir}/{mod_name}.py", [sub_dir]),
+                    [mod_name],
+                )
+            finally:
```
Pierre-Sassoulas, line 210: I think elsewhere we use a context manager for sys path modification in tests. Sorry I can't search as I'm on mobile.

akamat10: Will check. Don't remember seeing one in `astroid` repo. There is one in `pylint` repo. Perhaps you are remembering that one? If there isn't one, you want me to create one similar to that?

akamat10: Didn't find it. Will create one.

Pierre-Sassoulas: Probably this:

https://github.com/pylint-dev/pylint/blob/d9b06c6d5ba0fd9b74b5d8441accedc6bdbecfd8/pylint/lint/utils.py#L128-L135

Pierre-Sassoulas: Sorry I did not realize it was an astroid PR 😅

akamat10: Updated

Pierre-Sassoulas (approved):
👌

astroid/util.py
```diff
@@ -157,3 +160,26 @@ def safe_infer(
...
+        if additional_path not in seen:
+            changes.append(additional_path)
+            seen.add(additional_path)
+
+    sys.path[:] = changes + sys.path
+    return original
+
+
+@contextlib.contextmanager
+def augmented_sys_path(additional_paths: Sequence[str]) -> Iterator[None]:
```
DanielNoord, line 179: Could we do a follow up where we move this into the test directory. By putting it here we add it to our global API, whereas it is only used in tests and we don't want to have to keep supporting it.

akamat10: Good point. I saw that it is public for pylint and followed that option. However, what you are saying makes sense. Another option is to mark it private by pretending it with _. That would then be available for other modules in astroid. Open to either one.

DanielNoord: Based on the experience I had with deprecating stuff with astroid and pylint 2.x I'd say move it to tests. If we ever need it we can always move it :)

akamat10: Sounds good! Will change.

PR #2595, opened 2024-10-02
Remove unused numpy utility functions

correctmost:
## Type of Changes

## Description

Follow-up to c7ea1e941b4b88ec and https://github.com/pylint-dev/astroid/pull/2550.

I wasn't sure if more documentation or labeling was needed for changes that could impact backward compatibility.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,9 @@ What's New in astroid 4.0.0?
 ============================
 Release date: TBA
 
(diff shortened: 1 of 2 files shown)


DanielNoord (dismissed):
Since we are on `4.x` and I don't think this adds much I think we can just remove.

Let's wait for another maintainers opinion though.

Pierre-Sassoulas (approved):
Yeah, let's not bother with deprecation periods in astroid (unless this is a widely used (™️) API).

PR #2602, opened 2024-10-07
make constants have synthetic root as their parent

temyurchenko:
that fixes a bunch of tests in pylint.

We also make the synthetic root a singleton, as opposed to a module
   built in the astroid manager. That allows us to use it as a default
   value in Const constructors.

Note the changes to the test. Before, we in fact tested that constants
   don't have a parent. Is that reasonable? In fact, it seems like
   there are a variety of interpretations and we shouldn't commit to
   one by explicitly testing for a certain parent or lack thereof.

## Type of Changes

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -40,14 +40,7 @@ from astroid.exceptions import (
 from astroid.interpreter.dunder_lookup import lookup
 from astroid.interpreter.objectmodel import ClassModel, FunctionModel, ModuleModel
 from astroid.manager import AstroidManager
-from astroid.nodes import (
-    Arguments,
-    Const,
-    NodeNG,
-    _base_nodes,
-    const_factory,
-    node_classes,
-)
+from astroid.nodes import _base_nodes, node_classes
 from astroid.nodes.scoped_nodes.mixin import ComprehensionScope, LocalsDictNodeNG
 from astroid.nodes.scoped_nodes.utils import builtin_lookup
 from astroid.nodes.utils import Position
@@ -60,6 +53,7 @@ from astroid.typing import (
 
 if TYPE_CHECKING:
     from astroid import nodes, objects
+    from astroid.nodes import Arguments, Const, NodeNG
     from astroid.nodes._base_nodes import LookupMixIn
 
 
@@ -353,7 +347,7 @@ class Module(LocalsDictNodeNG):
         if name in self.special_attributes and not ignore_locals and not name_in_locals:
             result = [self.special_attributes.lookup(name)]
             if name == "__name__":
-                main_const = const_factory("__main__")
+                main_const = node_classes.const_factory("__main__")
                 main_const.parent = AstroidManager().builtins_module
                 result.append(main_const)
         elif not ignore_locals and name_in_locals:
@@ -605,6 +599,14 @@ class Module(LocalsDictNodeNG):
         self, context: InferenceContext | None = None, **kwargs: Any
     ) -> Generator[Module]:
         yield self
+
+
+class __SyntheticRoot(Module):
+    def __init__(self):
+        super().__init__("__astroid_synthetic", pure_python=False)
+
+
+SYNTHETIC_ROOT = __SyntheticRoot()
 
 
 class GeneratorExp(ComprehensionScope):
@@ -1572,7 +1574,7 @@ class FunctionDef(
             and len(self.args.args) == 1
             and self.args.vararg is not None
         ):
-            if isinstance(caller.args, Arguments):
+            if isinstance(caller.args, node_classes.Arguments):
                 assert caller.args.args is not None
                 metaclass = next(caller.args.args[0].infer(context), None)
             elif isinstance(caller.args, list):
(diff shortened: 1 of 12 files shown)


temyurchenko:
This has in fact uncovered several incorrect tests in pylint, a PR for which I'll make later

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1583,17 +1585,13 @@ def infer_call_result(
...
-                            b
-                            for b in arg.infer(
-                                context=context.clone() if context else context
-                            )
-                            if not (isinstance(b, Const) and b.value is None)
-                        )
-                        for arg in caller.args[1:]
-                    ]
+                    # Find the first non-None inferred base value
+                    get_base = lambda arg: next(  # pylint: disable=unnecessary-lambda-assignment # noqa
```
jacobtylerwalls, line 1589: Hi, sorry for the drive-by review. Just wondering: can we do this without the lambda & disable?

temyurchenko: Absolutely, but the formatter then turns it into a monstrosity with several single word lines. This one looks clearer. However, it is subjective and I will have no problems inlining the lambda.

jacobtylerwalls: What about without a lambda at all?

temyurchenko: > What about without a lambda at all?

Oh, yes, that's what I meant. I'm comparing the situation without a lambda = lambda inlined (weird formatting) vs with a lambda (the present).

jacobtylerwalls: Sorry, I guess I was wondering about extracting a function? That's the recommended prescription for this lint message.

temyurchenko: > Sorry, I guess I was wondering about extracting a function? That's the recommended prescription for this lint message.

Done!

temyurchenko: AFAICS, the underlying issue that necessitates that function is the `six.with_metaclass` issue (#5935). It's caused by the fact that `Enum` is defined two times in the `enum` module: first as `None`, second as the actual class.

So, the problem is that symbols can be defined several times. This workaround with skipping `None` values is unsatisfactory, because it one hand behaves too generally for handling just the `Enum` issue, but not generally enough to handle the wider problem of redefinition of symbols.

Perhaps, if we wanted to solve the specific Enum problem, we could do it with an `Enum` transform.
However, if we wanted to solve the larger problem, there is one heuristic that seems to make sense to me. When you import a symbol, only the last redefinition of that symbol is imported. Thus, we shouldn't be looking for the first «non-None», we should be looking for the _last_ definition (specifically, when we are inferring imports). 

These are just thoughts, but do tell me if you think there is some worth in it.

jacobtylerwalls: Absolutely. Elsewhere for functions we (as of recently) take the last definition unless it's a property and decorated with setter. On mobile so don't have a link. But you're right on.

temyurchenko: Got it. Based on our conversation, I've changed the the `infer_first_not_none` to `infer_last`, although it's better be done in the inference function for imports. That would be too much out of scope of this commit though.

tests/test_regrtest.py
```diff
@@ -248,7 +248,7 @@ def with_metaclass(meta, *bases):
             class metaclass(meta):
                 def __new__(cls, name, this_bases, d):
                     return meta(name, bases, d)
-        return type.__new__(metaclass, 'temporary_class', (), {})
+            return type.__new__(metaclass, 'temporary_class', (), {})
```
jacobtylerwalls, line 251: Hopefully the crash referred to in the bitbucket issue that has been lost to time didn't depend on this syntax error. 🙏

jacobtylerwalls (approved):
Thanks for this important cleanup and for the follow-through with the pylint PR to fix tests. By new year or so I imagine we'll issue an alpha of astroid-4.0 so we can use it on pylint main and merge those test fixes.

jacobtylerwalls (approved):

temyurchenko:
Give me a moment to squash the fixup commit

tests/test_regrtest.py
temyurchenko: From my understanding, the only part that mattered was the «six.with_metaclass» in the line below. The piece of code above could have been either simplified significantly or omitted altogether

jacobtylerwalls:
Sure, I'll disable auto-squash, but just so you know my plan was to squash all three commits and use the original message minus the question about the test. Would that have been okay?

temyurchenko:
> Sure, I'll disable auto-squash, but just so you know my plan was to squash all three commits and use the original message minus the question about the test. Would that have been okay?

Ah, absolutely fine.

tests/test_regrtest.py
jacobtylerwalls: Ugh. Thanks for looking into it.

temyurchenko:
> By new year or so I imagine we'll issue an alpha of astroid-4.0 so we can use it on pylint main and merge those test fixes.

Hey, can we somehow mark the pylint PR (https://github.com/pylint-dev/pylint/pull/10013) as being necessary to merge with astroid 4.0? Otherwise, there will be confusion about why the tests are failing and there is no obvious indication that they are fixed by that PR.

DanielNoord:
I have added labels that should help identify it as being necessary :)

temyurchenko:
> I have added labels that should help identify it as being necessary :)

Thank you!

PR #2615, opened 2024-10-17
Update setuptools to 75.2.0

cdce8p:
## Description
https://setuptools.pypa.io/en/stable/history.html#v75-2-0

Since setuptools `70.1` the `bdist_wheel` command is shipped with setuptools directly. It's no longer necessary to specify `wheel` as a build system requirement.


Pierre-Sassoulas (approved):
Great !

PR #2617, opened 2024-10-19
nonlocal in locals

### LINKED ISSUE #2616: nonlocal variables are included in `locals` while global variables not
Hello! I'm trying `locals` to get the variables (or names) in a function. However I've found that global variables are excluded, while nonlocal variables are here:

```python
import astroid.nodes

module = astroid.parse(
"""
x1 = 1                      # Line 2
def f1():                   # Line 3
    x2 = 2                  # Line 4
    def f2():               # Line 5
        global x1           # Line 6
        nonlocal x2         # Line 7
        x1 = 1              # Line 8
        x2 = 2              # Line 9
""")

f2 = module.locals["f1"][0].locals["f2"][0]
print(f2.locals)

# Output:
# {'x2': [<AssignName.x2 l.9 at 0x1ddffab04d0>]}
```

That's really weird to me. Is it or bug or intentionally designed? If this behavior is expected, what is the recommended solution to exclude those nonlocal variables? (and include global ones?)

I have found #250 , which however was moved from bitbucket in 2015, so I'm not sure whether it is outdated.

Thanks in advance.

yueyinqiu:
## Type of Changes

## Description

Closes #2616

I did this by adding a `_nonlocal_names` property in `TreeRebuilder` to store those nonlocal names, like what `_global_names` does. However it stores `FunctionDef`s where the variables are created, instead of `Nonlocal` nodes. (The saved `Global` nodes in `_global_names` is not used for as well, but I didn't touch it, in case we may need that one day.)

The items of `_nonlocal_names` are added in `visit_nonlocal`, which go through the tree and find the nearest `FunctionDef` whose `locals` contains that name.

And it is used in `_save_assignment`. If `node.name` is in `self._nonlocal_names[-1]`, then it will be put into `self._nonlocal_names[-1][node.name]` (the found `FunctionDef`) rather than into `node.parent`.

I have also added a unit test `test_locals_with_global_and_nonlocal` in `test_builder.py`, but I'm not really sure if I put it in the correct place.

There might be some other problems since it's my first PR for this project. Really need someone to check it before merging.


astroid/rebuilder.py
```diff
@@ -1383,14 +1393,32 @@ def visit_name(
...
             end_lineno=node.end_lineno,
             end_col_offset=node.end_col_offset,
             parent=parent,
         )
+        names = set(newnode.names)
+        # Go through the tree and find where those names are created
+        scope = newnode
+        while len(names) != 0:
+            scope = scope.parent
+            if not scope:
```
yueyinqiu, line 1409: I just `break` if it's not inside a nested function, or there are no variables with that name. I'm not sure whether it's an acceptable behavior.

But I did this because there is a similar case:

```python
x = 1
def f():
    x = 2
    global x
```

It should be `SyntaxError` in Python, but `astroid` could parse it without raising an exception.

So I think it's ok to just `break` here.

astroid/rebuilder.py
```diff
@@ -1383,14 +1393,32 @@ def visit_name(
...
+                # It's not inside a nested function or there are no variables with that name.
+                # Just ignore it as visit_global does when global is used in module scope.
+                break
```
yueyinqiu, line 1412: I just `break` if it's not inside a nested function, or there are no variables with that name. I'm not sure whether it's an acceptable behavior.

But I did this because of this similar case:

```python
x = 1
def f():
    x = 2
    global x
```

It should be `SyntaxError` in Python, but currently `astroid` could parse it without raising an exception.

So I think it's ok to do anything if there is a syntax error, so I just `break` here.

astroid/rebuilder.py
```diff
@@ -451,6 +456,9 @@ def _save_assignment(self, node: nodes.AssignName | nodes.DelName) -> None:
         """Save assignment situation since node.parent is not available yet."""
```
yueyinqiu, line 456: By the way, I'm a bit confused by this comment. What does it mean by `node.parent is not available`?

yueyinqiu:
Hmmm... Why pylint consider `scope` as `Nonlocal` since it's inside `if isinstance(scope, nodes.FunctionDef)`...

astroid/rebuilder.py
```diff
@@ -61,6 +61,11 @@ def __init__(
         self._manager = manager
         self._data = data.split("\n") if data else None
         self._global_names: list[dict[str, list[nodes.Global]]] = []
+        # In _nonlocal_names,
```
DanielNoord, line 64: This comment is formatted a bit strange. Does it fit on two lines?

astroid/rebuilder.py
DanielNoord: I don't think so. This should be okay.

astroid/rebuilder.py
DanielNoord: Yes! We shouldn't crash on invalid code so this is the right approach

DanielNoord (commented):
I also don't understand the `pylint` issue. Pinging @jacobtylerwalls to see if this sounds familiar, they have a very good grasp of recurring `pylint` issues

jacobtylerwalls (changes requested):
Thanks for working up the PR, but if you don't mind I'd like to pause to consider this in a little more detail first. My first impression is to wonder if we need to make a change here at all, see [comment](https://github.com/pylint-dev/astroid/issues/2616#issuecomment-2425213901) on issue.

astroid/rebuilder.py
yueyinqiu: I've modified it to

```python
# In _nonlocal_names saves the FunctionDef where the variable is created,
# rather than Nonlocal, since we don't really need the Nonlocal statement.
```

jacobtylerwalls:
Thanks for engaging in discussion. Let us know if the nodes_of_class() solution doesn't work well for you.

PR #2619, opened 2024-10-21
remove no-op if

temyurchenko:
Note the `if` condition above, the `values` are asserted to be empty,
  so adding them to the result doesn't do anything.

Also remove small inefficiencies

## Type of Changes

## Description

I just noticed a few changes left over after my work on previous PRs and they seemed reasonable enough.

### DIFF
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2354,26 +2354,24 @@ class ClassDef(
 
         if name in self.special_attributes and class_context and not values:
             result = [self.special_attributes.lookup(name)]
-            if name == "__bases__":
-                # Need special treatment, since they are mutable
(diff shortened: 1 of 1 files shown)


DanielNoord (commented):
I don't really understand the change with `__bases__` , could you explain?

The rest makes sense. Not doing the `copy()` but having `result` is much better

temyurchenko:
> I don't really understand the change with `__bases__` , could you explain?

Sure. Here's the code in question:

        if bla-bla and **not values**:
            ...
            if name == "__bases__":
                result += **values**

Note that inside the body of the first if, `values` is empty. Thus, adding it to anything is a no-op. Thus, we can just remove that statement.

DanielNoord (approved):
Awesome! Thanks for that explanation, probably needed some more coffee when I reviewed this :smile:

PR #2648, opened 2024-12-08
Add missing docstring on enum name property

AleksMat:
## Type of Changes

## Description

I maintain a custom lint rule implementation that checks that every class property has a docstring. It works well, except for  enum classes. The problem is that a `ClassDef` representation of an enum class always contains this mysterious  property `name`, even though `name` is only defined in the base class, and the property doesn't have a docstring. Hence, my lint rule keeps reporting a missing docstring for every enum class.

The reason for the issue is that `astroid/brain/brain_namedtuple_enum.py` artificially adds `name` property to a node representation for every enum class. I'm not sure if such implementation is the best, but at least the added `name` property should have a docstring.

This PR therefore adds a docstring to the `name` property, which is the same docstring used in the Python standard library ([here](https://github.com/python/cpython/blob/v3.13.1/Lib/enum.py#L1330)). It also adds a unit test that checks the `name` property really has the docstring.

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -516,11 +516,12 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
         # know that it should be a string, so infer that as a guess.
         if "name" not in target_names:
             code = dedent(
-                """
-            @property
-            def name(self):
-                return ''
-            """
+                '''
+                @property
+                def name(self):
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
I think it's ok to do this (even though it should probably be a special case in the dowstream library in the general case). Waiting for another maintainer opinion before merging.

Pierre-Sassoulas (approved):

DanielNoord (approved):

PR #2656, opened 2024-12-12
Fix `collections.abc` import error

zenlyj:
## Type of Changes

## Description

I was looking into the recent `collections.abc` import error with Python 3.13.1 and wanted to share my findings, regardless of whether this gets merged or not. Hopefully this helps in some way.

Did some testing locally with Linux environment and noticed the following:

- In 3.12.8, `collections.abc` is not a frozen module and origin is at `/usr/lib/python3.12/collections/abc.py`
- In 3.13.1, `collections.abc` is a frozen module

image

Now in the `spec._find_spec`, it is searching in the following `submodule_path` - `/usr/lib/python3.13/collections`, and giving error when not found. However, we expect it to be unfound as it is a frozen module.

There is already existing logic for handling frozen modules in the `ImportlibFinder` class, so perhaps we can utilize that?

Closes #XXXX

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -127,6 +127,27 @@ class ImportlibFinder(Finder):
     )
 
     @staticmethod
+    def find_frozen_module(module: str) -> ModuleSpec | None:
+        try:
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", category=UserWarning)
+                spec = importlib.util.find_spec(module)
+            if (
+                spec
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
Thank you for the investigation. Jacob tried to add an astroid brain for collections.abc in astroid 3.3.6 but it's not used during import for reasons I don't understand yet. (I'm on mobile si I can't link the issue right now but there's not much more information than what you also saw)

DanielNoord:
Superseded by https://github.com/pylint-dev/astroid/pull/2657. Thanks for taking a stab at this, you correctly identified this as being an issue with our checks for submodules.

PR #2657, opened 2024-12-18
Fix `collections.abc` imports on Python 3.13.0 and 3.13.1

DanielNoord:
Supersedes https://github.com/pylint-dev/astroid/pull/2656
References https://github.com/pylint-dev/pylint/issues/10112
Reverts and improves on https://github.com/pylint-dev/astroid/pull/2647
Is likely related to https://github.com/python/cpython/pull/125415

The issue was partially identified in https://github.com/pylint-dev/astroid/pull/2656 but not fixed correctly there I think.

The issue `3.13.1` introduces is that submodules of stdlib modules can now also be frozen and even change their name. This is somewhat annoying, but fixable.
However, while debugging I noticed that `find_spec` is actually pretty dangerous to call. It imports the full module and can therefore raise `warnings` etc.

In this PR I take a different approach. Only if we know we're dealing with a stdlib module or stdlib submodule we try and retrieve the `frozen` module. Otherwise, don't import the module as that is just costly for no additional gain.
The issue with that approach is that it only works on Python 3.10+. To work around this I just copied to old code to an `else` branch that is only executed on `3.9` and below. This is fine as these versions don't deal with submodules being frozen.

As can be seen from the test runs the tests does serve as a regression. It fails on `main`, but passes with the fixes in this PR.
The PyPy failure is unrelated. The checks job won't pass until `pylint` itself is fixed.

There was also already a Changelog entry for `3.3.7`, but I just reused that. I think a rebase went wrong there.

I tested this locally against `3.13.0` and `3.13.1`. I did see some failures on `pylint` but my local setup is very broken currently so I can't really determine whether it is related to this PR. Anyhow, I think this fix by itself is worth merging as it clearly fixes a test that should pass. We can worry about `pylint` later.

Lastly, @Pierre-Sassoulas @jacobtylerwalls I know I won't have time for a release of `astroid` and `pylint` in the coming days/weeks. If you approve the PR would you mind doing the release? I can answer questions, but my time off ends today so I need to get back to $work.

### DIFF
--- a/astroid/interpreter/_import/spec.py
+++ b/astroid/interpreter/_import/spec.py
@@ -134,32 +134,62 @@ class ImportlibFinder(Finder):
         processed: tuple[str, ...],
         submodule_path: tuple[str, ...] | None,
     ) -> ModuleSpec | None:
-        if submodule_path is not None:
-            search_paths = list(submodule_path)
-        elif modname in sys.builtin_module_names:
+        # Although we should be able to use `find_spec` this doesn't work on PyPy for builtins.
+        # Therefore, we use the `builtin_module_nams` heuristic for these.
+        if submodule_path is None and modname in sys.builtin_module_names:
             return ModuleSpec(
                 name=modname,
                 location=None,
                 type=ModuleType.C_BUILTIN,
             )
-        else:
-            try:
-                with warnings.catch_warnings():
-                    warnings.filterwarnings("ignore", category=UserWarning)
-                    spec = importlib.util.find_spec(modname)
+
+        # sys.stdlib_module_names was added in Python 3.10
+        if PY310_PLUS:
+            # If the module is a stdlib module, check whether this is a frozen module. Note that
+            # `find_spec` actually imports the module, so we want to make sure we only run this code
+            # for stuff this is can be expected to be frozen. For now this is only stdlib.
+            if modname in sys.stdlib_module_names or (
+                processed and processed[0] in sys.stdlib_module_names
+            ):
+                spec = importlib.util.find_spec(".".join((*processed, modname)))
                 if (
                     spec
                     and spec.loader  # type: ignore[comparison-overlap] # noqa: E501
                     is importlib.machinery.FrozenImporter
                 ):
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas:
Nice 😌, I was running in circle with this bug !  (I don't have time for a full review atm, but should we temporarily test in CI for python 3.13.0 and 3.13.1 ?) I can do the releases, sure.

DanielNoord:
> Nice 😌, I was running in circle with this bug ! (I don't have time for a full review atm, but should we temporarily test in CI for python 3.13.0 and 3.13.1 ?) I can do the releases, sure.

I'm 100% sure this fixes the issue for `astroid` on both `.0` and `.1`. I tested this extensively locally.

As for `pylint` I'm not sure, but we could.

astroid/interpreter/_import/spec.py
```diff
@@ -134,32 +134,62 @@ def find_module(
...
-            try:
-                with warnings.catch_warnings():
-                    warnings.filterwarnings("ignore", category=UserWarning)
-                    spec = importlib.util.find_spec(modname)
+
+        # sys.stdlib_module_names was added in Python 3.10
+        if PY310_PLUS:
+            # If the module is a stdlib module, check whether this is a frozen module. Note that
+            # `find_spec` actually imports the module, so we want to make sure we only run this code
+            # for stuff this is can be expected to be frozen. For now this is only stdlib.
```
Pierre-Sassoulas, line 150:
```suggestion
            # for stuff that can be expected to be frozen. For now this is only stdlib.
```

Pierre-Sassoulas (dismissed):
Great !

DanielNoord:
@Pierre-Sassoulas could you reapprove and merge? :)

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Working on the manual cherry-pick atm.

Pierre-Sassoulas:
The cherry-pick is hard, maybe we can release 3.4.0 directly ?

PR #2661, opened 2024-12-21
Skip flaky recursion test on PyPy

jacobtylerwalls:
Follow-up to 03cd54c.

No one has the time or inclination to find a stable recursion limit that is low enough to reliably trigger the UserWarning without actually causing core dumps or other problems, esp. when running with coverage.


Pierre-Sassoulas (approved):
Strongly agree with this decision

PR #2665, opened 2024-12-23
Another attempt at fixing the `collections.abc` issue for Python 3.13

DanielNoord:
Well, https://github.com/pylint-dev/astroid/pull/2658 did not work.

Let's try to fix https://github.com/pylint-dev/pylint/issues/10112 again with a new fix.
It is basically the same fix as last time, but this time with some more `try ... except` and changing the order around. First do the normal logic, then try the frozen module logic.

Note that this targets the `3.3.x` branch as I couldn't get the CI of `pylint` to pass from the `4.x` branch. Since this needs to be back ported anyway I think that is fine.


astroid/interpreter/_import/spec.py
```diff
@@ -150,7 +150,15 @@ def find_module(
             if modname in sys.stdlib_module_names or (
                 processed and processed[0] in sys.stdlib_module_names
             ):
-                spec = importlib.util.find_spec(".".join((*processed, modname)))
+                try:
+                    with warnings.catch_warnings():
+                        warnings.filterwarnings("ignore", category=Warning)
+                        spec = importlib.util.find_spec(".".join((*processed, modname)))
+                except ValueError:
```
jacobtylerwalls, line 157: The underlying module is still imported and can raise any exception, e.g. `django.core.exceptions.ImproperlyConfigured`:

```py
  File "/Users/jwalls/django/django/contrib/gis/gdal/libgdal.py", line 62, in <module>
    raise ImproperlyConfigured(
    ...<3 lines>...
    )
django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal", "GDAL", "gdal3.9.0", "gdal3.8.0", "gdal3.7.0", "gdal3.6.0", "gdal3.5.0", "gdal3.4.0", "gdal3.3.0", "gdal3.2.0", "gdal3.1.0"). Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings.
```

jacobtylerwalls (changes requested):
In 3.3.6, the django submodule wasn't imported, but even with this PR it's still being imported.

(Also major 🙏 🙏 for devoting debugging time to this, much appreciated!)

DanielNoord:
But on `3.3.5` we did import the module right? Considering `3.3.6` also didn't handle `collections` properly I don't think it is the right benchmark

DanielNoord:
Never mind, of course on `3.3.5` we did not import them. That was the whole issue 😅 
_Not trying to find frozen modules for submodules._

I have pushed another change, let's see if this works. First default to the normal behaviour, after it, try and see if we can frozen modules.

jacobtylerwalls:
> Note that this targets the 3.3.x branch as I couldn't get the CI of pylint to pass from the 4.x branch. Since this needs to be back ported anyway I think that is fine.

Thanks. Once we have a clean pylint run, let's rebase on main and follow the usual process. Around new year I was planning to cut an alpha of astroid and bring pylint up to date, sorry it hasn't happened yet. We already have a contributor who's ready to merge in the test fixes.

DanielNoord:
> > Note that this targets the 3.3.x branch as I couldn't get the CI of pylint to pass from the 4.x branch. Since this needs to be back ported anyway I think that is fine.
> 
> Thanks. Once we have a clean pylint run, let's rebase on main and follow the usual process. Around new year I was planning to cut an alpha of astroid and bring pylint up to date, sorry it hasn't happened yet. We already have a contributor who's ready to merge in the test fixes.

The clean run is going now :)

I'm not sure what the usual process would be here? Can't we merge this into the maintenance branch and then resolve any issues we get with merging it into `4.x` later?
That way we focus on the issue at hand (we don't have a compatible `astroid` version) and can worry about future versions later.

astroid/interpreter/_import/spec.py
```diff
@@ -150,7 +175,16 @@ def find_module(
             if modname in sys.stdlib_module_names or (
                 processed and processed[0] in sys.stdlib_module_names
             ):
```
jacobtylerwalls, line 177: I guess I'm still wondering if this can/should be:
```suggestion
            if processed and processed[0] in sys.stdlib_module_names:
```

Given that:
```sh
mkdir foo
pylint foo.io
```
Will unexpectedly evaluate True here.

jacobtylerwalls:
> I'm not sure what the usual process would be here? Can't we merge this into the maintenance branch and then resolve any issues we get with merging it into 4.x later?

My thought was:
Merge this into main (is there a blocker to doing this?)
Backport to 3.3.x and release 3.3.8.
Update the https://github.com/pylint-dev/pylint/pull/10125 to 3.3.8

ChangeLog
```diff
@@ -13,12 +13,18 @@ What's New in astroid 3.3.8?
 ============================
 Release date: TBA
 
+* Fix inability to import `collections.abc` in python 3.13.1. The reported fixes in astroid 3.3.6
```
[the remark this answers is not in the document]
DanielNoord: Yeah.. Kind of? But considering we yanked the release I don't think we should really call the previous fix a fix? I don't know.. Happy to change if you have suggestions.

astroid/interpreter/_import/spec.py
DanielNoord: No, because then `collections` will never be considered frozen. We need both `collections` and `collections.abc` to pass this check.

astroid/interpreter/_import/spec.py
```diff
@@ -150,7 +175,16 @@ def find_module(
...
-                spec = importlib.util.find_spec(".".join((*processed, modname)))
+                try:
+                    with warnings.catch_warnings():
+                        warnings.filterwarnings("ignore", category=Warning)
```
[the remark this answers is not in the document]
DanielNoord: I ran into issues with `DeprecationWarnings` for `lib2to3` in the `stdlib` primer for `pylint` which is why I had to bump this to `Warning`.

astroid/interpreter/_import/spec.py
```diff
@@ -150,7 +175,16 @@ def find_module(
...
+                        spec = importlib.util.find_spec(".".join((*processed, modname)))
+                except ValueError:
+                    # `find_spec` raises a ValueError for modules that are missing their `__spec__`
+                    # attributes. We don't really understand why, but this is likely caused by
+                    # us re-implementing the import behaviour but not correctly.
```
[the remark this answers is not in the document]
DanielNoord: I really have no clue. The only reason I know that we need to catch this is because it was part of the stack of the original issue reporter. I still don't really know how to reproduce this case.

DanielNoord:
> > I'm not sure what the usual process would be here? Can't we merge this into the maintenance branch and then resolve any issues we get with merging it into 4.x later?
> 
> My thought was: Merge this into main (is there a blocker to doing this?) Backport to 3.3.x and release 3.3.8. Update the [pylint-dev/pylint#10125](https://github.com/pylint-dev/pylint/pull/10125) to 3.3.8

It now targets `main` but that immediately causes merge conflicts...

Considering I won't have any time during Christmas I'd personally prefer to target this on the maintenance branch and just get the fix out.

ChangeLog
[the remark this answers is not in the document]
jacobtylerwalls: Good point. Resolving!

astroid/interpreter/_import/spec.py
jacobtylerwalls: Got it -- then let's update the comment, since it sounds like we know whether something is a stdlib module if this test passes.

DanielNoord: We don't 😅 

The issue some reporter/primer encountered was that any module called `io` would also pass this check, which makes sense. It is just a `str` lookup. That's why I have changed the logic around now. If we can find the file we use that, if we can't (such as for frozen modules) we try that logic before giving up.

astroid/interpreter/_import/spec.py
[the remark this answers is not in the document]
jacobtylerwalls: That's fine, but let's just remove the comment. The function is documented to raise ValueError, so let's just catch it without a comment.

astroid/interpreter/_import/spec.py
```diff
@@ -142,6 +142,31 @@ def find_module(
...
+                    )
+            for suffix, type_ in ImportlibFinder._SUFFIXES:
+                file_name = modname + suffix
+                file_path = os.path.join(entry, file_name)
+                if os.path.isfile(file_path):
+                    return ModuleSpec(name=modname, location=file_path, type=type_)
+
         # sys.stdlib_module_names was added in Python 3.10
         if PY310_PLUS:
             # If the module is a stdlib module, check whether this is a frozen module. Note that
```
jacobtylerwalls, line 172:
```suggestion
            # If the module name might be a stdlib module, check whether this is a frozen module. Note that
```

astroid/interpreter/_import/spec.py
jacobtylerwalls: Right, we don't -- but the comment makes it sound like we do. I added another comment with a suggestion.

astroid/interpreter/_import/spec.py
jacobtylerwalls: I can't comment on the line directly below, but I'd change `imports the module` -> `imports parent modules`

astroid/interpreter/_import/spec.py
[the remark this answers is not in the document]
jacobtylerwalls: Alright. It seems like a little bit of a regression that we're importing stdlib modules in more cases now, but I guess we have no choice really.

jacobtylerwalls (dismissed):
Thanks for giving me a little time to grok this, and thanks for such a diligent follow-up. Happy new year!

jacobtylerwalls:
(will reapprove after merge conflicts spruced up)

astroid/interpreter/_import/spec.py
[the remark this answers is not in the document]
DanielNoord: Yeah.. I really don't know of any other way to fix this.

astroid/interpreter/_import/spec.py
```diff
@@ -143,15 +143,45 @@ def find_module(
...
+                 if cached_os_path_isfile(file_path):
+                     return ModuleSpec(
+                         name=modname,
+                         location=package_directory,
+                         type=ModuleType.PKG_DIRECTORY,
+                     )
+             for suffix, type_ in ImportlibFinder._SUFFIXES:
+                 file_name = modname + suffix
+                 file_path = os.path.join(entry, file_name)
+                 if cached_os_path_isfile(file_path):
```
DanielNoord, line 166: @jacobtylerwalls Bit afraid of this change... The `3.3.x` branch doesn't have it so I'm not sure if this introduce more issues.

But if you approve I guess we can just try to make it work.

jacobtylerwalls (approved):
Good to know, that will help me resolve the merge conflict on the backport.

Pierre-Sassoulas (approved):
Thanks a lot for handling this Daniel, amazing !

PR #2687, opened 2025-02-16
Fix RuntimeError caused by analyzing live objects with `__getattribute__` or descriptors

### LINKED ISSUE #2686: `RuntimeError` when analyzing `sys.modules` live object
This issue details my findings from investigating https://github.com/pylint-dev/pylint/issues/8589 (see that issue for more information and steps to reproduce) after my PR couldn't pass checks.
I found the (quite arcane) cause and some different potential fixes.

### Bug
Under rare circumstances, pylint crashes with `RuntimeError: dictionary changed size during iteration`.

### Cause
For this bug, during pylint checking, one of the modules in `sys.modules` is an [`importlib.util.LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) object. These 'lazy modules' only execute their contents the first time an attribute is accessed on the module.
The lazy loader module is put in there by any sort of package that needs to be loaded during pylint execution: the package must be installed and importable. One reason why pylint would need to load the package is being included in pylint's [`extension-pkg-whitelist`](https://pylint.pycqa.org/en/stable/user_guide/configuration/all-options.html#extension-pkg-whitelist) (for `pygame-ce`'s case).

During checking (of any arbitrary code that needs the package to be loaded), pylint analyzes the `sys` module live (live because there is no python source to analyze statically), and builds child nodes for its members, including for `sys.modules` dict at [`astroid.raw_building.InspectBuilder.object_build()`](https://github.com/pylint-dev/astroid/blob/7318c4368377b02821e67a3af7f3d0b0525dfac7/astroid/raw_building.py#L508).
Since `sys.modules` is a dict, pylint also tries to create nodes for the keys (strings) and values (module objects) at [`astroid.nodes.node_classes._create_dict_items()`](https://github.com/pylint-dev/astroid/blob/7318c4368377b02821e67a3af7f3d0b0525dfac7/astroid/nodes/node_classes.py#L5531), by iterating over the dict's `.items()` and calling `const_factory()` on each key and value. The actual live `sys.modules` dict is iterated over, not a copy.
Eventually the iteration reaches the `LazyLoader` module object. Inside [`astroid.nodes.node_classes.const_factory()`](https://github.com/pylint-dev/astroid/blob/7318c4368377b02821e67a3af7f3d0b0525dfac7/astroid/nodes/node_classes.py#L5537), the `.__class__` attribute of the lazy module is accessed, either implicitly by the `isinstance()` assertion, or explicitly on the next line.

Since an attribute was accessed on the lazy module object for the first time, it's contents get executed. If it happens to import modules and packages during it's loading, new modules are put into `sys.modules` dict by the python import system, changing its size.
After the lazy module is fully loaded, code execution eventually gets back to where `sys.modules`'s `.items()` are being iterated. Upon the next iteration, python realizes the dict size has changed and raises `RuntimeError` at that point in the code.

#### Comments
- The packages where the bug were experienced: `metaflow`, `pyjanitor`, `pygame-ce` (lazy module branch), `django`. From what I see, these all utilize lazy loading and `importlib`.
- The error was raised at the loop instead of when a lazy module was being loaded, making the traceback less useful.
- The condition where a package needed to be loaded during pylint execution made the crash difficult to reproduce. It explains why it often happens in CI checks of the package's repo.

### Possible Fixes
1. Shallow copy live mutable collections before iterating. Inefficient but simple and sound fix.
2. Special-case `sys.modules` dict for live analysis. Technically a crash is still possible though, only addresses the common case.
3. Avoid all attribute accesses of `value` inside `const_factory()`. I checked and it is easy: replace two `value.__class__` with `type(value)` (bypassing attribute access), and use `issubclass()` instead of `isinstance()` (because `isinstance()` uses `__class__`) in one assertion, all in `const_factory()` code. Keeps original efficiency, but has side effects: harder to maintain, `__class__` is no longer be able to be proxied (possibly a good thing?), and `__getattribute__` is no longer invoked by pylint (probably good but is another change in behavior).

`astroid 3.3.8`

aatle:
## Type of Changes

## Description

During checking, lazy module objects may get loaded during analysis of `sys.modules`, causing a crash. (In general, when an object executes code from an attribute access, making its containing collection change size.)
Avoid some attribute accesses in `const_factory()` to avoid executing `__getattribute__` code.

Closes #2686
Closes pylint-dev/pylint#8589

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -5514,6 +5514,7 @@ def _create_basic_elements(
     """Create a list of nodes to function as the elements of a new node."""
     elements: list[NodeNG] = []
     for element in value:
+        # NOTE: avoid accessing any attributes of element in the loop.
         element_node = const_factory(element)
         element_node.parent = node
         elements.append(element_node)
@@ -5526,6 +5527,7 @@ def _create_dict_items(
     """Create a list of node pairs to function as the items of a new dict node."""
     elements: list[tuple[SuccessfulInferenceResult, SuccessfulInferenceResult]] = []
     for key, value in values.items():
+        # NOTE: avoid accessing any attributes of both key and value in the loop.
         key_node = const_factory(key)
         key_node.parent = node
         value_node = const_factory(value)
@@ -5536,18 +5538,23 @@ def _create_dict_items(
 
 def const_factory(value: Any) -> ConstFactoryResult:
     """Return an astroid node for a python value."""
-    assert not isinstance(value, NodeNG)
+    # NOTE: avoid accessing any attributes of value until it is known that value
+    # is of a const type, to avoid possibly triggering code for a live object.
+    # Accesses include value.__class__ and isinstance(value, ...), but not type(value).
+    # See: https://github.com/pylint-dev/astroid/issues/2686
+    value_type = type(value)
+    assert not issubclass(value_type, NodeNG)
 
     # This only handles instances of the CONST types. Any
     # subclasses get inferred as EmptyNode.
     # TODO: See if we should revisit these with the normal builder.
-    if value.__class__ not in CONST_CLS:
+    if value_type not in CONST_CLS:
         node = EmptyNode()
         node.object = value
         return node
 
     instance: List | Set | Tuple | Dict
-    initializer_cls = CONST_CLS[value.__class__]
+    initializer_cls = CONST_CLS[value_type]
     if issubclass(initializer_cls, (List, Set, Tuple)):
         instance = initializer_cls(
             lineno=None,
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,6 +42,10 @@ Release date: TBA
 
   Closes pylint-dev/pylint#10112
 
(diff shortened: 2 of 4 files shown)


astroid/nodes/node_classes.py
```diff
@@ -5514,6 +5514,7 @@ def _create_basic_elements(
     """Create a list of nodes to function as the elements of a new node."""
     elements: list[NodeNG] = []
     for element in value:
+        # NOTE: avoid accessing any attributes of element in the loop.
```
DanielNoord, line 5517: Let's remove this comment and the one below. If the implementation of `const_factory` ever changes this is outdated. Imo, comments should be at the offending code (which you already did :smile:)

DanielNoord (commented):
Is there a chance we can add a test for this? Or is that too difficult?

DanielNoord (approved):
@jacobtylerwalls Anything to add?

jacobtylerwalls (approved):
Thanks!

PR #2699, opened 2025-03-09
Merge maintenance in main following 3.3.9 release

Pierre-Sassoulas:


.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
       - id: end-of-file-fixer
         exclude: tests/testdata
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.9.9"
+    rev: "v0.9.10"
```
cdce8p, line 13: Wondering how that got in there 👀
Doesn't really matter in the end, pre-commit seems to be fine so we might as well do the update here.

Pierre-Sassoulas: It got in here because I'm sneaky :p

PR #2702, opened 2025-03-09
Avoid importing submodules sharing names with standard library modules

### LINKED ISSUE #2684: Astroid >=3.3.7 imports and executes modules that contain any similarly named identifier as some stdlib module
The frozen module finding logic in astroid >= 3.3.7 (https://github.com/pylint-dev/astroid/blob/main/astroid/interpreter/_import/spec.py#L177) seems to interfere with non-module names that match with names in `sys.stdlib_module_names`. 

The following example is done with pylint (using pylint 3.3.4, astroid 3.3.8 and Python 3.12.8), but I'm still posting it here because it seems to be astroid-related issue. The most simple reproduction I was able to figure out is with a module that is imported

#### sayhi.py
```
print("hello")

def trace():
    pass
```

#### main.py:
```
from sayhi import trace
```

output from `pylint main.py` is

```
hello
************* Module main
main.py 1: Unused trace imported from sayhi (unused-import)
```

The `trace` here is also name of a module in stdlib, but that should be valid Python code anyway.

Expected outcome is that "hello" is not printed, as `sayhi.py` shouldn't be imported or executed by astroid (or pylint). That is also the case with astroid 3.3.6.

jacobtylerwalls:
## Type of Changes

## Description
Closes #2684
Avoid importing `user_module.std_lib_name` if we can help it.


tests/test_modutils.py
```diff
@@ -628,3 +628,18 @@ def test_find_setuptools_pep660_editable_install():
     with unittest.mock.patch.object(sys, "meta_path", new=[_EditableFinder]):
         assert spec.find_spec(["example"])
         assert spec.find_spec(["example", "subpackage"])
+
+
+def test_no_import_done_for_submodule_sharing_std_lib_name() -> None:
+    sys.path.insert(0, resources.find("data"))
+    try:
+        with pytest.raises(ImportError):
+            spec._find_spec_with_path(
```
jacobtylerwalls, line 637: I would have preferred to find a public interface to test, but after almost an hour of fiddling I threw in the towel.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #2715, opened 2025-03-25
[setuptools] Force the upgrade of setuptools > 77

Pierre-Sassoulas:
## Type of Changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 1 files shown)


cdce8p (approved):

cdce8p:
It seems we didn't backport two previous PRs this one relied on:
- #2681
- #2707

Not sure if it's worth it at this point to do the backport manually. The bug which was fixed here won't happen on the backport branch since that hasn't adopted PEP 639 yet. What do you think @Pierre-Sassoulas ?

PR #2722, opened 2025-04-12
Pick correct file if two files with the same name but with different extensions exist

Chasarr:
## Type of Changes

|   | :hammer: Refactoring   |
|   | :scroll: Docs          |

## Description
To replicate the error that this PR fixes, you need to create two files:
```bash
test.py    # Contains Pylint errors
test.py2   # Does not contain Pylint errors
```
and then I run `pylint test.py2` it reports Pylint errors from test.py. (the wrong file)

I believe this was due to a typo/logical error. Before this PR, I don't think the prefer_stubs argument even did anything.

Closes https://github.com/pylint-dev/pylint/issues/3631

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,10 @@ Release date: TBA
 
   Closes #2684
 
+* Fix bug where ``pylint code.custom_extension`` would analyze ``code.py`` or ``code.pyi`` instead if they existed.
+
+	Closes pylint-dev/pylint#3631
+
 
 What's New in astroid 3.3.9?
 ============================
--- a/tests/test_modutils.py
+++ b/tests/test_modutils.py
@@ -48,7 +48,9 @@ class ModuleFileTest(unittest.TestCase):
                 del sys.path_importer_cache[k]
 
     def test_find_zipped_module(self) -> None:
-        found_spec = spec.find_spec([self.package], [resources.find("data/MyPyPa-0.1.0-py2.5.zip")])
+        found_spec = spec.find_spec(
+            [self.package], [resources.find("data/MyPyPa-0.1.0-py2.5.zip")]
+        )
         self.assertEqual(found_spec.type, spec.ModuleType.PY_ZIPMODULE)
         self.assertEqual(
             found_spec.location.split(os.sep)[-3:],
@@ -56,7 +58,9 @@ class ModuleFileTest(unittest.TestCase):
         )
 
     def test_find_egg_module(self) -> None:
-        found_spec = spec.find_spec([self.package], [resources.find("data/MyPyPa-0.1.0-py2.5.egg")])
+        found_spec = spec.find_spec(
+            [self.package], [resources.find("data/MyPyPa-0.1.0-py2.5.egg")]
+        )
         self.assertEqual(found_spec.type, spec.ModuleType.PY_ZIPMODULE)
         self.assertEqual(
             found_spec.location.split(os.sep)[-3:],
@@ -74,7 +78,9 @@ class LoadModuleFromNameTest(unittest.TestCase):
         self.assertEqual(modutils.load_module_from_name("os.path"), os.path)
 
     def test_raise_load_module_from_name_1(self) -> None:
-        self.assertRaises(ImportError, modutils.load_module_from_name, "_this_module_does_not_exist_")
+        self.assertRaises(
+            ImportError, modutils.load_module_from_name, "_this_module_does_not_exist_"
+        )
 
 
 def test_import_dotted_library(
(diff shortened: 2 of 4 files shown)


Chasarr:
Hmm, the CI isn't too happy. Investigating...

Chasarr:
All tests pass on my machine now, so it should work

Pierre-Sassoulas (commented):
Thank you for the fix @Chasarr, do you mind adding some automated tests please ?

Chasarr:
Done :)

tests/test_modutils.py
```diff
@@ -3,6 +3,7 @@
 # Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
 
 """Unit tests for module modutils (module manipulation utilities)."""
+
```
Pierre-Sassoulas, line 6:
```suggestion
```

Pierre-Sassoulas (changes requested):
Thank you for the automated tests, it looks great. Do you mind adding a changelog entry too ? You probably need to specify that the issue you're closing come from pylint, like this:

https://github.com/pylint-dev/astroid/blob/df1077718313228edb0129c0c65b3ab6880aafc2/ChangeLog#L52

And if you want to have an aliase for the contributors list you can do it in https://github.com/pylint-dev/astroid/blob/main/script/.contributors_aliases.json#L4

Chasarr:
Done! Thanks for the guidance!

tests/test_modutils.py
```diff
@@ -335,6 +335,22 @@ def test_pyi_preferred(self) -> None:
             os.path.normpath(module) + "i",
         )
 
+    def test_nonstandard_extension(self) -> None:
+        package = resources.find("pyi_data/find_test")
+        modules = [
+            os.path.join(package, "__init__.weird_ext"),
+            os.path.join(package, "standalone_file.weird_ext"),
```
Pierre-Sassoulas, line 342:
```suggestion
            os.path.join(package, "__init__.weird_ext"),
            os.path.join(package, "standalone_file.weird_ext"),
            os.path.join(package, "standalone_file.py"),
```
I might have reviewed this still half asleep the first time* but shouldn't it be this instead (issue talks about 'two file with the same name but different extension') ? I suppose being able to also handle 'two file with the same extension but different name is also valuable.

* (just found and read the pylint issue to close it automatically when we merge this and it doesn't match)

ChangeLog
```diff
@@ -41,6 +41,10 @@ Release date: TBA
 
   Closes #2684
 
+* Fix bug where ``pylint code.custom_extension`` would analyze ``code.py`` or ``code.pyi`` instead if they existed.
+
+	Closes pylint-dev/pylint#3631
+
```
Pierre-Sassoulas, line 47:
```suggestion
* Fix a bug where astroid could not parse ``code.pyi`` if ``code.py`` existed.

	Closes pylint-dev/pylint#3631

```

Chasarr: Hmm, not quite, this should not affect the behaviour of .pyi, .py or .pyw extensions. They worked as intended before my PR.

These three extensions seem to be picked according to some kind of hierarchy that can be changed with prefer_stubs = True/False.

With prefer_stubs = False:
  .py, .pyw, .pyi (.pyw is only valid picked on Windows machines)
With prefer_stubs = True:
  .pyi, .pyw, .py

So if I ran `pylint mycode.pyi` it would actually check `mycode.py`, if it exists. I’m not sure why it works in this way, though.

The bug that I fixed appeared when using a file with another file extension than these three, such as .weird_ext, where it would completely disregard my non-standard file in favor of the correctly named file. E.g. `pylint mycode.weird_ext` would in reality analyze `mycode.py` instead if it existed.

I don’t think I want to go too deep into what bug was fixed in the changelog, so perhaps the first explanation was sufficient? I can’t think of a better way to explain it in one sentence.

tests/test_modutils.py
Chasarr: There already exists a __init__.py, so I figured I could just reuse that file for my tests too. And I also threw in a standalone file without any weird naming and without any other file sharing it's base name just because  ¯\_(ツ)_/¯

Chasarr: Should we perhaps test for `__init__.py` too? I just realized there doesn't seem to be a test for that. Most likely other test would catch that error though.

DanielNoord:
@Pierre-Sassoulas Are you able to review this? Before it goes stale 😄

tests/test_modutils.py
Pierre-Sassoulas: It's okay let's test once

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Sorry for the delay, thank you for the PR, congrats on becoming a contributor :)

Chasarr:
Thank you, this is the first piece of code to a big open source project I have contributed to! Feels good!

Chasarr:
Should I do anything to backport the commit to 3.3.X?

Pierre-Sassoulas:
Don't worry I'm going to do it :)

PR #2730, opened 2025-05-06
Bump astroid to 3.3.10, update changelog

### LINKED ISSUE #2684: Astroid >=3.3.7 imports and executes modules that contain any similarly named identifier as some stdlib module

Pierre-Sassoulas:
* Avoid importing submodules sharing names with standard library modules.

  Closes #2684

* Fix bug where ``pylint code.custom_extension`` would analyze ``code.py`` or ``code.pyi`` instead if they existed.
   Closes pylint-dev/pylint#3631

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,9 +9,15 @@ Release date: TBA
 
 
 
-What's New in astroid 3.3.10?
+What's New in astroid 3.3.11?
 =============================
 Release date: TBA
+
+
+
+What's New in astroid 3.3.10?
+=============================
+Release date: 2025-05-07
 
 * Avoid importing submodules sharing names with standard library modules.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,7 +17,7 @@ Release date: TBA
 
 What's New in astroid 3.3.10?
 =============================
-Release date: 2025-05-07
+Release date: 2025-05-10
 
 * Avoid importing submodules sharing names with standard library modules.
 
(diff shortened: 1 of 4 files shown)


ChangeLog
```diff
@@ -9,10 +9,16 @@ Release date: TBA
...
-What's New in astroid 3.3.10?
+What's New in astroid 3.3.11?
 =============================
 Release date: TBA
 
+
+
+What's New in astroid 3.3.10?
+=============================
+Release date: 2025-05-07
```
Pierre-Sassoulas, line 20:
```suggestion
Release date: 2025-05-10
```

jacobtylerwalls (approved):

PR #2731, opened 2025-05-08
[ci] Start testing on python 3.14-beta

Pierre-Sassoulas:
## Type of Changes

### DIFF
--- a/tests/testdata/python3/data/module.py
+++ b/tests/testdata/python3/data/module.py
@@ -59,10 +59,10 @@ class YOUPI(YO):
                 return 'hehe'
             global_access(local, val=autre)
         finally:
-            # return in finally was previousely tested here but became a syntax error
+            # return in finally was previously tested here but became a syntax error
             # in 3.14 and this file is used in 188/1464 tests
             a = local
-
+    
     def static_method():
         """static method test"""
         assert MY_DICT, '???'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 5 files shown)


tests/testdata/python3/data/module.py
```diff
@@ -2,6 +2,9 @@
 """
 
 __revision__ = '$Id: module.py,v 1.2 2005-11-02 11:56:54 syt Exp $'
+
+import sys
+
```
Pierre-Sassoulas, line 7:
```suggestion
```

cdce8p:
I've been looking into the `brain` errors. Will open a PR to fix them later today.

cdce8p:
Rebased the PR and fixed the remaining two test cases. Just wondering if we need to split the `as_string` case.
On my machine (macOS ARM) I'm also seeing outright Python crashes for two test cases. The CI run seems to be fine though.?

Both are already ignored for PyPy. When I run them locally, Python crashes with `Abort trap: 6`
- `tests/test_inference.py::InferenceTest::test_no_infinite_ancestor_loop`
- `tests/test_inference.py::test_recursion_on_inference_tip`

--
Also seeing a crash for one pylint test
- `tests/test_functional.py::test_functional[regression_5408]`

Pierre-Sassoulas:
> Just wondering if we need to split the as_string case.

I was trying to keep the return in finally for python < 3.14 but it's near impossible (very hard as we have 188 tests using it, we can't remove them all for python 3.14). Probably not something we're going to break by mistake, so probably safe to skip.

Pierre-Sassoulas:
My current guess with the last issue is either a fail of pytest under windows and python 3.14 or a cache issue on our side.

Pierre-Sassoulas:
Added backport label because we'll need 3.14 compat for pylint faster than we can properly release 4.0.0.

PR #2735, opened 2025-05-12
Include subclasses of standard property classes as property decorators

mharding-hpe:
## Type of Changes

## Description

This is gets `astroid` to automatically recognize subclasses of standard Python `property` classes (like `functools.cached_property`) as `property` decorators. This change corrects the problem I reported in [issue 10377](https://github.com/pylint-dev/pylint/issues/10377).

I also wasn't sure if I should describe this as a bug fix or a new feature. :) 

This PR also corrects an inaccuracy regarding when `enum.property` was added to Python.

Refs #2306
Closes [#10377](https://github.com/pylint-dev/pylint/issues/10377)

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
     from astroid.constraint import Constraint
 
 
-PROPERTIES = {"builtins.property", "abc.abstractproperty"}
+PROPERTIES = {"builtins.property", "abc.abstractproperty", "functools.cached_property"}
 if PY310_PLUS:
     PROPERTIES.add("enum.property")
 
@@ -79,24 +79,30 @@ def _is_property(
     if any(name in stripped for name in POSSIBLE_PROPERTIES):
         return True
 
-    # Lookup for subclasses of *property*
     if not meth.decorators:
         return False
+    # Lookup for subclasses of *property*
     for decorator in meth.decorators.nodes or ():
         inferred = safe_infer(decorator, context=context)
         if inferred is None or isinstance(inferred, UninferableBase):
             continue
         if isinstance(inferred, nodes.ClassDef):
+            # Check for a class which inherits from a standard property type
+            if any(inferred.is_subtype_of(pclass) for pclass in PROPERTIES):
+                return True
             for base_class in inferred.bases:
-                if not isinstance(base_class, nodes.Name):
+                # Check for a class which inherits from functools.cached_property
+                # and includes a subscripted type annotation
+                if isinstance(base_class, nodes.Subscript):
+                    value = safe_infer(base_class.value, context=context)
+                    if not isinstance(value, nodes.ClassDef):
+                        continue
+                    if value.name != "cached_property":
+                        continue
+                    module, _ = value.lookup(value.name)
+                    if isinstance(module, nodes.Module) and module.name == "functools":
+                        return True
                     continue
-                module, _ = base_class.lookup(base_class.name)
-                if (
-                    isinstance(module, nodes.Module)
-                    and module.name == "builtins"
-                    and base_class.name == "property"
-                ):
-                    return True
 
     return False
 
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -13,7 +13,7 @@ from collections.abc import Iterable, Iterator
 from typing import TYPE_CHECKING, Any, Literal
 
 from astroid import decorators, nodes
-from astroid.const import PY310_PLUS
+from astroid.const import PY311_PLUS
 from astroid.context import (
     CallContext,
     InferenceContext,
@@ -39,7 +39,8 @@ if TYPE_CHECKING:
 
 
 PROPERTIES = {"builtins.property", "abc.abstractproperty", "functools.cached_property"}
-if PY310_PLUS:
+# enum.property was added in Python 3.11
+if PY311_PLUS:
     PROPERTIES.add("enum.property")
 
 # List of possible property names. We use this list in order
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -966,7 +966,7 @@ class BoundMethodNodeTest(unittest.TestCase):
         ):
             self._is_property(ast, prop)
 
-    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property")
+    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property introduced in 3.11")
     def test_is_standard_property_py311(self) -> None:
         # Test to make sure the Python-provided property decorators
         # are properly interpreted as properties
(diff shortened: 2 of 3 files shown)


DanielNoord (commented):
Do you mind adding a test for this? There should be other tests for `property` that we can adapt :)

mharding-hpe:
Actually, I have hit a problem with this PR as written. It works if my class inherits from `cached_property`, but not if it inherits from `cached_property[T]`. Will need to dig further. I'll move this PR into draft state in the meantime.

Pierre-Sassoulas (commented):
Looks great so far, thank you for digging into this !

mharding-hpe:
I have a possible fix to the subscripting issue. I have started working on the tests.

mharding-hpe:
Ok, I've updated my original fix so that it handles subclasses of `functools.cached_property`, whether or not it is subscripted. I added to the existing property tests in `test_nodes.py` to cover these cases, as well as a couple of others that were not currently being covered by that test.

mharding-hpe:
I ran the full test suite without my PR, and got:
```
1653 passed, 50 skipped, 15 xfailed, 1 xpassed
```

And with my PR (including the 1 additional test (`test_is_property_py312`) it created):
```
1654 passed, 50 skipped, 15 xfailed, 1 xpassed
```

mharding-hpe:
I see the tests don't cover a few of the non-property cases in the code. I think I know which ones I can add to address this. Hold, please.

mharding-hpe:
In working on filling out the code coverage, I ran into a question about how we want `pylint` to behave in a particular case.

Some context -- from looking at the `_is_property` function in `bases.py`, it is clear that `pylint` is willing to decide that a decorator is a property decorated just based on its name. For example, if the decorator is named `@cached_property` or `@reify`, then it will be assumed to be a property, even in cases where this is clearly not the case (like in the existing test, where an empty function is named `cached_property` and used successfully as a property decorator).

I am assuming that this is the desired philosophy into which my PR should fit. But I realized that my PR is making pylint pickier in some sense. Not in a regression sense -- my changes don't break any current behavior. But for example, consider the following code:

```python
import functools

class cached_property:
    pass

class user_property(cached_property):
    pass

class user_property_from_functools(functools.cached_property):
    pass

class A:
    @cached_property
    def cached_property(self): return 42
    @user_property
    def user_property(self): return 42
    @user_property_from_functools
    def user_property_from_functools(self): return 42
```

Without my PR in place, then only `A.cached_property` is considered by `pylint` to be a property.
With my PR in place, `A.cached_property` and `A.user_property_from_functools` are considered by `pylint` to be properties, but `A.user_property` is not. In other words, when I am validating the parent classes, I am not being as generous as `pylint` normally has been. This new behavior feels inconsistent, since I can use an empty class named `cached_property` and have it be considered a property, but if I subclass it, then it no longer works.

If we want to retain this idea that `pylint` will use the name as an assurance that it is a property, then I can easily change my PR so that in the above example, all of those will work. BUT, there is another case. Consider:

```python
cached_property = { int: float }
class user_property(cached_property[int]):
    pass
```

Do we want `user_property` here to also be interpreted as a property if used as a decorator? In this case, `cached_property` is not even referring to a class, but instead to a dict. I do see that `pylint` will not interpret THIS `cached_property` as a property decorator by itself, so maybe it's okay to also forbid the subclassing to be considered as such.

I'm going to modify my PR to allow all 3 cases of my first example, but I will retain the fact that it balks at that last example. Let me know if you disagree with any of that, or if I'm not being clear about it.

mharding-hpe:
Ok, I've added tests to cover all of the situations I described in my earlier comments. I also added tests for other things that had been missing, like for the `enum.property` class.

mharding-hpe:
I see that the 3.10 tests are failing with the new `enum.property` tests I added.
When looking into this failure, I noticed that `enum.property` was actually not added until Python 3.11.
I had been using checks for Python 3.10 because that it what was present in the existing `astroid` code in `bases.py`.
I think that needs to be changed to reflect when it was actually added. I can cover that with this same PR.

mharding-hpe:
[This PR](https://github.com/pylint-dev/astroid/pull/1028) seems to be the one which claimed `enum.properties` was added in 3.10.
[The Python docs](https://docs.python.org/3/library/enum.html#enum.property) say 3.11, which matches what I see when I try to import it under 3.10.

mharding-hpe:
I've added a commit to change the tests and `bases.py` to reflect that `enum.property` was added in Python 3.11, not 3.10. I'll also add the original PR author as a reviewer in case there is something going on here that is not apparent to me.

nelfin:
LGTM. For context, 3.10 wasn't released in Jul of 2021, that change was for
the alpha release. enum.property was removed in
https://github.com/python/cpython/pull/27010 commit b58dc58

mharding-hpe:
Aha -- that makes sense. I should have paid closer attention to the date of the original commit. Thank you for the insight!

> LGTM. For context, 3.10 wasn't released in Jul of 2021, that change was for
> the alpha release. enum.property was removed in
> https://github.com/python/cpython/pull/27010 commit b58dc58

mharding-hpe:
Working on this PR gave me the thought that there may be other cases that are currently overlooked in the tests (like validating the other standard property classes). @DanielNoord @Pierre-Sassoulas are you open to further PRs that are purely test improvements/additions, if I see anything worth doing?

Pierre-Sassoulas:
We're open to any PR making the code base better :) your work and in depth digging into issues is appreciated

DanielNoord (approved):
This looks awesome!

@Pierre-Sassoulas Do you want to review as well?

astroid/bases.py
```diff
@@ -38,8 +38,9 @@
     from astroid.constraint import Constraint
 
 
-PROPERTIES = {"builtins.property", "abc.abstractproperty"}
-if PY310_PLUS:
+PROPERTIES = {"builtins.property", "abc.abstractproperty", "functools.cached_property"}
+# enum.property was added in Python 3.11
+if PY311_PLUS:
     PROPERTIES.add("enum.property")
```
Pierre-Sassoulas, line 44: Should this be based on pylint's ``py-version`` instead ? Not sure if we have the ability to propagate this option to astroid yet. If not I don't know what exactly the design should be. (this is more a question for @DanielNoord)

tests/test_nodes.py
```diff
@@ -924,67 +924,274 @@ def test(self):
...
+        )
+        for prop in (
+            "builtin_p",
+            "abc_p",
+            "abstractmethod_p",
+            "functools_p",
+        ):
+            self._is_property(ast, prop)
+
+    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property")
```
Pierre-Sassoulas, line 969:
```suggestion
    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property introduced in 3.11")
```

tests/test_nodes.py
```diff
@@ -924,67 +924,274 @@ def test(self):
...
+        )
+        for prop in (
+            "user_p",
+            "user_abc_p",
+            "user_functools_p",
+            "annotated_user_functools_p",
+        ):
+            self._is_property(ast, prop)
+
+    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property")
```
Pierre-Sassoulas, line 1105:
```suggestion
    @pytest.mark.skipif(not PY311_PLUS, reason="Uses enum.property introduced in 3.11")
```

Pierre-Sassoulas (commented):
This is great !

astroid/bases.py
DanielNoord: We don't really have an option to...
Since we haven't had anybody complain about this for 2-3 years after making the mistake I wouldn't bother with adding complicated support for it.

DanielNoord (approved):

Issue #2743, opened 2025-05-13
`eval.__get__(1)` should be an inference error


temyurchenko:
Right now, it infers to `eval` (the bound method of `builtins) itself, which is wrong. Note that `eval.__get()__` correctly resolves to an inference error. 

Found while investigating #2672

PR #2739, opened 2025-05-13
[fix] Fix a crash when the root of a node is not a module but is unknown

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #2672

PR #2744, opened 2025-05-13
Fix another non module root

### LINKED ISSUE #120: Raise an error if .root() is not a module

---

.root() should guarantee that what it returns is a module. Unfortunately, I found a couple of cases, which involved transforms, where .root() return classes instead of modules. We should raise an error in that case (and probably we should fix whatever was returning a class to return a module instead).

---
- Bitbucket: https://bitbucket.org/logilab/astroid/issue/120

temyurchenko:
This fixes Unknown nodes that are created in ObjectModel without a parent. Furthermore, it forbids creation of Unknown nodes without a parent.

It also creates an xfailing test for #2743.

## Type of Changes

## Description

Refs #2739, #2743.

Closes #120

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -228,17 +228,17 @@ class ModuleModel(ObjectModel):
     @property
     def attr___spec__(self):
         # No handling for now.
-        return node_classes.Unknown()
+        return node_classes.Unknown(parent=self._instance)
 
     @property
     def attr___loader__(self):
         # No handling for now.
-        return node_classes.Unknown()
+        return node_classes.Unknown(parent=self._instance)
 
     @property
     def attr___cached__(self):
         # No handling for now.
-        return node_classes.Unknown()
+        return node_classes.Unknown(parent=self._instance)
 
 
 class FunctionModel(ObjectModel):
@@ -462,7 +462,7 @@ class FunctionModel(ObjectModel):
     # These are here just for completion.
     @property
     def attr___ne__(self):
-        return node_classes.Unknown()
+        return node_classes.Unknown(parent=self._instance)
 
     attr___subclasshook__ = attr___ne__
     attr___str__ = attr___ne__
@@ -493,8 +493,8 @@ class ClassModel(ObjectModel):
         super().__init__()
 
     @property
-    def attr___annotations__(self) -> node_classes.Unkown:
-        return node_classes.Unknown()
+    def attr___annotations__(self) -> node_classes.Unknown:
+        return node_classes.Unknown(parent=self._instance)
 
     @property
     def attr___module__(self):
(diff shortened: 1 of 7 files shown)


astroid/nodes/node_classes.py
```diff
@@ -4963,9 +4963,9 @@ class Unknown(_base_nodes.AssignTypeNode):
 
     def __init__(
         self,
+        parent: NodeNG,
         lineno: None = None,
         col_offset: None = None,
-        parent: None = None,
         *,
```
Pierre-Sassoulas, line 4969: Should we consider this a breaking change ?

astroid/interpreter/objectmodel.py
```diff
@@ -493,8 +493,8 @@ def __init__(self):
         super().__init__()
 
     @property
-    def attr___annotations__(self) -> node_classes.Unkown:
-        return node_classes.Unknown()
+    def attr___annotations__(self) -> node_classes.Unknown:
```
Pierre-Sassoulas, line 496: Nice catch, strange that this typo was not a problem anywhere.

Pierre-Sassoulas (commented):
Thank you for chiming in and investigating the root cause ! LGTM, just wondering if we should backport or wait for 4.0.

astroid/nodes/node_classes.py
temyurchenko: You are right, it is.

astroid/interpreter/objectmodel.py
temyurchenko: My local checker (not mypy, pyright) found other typing issues in that file. Maybe mypy isn't running on this file for whatever reason.

astroid/nodes/node_classes.py
temyurchenko: What is the procedure for handling this?

Pierre-Sassoulas: We'll wait for 4.0 to release it, and in astroid we just add a mildly strongly worded changelog entry. Or maybe we're not going to, I'm sure @DanielNoord will have an opinion about it :)

astroid/interpreter/objectmodel.py
Pierre-Sassoulas: Yeah, mypy doesn't run on a lot of file at the moment. Lots of dynamic things that are very hard to type. I would have thought that actual import of something that does not exist would have been an error, but it seems it's not.

temyurchenko: I think it's due to `from __future__ import annotations` at the top. It makes python treat every annotation as a string. Without it, it would have been checked.

DanielNoord (approved):

Pierre-Sassoulas:
@DanielNoord release in 4.0.0 as breaking or backport in 3.3.11 ?

DanielNoord:
I think back porting is fine :)

PR #2756, opened 2025-05-26
Fix modutils._get_relative_base_path returns incorrect result when directory name starts with directory name in path_to_check

### LINKED ISSUE #2608: modutils._get_relative_base_path returns incorrect result when directory name starts with directory name in path_to_check
Consider following call: `astroid.modutils._get_relative_base_path('something', os.path.join(os.getcwd(), 'some'))`.
I would expect this to return `None` by design. However, this returns `'thing'` instead.

This results in issues with running pylint on a project with following directory structure:
```
some
\- __init__.py
something
\- __init__.py
thing
\- __init__.py
```
I have observed this in astroid version 3.1.0, but it seems to me that the code in master is still the same.

pavan-msys:
## Type of Changes

## Description

This PR fixes a bug in `astroid.modutils._get_relative_base_path` where the function would incorrectly return a partial subpath if the target path started with the same prefix as `path_to_check`, but was not actually a subdirectory.

Closes #2608

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,10 @@ Release date: TBA
 
 * Modify ``astroid.bases`` and ``tests.test_nodes`` to reflect that `enum.property` was added in Python 3.11, not 3.10
 
+* Fix incorrect result in `_get_relative_base_path` when the target directory name starts with the base path
+
+  Closes #2608
+
 What's New in astroid 3.3.11?
 =============================
 Release date: TBA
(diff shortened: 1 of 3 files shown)


DanielNoord (approved):
Awesome, thanks!

DanielNoord:
@pavan-msys Do you think it is possible to get full coverage on this diff?

pavan-msys:
Hi @DanielNoord 

Thanks for the feedback! I added the missing edge test case after the initial coverage was completed, so the current diff should now have full coverage.

DanielNoord (approved):
My bad, I totally missed the fact I still had to approve the run.

Thanks!

PR #2765, opened 2025-06-11
[maintenance] Drop the brain for nose, filter warning on tests for old style namespace packages

Pierre-Sassoulas:
## Description

The faster way to fix the pipeline.


DanielNoord:
Pipeline still fails :smile: 

What's the error we're fixing?

Pierre-Sassoulas:
```
 ERROR tests/brain/test_nose.py - UserWarning: pkg_resources is deprecated as an API. 
See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package
 is slated for removal as early as 2025-11-30. Refrain from using this package or pin to 
Setuptools<81.
 ```
Seen here: https://github.com/pylint-dev/astroid/pull/2763

File: ChangeLog [hunk 1]
```diff
@@ -42,6 +42,10 @@ Release date: TBA
 
   Closes #2608
 
+* The brain for nose was dropped. It's been deprecated for 10 years and required some maintenance.
```
```suggestion
* The brain for nose was dropped. nose has been deprecated for 10 years and the brain required some maintenance.
```

DanielNoord (dismissed):
I agree with your suggestion for the changelog, rest LGTM!

Pierre-Sassoulas:
The test was introduced in https://github.com/pylint-dev/astroid/pull/1536, it seems the warning is raised from the analysed code and not the astroid code.

PR #2786, opened 2025-07-22
Fix false positive no-member in except * handler

matusvalo:
## Type of Changes

## Description

This PR allows astroid to understand that `except *` produces `ExceptionGroup` instead of specified exception. 

> [!NOTE] 
> This PR is mainly interested in `ExceptionGroup.exceptions` attribute. Other attributes/methods are left for future improvements.

Refs https://github.com/pylint-dev/pylint/issues/9056


astroid/protocols.py
```diff
@@ -527,11 +527,30 @@ def excepthandler_assigned_stmts(
...
+    def _generate_assigned():
+        for assigned in node_classes.unpack_infer(self.type):
+            if isinstance(assigned, nodes.ClassDef):
+                assigned = objects.ExceptionInstance(assigned)
 
+            yield assigned
+
+    if isinstance(self.parent, node_classes.TryStar):
+        # except * handler has assigned ExceptionGroup
+        eg = nodes.ClassDef(
```
DanielNoord: I'm wondering if we can just get this from `builtins` and infer it from there? Why do we need to create the `ClassDef` ourselves?

matusvalo: I will try that.

matusvalo: OK `ExceptionGroup` constructed and infered from `builtins` module...

Pierre-Sassoulas:
Let's noqa pylint

Pierre-Sassoulas (approved):

astroid/protocols.py
```diff
@@ -527,11 +528,34 @@ def excepthandler_assigned_stmts(
...
 
+            yield assigned
+
+    if isinstance(self.parent, node_classes.TryStar):
+        # except * handler has assigned ExceptionGroup with caught
+        # exceptions under exceptions attribute
+        # pylint: disable-next=stop-iteration-return
+        eg = next(
+            node_classes.unpack_infer(
+                extract_node(
```
Pierre-Sassoulas, line 544: That's clever but I'm not sure it's the right approach (maybe there's something more efficient ?). Maybe another maintainer will have ideas about this.

matusvalo: I used it since it seemed to be standard way how to do it - e.g. here: https://github.com/pylint-dev/astroid/blob/8e19277598d5764f5afd821d7bb6104b701ac88f/astroid/brain/brain_collections.py#L120C32-L120C55

cdce8p:
@matusvalo Please make sure to select `squash` for merging. It helps keep the commit history clean. Just noticed a "Merge 'main' into .." commit on main and it would be great if we could avoid that. I'd also recommend not to enable auto-merge too early. Good rule of thumb is usually to just do it if all comments have been addressed + at least one approving review + only waiting for CI tests to pass.

Issue #2789, opened 2025-07-28
TemplateString: Proper inference similar to JoinedStr._infer and additional tests for the as_string()


jacobtylerwalls:
Fix this error in pylint with 3.14:

-  AttributeError: 'TreeRebuilder' object has no attribute 'visit_templatestr'

https://docs.python.org/3.14/library/ast.html#ast.TemplateStr

cdce8p:
Initial support added in #2793 and released as part of `4.0.0b2`.

#### Open issues
- Additional tests for the `as_string()` logic. The basics work but there might be a lot of edge case which need to be handled properly. The existing f-string tests could provide a good starting point.
https://github.com/pylint-dev/astroid/blob/f9e109eedf75fcc900e9209a9d42ec516599e301/tests/test_nodes.py#L295-L307
- Add proper inference similar to `JoinedStr._infer`.
https://github.com/pylint-dev/astroid/blob/f9e109eedf75fcc900e9209a9d42ec516599e301/astroid/nodes/node_classes.py#L4775-L4781

PR #2793, opened 2025-08-06
Add basic support for TemplateStr

cdce8p:
## Description

Add the basic buildings blocks for `ast.TemplateStr` and `ast.Interpolation` nodes. Most of it is pretty similar to the f-string nodes.

The implementation isn't complete but good enough to not outright fail if a t-string is used anywhere in a file. Future PRs should look to address the following
- Additional tests for the `as_string()` logic. The basics work but there might be a lot of edge case which need to be handled properly. The existing f-string tests could provide a good starting point.
https://github.com/pylint-dev/astroid/blob/f9e109eedf75fcc900e9209a9d42ec516599e301/tests/test_nodes.py#L295-L307
- Add proper inference similar to `JoinedStr._infer`.
https://github.com/pylint-dev/astroid/blob/f9e109eedf75fcc900e9209a9d42ec516599e301/astroid/nodes/node_classes.py#L4775-L4781

Ref #2789


astroid/nodes/node_classes.py
```diff
@@ -5484,6 +5484,114 @@ def postinit(self, *, patterns: list[Pattern]) -> None:
...
+            end_lineno=end_lineno,
+            end_col_offset=end_col_offset,
+            parent=parent,
+        )
+
+    def postinit(
+        self,
+        *,
+        value: NodeNG,
+        str: str,  # pylint: disable=redefined-builtin
```
cdce8p, line 5580: @Pierre-Sassoulas Should we change the attribute name here? `str` is the one used by the ast node but it's a bit unfortunate in my opinion since it shadows the builtin str.

Not sure using an alias here would be better though.

https://docs.python.org/3.14/library/ast.html#ast.Interpolation

Pierre-Sassoulas: I agree, not ideal, but what should we use instead ? Any non adherence to ast's API is mental load on astroid's user and documentation burden on us.

cdce8p: Just wanted to discuss it first. I also think that `str` while not ideal is still the best option.

Pierre-Sassoulas: Before you merged I was writing: Let's wait to read other opinions. Some maintainers might still be ambitious enough to keep astroid a wrapper above the AST and its multiple interpreters :D 

It's still possible to change this later if someone want, we're in alpha/betas atm. Being able to test this in pylint fast is good too.

cdce8p: Sorry about that. Guess I was too fast this time 😅

Yeah, we can still change it. Furthermore I think the biggest issue is actually in the `postinit` arg name as that's used on it's own. Accessing the attribute itself shouldn't really be an issue with `node.str`.

cdce8p:
> > Do you want me to release an astroid beta ?
> 
> I still need to debug at least one astroid issue for 3.14. Might make sense to wait a bit.

Looks like what I was seeing is an existing issue with astroid not being able to infer `sys.version_info` properly.
_Even found an old issue for it #113. Maybe that's something I'll look into at some point._

Feel free to do a new beta release if you've some time.

Pierre-Sassoulas:
I can't find the conversation about it but I remember saying somewhere that this kind of call should be uninferable (you don't know at static analysis time what is the value for the interpreter or the os, or...). If we want to do better, we would need to use a different constant for each, maybe we can discuss this in 113.

PR #2797, opened 2025-08-09
Add support for type parameter defaults + string repr improvements

cdce8p:
## Description
Type parameter defaults were added in Python 3.13.
Also improved the string representation of various TypeVar / type param related nodes.

https://peps.python.org/pep-0696
https://peps.python.org/pep-0695

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -60,6 +60,11 @@ Release date: TBA
 
   refs #2789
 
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (approved):
Look great, let's wait for another reviewer's opinion before merging.

PR #2811, opened 2025-09-02
Update `attrs` brain to exclude `ClassVar` annotated attributes

zenlyj:
## Type of Changes

## Description

This is related to [pylint #10525](https://github.com/pylint-dev/pylint/issues/10525). When a class is decorated with `attrs.define`, attributes are treated as instance variables, unless they are explicitly annotated with `typing.ClassVar`. However, the `attrs` brain transforms all class attributes into instance attributes, including those with `ClassVar` type annotation. 

This update fixes that behavior by ensuring that attributes annotated with `ClassVar` are skipped during the transformation.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 4.0.0?
 ============================
 Release date: TBA
 
+* Fix false positive `invalid-name` on `attrs` classes with `ClassVar` annotated variables.
+
+  Closes pylint-dev/pylint#10525
+
 * Prevent crash when parsing deeply nested parentheses causing MemoryError in python's built-in ast.
 
   Closes #2643


astroid/brain/brain_dataclasses.py
```diff
@@ -550,16 +550,6 @@ def _get_field_default(field_call: nodes.Call) -> _FieldDefaultReturn:
     return None
 
 
-def _is_class_var(node: nodes.NodeNG) -> bool:
```
zenlyj, line 553: I came across this method while looking for a way to detect `ClassVar` annotations, moved it to `utils` to make it reusable, but let me know if there is a more suitable place for it.

astroid/util.py
```diff
@@ -157,3 +157,13 @@ def safe_infer(
...
+
+
+def is_class_var(node: nodes.NodeNG) -> bool:
+    """Return True if node is a ClassVar, with or without subscripting."""
+    try:
+        inferred = next(node.infer())
+    except (InferenceError, StopIteration):
+        return False
+
+    return getattr(inferred, "name", "") == "ClassVar"
```
DanielNoord: This is not super safe of course, especially with the `getattr`.

Perhaps we can move it to `helpers.py` in `astroid/brain`? In general I think we have a bit of a higher standard for code in `astroid/*` than in `astroid/brain/*`.

zenlyj: Thanks for the suggestion! Agree that it is a little out of place in `astroid/*`, I've moved this to `astroid/brain/helpers.py`. 

Also, I added some tests for the `is_class_var` method. I had to create a new file for this as I did not see any test file used for `brain/helpers.py`.

DanielNoord (approved):

PR #2823, opened 2025-09-19
Support boolean truthiness constraints in inference

zenlyj:
## Type of Changes

## Description

This PR builds on #1189 by implementing the constraint interface to handle boolean constraints, similar to how `None` constraints are currently treated.

Previously, constraints only accounted for `None` checks:
```
x = 1

if x is None:
    x    # Uninferable, `1` is filtered out because x is not None
```

These changes support constraints derived from boolean truthiness:

1. If condition is `True`, keep the inferred value.
```
x = 1    # truthy

if x:
    x    # inferred as `1`
```

2. If condition is `False`, filter out the inferred value.
```
x = []    # falsy

if x:
    x    # Uninferable, `[]` filtered out
```

3. Negation rules are the same as `None` constraints.
```
x = 1

if x:
    pass
else:
    x    # Uninferable, `1` filtered out


y = []

if not y:
    y    # inferred as `[]`
```

4. If the boolean value cannot be inferred, assume constraint is satisfied.

Closes https://github.com/pylint-dev/pylint/issues/9515

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -5491,7 +5491,7 @@ class CallSiteTest(unittest.TestCase):
             else:
                 kwargs = {}
 
-            if nums:
+            if nums is not None:
                 add(*nums)
                 print(**kwargs)
         """
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 4.0.0?
 ============================
 Release date: TBA
 
+* Add support for boolean truthiness constraints (`x`, `not x`) in inference.
+
+  Closes pylint-dev/pylint#9515
+
 * Fix false positive `invalid-name` on `attrs` classes with `ClassVar` annotated variables.
 
   Closes pylint-dev/pylint#10525
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -84,6 +84,47 @@ class NoneConstraint(Constraint):
         return self.negate ^ _matches(inferred, self.CONST_NONE)
 
 
+class BooleanConstraint(Constraint):
+    """Represents an "x" or "not x" constraint."""
+
+    @classmethod
+    def match(
+        cls, node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
+    ) -> Self | None:
+        """Return a new constraint for node if expr matches one of these patterns:
+
+        - direct match (expr == node): use given negate value
+        - negated match (expr == `not node`): flip negate value
+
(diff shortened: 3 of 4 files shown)


tests/test_inference.py
```diff
@@ -5495,12 +5495,14 @@ def add(x, y):
...
-        # Test that `*nums` argument should be Uninferable
         ast = parse(code, __name__)
         *_, add_call, print_call = list(ast.nodes_of_class(nodes.Call))
         nums_arg = add_call.args[0]
         add_call_site = self._call_site_from_call(add_call)
-        self.assertEqual(add_call_site._unpack_args([nums_arg]), [Uninferable])
+        add_call_site_args = add_call_site._unpack_args([nums_arg])
+        self.assertEqual(len(add_call_site_args), 2)
+        self.assertTrue(all(isinstance(arg, nodes.Const) for arg in add_call_site_args))
+        self.assertEqual([arg.value for arg in add_call_site_args], [1, 2])
```
zenlyj: This assertion broke and I updated it. It is related to https://github.com/pylint-dev/astroid/pull/2117. From my understanding, `nums` is inferred as `()` and `(1, 2)`. Previously, it was asserted as `Uninferable` because of `safe_infer`. With these changes, the `if nums` guard filters out falsy `()`, and `safe_infer` should now return `(1, 2)`.

Pierre-Sassoulas (approved):
Looks amazing !

DanielNoord (approved):
Perhaps we should add a Changelog entry? Rest LGTM!

tests/test_inference.py
```diff
@@ -5495,12 +5495,14 @@ def add(x, y):
                 add(*nums)
```
jacobtylerwalls, line 5495: We should be able to leave the sense of this test as is without changing the assertions by:

```suggestion
            if nums is not None:
                add(*nums)
```

jacobtylerwalls (changes requested):
Lovely, I'm glad more knowledge of the constraints interface is starting to spread 🪐

tests/test_inference.py
zenlyj: Makes sense, I've updated it with the latest commits

zenlyj:
Thanks for the reviews! A lot of thought and effort went into the constraints interface, and it deserves to be built upon rather than left unused. More extensions coming soon in separate PRs!

jacobtylerwalls (approved):

PR #2827, opened 2025-09-22
Revert the botched bump to 4.0.0b4-dev0

Pierre-Sassoulas:
Sorry @jacobtylerwalls ! (Let's not squash this time)

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -41,10 +41,11 @@ Contributors
 - Eevee (Alex Munroe) <amunroe@yelp.com>
 - David Gilman <davidgilman1@gmail.com>
 - Tushar Sadhwani <tushar.sadhwani000@gmail.com>
+- Matus Valo <matusvalo@users.noreply.github.com>
 - Julien Jehannet <julien.jehannet@logilab.fr>
+- Hugo van Kemenade <hugovk@users.noreply.github.com>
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):

Pierre-Sassoulas:
The rebase option disapeared ? It leads to bad history with unnecessary context of what was done in parallel (here in pylint). 

image

DanielNoord:
@Pierre-Sassoulas I have no clue why that is? Perhaps somebody changed an organization setting?

DanielNoord:
Documentation also doesn't seem very helpful: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#rebase-and-merge-your-commits

Perhaps you can try via the CLI to see if that gives you a useful error?

jacobtylerwalls:
Yeah, I had to fix our repo setting.

Pierre-Sassoulas:
Thank you @jacobtylerwalls !

cdce8p:
> Perhaps somebody changed an organization setting?

🙋🏻‍♂️ Guess I was the one. Tbh I don't think we should ever really use rebase merges. If that's necessary, it's probably best to rebase the branch locally first. Ideally we would only allow squash merges. That's just safer. Let's remember that we four are not the only ones with merge privileges and not everyone does the right thing. _Mixup do happen._

PR #2830, opened 2025-09-23
Fix a false positive for ``not-an-iterable`` in Pylint.

mbyrnepr2:
## Type of Changes

## Description

Fix a false positive for ``not-an-iterable`` in an ``ExceptionGroup``context.

Refs #XXXX

Closes pylint-dev/pylint#8985
Closes pylint-dev/pylint#10558

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -85,7 +85,7 @@ Release date: TBA
 * Improve ``as_string()`` representation for ``TypeVar``, ``ParamSpec`` and ``TypeVarTuple`` nodes, as well as
   type parameter in ``ClassDef``, ``FuncDef`` and ``TypeAlias`` nodes (PEP 695).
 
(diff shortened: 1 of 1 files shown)


tests/test_group_exceptions.py
```diff
@@ -21,6 +22,22 @@
 from astroid.nodes import Expr, Raise, TryStar
 
 
+@pytest.mark.skipif(not PY311_PLUS, reason="Requires Python 3.11 or higher")
```
Pierre-Sassoulas:
```suggestion
@pytest.mark.skipif(not PY311_PLUS, reason="Exception group introduced in Python 3.11")
```

DanielNoord (dismissed):
I'd probably mention the fact that we now correctly support the `exceptions` attribute for `ExceptionGroup` instead, as that is the real `astroid` future. But I don't mind too much, so feel free to merge!

PR #2831, opened 2025-09-24
Handle deprecated boolean cast of `NotImplemented`

zenlyj:
## Type of Changes

## Description

Related to https://github.com/pylint-dev/pylint/pull/10585, to address CI test failures encountered when linting Python `stdlib`. From my understanding, the failures were caused by `DeprecatedWarning` emissions when casting `NotImplemented` to boolean in `Const` nodes.

These changes introduce a guard clause to explicitly return `True` when calling `bool_value` on `NotImplemented` nodes. It seems like future Python versions will raise an error when evaluating `NotImplemented` in a boolean context, but the current behavior is to resolve to `True`.

Refer to https://docs.python.org/3.13/library/constants.html#NotImplemented.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -2987,7 +2987,10 @@ class InferenceTest(resources.SysPathSetup, unittest.TestCase):
     def test_bool_value_not_implemented(self) -> None:
         node = extract_node("""NotImplemented""")
         inferred = next(node.infer())
-        self.assertIs(inferred.bool_value(), True)
+        if PY314_PLUS:
+            self.assertIs(inferred.bool_value(), util.Uninferable)
+        else:
+            self.assertIs(inferred.bool_value(), True)
 
     def test_infer_coercion_rules_for_floats_complex(self) -> None:
         ast_nodes = extract_node(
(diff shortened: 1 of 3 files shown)


DanielNoord:
> Ahh I completely missed that, let's raise `TypeError` then to keep things consistent. I originally thought of returning `True` because `bool(NotImplemented)` evaluates to `True` in runtime (for versions under 3.14).

Perhaps we can do it based on `sys.version`?

zenlyj:
That works, to keep compatibility with older versions. I was thinking whether we should be more conservative and return `Uninferable` instead of raising an error, there are some nodes already returning `Uninferable` in `bool_value`

DanielNoord (approved):

tests/test_inference.py
```diff
@@ -2984,6 +2984,14 @@ def __bool__(self):
         inferred = next(instance.infer())
         self.assertIs(inferred.bool_value(), util.Uninferable)
 
+    def test_bool_value_not_implemented(self) -> None:
+        node = extract_node("""NotImplemented""")
```
Pierre-Sassoulas, line 2988: Shouldn't it be 
```suggestion
        node = extract_node("""bool(NotImplemented)""")
```

Pierre-Sassoulas (commented):
Look great, I don't understand something in the test.

tests/test_inference.py
DanielNoord: `bool_value` returns what would have happened if you called `bool` on the `Const` node :)

So, test is correct. You get the `Const` as normal and then check what calling `bool` would do.

Pierre-Sassoulas: Thank you for explaining :);

PR #2832, opened 2025-09-25
Add brain module for statistics inference

emmanuel-ferdman:
## Type of Changes

## Description

My first PR to `astroid` - hope the PR was implemented correctly.

This PR adds a new brain module `brain_statistics.py` that improves inference for `statistics.quantiles()` function calls. The `statistics.quantiles()` function has complex runtime behavior that returns lists with sizes dependent on the n parameter (specifically n-1 elements), but static analysis can only see the empty list initializations (`result = []`) in the function body and cannot follow the control flow that populates these lists. This causes astroid to infer empty lists, leading to false positives in pylint's unbalanced-tuple-unpacking checker. The brain module detects calls to `statistics.quantiles()` and returns `Uninferable` instead, allowing downstream tools to handle the "unknown size" case appropriately rather than acting on incorrect inference results.

Example:
```python
import statistics

# This should work - statistics.quantiles(data, n=4) returns 3 quantiles
q1, q2, q3 = statistics.quantiles(list(range(100)), n=4)
```
Relevant Pylint issue: https://github.com/pylint-dev/pylint/issues/10317

### DIFF
--- a/astroid/brain/brain_statistics.py
+++ b/astroid/brain/brain_statistics.py
@@ -1,0 +1,75 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
+
+"""Astroid hooks for understanding statistics library module.
+
+Provides inference improvements for statistics module functions that have
+complex runtime behavior difficult to analyze statically.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from astroid.context import InferenceContext
+from astroid.inference_tip import inference_tip
+from astroid.manager import AstroidManager
+from astroid.nodes.node_classes import Attribute, Call, ImportFrom
+from astroid.util import Uninferable
+
+if TYPE_CHECKING:
+    from astroid.typing import InferenceResult
+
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+    from astroid.typing import InferenceResult
+
+
+def _looks_like_statistics_quantiles(node: Call) -> bool:
+    """Check if this is a call to statistics.quantiles."""
+    # Case 1: statistics.quantiles(...)
+    if isinstance(node.func, Attribute):
+        if node.func.attrname != "quantiles":
+            return False
+        if hasattr(node.func, "expr") and hasattr(node.func.expr, "name"):
```
DanielNoord: Ideally we don't use `hasattr` here but an `isinstance` on the type of `node.func`. Would that be possible?

astroid/brain/brain_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+            if node.func.expr.name == "statistics":
+                return True
+
+    # Case 2: from statistics import quantiles; quantiles(...)
+    if hasattr(node.func, "name") and node.func.name == "quantiles":
```
DanielNoord: Same here about the use of `hasattr`

astroid/brain/brain_statistics.py
emmanuel-ferdman: Yes, you are right - changed and tested 😄 thanks for the review!

PR #2833, opened 2025-09-26
Update astroid nodes import

cdce8p:
## Description
https://github.com/pylint-dev/pylint/pull/10592

Update node imports to use `from astroid import nodes` instead.

If we want to, we could consider deprecating the re-export in `astroid/__init__.py` at some point.
https://github.com/pylint-dev/astroid/blob/7c3b481d95456a5bbc5a314db2696af3664b6942/astroid/__init__.py#L83-L167

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -14,7 +14,6 @@ from astroid import nodes
 
 if TYPE_CHECKING:
     from astroid import objects
-    from astroid.nodes.node_ng import NodeNG
 
 # pylint: disable=unused-argument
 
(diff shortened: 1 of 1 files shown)


astroid/nodes/as_string.py
```diff
@@ -14,22 +14,6 @@
...
-        MatchClass,
-        MatchMapping,
-        MatchOr,
-        MatchSequence,
-        MatchSingleton,
-        MatchStar,
-        MatchValue,
-        TemplateStr,
-        Unknown,
-    )
```
Pierre-Sassoulas, line 32: Risky to do this in the astroid.nodes package as it's going to easily create circular import, no ?

cdce8p: Good point, though I believe it works in this case. We already use the module import.
https://github.com/pylint-dev/astroid/blob/036714105503e7a1fdb3859e0a49b68936757441/astroid/nodes/as_string.py#L13-L16

_I don't believe it will but if something crashes, we can still move the import into the type checking block. We'd only need to add back the in-block import for the `isinstance` check in `visit_matchas`._

Pierre-Sassoulas (approved):

PR #2837, opened 2025-09-29
Deprecate nodes import from astroid

cdce8p:
## Description
Followup to https://github.com/pylint-dev/astroid/pull/2833

Deprecate the import of node classes from toplevel `astroid`.
Import them from `astroid.nodes` instead.


astroid/__init__.py
```diff
@@ -174,3 +174,69 @@
 # isort: on
 
 from astroid.util import Uninferable
+
+__all__ = [
```
cdce8p, line 178: Added `__all__` to not expose the `_DEPRECATED_*` classes here.

tests/test_nodes.py
```diff
@@ -2311,3 +2312,17 @@ def test_arguments_default_value():
...
+    with pytest.raises(
+        DeprecationWarning, match="importing 'For' from 'astroid' is deprecated"
+    ):
+        from astroid import For
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+        from astroid import For
+
+        assert For is nodes.For
```
Pierre-Sassoulas, line 2328: We could also test that non deprecated import like MANAGER works and do not raise deprecation warning.

Pierre-Sassoulas (commented):
Looks pretty good, clever trick.

tests/test_nodes.py
cdce8p: Added a test for one (not deprecated) import from `nodes`. Don't think it makes much sense to test everything. In the end `__getattr__` is only called if no name is found anyway.

Pierre-Sassoulas: Yeah sure, I meant one, thank you

Pierre-Sassoulas (approved):

PR #2845, opened 2025-10-02
Improve conditionals

cdce8p:
Improvements identified by the `improve-conditionals` check from https://github.com/pylint-dev/pylint/pull/10600.
After the pylint `v4` release, the check should be added to `MESSAGE CONTROL.enable`.

https://github.com/pylint-dev/astroid/blob/a14cc3c6c8c112ea4b8204351cebe2680f4c021b/pylintrc#L115

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -104,7 +104,7 @@ def object_type(
         types = set(_object_type(node, context))
     except InferenceError:
         return util.Uninferable
-    if len(types) > 1 or not types:
+    if len(types) != 1:
         return util.Uninferable
     return next(iter(types))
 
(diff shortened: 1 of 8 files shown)


astroid/helpers.py
```diff
@@ -104,7 +104,7 @@ def object_type(
         types = set(_object_type(node, context))
     except InferenceError:
         return util.Uninferable
-    if len(types) > 1 or not types:
+    if len(types) != 1:
```
Pierre-Sassoulas, line 107: Nice !

Pierre-Sassoulas (approved):

PR #2846, opened 2025-10-04
Support `isinstance` constraints in inference

zenlyj:
## Type of Changes

## Description

These changes implements the constraint interface to add a new constraint - `TypeConstraint` to improve the accuracy of inference for objects used in type-narrowing guards.

Currently, the implementation only supports the `isinstance` pattern, but it should be relatively straightforward to extend it to support the negated case (`not isinstance`) in the future.

To determine whether an object is an instance of the given types, the implementation reuses existing inference mechanism for `isinstance` in `brains`.

Refs https://github.com/pylint-dev/pylint/issues/1162
Refs https://github.com/pylint-dev/pylint/issues/4635
Refs https://github.com/pylint-dev/pylint/issues/10469

### DIFF
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -10,7 +10,8 @@ from abc import ABC, abstractmethod
 from collections.abc import Iterator
 from typing import TYPE_CHECKING
 
-from astroid import nodes, util
+from astroid import helpers, nodes, util
+from astroid.exceptions import AstroidTypeError, InferenceError, MroError
 from astroid.typing import InferenceResult
 
 if sys.version_info >= (3, 11):
@@ -125,6 +126,55 @@ class BooleanConstraint(Constraint):
         return self.negate ^ inferred_booleaness
 
 
+class TypeConstraint(Constraint):
+    """Represents an "isinstance(x, y)" constraint."""
+
+    def __init__(
+        self, node: nodes.NodeNG, classinfo: nodes.NodeNG, negate: bool
+    ) -> None:
+        super().__init__(node=node, negate=negate)
+        self.classinfo = classinfo
+
+    @classmethod
+    def match(
+        cls, node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
+    ) -> Self | None:
+        """Return a new constraint for node if expr matches the
+        "isinstance(x, y)" pattern. Else, return None.
+        """
+        is_instance_call = (
+            isinstance(expr, nodes.Call)
+            and isinstance(expr.func, nodes.Name)
+            and expr.func.name == "isinstance"
+            and not expr.keywords
+            and len(expr.args) == 2
+        )
+        if is_instance_call and _matches(expr.args[0], node):
+            return cls(node=node, classinfo=expr.args[1], negate=negate)
+
+        return None
+
+    def satisfied_by(self, inferred: InferenceResult) -> bool:
+        """Return True for uninferable results, or depending on negate flag:
+
+        - negate=False: satisfied when inferred is an instance of the checked types.
+        - negate=True: satisfied when inferred is not an instance of the checked types.
+        """
+        if isinstance(inferred, util.UninferableBase):
+            return True
+
+        try:
+            types = helpers.class_or_tuple_to_container(self.classinfo)
+            matches_checked_types = helpers.object_isinstance(inferred, types)
+
+            if isinstance(matches_checked_types, util.UninferableBase):
+                return True
+
+            return self.negate ^ matches_checked_types
+        except (InferenceError, AstroidTypeError, MroError):
+            return True
+
+
 def get_constraints(
     expr: _NameNodes, frame: nodes.LocalsDictNodeNG
 ) -> dict[nodes.If | nodes.IfExp, set[Constraint]]:
@@ -159,6 +209,7 @@ ALL_CONSTRAINT_CLASSES = frozenset(
     (
         NoneConstraint,
         BooleanConstraint,
+        TypeConstraint,
     )
 )
 """All supported constraint types."""
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -170,6 +170,30 @@ def object_issubclass(
     return _object_type_is_subclass(node, class_or_seq, context=context)
 
 
+def class_or_tuple_to_container(
+    node: InferenceResult, context: InferenceContext | None = None
+) -> list[InferenceResult]:
+    # Move inferences results into container
+    # to simplify later logic
+    # raises InferenceError if any of the inferences fall through
+    try:
+        node_infer = next(node.infer(context=context))
+    except StopIteration as e:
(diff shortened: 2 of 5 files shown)


zenlyj:
Seems like test coverage is lacking. Shouldn't be an issue to get it up to 100%, but I would like a review first to see if this implementation makes sense.

astroid/constraint.py
```diff
@@ -125,6 +126,55 @@ def satisfied_by(self, inferred: InferenceResult) -> bool:
...
+
+        return None
+
+    def satisfied_by(self, inferred: InferenceResult) -> bool:
+        """Return True for uninferable results, or depending on negate flag:
+
+        - negate=False: satisfied when inferred is an instance of the checked types.
+        - negate=True: satisfied when inferred is not an instance of the checked types.
+        """
+        if isinstance(inferred, util.UninferableBase):
```
jacobtylerwalls: Not a blocker, but would be nice to get a microbenchmark to prove whether we should be doing `isinstance` or `is`. I'm assuming `is Uninferable` is faster.

tests/test_constraint.py
```diff
@@ -773,3 +776,134 @@ def method(self, x = {fail_val}):
...
+        x  #@
+    """
+    )
+
+    inferred = node.inferred()
+    assert len(inferred) == 1
+    assert inferred[0] is Uninferable
+
+
+def test_isinstance_supertype():
```
jacobtylerwalls, line 850: Couple ideas for test cases at your option:
- class A(B, C): pass; `isinstance(a(), C)`
- diamond inheritance

DanielNoord:
> I wondered as well, happy to wait for 4.1. Let's release 4.0 final tomorrow in that case.

Let's do so, especially as we haven't tested this with `pylint` yet. `4.0.0` has enough new features already :)

Pierre-Sassoulas (approved):
This is neat ! Glad to see that the constraint api is getting some love.

astroid/constraint.py
jacobtylerwalls:
```py
>>> from astroid.util import UninferableBase
>>> from timeit import timeit
>>> uninf = UninferableBase()
>>> x = UninferableBase()
>>> timeit("x is uninf", globals=globals())
0.04412866701022722
>>> timeit("isinstance(x, UninferableBase)", globals=globals())
0.051591374998679385

File: astroid/helpers.py [hunk 3]
```diff
@@ -170,6 +170,30 @@ def object_issubclass(
...
+        raise InferenceError(node=node, context=context) from e
+    # arg2 MUST be a type or a TUPLE of types
+    # for isinstance
+    if isinstance(node_infer, nodes.Tuple):
+        try:
+            class_container = [
+                next(node.infer(context=context)) for node in node_infer.elts
+            ]
+        except StopIteration as e:
+            raise InferenceError(node=node, context=context) from e
```
zenlyj: The `StopIteration` handlers are not covered by tests, wondering if it is reasonable to exclude them. From what I understand `.infer()` already throws `InferenceError` and we won't get to these blocks?

Pierre-Sassoulas: Isn't it possible for ``node.infer(context=context)`` to be inferable but an empty iterable ?

astroid/constraint.py
```diff
@@ -117,14 +118,61 @@ def satisfied_by(self, inferred: InferenceResult) -> bool:
...
+        """
+        if inferred is util.Uninferable:
+            return True
+
+        try:
+            types = helpers.class_or_tuple_to_container(self.classinfo)
+            matches_checked_types = helpers.object_isinstance(inferred, types)
+
+            if matches_checked_types is util.Uninferable:
+                return True
```
zenlyj, line 169: This `Uninferable` check is not covered by tests, it should be unreachable in runtime. I am including it as a safe guard because `helpers.object_isinstance()` returns `bool | Uninferable`, but I am also fine with removing it. Thoughts?

astroid/helpers.py
zenlyj: That was what I thought as well, but when we get an empty iterable, this decorator catches `StopIteration` and raises `InferenceError`? Not sure if there are cases where this decorator is not used.

https://github.com/pylint-dev/astroid/blob/ab119c2238870f5bf2e0efbd5fab7bf4866e431d/astroid/decorators.py#L75-L96

jacobtylerwalls: > Not sure if there are cases where this decorator is not used.

That worries me also (plugins?)

A pragma to exclude coverage seems reasonable.

astroid/constraint.py
jacobtylerwalls: Can you share more details about why this is unreachable? I see that we short-circuit if inferred is uninferable, but it wasn't apparent to me why inferring again wouldn't return uninferable.

zenlyj: Looked into `object_isinstance()`, there are two main cases where it can return `Uninferable`:

- When the type of the object cannot be inferred, happens when:
  - The object resolves to multiple types. This should not occur since `inferred` is an item within a list of inferred results. 
  - Inference error is raised while doing `inferred.infer()`. This may occur when `inferred` itself is `Uninferable`, but as you mentioned there is already a short-circuit check above.

- When the type can be inferred but is not an instance of `nodes.ClassDef`. However, it seems that the inferred type always appears to be a ClassDef, or wrapped in a ClassDef proxy.

jacobtylerwalls: > This should not occur since inferred is an item within a list of inferred results.

That's what I'm unsure about, inferring an item may return something besides itself.

zenlyj: Trying to think of a scenario where inferring a non-Uninferable inference result returns multiple results. Not sure if that's a common case but I'd be interested to know if you have any concrete examples in mind.

An alternative is to mock the method to return `Uninferable`:
```
def test_isinstance_uninferable():
    node = builder.extract_node(
        """
    x = 3

    if isinstance(x, str):
        x  #@
    """
    )

    with patch.object(helpers, "object_isinstance", return_value=Uninferable):
        inferred = node.inferred()
        assert len(inferred) == 1
        assert isinstance(inferred[0], nodes.Const)
        assert inferred[0].value == 3
```

jacobtylerwalls: I'm good with that. Sorry I don't have the time to construct a concrete example.

zenlyj: No worries, I will go with the mocking approach for now. If anyone has encountered this before, feel free to chime in!

astroid/helpers.py
zenlyj: I excluded coverage for those blocks, it is worth noting that they were previously uncovered as well. These changes only shift the method to a different file.

jacobtylerwalls:
Hey @zenlyj :wave:

I think this is mergeable. What do we need for the pylint job to pass, do we need a separate PR to unblock, or is it just a rebase?

Issue #2741, opened 2025-05-13
Add necessary attributes to ObjectModel


temyurchenko:
The majority of FunctionModel attributes (e.g. `"__lt__"`, `"__eq__"`,` "__class__"`) should be present in ObjectModel (see https://docs.python.org/3/reference/datamodel.html#object.__lt__); and inherited by FunctionModel. 

This is found while investigating #2672

jacobtylerwalls:
Closed by #2847

PR #2847, opened 2025-10-04
Move object dunders from `FunctionModel` to `ObjectModel`

### LINKED ISSUE #2742: Add `__le__`, `__ge__` and other missing attributes to FunctionModel and ObjectModel
Right now they are missing from the both. Related to #2741

emmanuel-ferdman:
## Type of Changes

## Description

This PR does:

This PR moves object dunder methods from `FunctionModel` to `ObjectModel` as `Unknown` placeholders, making them available to all object types.

Fixes #2742 #2741.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,12 @@ What's New in astroid 4.1.0?
 ============================
 Release date: TBA
 
+* Move object dunder methods from ``FunctionModel`` to ``ObjectModel`` to make them
+  available on all object types, not just functions.
+
+  Closes #2742
+  Closes #2741
+  Closes pylint-dev/pylint#6094
 
 
 What's New in astroid 4.0.1?
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4966,9 +4966,10 @@ class NamedExpr(_base_nodes.AssignTypeNode):
 
 class Unknown(_base_nodes.AssignTypeNode):
     """This node represents a node in a constructed AST where
-    introspection is not possible.  At the moment, it's only used in
-    the args attribute of FunctionDef nodes where function signature
-    introspection failed.
+    introspection is not possible.
+
+    Used in the args attribute of FunctionDef nodes where function signature
+    introspection failed, and as a placeholder in ObjectModel.
     """
 
     name = "Unknown"


DanielNoord (commented):
As can be seen from the docs you link, most of these should live on `ObjectModel` instead as they come from `object`.

We tried to do so before in https://github.com/pylint-dev/astroid/pull/1519 but failed. Perhaps you want to have another look at that PR and see if you can fix the issue we faced there?

emmanuel-ferdman:
@DanielNoord Thanks! I’ll take a look at this. From that thread, it seems you were close to a solution, except for one test case that didn’t pass. Do you remember which test it was?

emmanuel-ferdman:
@DanielNoord I've put together an initial solution for moving object dunders from `FunctionModel` to `ObjectModel`. I've tested it against both the astroid test suite and the pylint test suite (all passing). Since I'm still learning the codebase, I'd appreciate any guidance on whether this approach is correct.

The solution uses a placeholder pattern: `ObjectModel` provides `Unknown` placeholders for all 24 object dunders, and the lookup logic skips these placeholders to find actual implementations. 
 Key changes:
- `ObjectModel` now has all object dunders as `Unknown` placeholders that act as fallbacks when no actual implementation exists
- Lookup logic in `BaseInstance.getattr()` and `ClassDef.getattr()` skips `Unknown` placeholders and continues searching
- Builtin dunders (from the builtins module) return `Uninferable` instead of raising `InferenceError` since we can't infer their result without executing C code
- Special case for `__hash__ = None` to ensure unhashable types (list, dict, set) properly override object's `__hash__`
- Enhanced `type.__new__()` validation to raise descriptive `InferenceError` instead of silently returning `None`

 How it works:
- When a class overrides a dunder: lookup finds the actual implementation and returns the bound method
- When a class doesn't override: lookup finds the `Unknown` placeholder, skips it, continues searching, eventually returns the placeholder which yields `Uninferable` when called
- For unhashable builtin types (list, dict, set): `__hash__` is explicitly set to `None` in locals.

Thanks for any feedback 🙌

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1618,6 +1618,17 @@ def infer_call_result(
...
+            # Builtin dunder methods have empty bodies, return Uninferable.
+            if (
+                self.root().qname() == "builtins"
+                and self.name.startswith("__")
+                and self.name.endswith("__")
+                and self.parent
+                and self.parent.__class__.__name__ == "ClassDef"
+            ):
+                yield util.Uninferable
+                return
```
DanielNoord, line 1629: I think I sort of understand this change, but don't understand why we need to special case it like this.

Can we replace the checks with a look up in `special_attributes`? Or doesn't that work?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2346,8 +2357,12 @@ def getattr(
...
 
         if name in self.special_attributes and class_context and not values:
-            result = [self.special_attributes.lookup(name)]
-            return result
+            special_attr = self.special_attributes.lookup(name)
+            if not isinstance(
+                special_attr, (util.UninferableBase, node_classes.Unknown)
+            ):
+                result = [special_attr]
+                return result
```
DanielNoord, line 2362: What is the effect of this? What will we eventually return if the `if` is not `True`?

astroid/bases.py
```diff
@@ -250,7 +250,9 @@ def getattr(
             values = self._proxied.instance_attr(name, context)
         except AttributeInferenceError as exc:
             if self.special_attributes and name in self.special_attributes:
-                return [self.special_attributes.lookup(name)]
+                special_attr = self.special_attributes.lookup(name)
```
DanielNoord, line 253: Similar question as above

astroid/bases.py
```diff
@@ -571,10 +573,14 @@ def _infer_type_new_call(
             raise InferenceError(context=context) from e
         if not isinstance(mcs, nodes.ClassDef):
             # Not a valid first argument.
-            return None
+            raise InferenceError(
```
DanielNoord, line 576: This feels like it could be merged on its own, is that True? In that case I'd probably split this out to ensure we don't do too much in a single PR.

astroid/raw_building.py
```diff
@@ -78,7 +78,10 @@ def attach_const_node(node, name: str, value) -> None:
     """create a Const node and register it in the locals of the given
     node with the specified name
     """
-    if name not in node.special_attributes:
+    # Special case: __hash__ = None overrides ObjectModel for unhashable types.
+    if name == "__hash__" and value is None:
+        _attach_local_node(node, nodes.const_factory(value), name)
```
DanielNoord, line 84: Should this be the case for all attributes where `value is None`? Or only `hash`?

astroid/raw_building.py
```diff
@@ -507,7 +510,10 @@ def object_build(
             elif inspect.isdatadescriptor(member):
                 child = object_build_datadescriptor(node, member)
             elif isinstance(member, tuple(node_classes.CONST_CLS)):
-                if alias in node.special_attributes:
+                # Special case: __hash__ = None overrides ObjectModel for unhashable types.
```
DanielNoord, line 514: Same question as above

tests/test_inference.py
```diff
@@ -5618,8 +5618,9 @@ def test_cannot_infer_call_result_for_builtin_methods() -> None:
     )
     inferred = next(node.infer())
     lenmeth = next(inferred.igetattr("__len__"))
-    with pytest.raises(InferenceError):
-        next(lenmeth.infer_call_result(None, None))
+    # Builtin dunder methods now return Uninferable instead of raising InferenceError
+    result = next(lenmeth.infer_call_result(None, None))
+    assert result is util.Uninferable
```
DanielNoord, line 5623: Ah, I think the changes above now make more sense to me.

`pylint` works correctly with this change? Thanks for testing that.

tests/test_object_model.py
```diff
@@ -897,3 +906,112 @@ class Apple(TypedDict):
...
+        class MyClass:
+            def __eq__(self, other):
+                return "custom equality"
+
+        MyClass().__eq__(None)  #@
+        """
+    )
+    inferred = next(eq_result.infer())
+    assert isinstance(inferred, nodes.Const)
+    assert inferred.value == "custom equality"
```
DanielNoord, line 925: Nice!

DanielNoord (commented):
This is amazing, very well written PR and a nice set of tests.

Well done on getting this to pass all tests. I have left some comments, but would really like to help you push this over the line :)

astroid/nodes/scoped_nodes/scoped_nodes.py
emmanuel-ferdman: We need this case because `infer_call_result()` has to decide what a call returns, and for built-in dunder methods there is no Python body to inspect (since they're implemented in C). If we tried to parse `return` nodes we would either fail or produce incorrect results. 

`special_attributes` only tells us that an attribute exists (a placeholder), not whether there is a Python body to infer a return value from. Therefore it cannot replace the empty-body/builtins check used here.

But on second thought, I simplified the case by explicitly checking `len(self.body) == 0`.

astroid/nodes/scoped_nodes/scoped_nodes.py
emmanuel-ferdman: This block is a short-circuit used when there are no concrete locals/ancestor definitions: originally it always returned `special_attributes.lookup(name)` (even if that was `Unknown` or `Uninferable`). The new behavior only returns the `special_attr` when it is a concrete value (not `node_classes.Unknown` or `util.UninferableBase`), preventing a placeholder from being returned prematurely and masking an override in a metaclass/base class. If the `if` is not true we continue the normal lookup (metaclass lookup, collect/filter locals/ancestors) and ultimately return any real definitions found - otherwise an `AttributeInferenceError` is raised. Placeholders (`Unknown`/`Uninferable`) therefore mean “keep looking,” not “return this as the final result.”

tests/test_inference.py
emmanuel-ferdman: Yes - `pylint` works correctly (verified by running `pylint` test suite with astroid changes). `pylint` handles `Uninferable` in two ways: (1) explicit `isinstance(value, util.UninferableBase)` checks (e.g., [`pylint/checkers/typecheck.py:1365-1369`](https://github.com/pylint-dev/pylint/blob/5f0298dcf642742e8980266f6e243c760ffd0865/pylint/checkers/typecheck.py#L1364-L1369), and (2) `Uninferable` is falsy like `None`, so if inferred: checks skip both. The behavior is identical - when builtin dunders can't be inferred, pylint skips type checking whether it gets `None` from an exception or `Uninferable` from the return value.

astroid/raw_building.py
emmanuel-ferdman: I believe only for `__hash__`. [In Python `__hash__ = None` has a special semantic meaning](https://docs.python.org/3/reference/datamodel.html#object.__hash__) - it explicitly marks a type as unhashable and must override the inherited `object.__hash__`. Other attributes set to `None` do not carry this override semantics and should not implicitly replace `special_attributes` entries, because that could hide real implementations or metadata. So I belive we should have this special-case restricted to `__hash__` only.

astroid/bases.py
emmanuel-ferdman: This change is independent of the object dunder work. It's about improving `type.new()` validation error messages - changing silent `None` returns to descriptive `InferenceError` exceptions. The existing tests already expect `InferenceError`, so it's fully compatible. I'm happy to split it into a separate PR to keep this one focused on the object dunder changes. Should I remove it from this PR and create a new one, or would you prefer I keep it here as a small related improvement?

astroid/bases.py
emmanuel-ferdman: Same pattern as the `ClassDef.getattr()` check - when `special_attr` is `Unknown` or `Uninferable`, we skip the early return and continue to the if lookupclass: block which searches the class for the attribute. This ensures we find actual implementations in classes rather than stopping at the `Unknown` placeholder.

astroid/raw_building.py
emmanuel-ferdman: Same reasoning as the previous hash question - `hash = None` is the only attribute where `None` has special semantic meaning in Python (marks unhashable types). Other `None` values don't need this override behavior.

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Wouldn't it be okay if we always did this for `len(self.body) == 0`? Is there a function where `len(self.body) == 0` that is inferable? Or is as `...` or `pass` as body also `len(self.body) == 0`?

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: `Unkown` feels like a nice placeholder for "keep looking", `Uninferable` generally means "stop inferring, you won't be able to". Can we make this only check for `Unknown`? Or does that not work?

astroid/bases.py
DanielNoord: Please split it! I'll make sure to review it asap, but generally I and other maintainers need quite a lot of mental energy to review `astroid` as it all quite complex. Generallty, the smaller the PR the easier I can find that energy.

astroid/raw_building.py
DanielNoord: Perhaps we can link that documentation here? It would prevent me from wondering why this special case is here the next time.


And thanks for the link, TIL about something I knew implicitly but now know where to find the documentation for that feature 😄 Appreciated!

astroid/nodes/scoped_nodes/scoped_nodes.py
emmanuel-ferdman: Yes, good idea! `ObjectModel` only returns `Unknown` as placeholders, never `Uninferable`. I changed both checks (in `ClassDef.getattr()` and `BaseInstance.getattr()`) to only check for `Unknown`. If `Uninferable` somehow appeared from `special_attributes`, we should treat it as a final value and return it, not skip it. All tests pass with this change.

astroid/raw_building.py
emmanuel-ferdman: Good idea - added the documentation link to both places 😄

astroid/nodes/scoped_nodes/scoped_nodes.py
emmanuel-ferdman: I don't think we can't simplify to just `len(self.body) == 0` because both builtin dunders (`__eq__`, `__hash__`) and non-dunder builtins (`getattr`, `len`, `print`) have empty bodies. If we return `Uninferable` for all empty-body functions, it breaks tests - specifically `getattr(int, 'missing_attr')` returns `Uninferable` instead of raising `InferenceError` (test `test_attribute_missing` fails). Regarding `pass/...` - they create `len(body) = 1`, not 0. Only C builtins have truly empty bodies.

What I think I can simplify is to drop the `self.root().qname() == "builtins"` check since `len(body) == 0` already guarantees it's a `builtin`. All tests pass (both in astroid and pylint) with this simplification. However, I'm not 100% certain there isn't some corner case we're missing with that removal. Let me know if you prefer me to drop this check 😄

DanielNoord: Hmm, I think it is indeed likely that we should only do this for `builtins`. Let's keep as is, thanks for the extensive research!

astroid/nodes/scoped_nodes/scoped_nodes.py
DanielNoord: Perhaps we should update the docstring of `Unknown` to reflect this?

jacobtylerwalls:
<details>
<summary>Full details</summary>

```
First, please verify that the bug is not already filled:
https://github.com/pylint-dev/pylint/issues/

Then create a new issue:
https://github.com/pylint-dev/pylint/issues/new?labels=Crash 💥%2CNeeds triage 📥



Issue title:
Crash ```` (if possible, be more specific about what made pylint crash)

### Bug description

When parsing the following ``a.py``:

<!--
 If sharing the code is not an option, please state so,
 but providing only the stacktrace would still be helpful.
 -->

```python
# Copyright 2009 Brian Quinlan. All Rights Reserved.
# Licensed to PSF under a Contributor Agreement.

"""Implements ThreadPoolExecutor."""

__author__ = 'Brian Quinlan (brian@sweetapp.com)'

from concurrent.futures import _base
import itertools
import queue
import threading
import types
import weakref
import os


_threads_queues = weakref.WeakKeyDictionary()
_shutdown = False
# Lock that ensures that new workers are not created while the interpreter is
# shutting down. Must be held while mutating _threads_queues and _shutdown.
_global_shutdown_lock = threading.Lock()

def _python_exit():
    global _shutdown
    with _global_shutdown_lock:
        _shutdown = True
    items = list(_threads_queues.items())
    for t, q in items:
        q.put(None)
    for t, q in items:
        t.join()

# Register for `_python_exit()` to be called just before joining all
# non-daemon threads. This is used instead of `atexit.register()` for
# compatibility with subinterpreters, which no longer support daemon threads.
# See bpo-39812 for context.
threading._register_atexit(_python_exit)

# At fork, reinitialize the `_global_shutdown_lock` lock in the child process
if hasattr(os, 'register_at_fork'):
    os.register_at_fork(before=_global_shutdown_lock.acquire,
                        after_in_child=_global_shutdown_lock._at_fork_reinit,
                        after_in_parent=_global_shutdown_lock.release)
    os.register_at_fork(after_in_child=_threads_queues.clear)


class WorkerContext:

    @classmethod
    def prepare(cls, initializer, initargs):
        if initializer is not None:
            if not callable(initializer):
                raise TypeError("initializer must be a callable")
        def create_context():
            return cls(initializer, initargs)
        def resolve_task(fn, args, kwargs):
            return (fn, args, kwargs)
        return create_context, resolve_task

    def __init__(self, initializer, initargs):
        self.initializer = initializer
        self.initargs = initargs

    def initialize(self):
        if self.initializer is not None:
            self.initializer(*self.initargs)

    def finalize(self):
        pass

    def run(self, task):
        fn, args, kwargs = task
        return fn(*args, **kwargs)


class _WorkItem:
    def __init__(self, future, task):
        self.future = future
        self.task = task

    def run(self, ctx):
        if not self.future.set_running_or_notify_cancel():
            return

        try:
            result = ctx.run(self.task)
        except BaseException as exc:
            self.future.set_exception(exc)
            # Break a reference cycle with the exception 'exc'
            self = None
        else:
            self.future.set_result(result)

    __class_getitem__ = classmethod(types.GenericAlias)


def _worker(executor_reference, ctx, work_queue):
    try:
        ctx.initialize()
    except BaseException:
        _base.LOGGER.critical('Exception in initializer:', exc_info=True)
        executor = executor_reference()
        if executor is not None:
            executor._initializer_failed()
        return
    try:
        while True:
            try:
                work_item = work_queue.get_nowait()
            except queue.Empty:
                # attempt to increment idle count if queue is empty
                executor = executor_reference()
                if executor is not None:
                    executor._idle_semaphore.release()
                del executor
                work_item = work_queue.get(block=True)

            if work_item is not None:
                work_item.run(ctx)
                # Delete references to object. See GH-60488
                del work_item
                continue

            executor = executor_reference()
            # Exit if:
            #   - The interpreter is shutting down OR
            #   - The executor that owns the worker has been collected OR
            #   - The executor that owns the worker has been shutdown.
            if _shutdown or executor is None or executor._shutdown:
                # Flag the executor as shutting down as early as possible if it
                # is not gc-ed yet.
                if executor is not None:
                    executor._shutdown = True
                # Notice other workers
                work_queue.put(None)
                return
            del executor
    except BaseException:
        _base.LOGGER.critical('Exception in worker', exc_info=True)
    finally:
        ctx.finalize()


class BrokenThreadPool(_base.BrokenExecutor):
    """
    Raised when a worker thread in a ThreadPoolExecutor failed initializing.
    """


class ThreadPoolExecutor(_base.Executor):

    BROKEN = BrokenThreadPool

    # Used to assign unique thread names when thread_name_prefix is not supplied.
    _counter = itertools.count().__next__

    @classmethod
    def prepare_context(cls, initializer, initargs):
        return WorkerContext.prepare(initializer, initargs)

    def __init__(self, max_workers=None, thread_name_prefix='',
                 initializer=None, initargs=(), **ctxkwargs):
        """Initializes a new ThreadPoolExecutor instance.

        Args:
            max_workers: The maximum number of threads that can be used to
                execute the given calls.
            thread_name_prefix: An optional name prefix to give our threads.
            initializer: A callable used to initialize worker threads.
            initargs: A tuple of arguments to pass to the initializer.
            ctxkwargs: Additional arguments to cls.prepare_context().
        """
        if max_workers is None:
            # ThreadPoolExecutor is often used to:
            # * CPU bound task which releases GIL
            # * I/O bound task (which releases GIL, of course)
            #
            # We use process_cpu_count + 4 for both types of tasks.
            # But we limit it to 32 to avoid consuming surprisingly large resource
            # on many core machine.
            max_workers = min(32, (os.process_cpu_count() or 1) + 4)
        if max_workers <= 0:
            raise ValueError("max_workers must be greater than 0")

        (self._create_worker_context,
         self._resolve_work_item_task,
         ) = type(self).prepare_context(initializer, initargs, **ctxkwargs)

        self._max_workers = max_workers
        self._work_queue = queue.SimpleQueue()
        self._idle_semaphore = threading.Semaphore(0)
        self._threads = set()
        self._broken = False
        self._shutdown = False
        self._shutdown_lock = threading.Lock()
        self._thread_name_prefix = (thread_name_prefix or
                                    ("ThreadPoolExecutor-%d" % self._counter()))

    def submit(self, fn, /, *args, **kwargs):
        with self._shutdown_lock, _global_shutdown_lock:
            if self._broken:
                raise self.BROKEN(self._broken)

            if self._shutdown:
                raise RuntimeError('cannot schedule new futures after shutdown')
            if _shutdown:
                raise RuntimeError('cannot schedule new futures after '
                                   'interpreter shutdown')

            f = _base.Future()
            task = self._resolve_work_item_task(fn, args, kwargs)
            w = _WorkItem(f, task)

            self._work_queue.put(w)
            self._adjust_thread_count()
            return f
    submit.__doc__ = _base.Executor.submit.__doc__

    def _adjust_thread_count(self):
        # if idle threads are available, don't spin new threads
        if self._idle_semaphore.acquire(timeout=0):
            return

        # When the executor gets lost, the weakref callback will wake up
        # the worker threads.
        def weakref_cb(_, q=self._work_queue):
            q.put(None)

        num_threads = len(self._threads)
        if num_threads < self._max_workers:
            thread_name = '%s_%d' % (self._thread_name_prefix or self,
                                     num_threads)
            t = threading.Thread(name=thread_name, target=_worker,
                                 args=(weakref.ref(self, weakref_cb),
                                       self._create_worker_context(),
                                       self._work_queue))
            t.start()
            self._threads.add(t)
            _threads_queues[t] = self._work_queue

    def _initializer_failed(self):
        with self._shutdown_lock:
            self._broken = ('A thread initializer failed, the thread pool '
                            'is not usable anymore')
            # Drain work queue and mark pending futures failed
            while True:
                try:
                    work_item = self._work_queue.get_nowait()
                except queue.Empty:
                    break
                if work_item is not None:
                    work_item.future.set_exception(self.BROKEN(self._broken))

    def shutdown(self, wait=True, *, cancel_futures=False):
        with self._shutdown_lock:
            self._shutdown = True
            if cancel_futures:
                # Drain all work items from the queue, and then cancel their
                # associated futures.
                while True:
                    try:
                        work_item = self._work_queue.get_nowait()
                    except queue.Empty:
                        break
                    if work_item is not None:
                        work_item.future.cancel()

            # Send a wake-up to prevent threads calling
            # _work_queue.get(block=True) from permanently blocking.
            self._work_queue.put(None)
        if wait:
            for t in self._threads:
                t.join()
    shutdown.__doc__ = _base.Executor.shutdown.__doc__

```

### Command used

```shell
pylint a.py
```

### Pylint output

<details open>
    <summary>
        pylint crashed with a ``AstroidError`` and with the following stacktrace:
    </summary>

```python
  File "/Users/jwalls/pylint/pylint/lint/pylinter.py", line 845, in _lint_file
    check_astroid_module(module)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/jwalls/pylint/pylint/lint/pylinter.py", line 1077, in check_astroid_module
    retval = self._check_astroid_module(
        ast_node, walker, rawcheckers, tokencheckers
    )
  File "/Users/jwalls/pylint/pylint/lint/pylinter.py", line 1129, in _check_astroid_module
    walker.walk(node)
    ~~~~~~~~~~~^^^^^^
  File "/Users/jwalls/pylint/pylint/utils/ast_walker.py", line 90, in walk
    self.walk(child)
    ~~~~~~~~~^^^^^^^
  File "/Users/jwalls/pylint/pylint/utils/ast_walker.py", line 90, in walk
    self.walk(child)
    ~~~~~~~~~^^^^^^^
  File "/Users/jwalls/pylint/pylint/utils/ast_walker.py", line 87, in walk
    callback(astroid)
    ~~~~~~~~^^^^^^^^^
  File "/Users/jwalls/pylint/pylint/extensions/redefined_variable_type.py", line 103, in visit_assign
    (node, _type.pytype())
           ^^^^^^^^^^^^
AttributeError: 'Unknown' object has no attribute 'pytype'

  File "/Users/jwalls/pylint/pylint/lint/pylinter.py", line 808, in _lint_files
    self._lint_file(fileitem, module, check_astroid_module)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jwalls/pylint/pylint/lint/pylinter.py", line 847, in _lint_file
    raise astroid.AstroidError from e

```


</details>

### Expected behavior

No crash.

### Pylint version

```shell
pylint 4.1.0-dev0
astroid 4.1.0-dev0
Python 3.14.3 (v3.14.3:323c59a5e34, Feb  3 2026, 11:41:37) [Clang 16.0.0 (clang-1600.0.26.6)]
```

### OS / Environment

darwin (Darwin)

### Additional dependencies


```
</details>

PR #2851, opened 2025-10-07
Make `type.__new__()` raise clear errors instead of returning `None`

emmanuel-ferdman:
## Type of Changes

## Description

This PR improves error reporting for `type.__new__()` validation by replacing silent `None` returns with descriptive `InferenceError` exceptions. Previously, when `type.__new__()` received invalid arguments (wrong metaclass type, invalid name, bad bases, etc.), the code would return `None` without explanation. Now it raises `InferenceError` with clear messages like `"type.new() requires a class for metaclass"` or `"type.new() requires 4 arguments, got 3"`. This makes debugging easier and brings the validation in line with other astroid error handling patterns. All existing tests pass without modification since they already expected `InferenceError` for these cases.

This is part of the on-going work in PR #2847.

### DIFF
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -571,10 +571,14 @@ class BoundMethod(UnboundMethod):
             raise InferenceError(context=context) from e
         if not isinstance(mcs, nodes.ClassDef):
             # Not a valid first argument.
-            return None
+            raise InferenceError(
+                "type.__new__() requires a class for metaclass", context=context
+            )
         if not mcs.is_subtype_of("builtins.type"):
             # Not a valid metaclass.
-            return None
+            raise InferenceError(
+                "type.__new__() metaclass must be a subclass of type", context=context
+            )
 
         # Verify the name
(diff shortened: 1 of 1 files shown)


emmanuel-ferdman:
@DanielNoord The existing tests (`test_type__new__invalid_mcs_argument`, `test_type__new__invalid_name`, `test_type__new__invalid_bases`) cover these changes. While we added descriptive error messages in the code for developer clarity, they (at least currently) don't propagate to end users because `Call._infer()` catches all `InferenceError` exceptions. Even `pylint` currently [siliently ignores](https://github.com/pylint-dev/pylint/blob/5f0298dcf642742e8980266f6e243c760ffd0865/pylint/checkers/base/basic_checker.py#L371) the `InferenceError` error. The important improvement is that we now raise errors immediately instead of returning `None` and falling through to potentially incorrect inference paths.

I have added the changelog, hopefully in the correct place 😄

PR #2863, opened 2025-10-14
Refactor Qt brain transforms to handle PyQt and PySide

emmanuel-ferdman:
## Type of Changes

## Description

This PR rewrites `astroid/brain/brain_qt.py` to correctly infer Qt signals for both PyQt and PySide, regardless of how the binding exposes them: Function-style signals (descriptor) or Class-style signals. Main changes:
- Reorganized `astroid/brain/brain_qt.py` to share explicit PyQt/PySide signal templates, restoring `connect`/`disconnect`/`emit` inference for both toolkits. I believe it also improves readability of code.
- Handle function-style vs class-style signals: Some bindings (e.g., PySide ≥ 5.15.2 / PySide6) expose `Signal` via the descriptor protocol, so it infers as a `FunctionDef`. The transform now catches that path and attaches the expected members. See Example 1 (subclassed `QTimer.timeout`) and Example 2 (class-style `Signal()` instance) below. Other bindings (e.g., PyQt5/6) expose `pyqtSignal` as a class, so it infers as a `ClassDef`. We register a matching class transform there as well.
- Enabled the PyQt6 brain tests across every supported Python version (Previously it was only for 3.10).

Please note that I tried to add tests to `PySide6` but it looks like PyQt6 and PySide6 cannot coexist in a single Linux process because their wheels bundle different Qt 6 library revisions (similiary to concolution in work made in PR pylint-dev/astroid#1654). Due to this limiation, I added `pragma: no cover` for the `PySide` code branches.

Previously failing examples - Example 1:

```python
from PySide6.QtCore import QTimer


class CustomTimer(QTimer):
    def __init__(self) -> None:
        super().__init__()
        self.timeout.connect(self.on_timeout)

    def on_timeout(self) -> None:
        pass


def build_timer() -> CustomTimer:
    timer = CustomTimer()
    timer.timeout.disconnect()
    return timer
```

Used to emit:

```
input.py:8:8: E1101: Method '<no-name>' has no 'connect' member (no-member)
input.py:16:4: E1101: Method '<no-name>' has no 'disconnect' member (no-member)
```

Example 2:

```python
from PySide6.QtCore import Signal


sig = Signal()
sig.connect(lambda: None)
```

Used to emit:

```
input.py:5:0: E1101: Instance of 'Signal' has no 'connect' member (no-member)
```

The same pattern affected PyQt5 and PyQt6 as well (with their appropriate API/syntax).

Fixes pylint-dev/pylint-qt#1.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,7 +98,6 @@ jobs:
           allow-prereleases: true
           check-latest: true
       - name: Install Qt
-        if: ${{ matrix.python-version == '3.10' }}
         run: |
           sudo apt-get update
           sudo apt-get install build-essential libgl1-mesa-dev
--- a/tests/brain/test_qt.py
+++ b/tests/brain/test_qt.py
@@ -8,7 +8,6 @@ import pytest
 
 from astroid import Uninferable, extract_node
 from astroid.bases import UnboundMethod
-from astroid.const import PY312_PLUS
 from astroid.manager import AstroidManager
 from astroid.nodes import FunctionDef
 
@@ -16,8 +15,6 @@ HAS_PYQT6 = find_spec("PyQt6")
 
 
 @pytest.mark.skipif(HAS_PYQT6 is None, reason="These tests require the PyQt6 library.")
-# TODO: enable for Python 3.12 as soon as PyQt6 release is compatible
-@pytest.mark.skipif(PY312_PLUS, reason="This test was segfaulting with Python 3.12.")
 class TestBrainQt:
     AstroidManager.brain["extension_package_whitelist"] = {"PyQt6"}  # noqa: RUF012
 
@@ -73,3 +70,17 @@ class TestBrainQt:
             pytest.skip("PyQt6 C bindings may not be installed?")
         assert isinstance(attribute_node, FunctionDef)
         assert attribute_node.args.defaults
+
+    @staticmethod
+    def test_pyqt_signal_instance_connect_available() -> None:
+        """Test pyqtSignal() instances expose connect."""
+        src = """
+        from PyQt6.QtCore import pyqtSignal
+        sig = pyqtSignal()
+        sig.connect  #@
+        """
+        node = extract_node(src)
+        attribute_node = node.inferred()[0]
+        if attribute_node is Uninferable:
+            pytest.skip("PyQt6 C bindings may not be installed?")
+        assert isinstance(attribute_node, FunctionDef)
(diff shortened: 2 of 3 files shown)


tests/brain/test_qt.py
```diff
@@ -8,16 +8,13 @@
...
-from astroid.const import PY312_PLUS
 from astroid.manager import AstroidManager
 from astroid.nodes import FunctionDef
 
 HAS_PYQT6 = find_spec("PyQt6")
 
 
 @pytest.mark.skipif(HAS_PYQT6 is None, reason="These tests require the PyQt6 library.")
-# TODO: enable for Python 3.12 as soon as PyQt6 release is compatible
-@pytest.mark.skipif(PY312_PLUS, reason="This test was segfaulting with Python 3.12.")
```
Pierre-Sassoulas, line 20: 👍

.github/workflows/ci.yaml
```diff
@@ -98,7 +98,6 @@ jobs:
           allow-prereleases: true
           check-latest: true
       - name: Install Qt
-        if: ${{ matrix.python-version == '3.10' }}
         run: |
           sudo apt-get update
           sudo apt-get install build-essential libgl1-mesa-dev
```
Pierre-Sassoulas, line 103: Should we add a matrix for pyqt 5 / 6 ?

emmanuel-ferdman: @Pierre-Sassoulas I think I can add a matrix for PyQt5/6 if you think that’s the right approach - though they share most of the same code paths now. I could also try adding a PySide6 matrix, which might be more valuable since it goes through a different branch than PyQt, but it would require installing different libraries in the ci. 
Also, currently, we have PyQt6 in the requirements file. If we decide to go with a matrix, I’m not sure whether the right approach is to include both PyQt6 and PySide6 in the requirements or to install them dynamically during CI. Let me know what you prefer, and I’ll implement it.

Pierre-Sassoulas: Would you be open to create a proof of concept of astroid plugin ? I.e. a repo containing the pyqt brain that we add as an optional dependency to pylint or astroid later (*) and that we extensively test in it's own repo with as much dependency as required (pyqt from 1 to 6, anything is possible because the test won't slow down the main repo). If you are then the question on the other comment is moot because this is clearly the superior way to do thing, we're just limited by available maintainer time to do it.

(*) Not sure for this part the brain need astroid so if astroid has an optional dependency to it it's going to be a circular one ?

emmanuel-ferdman: @Pierre-Sassoulas Sure! I’d be happy to look into this. Are there any existing Astroid plugins I could review to better understand how this should be implemented?

Pierre-Sassoulas: It's a proof of concept so it would be the first astroid plugin. Let's discuss in https://github.com/pylint-dev/astroid/issues/1312 if required :)

PR #2880, opened 2025-11-07
Handle FunctionDef blockstart_tolineno edge cases

lowzhao:
## Type of Changes

## Description

Getting the lineno of the start of the block for function definition(`FunctionDef.blockstart_tolineno`) can be quite tricky. Take below example:

```python
# Case A
def foo(bar: str) -> None: 
    pass
# should returns line=1

# Case B
def foo(
        bar:str): 
    pass
# should returns line=2

# Case C
def foo(
    bar:str
) -> None: 
    pass
# should returns line=3

# Case D
def foo(
    bar:str
): 
# should returns line=3
    pass
```

Currently we only handled Case A, B. With this commit we can cover case C. 

But for Case D, we will need a better solution

### DIFF
--- a/tests/test_scoped_nodes.py
+++ b/tests/test_scoped_nodes.py
@@ -1020,7 +1020,6 @@ class FunctionNodeTest(ModuleLoader, unittest.TestCase):
         # Unimplemented, will return line 14 for now.
         # assert ast_nodes[3].blockstart_tolineno == 15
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
LGTM, just waiting for the coverage job to show up :)

Pierre-Sassoulas:
Great first contribution to astroid !

PR #2900, opened 2025-12-12
Improve self argument typing

cdce8p:
`typing.Self` was added in Python 3.11. It is available in `typing_extensions` since `v4.0.0`.
https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.Self

https://github.com/pylint-dev/astroid/blob/29667191d106ea7248160ba91b32652c0f4b6977/pyproject.toml#L35-L37


Pierre-Sassoulas (approved):
Great !

PR #2901, opened 2025-12-12
Fix decorator frame

cdce8p:
This partially fixes pylint-dev/pylint#8425. The additional changes should likely be done in pylint itself.

This does impact the test output for pylint a bit as errors on decorators no longer belong to the function / class now. Not sure that can be avoided though. In any case, I think it's better to do this in `4.1.0` and not in a bugfix release.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -55,8 +55,7 @@ else:
 if TYPE_CHECKING:
     from astroid import nodes
     from astroid.nodes import LocalsDictNodeNG
-
-    _FrameType = nodes.FunctionDef | nodes.Module | nodes.ClassDef | nodes.Lambda
+    from astroid.nodes.node_ng import FrameType
 
 
 def _is_const(value) -> bool:
@@ -2248,7 +2247,7 @@ class Decorators(NodeNG):
             raise ParentMissingError(target=self.parent)
         return self.parent.parent.scope()
 
-    def frame(self) -> _FrameType:
+    def frame(self) -> FrameType:
         """The first parent node defining a new frame."""
         # skip the function or class node to go directly to the upper level frame
         if not self.parent:
@@ -4938,7 +4937,7 @@ class NamedExpr(_base_nodes.AssignTypeNode):
     See astroid/protocols.py for actual implementation.
     """
 
-    def frame(self) -> nodes.FunctionDef | nodes.Module | nodes.ClassDef | nodes.Lambda:
+    def frame(self) -> FrameType:
         """The first parent frame node.
 
         A frame node is a :class:`Module`, :class:`FunctionDef`,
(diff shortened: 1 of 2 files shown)


tests/test_scoped_nodes.py
```diff
@@ -2909,6 +2909,36 @@ def method():
...
+        """
+        )
+        name_expr_node1, func_node, name_expr_node2, class_node = code
+        module = func_node.root()
+        assert name_expr_node1.scope() == module
+        assert name_expr_node1.frame() == module
+        assert name_expr_node2.scope() == module
+        assert name_expr_node2.frame() == module
+        assert module.locals.get("x") == [name_expr_node1.target]
+        assert module.locals.get("y") == [name_expr_node2.target]
```
cdce8p, line 2941: The `frame` is used for the scope lookup to determine where to assign the variables to. Without the changes in this PR, any variables get assigned to the function / class scope instead even though decorators are evaluated in the module scope.

--
A similar workaround might be required for function argument defaults. Though I haven't searched the issues tracker to see if anyone has complained yet.

astroid/nodes/node_classes.py
```diff
@@ -56,6 +56,8 @@
     from astroid import nodes
     from astroid.nodes import LocalsDictNodeNG
 
+    _FrameType = nodes.FunctionDef | nodes.Module | nodes.ClassDef | nodes.Lambda
```
jacobtylerwalls: Looks good -- looser than we really need for decorators (lambda?), but totally fine.

tests/test_scoped_nodes.py
```diff
@@ -2909,6 +2909,36 @@ def method():
...
+    def test_frame_node_for_decorators():
+        code = builder.extract_node(
+            """
+            def deco(var): ...
+
+            @deco(
+                x := 1  #@
+            )
+            def func():  #@
+                ...
```
jacobtylerwalls: This example raises a syntax error, unless I'm missing something:
```py
>>> def deco(var): ...
>>> @deco(
...     x := 1  #@
... )
... def func():  #@
...     ...
... 
    @deco(
     ~~~~^
        x := 1  #@
        ^^^^^^^^^^
    )
    ^
TypeError: 'NoneType' object is not callable
```

astroid/nodes/node_classes.py
cdce8p: > looser than we really need for decorators (lambda?)

It's actually necessary since the method returns `parent.frame()` which resolves to the base `NodeNG` implementation.
https://github.com/pylint-dev/astroid/blob/dbb5054594ddb34e86d91c0d269f3525f421886b/astroid/nodes/node_ng.py#L287-L288

tests/test_scoped_nodes.py
cdce8p: Only a `TypeError` at runtime. Astroid only fails for SyntaxErrors. Added a stub inner function to resolve it.

PR #2914, opened 2025-12-16
Wrong inference with default argument values

### LINKED ISSUE #2899: Wrong inference with default argument values
Got the unexpected result: default value of function argument is treated as the 'only' value.

Minimal example:
```python

bad, good = extract_node("""
def bad(foo: str | None = None):
    a = 1 if foo else ""  #@

def good(foo: str):
    a = 1 if foo else ""  #@
""")

print(bad.value.inferred())
print(good.value.inferred())
```

Result:
```python
[<Const.str l.3 at 0x27945e7eae0>]
[<Const.int l.6 at 0x27945e7ec90>, <Const.str l.6 at 0x27945e7ecc0>]
```

I expected the first `.inferred` to return two results. The second one is ok.

jkmnt:
This is a q&d draft PR fixing #2899: Wrong inference with default argument values.
I've added the same test as in reported issue. 

I'm not an expert on Astroid. Maybe it brokes something else. `pytest -k inference` is passing ok. 

Closes #2899


jacobtylerwalls (approved):
Thanks!

PR #2917, opened 2025-12-23
Add command line interface

cdce8p:
Especially for debugging pylint and astroid issues, it can be helpful to view the generated AST from astroid. There is already `NodeNG.repr_tree` but calling it requires having a script ready to read a source file putting that through `astroid.parse`.

This PR adds a simple command line interface to make it much easier. Just use
```bash
python -m astroid ast <file>
```

It uses subparsers so we can easily add new subcommand later if necessary.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ relative_files = true
 
 [report]
 omit =
+    */astroid/__main__.py
     */tests/*
     */tmp*/*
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
I had a nit, but strangely couldn't save it, then couldn't save the review either. Anyway great addition.

PR #2921, opened 2025-12-31
Multiline marker

### LINKED ISSUE #2566: Support for Multi-line Code Snippets with #@ Marker
### Steps to reproduce

1. Write a multi-line code snippet (e.g., a Keras Sequential model definition) and place the #@ marker at the end of the line.
2. Try splitting the code across multiple lines while keeping the #@ marker.
3. Run the checker that leverages the #@ marker to see how nodes are created.

### Current behavior
The code checker creates new nodes for each line when the #@ marker is used in multi-line code snippets, resulting in separate nodes rather than treating the code block as one unit. This reduces the clarity of scenarios and makes it challenging to handle more complex, multi-line code.

```
model = tf.keras.Sequential(layers=[
    tf.keras.layers.Dense(units=64, activation='relu'),
    tf.keras.layers.Dense(units=10)
])  #@
```

Additionally, having everything on a single line makes the code harder to read and maintain. For example, writing it all in one line like below is less readable and too long:

`model = tf.keras.Sequential(layers=[tf.keras.layers.Dense(units=64, activation='relu'), tf.keras.layers.Dense(units=10)])  #@
`

### Expected behavior
The code checker should treat multi-line code snippets marked with #@ as a single unit, allowing for improved readability and easier handling of more complex scenarios.

H-N-Chavda:
This PR fixes incorrect node extraction when using the `#@` marker on multi-line
statements in `astroid.builder.extract_node`.

Previously, when `#@` was placed at the end of a multi-line statement (for example,
a function call or object construction spanning multiple lines), the extractor
would match only the exact line number and could either return incomplete nodes
or create multiple nodes instead of treating the statement as a single unit.

This change updates the statement lookup logic to consider a node’s full
source span (`lineno` to `end_lineno`) when resolving `#@` markers. As a result,
multi-line statements are now correctly identified and extracted as one coherent
AST node.

The fix improves correctness for real-world code patterns where readability
requires breaking expressions across multiple lines, while preserving existing
behavior for single-line statements.

Closes #2566


ChangeLog
```diff
@@ -6,6 +6,8 @@ astroid's ChangeLog
 What's New in astroid 4.1.0?
 ============================
 Release date: TBA
```
DanielNoord, line 8:
```suggestion
Release date: TBA

```

test.sh
DanielNoord, line 1: Could you remove this file? :)

astroid/builder.py
```diff
@@ -370,15 +370,19 @@ def _find_statement_by_line(node: nodes.NodeNG, line: int) -> nodes.NodeNG | Non
     :rtype:  astroid.bases.NodeNG or None
     """
     if isinstance(node, (nodes.ClassDef, nodes.FunctionDef, nodes.MatchCase)):
-        # This is an inaccuracy in the AST: the nodes that can be
-        # decorated do not carry explicit information on which line
-        # the actual definition (class/def), but .fromline seems to
-        # be close enough.
-        node_line = node.fromlineno
+        start = node.fromlineno
+        end = getattr(node, "end_lineno", None)
```
DanielNoord, line 374: Why is `getattr` necessary here?

astroid/builder.py
```diff
@@ -370,15 +370,19 @@ def _find_statement_by_line(node: nodes.NodeNG, line: int) -> nodes.NodeNG | Non
...
     else:
-        node_line = node.lineno
+        start = node.lineno
+        end = getattr(node, "end_lineno", None)
```
DanielNoord, line 377: Same question

tests/test_builder.py
```diff
@@ -88,6 +88,47 @@ def test_callfunc_lineno(self) -> None:
...
+        assert isinstance(
+            node, nodes.Assign
+        ), f"Expected Assign node, got {type(node).__name__}"
+        value = node.value
+        assert isinstance(value, nodes.Call)
+        func = value.func
+        assert isinstance(func, nodes.Attribute)
+        assert func.attrname == "Sequential"
+        start = node.lineno
+        end = getattr(node, "end_lineno", None)
```
DanielNoord, line 115: Same question

DanielNoord:
Hmm never mind, the tests seems to fail pretty hard with this change. Perhaps you can have a look into why that is!

PR #2925, opened 2026-01-03
Bump astroid to 4.0.3, update changelog

### LINKED ISSUE #2899: Wrong inference with default argument values

jacobtylerwalls:
What's New in astroid 4.0.3?
============================
Release date: 2026-01-03

* Fix inference of ``IfExp`` (ternary expression) nodes to avoid prematurely narrowing
  results in the face of inference ambiguity.

  Closes #2899

* Fix base class inference for dataclasses using the PEP 695 typing syntax.

  Refs pylint-dev/pylint#10788

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,9 +9,15 @@ Release date: TBA
 
 
 
-What's New in astroid 4.0.3?
+What's New in astroid 4.0.4?
 ============================
 Release date: TBA
+
+
+
+What's New in astroid 4.0.3?
+============================
+Release date: 2026-01-03
 
 * Fix inference of ``IfExp`` (ternary expression) nodes to avoid prematurely narrowing
   results in the face of inference ambiguity.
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
LGTM ! (one of the issues listed is still open, might want to close it)

jacobtylerwalls:
My preliminary triage showed that https://github.com/pylint-dev/pylint/issues/10788 is still a bug.

PR #2927, opened 2026-01-04
Implement equality constraint

zenlyj:
## Type of Changes

## Description

Add a new `EqualityConstraint` to support narrowing of inferred values based on equality (`==`) and inequality (`!=`) comparisons.

The constraint currently handles:
- Comparison between constants
- Comparison between callables (e.g. classes, functions, lambdas)

Refs 
https://github.com/pylint-dev/pylint/issues/3632
https://github.com/pylint-dev/pylint/issues/3633

### DIFF
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -174,6 +174,62 @@ class TypeConstraint(Constraint):
             return True
 
 
+class EqualityConstraint(Constraint):
+    """Represents a "==" or "!=" constraint."""
+
+    def __init__(self, node: nodes.NodeNG, operand: nodes.NodeNG, negate: bool) -> None:
+        super().__init__(node=node, negate=negate)
+        self.operand = operand
+
+    @classmethod
+    def match(
+        cls, node: _NameNodes, expr: nodes.NodeNG, negate: bool = False
+    ) -> Self | None:
+        """Return a new constraint for node if expr matches one of these patterns:
+
+        - "node == operand" or "operand == node": use given negate value
+        - "node != operand" or "operand != node": flip negate value
+
+        Return None if no pattern matches.
+        """
+        if isinstance(expr, nodes.Compare) and len(expr.ops) == 1:
+            left = expr.left
+            op, right = expr.ops[0]
+            matches_left = _matches(left, node)
+
+            if op in {"==", "!="} and (matches_left or _matches(right, node)):
+                operand = right if matches_left else left
+                negate = (op == "==" and negate) or (op == "!=" and not negate)
+                return cls(node=node, operand=operand, negate=negate)
+
+        return None
+
+    def satisfied_by(self, inferred: InferenceResult) -> bool:
+        """Return True for uninferable/ambiguous results, or depending on negate flag:
+
+        - negate=False: satisfied when both operands are equal.
+        - negate=True: satisfied when both operands are not equal.
+
+        Only comparisons between constants and callables are supported.
+        """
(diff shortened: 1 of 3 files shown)


DanielNoord (commented):
@jacobtylerwalls Do you have time to review this? I consider you domain expert on the constraints.

It LGTM, but I might be missing something.

Thanks @zenlyj for this PR!

astroid/constraint.py
```diff
@@ -173,6 +173,62 @@ def satisfied_by(self, inferred: InferenceResult) -> bool:
...
+        if operand_inferred is util.Uninferable or operand_inferred is None:
+            return True
+
+        if isinstance(inferred, nodes.Const) and isinstance(
+            operand_inferred, nodes.Const
+        ):
+            return self.negate ^ (inferred.value == operand_inferred.value)
+
+        if inferred.callable() and operand_inferred.callable():
+            return self.negate ^ (inferred is operand_inferred)
```
jacobtylerwalls, line 228: Can you add a test similar to:
```py
x = Fraction(1, 3)
y = Fraction(1, 3)
if x == y:
  ...
```

I want to know whether it passes through this identity check.

jacobtylerwalls (commented):
Thanks 👍 

I was sitting on this one for a bit because I thought I might resuscitate my effort to adding a github workflow to run the pylint primer on astroid PRs, but I just didn't get there yet.

astroid/constraint.py
zenlyj: Done. In this case, `x` and `y` are neither constants nor callables, and should not be processed by this new constraint.

tests/test_constraint.py
```diff
@@ -999,3 +1003,127 @@ def test_isinstance_uninferable():
...
+
+    for node in (node1, node2):
+        inferred = node.inferred()
+        assert len(inferred) == 1
+        assert isinstance(inferred[0], nodes.Const)
+        assert inferred[0].value == 3
+
+
+def test_equality_fractions():
+    """Test that equality constraint is satisfied when both operands are fractions."""
```
jacobtylerwalls, line 1121: Doesn't the test show the equality constraint _not_ satisfied?

tests/test_constraint.py
```diff
@@ -999,3 +1003,127 @@ def test_isinstance_uninferable():
...
+    if x == y:
+        x  #@
+        y  #@
+
+    if x != y:
+        x  #@
+        y  #@
+    """)
+
+    for node in (node1, node2, node3, node4):
```
jacobtylerwalls, line 1137: Is there a pytest-idiomatic way to show the node that failed in the assertion?

jacobtylerwalls (commented):
Thanks 👍, just a teeny polish question.

tests/test_constraint.py
Pierre-Sassoulas: I usually add info about the node in the assert message when doing this

tests/test_constraint.py
jacobtylerwalls: Ah, sorry -- when there's ambiguity we say the constraint is satisfied, resolving!

tests/test_constraint.py
zenlyj: Good point, I've added some info to identify the node that fails. The message looks something like this:

image

Let me know if the message needs to be refined

jacobtylerwalls (approved):
Thanks!

PR #2928, opened 2026-01-08
Fix modutils._is_subpath to handle base with trailing os.sep

davidfstr:
## Type of Changes

## Description

Currently on Windows if PATH contains a UNC path like `r"\\Mac\Code\"` with a trailing slash, modutils.modpath_from_file won't be able to locate .py files like `r"\\Mac\Code\tests\test_resources.py"` inside such a path. Consequently, astroid will refuse to load the source code for .py files in such locations.

This change fixes the underlying modutils._is_subpath function used by modutils.modpath_from_file to handle paths with trailing os.sep correctly.

---

I would have preferred to update the ModPathFromFileTest test suite rather than creating a new IsSubpathTest suite, but ModPathFromFileTest is currently not passing on Windows.

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -238,9 +238,9 @@ def _is_subpath(path: str, base: str) -> bool:
     if not path.startswith(base):
         return False
     return (
-        (len(path) == len(base)) or 
-        (path[len(base)] == os.path.sep) or
-        (base.endswith(os.path.sep) and path[len(base) - 1] == os.path.sep)
+        (len(path) == len(base))
+        or (path[len(base)] == os.path.sep)
+        or (base.endswith(os.path.sep) and path[len(base) - 1] == os.path.sep)
     )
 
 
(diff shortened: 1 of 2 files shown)


davidfstr:
In case anyone else is impacted by this issue and needs a quick fix, here's what I'm doing:

```
def _patch_astroid() -> None:
    """
    Patches astroid 4.0.3 to work with PATH entries that end with os.path.sep.
    
    Backports the fix: https://github.com/pylint-dev/astroid/pull/2928
    """
    if sys.platform != 'win32':
        # Patch only useful on Windows
        return
    try:
        import astroid.modutils
    except ImportError:
        pass
    else:
        path = '\\\\Mac\\Code\\tests\\test_resources.py'
        base = '\\\\mac\\code\\'
        if not astroid.modutils._is_subpath(path, base):
            def _is_subpath2(path: str, base: str) -> bool:
                path = os.path.normcase(os.path.normpath(path))
                base = os.path.normcase(os.path.normpath(base))
                if not path.startswith(base):
                    return False
                return (
                    (len(path) == len(base)) or 
                    (path[len(base)] == os.path.sep) or
                    # Fix
                    (base.endswith(os.path.sep) and path[len(base) - 1] == os.path.sep)
                )
            astroid.modutils._is_subpath = _is_subpath2
        assert astroid.modutils._is_subpath(path, base)
```

PR #2930, opened 2026-01-09
Fix method classification for pygobject>=3.51.0

### LINKED ISSUE #2594: brain_gi generates incorrectly classifies some functions as methods
6 years ago, there was a bug report about pylint giving spurious errors for certain functions provided by the pygobject bindings:

https://gitlab.gnome.org/GNOME/pygobject/-/issues/217

A [workaround](https://gitlab.gnome.org/GNOME/pygobject/-/commit/5689b4ba2a2be95abc7f6439770191e90981ddf9) was committed that hid the problem by hiding the `FunctionInfo.__get__` method from Python while still making it act like a descriptor at the C level.

The next version of pygobject removes this workaround as it was causing some problems for other improvements I was making. So I decided to track down the root cause of the problem, which led me to the `astroid/brain/brain_gi.py` code.

It decides that anything that `inspect.ismethoddescriptor()` returns true for is a method and should be given the signature `(self, *args, **kwargs)`:

https://github.com/pylint-dev/astroid/blob/ba7df4acab0e985d1a6274def9f1798a9a7ce40d/astroid/brain/brain_gi.py#L134-L136

FunctionInfo objects meet this criteria since they have a `__get__` method but no `__set__` method. This signature is fine when the FunctionInfo object is being used to represent a method on a class, but is inappropriate when it represents a plain function at the module level. The fact that it has provides descriptor behaviour is irrelevant since it is stored in the module instance dict rather than being an attribute of the module type.

Using a `(*args, **kwargs)` signature would probably fix the bug. Alternatively, the next pygobject release will also support `inspect.signature()` on FunctionInfo objects, providing info about the actual function arguments.

### Steps to reproduce

1. Install current git version of pygobject
2. Run pylint on a program that calls a module level function with no arguments (such as `gi.repository.GLib.get_monotonic_time()`)

### Current behavior

See an E1120 error complaining about a missing "self" parameter.

### Expected behavior

The error should not be generated

3.2.4

### LINKED ISSUE #2929: Linting fails for GObject introspection bindings >=3.51.0
With the release of pygobjects 3.51.0, the workaround that made it possible to use `inspect.ismethoddescriptor()` to distinguish methods from functions has been disabled. Consequently, module-level functions are classified as methods and linting may fail for functions called without arguments due to the missing self argument.

Related to #2594

**Steps to reproduce**

- install pygobject >= 3.51.0
- run pylint on code that calls a module level function with no arguments

**Current behavior**

Calling those functions will wrongfully trigger
```
E1120: No value for argument 'self' in function call (no-value-for-parameter)
```

**Expected behavior**

The error should not be generated

oreiche:
## Type of Changes

## Description

The classification of functions and methods for pygobject version 3.51.0 and later leads to incorrect results. This is caused by the workaround that made it possible to use `inspect.ismethoddescriptor()` being disabled in newer versions. Instead, the newly supported `inspect.signature()` should be used.

Closes #2929, Closes #2594

### DIFF
--- a/astroid/brain/brain_gi.py
+++ b/astroid/brain/brain_gi.py
@@ -54,6 +54,35 @@ _special_methods = frozenset(
 )
 
 
+def _gi_supports_inspect_signature():
+    """
+    Indicates if pygobject supports inspect.signature().
+    """
+    import gi
+
+    try:
+        # inspect.signature() is supported since pygobject==3.51.0 (ee9558e4).
+        gi.check_version((3, 51, 0))
+        return True
+    except ValueError:
+        pass
+    return False
+
+
+def _gi_is_method_call(obj):
+    if _gi_supports_inspect_signature():
+        # Since inspect.signature() is supported, the workaround to use
+        # inspect.ismethoddescriptor() was disabled and cannot be used anymore
+        # to tell apart functions from methods.
+        # See https://github.com/pylint-dev/astroid/issues/2594
+        try:
+            sig = str(inspect.signature(obj))
+            return sig == "(self)" or sig.startswith("(self, ")
+        except Exception:  # pylint: disable=broad-except
+            return False
+    return inspect.ismethod(obj) or inspect.ismethoddescriptor(obj)
+
+
 def _gi_build_stub(parent):  # noqa: C901
     """
     Inspect the passed module recursively and build stubs for functions,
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (changes requested):
Thank you for opening a pull request. LGTM, but do you mind adding tests ? They should be skipped if pygoobject is not installed.

Pierre-Sassoulas (approved):

oreiche:
Thanks, rebased on latest main

Pierre-Sassoulas (approved):

PR #2936, opened 2026-01-14
Fix incorrect type inference for super() calls returning Self

### LINKED ISSUE #457: Incorrect super() inference for classmethods calling classmethods that return instances (constructors)
### Steps to reproduce

Run the following:

```python
n = extract_node('''
class A:

    @classmethod
    def new(cls):
        return cls()

class B(A):
    @classmethod
    def new(cls):
        inst = super().new()
        return inst

B.new()
''')
print(next(n.infer()))
```

### Current Behavior

Outputs `Instance of .A`

### Expected behavior

In Python, `B.new()` is an instance of B. However, the above prints `Instance of .A`, so this should print `Instance of .B`. Digging in, this seems to be due to the inferred type of the return of the call to `super().new()` from `B.new()` being an instance of A, rather than an instance of B.

This leads to bugs like pycqa/pylint#981, where the returns from overridden constructor methods are inferred to be the wrong type.

### Version Tested

1.5.3

### What I have done

I have managed to get the above code snippet to output `Instance of .B` by updating the following section of `objects.py`: https://github.com/PyCQA/astroid/blob/7bfb5e83b162d2126c9929b6d863724176269b6d/astroid/objects.py#L154-L167

I have updated it to read as follows, with the reasoning that B's classmethod is really calling A's new() method with its class being passed to `super()`, not its instance:

```python
                # We can obtain different descriptors from a super depending
                # on what we are accessing and where the super call is.
                if self._scope.type == 'classmethod' and inferred.type == 'classmethod':
                    yield bases.BoundMethod(inferred, self.type)
                elif inferred.type == 'classmethod':
                    yield bases.BoundMethod(inferred, cls)
                elif self._scope.type == 'classmethod' and inferred.type == 'method':
                    yield inferred
                elif self._class_based or inferred.type == 'staticmethod':
                    yield inferred
                elif bases._is_property(inferred):
                    # TODO: support other descriptors as well.
                    for value in inferred.infer_call_result(self, context):
                        yield value
                else:
                    yield bases.BoundMethod(inferred, cls)
```

This breaks one test in the test suite:

```
======================================================================
FAIL: test_super_bound_model (astroid.tests.unittest_objects.SuperTests)
----------------------------------------------------------------------
  File "/Users/Nautilus/Documents/Programming/astroid/astroid/tests/unittest_objects.py", line 285, in test_super_bound_model
    self.assertEqual(fourth.bound.name, 'First')
AssertionError: 'Super_Type_Type' != 'First'
- Super_Type_Type
+ First
```

Currently, I'm not certain if what I'm doing is correct. I intend to investigate further, to fix the broken test, and to add a test that illustrates the current pathological behavior. I don't know if I will have time this weekend (and I also don't really know what I'm doing!), so assistance is definitely welcome.

ywkim:
## Type of Changes

## Description

Closes #457

### Problem

When a child class calls `super().method()` and the method returns `cls()` or `self`, astroid incorrectly infers the parent class type instead of the child class type.

**Original reproduction from #457:**

```python
class A:
    @classmethod
    def new(cls):
        return cls()

class B(A):
    @classmethod
    def new(cls):
        inst = super().new()
        return inst

B.new()  # Incorrectly inferred as "Instance of A" instead of "Instance of B"
```

This causes pylint E1101 false positives in method chaining scenarios.

### Root Cause

In `Super.igetattr`, when creating a `BoundMethod`, the `bound` parameter is set to the parent class found in the MRO. Later, `BoundMethod.infer_call_result` uses `bind_context_to_node(context, self.bound)`, which causes `self` or `cls()` to be inferred as the parent type.

### Solution

Add an optional `original_caller` parameter to `BoundMethod` to track the actual caller (child instance/class) separate from the bound class (parent class from MRO). When inferring call results, use `original_caller` for context binding if available.

This preserves backward compatibility:
- `bound` attribute unchanged (existing tests pass)
- `original_caller` defaults to `None` (existing behavior for non-super calls)

### Changes

- `astroid/bases.py`: Add `original_caller` parameter to `BoundMethod.__init__` and use it in `infer_call_result`
- `astroid/objects.py`: Pass `self.type` as `original_caller` when creating `BoundMethod` from `Super.igetattr` (both instance methods and classmethods)
- `tests/test_objects.py`: Add regression tests
- `ChangeLog`: Add entry

### Verification

This fix resolves all cases mentioned in #457 and related issues:

| Case | Source | `main` | This PR |
|------|--------|--------|---------|
| classmethod `cls()` | [#457 original](https://github.com/pylint-dev/astroid/issues/457) | Instance of A ✗ | Instance of B ✓ |
| instance method `self.__class__()` | [#457 comment](https://github.com/pylint-dev/astroid/issues/457#issuecomment-1083712596) | Instance of A ✗ | Instance of B ✓ |
| `super().clone()` chaining | [pylint#6033](https://github.com/pylint-dev/pylint/issues/6033) | Instance of ParentClass ✗ | Instance of MyClass ✓ |
| classmethod with explicit super | [pylint#981](https://github.com/pylint-dev/pylint/issues/981) | Instance of Foo ✗ | Instance of Bar ✓ |
| `__deepcopy__` with super | [pylint#4639](https://github.com/pylint-dev/pylint/issues/4639) | Instance of ParentClass ✗ | Instance of ChildClass ✓ |

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,12 @@ Release date: TBA
 * Add simple command line interface for astroid to output generated AST.
   Use with ``python -m astroid``.
 
+* Fix incorrect type inference for ``super().method()`` calls that return ``Self``.
+  Previously, astroid would infer the parent class type instead of the child class type,
+  causing pylint E1101 false positives in method chaining scenarios.
+
+  Closes #457
+
 
 
 What's New in astroid 4.0.4?
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -541,9 +541,13 @@ class BoundMethod(UnboundMethod):
         self,
         proxy: nodes.FunctionDef | nodes.Lambda | UnboundMethod,
         bound: SuccessfulInferenceResult,
+        original_caller: SuccessfulInferenceResult | None = None,
     ) -> None:
         super().__init__(proxy)
         self.bound = bound
+        # For super() calls: the actual instance/class that called super()
+        # Used to correctly infer return type of methods returning Self
+        self.original_caller = original_caller
 
     def implicit_parameters(self) -> Literal[0, 1]:
         if self.name == "__new__":
@@ -674,7 +678,9 @@ class BoundMethod(UnboundMethod):
         caller: SuccessfulInferenceResult | None,
         context: InferenceContext | None = None,
     ) -> Iterator[InferenceResult]:
-        context = bind_context_to_node(context, self.bound)
+        # For super() calls, use original_caller to correctly infer Self return type
+        bound_node = self.original_caller if self.original_caller else self.bound
+        context = bind_context_to_node(context, bound_node)
         if (
             isinstance(self.bound, nodes.ClassDef)
             and self.bound.name == "type"
--- a/astroid/objects.py
+++ b/astroid/objects.py
@@ -194,7 +194,8 @@ class Super(node_classes.NodeNG):
                 # We can obtain different descriptors from a super depending
                 # on what we are accessing and where the super call is.
                 if inferred.type == "classmethod":
-                    yield bases.BoundMethod(inferred, cls)
+                    # Pass original caller for classmethod too
+                    yield bases.BoundMethod(inferred, cls, original_caller=self.type)
                 elif self._scope.type == "classmethod" and inferred.type == "method":
                     yield inferred
                 elif self._class_based or inferred.type == "staticmethod":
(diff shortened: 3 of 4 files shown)


Pierre-Sassoulas:
Thank you for this, it looks pretty promising. It seems like https://github.com/pylint-dev/pylint/issues/6033 is a duplicate of https://github.com/pylint-dev/pylint/issues/981 (old issue with high priority ❤️), do you mind checking/referencing that pylint issue too ?

DanielNoord (approved):
LGTM!

@Pierre-Sassoulas Shall we merge or do you want to review as well?

Pierre-Sassoulas (approved):
Just wanted to say great small fix to an issue that was open for a long time. We're probably going to have to check a lot of pylint issues to see if they were fixed by this.

PR #2937, opened 2026-01-14
Fix `random.sample` crash when cloning `ClassDef` and `FunctionDef` nodes

### LINKED ISSUE #2923: brain_random makes faulty assumption that all instance attributes for ClassDef can be passed to `__init__`
oss-fuzz (not public yet) found this:
```py
>>> klass = extract_node("class A: ...")
>>> klass
<ClassDef.A l.1 at 0x10449aa80>
>>> _clone_node_with_lineno(klass, None, 1)
  File "<python-input-15>", line 1, in <module>
    _clone_node_with_lineno(klass, None, 1)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/Users/<user>/astroid/astroid/brain/brain_random.py", line 35, in _clone_node_with_lineno
    new_node = cls(**init_params)
TypeError: ClassDef.__init__() got an unexpected keyword argument 'is_dataclass'
```

Not certain exactly how to trigger this from a public interface.

emmanuel-ferdman:
## Type of Changes

## Description

The `_clone_node_with_lineno` function assumed all `_other_fields` were valid `__init__` parameters, but fields like `is_dataclass` on `ClassDef` and position on `FunctionDef` are set inside `__init__`, not passed to it. This caused a crash when inferring for example `random.sample([dict] * 2, 1)`. The fix uses `inspect.signature` to determine which fields are actual `__init__` parameters and copies the rest via `setattr` after node creation.

Closes #2923

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -56,6 +56,10 @@ Release date: TBA
 
   Closes #2518
 
+* Fix ``random.sample`` crash when cloning ``ClassDef`` or ``FunctionDef`` nodes.
+
+  Closes #2923
+
 What's New in astroid 4.0.3?
 ============================
 Release date: 2026-01-03
--- a/astroid/brain/brain_random.py
+++ b/astroid/brain/brain_random.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import inspect
 import random
 
 from astroid import nodes
@@ -30,11 +31,19 @@ def _clone_node_with_lineno(node, parent, lineno):
         "end_col_offset": node.end_col_offset,
     }
     postinit_params = {param: getattr(node, param) for param in _astroid_fields}
-    if other_fields:
-        init_params.update({param: getattr(node, param) for param in other_fields})
+
+    valid_init_params = set(inspect.signature(cls.__init__).parameters)
+    for param in other_fields:
+        if param in valid_init_params:
+            init_params[param] = getattr(node, param)
+
     new_node = cls(**init_params)
     if hasattr(node, "postinit") and _astroid_fields:
         new_node.postinit(**postinit_params)
+
+    for param in other_fields:
+        if param not in valid_init_params:
+            setattr(new_node, param, getattr(node, param))
     return new_node
 
 
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -1108,6 +1108,23 @@ class RandomSampleTest(unittest.TestCase):
         assert isinstance(inferred.elts[0], nodes.ClassDef)
         assert inferred.elts[0].name == "dict"
 
+    def test_no_crash_on_functiondef_clone(self) -> None:
+        """Test that random.sample does not crash when cloning FunctionDef nodes.
+
+        Regression test for https://github.com/pylint-dev/astroid/issues/2923
+        """
+        node = astroid.extract_node(
+            """
+        import random
+        random.sample([len] * 2, 1)  #@
+        """
+        )
+        inferred = next(node.infer())
+        assert isinstance(inferred, nodes.List)
+        assert len(inferred.elts) == 1
+        assert isinstance(inferred.elts[0], nodes.FunctionDef)
+        assert inferred.elts[0].name == "len"
+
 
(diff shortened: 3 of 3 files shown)


Pierre-Sassoulas:
If it's an unreleased bug as I suspect no need for a changelog :)

Pierre-Sassoulas (dismissed):
Thank you for your work on random.sample 😄 LGTM !

ChangeLog
```diff
@@ -56,6 +56,10 @@ Release date: TBA
 
   Closes #2518
 
+* Fix ``random.sample`` crash when cloning ``ClassDef`` or ``FunctionDef`` nodes.
```
Pierre-Sassoulas, line 59: Is it both ClassDef and FunctionDef or only one ? The MR title only has ClassDef

emmanuel-ferdman: Yes, both are affected - Added a test for `FunctionDef` use-case and changed the title. Thank you!

Pierre-Sassoulas (approved):

astroid/brain/brain_random.py
```diff
@@ -30,11 +31,19 @@ def _clone_node_with_lineno(node, parent, lineno):
         "end_col_offset": node.end_col_offset,
     }
     postinit_params = {param: getattr(node, param) for param in _astroid_fields}
-    if other_fields:
-        init_params.update({param: getattr(node, param) for param in other_fields})
+
+    valid_init_params = set(inspect.signature(cls.__init__).parameters)
```
jacobtylerwalls, line 35: Is `inspect.signature` safe to call on Python 3.14 without `annotation_format=annotationlib.Format.FORWARDREF`?

tests/brain/test_brain.py
```diff
@@ -1091,6 +1091,40 @@ def test_no_crash_on_uninferable_element(self) -> None:
...
 
+    def test_no_crash_on_classdef_clone(self) -> None:
+        """Test that random.sample does not crash when cloning ClassDef nodes.
+
+        Regression test for https://github.com/pylint-dev/astroid/issues/2923
+        """
+        node = astroid.extract_node(
+            """
+        import random
+        random.sample([dict] * 2, 1)  #@
```
jacobtylerwalls, line 1102: A regression test with a class with stringified deferred annotations would be good, e.g.

```py
class MyClass:
    def __init__(my_param: "SomeClass"): ...
```

jacobtylerwalls (commented):
Thanks @emmanuel-ferdman 👍

I'm a little nervous about stringified annotations, could you check?

astroid/brain/brain_random.py
emmanuel-ferdman: Valid concern! From what I tested, it's safe for our specific use case. I tested this on Python 3.14.2 and confirmed the fix works. The reason is that all astroid node files (`scoped_nodes.py`, `node_classes.py`, etc.) use `from __future__ import annotations`, which stores annotations as strings rather than going through Python 3.14's new deferred evaluation. When we call `inspect.signature(ClassDef.__init__)`, the annotations are already pre-stringified, so no `NameError` can occur. That said, [pytest](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/compat.py#L68-L72) does use `annotation_format=Format.STRING` on Python 3.14+ as a defensive measure, and we could do the same if we want extra safety against edge cases like custom node subclasses that don't use the future import. If you think that's the correct approch, feel free to open an issue and assign me 😄

tests/brain/test_brain.py
emmanuel-ferdman: Thanks for the suggestion! I checked this scenario - a test with user code like you suggested wouldn't add coverage because we never call `inspect.signature` on the user's class. We only call it on astroid's internal node classes (`ClassDef.__init__`, `FunctionDef.__init__`). The user's forward reference annotations are just data stored in the node, they don't affect the cloning logic. The existing tests already cover the fix.

jacobtylerwalls: Wonderful. Resolving!

PR #2958, opened 2026-02-09
Fix `__doc__` on unbound method

DanielNoord:
## Type of Changes

## Description

Fixes the issue as described in https://github.com/pylint-dev/astroid/pull/2847

The reproducer for the crash is:

```python
class Executor:
    def submit(self): ...


class ThreadPoolExecutor(Executor):
    def submit(self): ...

    submit.__doc__ = Executor.submit.__doc__
```

That could be fixed in `pylint` (by not calling `pytype()` on `NodeNG` subclasses that don't have it), but that seems unnecessary.

Instead we can (I hope/think) better represent the actual object model by letting `UnboundMethodModel` inherit from `FunctionModel`. That feels like it would make sense? The test doesn't pass without this fix.

We could add `attr___doc__` on `UnboundMethodModel` but for me this feels like it better fits at least my own mental model of Python objects 😅

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -413,7 +413,7 @@ class FunctionModel(ObjectModel):
                     )
 
                 # The `func` can already be a Unbound or BoundMethod. If the former, make sure to
-                # wrap as a BoundMethod like we do below whe constructing the function from scratch.
+                # wrap as a BoundMethod like we do below when constructing the function from scratch.
                 if isinstance(func, bases.UnboundMethod):
                     yield bases.BoundMethod(proxy=func, bound=cls)
                     return
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -412,8 +412,12 @@ class FunctionModel(ObjectModel):
                         "Invalid class inferred", target=self, context=context
                     )
 
-                # For some reason func is a Node that the below
-                # code is not expecting
+                # The `func` can already be a Unbound or BoundMethod. If the former, make sure to
+                # wrap as a BoundMethod like we do below whe constructing the function from scratch.
+                if isinstance(func, bases.UnboundMethod):
+                    yield bases.BoundMethod(proxy=func, bound=cls)
+                    return
+
                 if isinstance(func, bases.BoundMethod):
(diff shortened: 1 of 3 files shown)


emmanuel-ferdman:
Hey @DanielNoord! I was just working on this same issue. I should've commented that but no worries 😄 .
I agree that `UnboundMethodModel` should extend `FunctionModel`. I would like to suggest this pr to also include fixes to two releated prolems that I think are still happening:
- ObjectModel-only dunders like `__eq__`, `__hash__`, `__repr__` still return `Unknown` instead of `Uninferable` because `FunctionModel` inherits from `ObjectModel` and doesn't override those.
```python
class A:
    def test(self): pass

A.test.__eq__  # Returns Unknown, not Uninferable
```
- `super().__doc__` has the same problem. 
```python
class Base:
    def method(self):
        return super().__doc__

Base().method()  # Returns Unknown, not Uninferable
```

I think the fix can be adding Unknown filtering to `UnboundMethod.getattr/igetattr` and `Super.igetattr` (same pattern PR #2847 used for `BaseInstance` and `ClassDef`) so Unknown placeholders fall through to the proxied object. I have [this implementation in a commit](https://github.com/emmanuel-ferdman/astroid/commit/377e096b05c4d783d05328c2b2a5346076139a63) on my fork branch - feel free to take the changes if they're relevant to your PR, or let me know if I should open a separate PR to address the additional cases.

astroid/interpreter/objectmodel.py
```diff
@@ -412,8 +412,12 @@ def infer_call_result(
                         "Invalid class inferred", target=self, context=context
                     )
 
-                # For some reason func is a Node that the below
-                # code is not expecting
+                # The `func` can already be a Unbound or BoundMethod. If the former, make sure to
+                # wrap as a BoundMethod like we do below whe constructing the function from scratch.
```
jacobtylerwalls: whe -> when

jacobtylerwalls (dismissed):
Looks good!

DanielNoord:
@jacobtylerwalls Unrelated to this PR, but this action failed: https://github.com/pylint-dev/astroid/actions/runs/22041554054

jacobtylerwalls (approved):

PR #2962, opened 2026-02-15
Add missing tag to test-pylint workflow

jacobtylerwalls:
https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow


DanielNoord (dismissed):
Change LGTM, but CI shows we need some more fixes 😅

jacobtylerwalls:
Now depends on pylint-dev/pylint#10850. If it's not too much churn, I suppose I could change this to point to that PR instead of pylint main.

DanielNoord (approved):

PR #2963, opened 2026-02-15
Prevent Unknown from leaking via special attribute lookup

emmanuel-ferdman:
## Type of Changes

## Description

PR #2847 added `Unknown` filtering to `BaseInstance.getattr()` and `ClassDef.getattr()` so that `Unknown` placeholders from `ObjectModel` don't leak out. But the same filtering was missing from `UnboundMethod.getattr/igetattr` and `Super.igetattr`.

ObjectModel-only dunders on unbound methods return `Unknown` instead of `Uninferable`:

```python
class A:
    def test(self): pass

A.test.__eq__   # Returns Unknown, should be Uninferable
A.test.__hash__ # Returns Unknown, should be Uninferable
A.test.__repr__ # Returns Unknown, should be Uninferable
```

Same problem with `super()`:

```python
class Base:
    def method(self):
        return super().__doc__

Base().method()  # Returns Unknown, should be Uninferable
```

cc @jacobtylerwalls @DanielNoord

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,9 @@ Release date: TBA
 * Let `UnboundMethodModel` inherit from `FunctionModel` to improve inference of
   dunder methods for unbound methods.
 
+* Filter ``Unknown`` from ``UnboundMethod`` and ``Super`` special attribute
+  lookup to prevent placeholder nodes from leaking during inference.
+
 What's New in astroid 4.1.0?
 ============================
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):
But please let @jacobtylerwalls review this as well :)

jacobtylerwalls (approved):

PR #2965, opened 2026-02-16
Return the ``FunctionDef`` node when a call is made to a decorated function.

### LINKED ISSUE #10848: pylint does not honor decorator return type when using @ syntax
### Bug description

```python
When a decorator changes the return type of a function, pylint correctly infers
the new type if the decorator is applied via explicit assignment
(`f = decorator(f_)`), but ignores the decorator's return type annotation when
applied via `@decorator` syntax. This causes a spurious `E1101 (no-member)` error.

Both `mypy` and `pyright` handle both forms correctly.

A complete, self-contained reproduction (including Dockerfile) is available as a gist:
https://gist.github.com/bnsh/e73e546469ec3d796fd22da0b7c9a31b

**`test_fail.py`** — uses `@type_changing_decorator` syntax (pylint fails):


from typing import Callable

def type_changing_decorator(func: Callable[[int], int]) -> Callable[[int], str]:
    """This is just a decorator to demonstrate the problem."""
    def wrapper(val: int) -> str:
        res = func(val)
        return f"the result is {res:d}"
    return wrapper

@type_changing_decorator
def add1(val: int) -> int:
    """This just adds 1 to its input."""
    return val+1

def main() -> None:
    """This is the main program."""
    res = add1(5)
    # This _works_, but pylint complains that int has no ".replace"
    print(res.replace("the", "a"))

if __name__ == "__main__":
    main()


**`test_succeed.py`** — uses explicit `add1 = type_changing_decorator(add1_)` (pylint succeeds):


from typing import Callable

def type_changing_decorator(func: Callable[[int], int]) -> Callable[[int], str]:
    """This is just a decorator to demonstrate the problem."""
    def wrapper(val: int) -> str:
        res = func(val)
        return f"the result is {res:d}"
    return wrapper

def add1_(val: int) -> int:
    """This just adds 1 to its input."""
    return val+1
add1 = type_changing_decorator(add1_)

def main() -> None:
    """This is the main program."""
    res = add1(5)
    # This _works_, and pylint accepts it, since I don't use
    # `type_changing_decorator` as a _decorator_.
    print(res.replace("the", "a"))

if __name__ == "__main__":
    main()
```

### Configuration

```ini
Default pylint configuration (no custom rcfile).
```

### Command used

```shell
pylint test_fail.py
```

### Pylint output

```python
************* Module test_fail
test_fail.py:27:10: E1101: Instance of 'int' has no 'replace' member (no-member)

-----------------------------------
Your code has been rated at 6.15/10


For comparison, `test_succeed.py` (identical logic, explicit call instead of `@` syntax):


Your code has been rated at 10.00/10


Both `mypy` and `pyright` pass cleanly on both files.
```

### Expected behavior

`pylint` should infer the return type of a decorated function from the decorator's
return type annotation. In this case, `type_changing_decorator` is annotated as
`(Callable[[int], int]) -> Callable[[int], str]`, so `add1` should be understood
as returning `str` regardless of whether the decorator is applied via `@` syntax
or explicit assignment.

### Pylint version

```shell
pylint 4.0.4
astroid 4.0.4
Python 3.14.3 (main, Feb  4 2026, 20:08:31) [GCC 14.2.0]
```

### OS / Environment

Debian GNU/Linux 13 (trixie) (python:3.14 Docker image)

### Additional dependencies

```python
None required to reproduce. The only installed packages are:


astroid==4.0.4
dill==0.4.1
isort==7.0.0
mccabe==0.7.0
platformdirs==4.5.1
pylint==4.0.4
tomlkit==0.14.0
```

mbyrnepr2:
## Type of Changes

## Description

Return the ``FunctionDef`` node when a call is made to a decorated function.

instead of returning the inferred ``FunctionDef`` node. This prevents ``no-member`` messages in Pylint when a decorated function is called and then is missing expected attributes; Pylint recognizes when a ``FunctionDef`` node has any decorators and suppresses the message (https://github.com/pylint-dev/pylint/blob/main/pylint/checkers/typecheck.py#L1133-L1139)

Closes pylint-dev/pylint#10848

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,14 @@ Release date: TBA
 * Let `UnboundMethodModel` inherit from `FunctionModel` to improve inference of
   dunder methods for unbound methods.
 
+* Return the ``FunctionDef`` node when a call is made to a decorated function
+  instead of returning the inferred ``FunctionDef`` node.
+  This prevents ``no-member`` messages in Pylint when a decorated function is
+  called and then is missing expected attributes; Pylint recognizes when a
+  ``FunctionDef`` node has any decorators and suppresses the message.
+
+  Closes pylint-dev/pylint#10848
+
 What's New in astroid 4.1.0?
 ============================
 Release date: 2026-02-08
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -1771,7 +1771,7 @@ class Call(NodeNG):
             callcontext.extra_context = self._populate_context_lookup(context.clone())
 
         for callee in self.func.infer(context):
-            if isinstance(callee, util.UninferableBase):
+            if isinstance(callee, util.UninferableBase) or callee.decorators:
                 yield callee
                 continue
             try:
--- a/tests/test_inference_calls.py
+++ b/tests/test_inference_calls.py
@@ -547,10 +547,10 @@ def test_chained_attribute_inherited() -> None:
 
 def test_decorated_function() -> None:
     """Test that a call to a function, which has a decorator,
-       returns the function itself.
+    returns the function itself.
 
-       Currently we do not consider the result of a function after it's
-       decorator is applied to it.
+    Currently we do not consider the result of a function after it's
+    decorator is applied to it.
     """
     node = builder.extract_node("""
     from typing import Callable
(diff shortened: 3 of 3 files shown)


DanielNoord:
@mbyrnepr2 Any interest in finsihing this?

mbyrnepr2:
@DanielNoord  if you'd like to commandeer the work please feel free.

DanielNoord:
No, I was mostly looking to close stale PRs in `astroid`. With the onslaught of AI PRs it is hard to keep up with which PRs require attention and which don't

PR #2968, opened 2026-02-19
Fix RecursionError in _compute_mro() on circular class hierarchies

worksbyfriday:
## Description

Fixes #2967
Closes pylint-dev/pylint#10821

When class inference encounters circular name bindings (e.g., a module-level name rebound to a subclass of itself), `_compute_mro()` recurses infinitely and hits Python's recursion limit.

### Reproducer

```python
from pdb import Pdb as StdlibPdb

class Pdb(StdlibPdb):
    pass

class PatchedPdb(Pdb):
    pass

if __name__ == '__main__':
    import pdb
    pdb.Pdb = PatchedPdb
```

### Root cause

`_compute_mro()` only guards against immediate self-loops (`if base is self: continue`) but not cycles through multiple classes. When inference resolves `pdb.Pdb` to `PatchedPdb` after the rebinding, the chain `PatchedPdb → Pdb → PatchedPdb` recurses without bound.

### Fix

Add a `_chain` set parameter to `_compute_mro()` that tracks which `ClassDef` nodes are currently being computed in the recursion chain. When a base class is found in the chain, `InconsistentMroError` is raised instead of recursing. The set uses add/discard around the computation to correctly handle diamond inheritance (where the same class appears in multiple branches but isn't a cycle).

All 20 existing MRO tests pass, plus a new test for the circular case. The broader test suite (902 tests across scoped_nodes, inference, helpers, objects, constraints) passes cleanly.

I'm the AI agent that filed #2967 — this is the fix DanielNoord suggested should live in astroid rather than pylint.

PR #2969, opened 2026-02-19
Fix inference cache ignoring kwargs like asname

worksbyfriday:
## Description

The inference cache key in `NodeNG.infer()` was `(node, lookupname, callcontext, boundnode)`, which did not include keyword arguments passed to `_infer()`. This caused cache collisions when the same node was inferred with different kwargs.

The concrete bug: `Import._infer()` accepts an `asname` parameter that fundamentally changes which module is resolved (`self.real_name(name)` vs `name` directly). When an aliased import like `import foo.bar as foo` was inferred first with `asname=True` (yielding `foo.bar`), the cached result was returned for a subsequent `asname=False` call (which should yield the `foo` package), producing incorrect inference results.

This caused pylint false positives like `no-name-in-module` when an import alias shadowed its parent package and a method name collided with a builtin (e.g., `.format()`). See [pylint#10193](https://github.com/pylint-dev/pylint/issues/10193) and the discussion on [pylint#10855](https://github.com/pylint-dev/pylint/pull/10855) where @DanielNoord identified this as an astroid-level issue.

## Changes

- **`astroid/nodes/node_ng.py`**: Include a frozen representation of `**kwargs` in the inference cache key, so calls with different keyword arguments are cached separately.
- **`astroid/context.py`**: Update the `_InferenceCache` type annotation to match the wider key tuple.
- **`tests/test_inference.py`**: Add regression test that verifies `asname=True` and `asname=False` on the same `Import` node produce different results (not cache-poisoned).

## Test

The new test (`test_import_as_infer_cache_with_different_asname`) demonstrates the bug:
1. Parse `import os.path as osp`
2. Infer with `asname=True` → correctly resolves to `os.path`
3. Infer with `asname=False` → should fail (no module named `osp`), but before the fix, returns cached `os.path`

All 438 existing inference tests pass. All 253 node/lookup tests pass.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,12 @@ Release date: TBA
 
   Closes #2762
 
+* Fix inference cache ignoring kwargs like ``asname``, causing incorrect
+  results when the same ``Import`` or ``ImportFrom`` node was inferred with
+  different ``asname`` values within the same context.
+
+  Closes pylint-dev/pylint#10193
+
 
 What's New in astroid 4.1.1?
 ============================
--- a/astroid/context.py
+++ b/astroid/context.py
@@ -17,7 +17,14 @@ if TYPE_CHECKING:
     from astroid import constraint, nodes
 
 _InferenceCache = dict[
-    tuple["nodes.NodeNG", str | None, str | None, str | None], Sequence["nodes.NodeNG"]
+    tuple[
+        "nodes.NodeNG",
+        str | None,
+        str | None,
+        str | None,
+        tuple[tuple[str, object], ...],
+    ],
+    Sequence["nodes.NodeNG"],
 ]
 
 _INFERENCE_CACHE: _InferenceCache = {}
--- a/astroid/nodes/node_ng.py
+++ b/astroid/nodes/node_ng.py
@@ -153,7 +153,8 @@ class NodeNG:
             except UseInferenceDefault:
                 pass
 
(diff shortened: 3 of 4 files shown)


DanielNoord (commented):
Add a changelog, make better tests, don't let stuff fail in CI.

@Fridayai700 Please fork the project you are working on and run the CI of each project in your fork. Don't make PRs that do not pass all CI steps.

DanielNoord (commented):
@jacobtylerwalls What do you think? The change itself seems sensible but I'm not a big fan of casting `kwargs` to `tuple`. Should `_infer` be allowed to return (an optional) cache key so that it is up to each Node to determine cache hit/miss?

worksbyfriday:
Rebased onto current main (now includes #2970 and #2972). Resolved the ChangeLog conflict — entry is now in the 4.1.2 section.

Regarding the `test-pylint-with-astroid-sha` failures: these are the same 4 import-order test failures that appear on #2968 (approved, awaiting merge) and on the main branch itself. They are not caused by this change.

On the architectural question about whether `_infer` should return an optional cache key instead of including `kwargs` in the tuple: I'm happy to implement whichever approach you prefer. The current approach (frozen `kwargs` in the cache key) is simpler and doesn't require changing the `_infer` interface, but the per-node cache key approach would be more flexible and let each node type decide what matters for caching. Let me know.

PR #2971, opened 2026-02-20
Fix ValueError in __str__/repr and error messages with extreme values

worksbyfriday:
## Description

Fix two related crashes where ValueError is raised during string formatting of nodes:

### 1. Long identifiers (#2764)

When a node has a very long name (e.g., a class with a 58+ character identifier), the pprint width calculation `80 - len(field) - alignment` can produce zero or negative values, causing `pprint.pformat()` to raise `ValueError('width must be != 0')`.

### 2. Large integer values (#2785)

When a node contains a very large integer value (e.g., `7**10000`), both `pprint.pformat()` and `str()` hit Python 3.11+'s integer-to-string conversion limit (4300 digits), raising `ValueError`. This affects `__str__` and error message formatting in `Const.getitem()`.

Both issues were discovered by OSS-Fuzz and crash pylint when it tries to format error messages for problematic nodes.

## Changes

- `astroid/nodes/node_ng.py` (`NodeNG.__str__`): `max(..., 1)` for width + `try/except ValueError`
- `astroid/interpreter/objectmodel.py` (`ObjectModel.__repr__`): same
- `astroid/context.py` (`InferenceContext.__str__`): same
- `astroid/nodes/node_classes.py` (`Const.getitem`): guard value stringification

## Reproducers

```python
# Issue #2764 — long identifier
class aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(c&()):
    _

# Issue #2785 — large integer
a, = 7 ** 10000
a.B
```

Closes #2764
Closes #2785

Issue #2785, opened 2025-07-22
ValueError when stringifying large int in AstroidTypeError message


correctmost:
### Steps to reproduce

The following code triggers a `ValueError` in astroid:

```python
a,=7**10000
a.B
```

This bug was discovered by OSS-Fuzz (an OSS-Fuzz bug report was not filed for it due to infrastructure issues).

### Current behavior
```
Traceback (most recent call last):
  File "pylint/pylint/checkers/utils.py", line 1364, in safe_infer
    value = next(infer_gen)
            ^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_ng.py", line 164, in infer
    for i, result in enumerate(self._infer(context=context, **kwargs)):
  File "astroid/astroid/decorators.py", line 81, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 44, in wrapped
    for res in _func(node, context, **kwargs):
  File "astroid/astroid/bases.py", line 184, in _infer_stmts
    for inf in stmt.infer(context=context):
  File "astroid/astroid/nodes/node_ng.py", line 164, in infer
    for i, result in enumerate(self._infer(context=context, **kwargs)):
  File "astroid/astroid/decorators.py", line 81, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 44, in wrapped
    for res in _func(node, context, **kwargs):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_classes.py", line 451, in _infer
    stmts = list(self.assigned_stmts(context=context))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 81, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/protocols.py", line 456, in assign_assigned_stmts
    yield from _resolve_assignment_parts(
  File "astroid/astroid/protocols.py", line 497, in _resolve_assignment_parts
    assigned = part.getitem(index_node, context)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_classes.py", line 2136, in getitem
    raise AstroidTypeError(f"{self!r} (value={self.value})")
[...]

Issue #2764, opened 2025-06-11
ValueError when pretty-printing node while handling RecursionError


correctmost:
### Steps to reproduce

The following code triggers a `ValueError` in astroid:

```python
class c(c&()):
    _
class aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(c&()):
    _
```

This bug was discovered by OSS-Fuzz:
https://issues.oss-fuzz.com/issues/423673973 (report not public yet)

### Current behavior
```
[...snip...]

RecursionError: maximum recursion depth exceeded

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "pylint/pylint/lint/pylinter.py", line 994, in get_ast
    return MANAGER.ast_from_file(filepath, modname, source=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/manager.py", line 165, in ast_from_file
    return AstroidBuilder(self).file_build(filepath, modname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 145, in file_build
    return self._post_build(module, builder, encoding)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 173, in _post_build
    module = self._manager.visit_transforms(module)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/manager.py", line 129, in visit_transforms
    return self._transform.visit(node)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/transforms.py", line 162, in visit
    return self._visit(node)
           ^^^^^^^^^^^^^^^^^
  File "astroid/astroid/transforms.py", line 84, in _visit
    visited = self._visit_generic(value)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/transforms.py", line 112, in _visit_generic
    return [self._visit_generic(child) for child in node]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2972, opened 2026-02-20
Fix MemoryError when inferring f-string with large width

### LINKED ISSUE #2762: MemoryError when inferring formatted string with large width value
### Steps to reproduce

The following code triggers a `MemoryError` in astroid:

```python
x=f'{0:11111111111}'
```

This bug was discovered by OSS-Fuzz:
https://issues.oss-fuzz.com/issues/421940232 (report not public yet)

### Current behavior
```
    callback(astroid)
  File "pylint/pylint/extensions/redefined_variable_type.py", line 100, in visit_assign
    _type = node_type(node.value)
            ^^^^^^^^^^^^^^^^^^^^^
  File "pylint/pylint/checkers/utils.py", line 1501, in node_type
    for var_type in node.infer():
  File "astroid/astroid/nodes/node_ng.py", line 165, in infer
  File "astroid/astroid/nodes/node_classes.py", line 4785, in _infer
    yield from self._infer_with_values(context)
  File "astroid/astroid/nodes/node_classes.py", line 4793, in _infer_with_values
    for inferred in self._infer_from_values(self.values, context):
  File "astroid/astroid/nodes/node_classes.py", line 4811, in _infer_from_values
    for node in cls._safe_infer_from_node(nodes[0], context, **kwargs):
  File "astroid/astroid/nodes/node_classes.py", line 4832, in _safe_infer_from_node
    yield from node._infer(context, **kwargs)
  File "astroid/astroid/nodes/node_classes.py", line 4702, in _infer
    formatted = format(value_to_format, format_spec.value)
MemoryError
```

### Expected behavior
No crash

### Version
580d17cd55af78

worksbyfriday:
## Description

Fix `MemoryError` crash when astroid infers an f-string with a huge format width specification (e.g., `f'{0:11111111111}'`).

The `FormattedValue._infer()` method calls Python's built-in `format()` to evaluate format specs during inference. A format spec with a large width value causes `format()` to attempt allocating a multi-gigabyte string, triggering a `MemoryError`. The existing exception handler only caught `ValueError` and `TypeError`.

### Changes

- Added `MemoryError` to the except clause in `FormattedValue._infer()` (alongside existing `ValueError`/`TypeError`)
- Added regression test

Closes #2762

### Type of Changes
- Bug fix

### Related Issue
This is another OSS-Fuzz discovery (like #2764 and #2785 from PR #2971), in the same family of "adversarial input crashes inference" bugs. This one is in f-string inference rather than `__str__`/`repr`.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -5262,6 +5262,28 @@ def test_formatted_fstring_inference(code, result) -> None:
         assert value_node.value == result
 
 
+def test_fstring_large_width_no_memory_error() -> None:
+    """MemoryError should not crash inference for f-strings with huge width.
+
+    Regression test for https://github.com/pylint-dev/astroid/issues/2762
+    """
+
+    class OOMInt(int):
+        """An int whose __format__ raises MemoryError, simulating f'{0:>11111111111}'."""
+
+        def __format__(self, spec: str) -> str:
+            raise MemoryError
+
+    node = extract_node("f'{0:>9}'")
+    # Replace the Const value with our OOMInt so format() raises MemoryError
+    # without actually allocating a huge string.
+    fmt_value = node.values[0]
+    fmt_value.value.value = OOMInt(0)
+    inferred = node.inferred()
+    assert len(inferred) == 1
+    assert inferred[0] is util.Uninferable
+
+
 def test_augassign_recursion() -> None:
     """Make sure inference doesn't throw a RecursionError.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,11 @@ What's New in astroid 4.1.2?
 ============================
 Release date: TBA
 
+* Catch ``MemoryError`` when inferring f-strings with extremely large format
+  widths (e.g. ``f'{0:11111111111}'``) so that inference yields ``Uninferable``
+  instead of crashing.
+
+  Closes #2762
 
 
 What's New in astroid 4.1.1?
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -4756,8 +4756,9 @@ class FormattedValue(NodeNG):
                         end_col_offset=self.end_col_offset,
                     )
                     continue
-                except (ValueError, TypeError):
-                    # happens when format_spec.value is invalid
+                except (ValueError, TypeError, MemoryError):
+                    # ValueError/TypeError: invalid format spec
+                    # MemoryError: format spec with huge width (e.g. f'{0:11111111111}')
                     yield util.Uninferable
                     uninferable_already_generated = True
                 continue


jacobtylerwalls (commented):
Super. A changelog in 4.1.1, please.

tests/test_inference.py
```diff
@@ -5262,6 +5262,19 @@ def test_formatted_fstring_inference(code, result) -> None:
...
 
 
+def test_fstring_large_width_no_memory_error() -> None:
+    """MemoryError should not crash inference for f-strings with huge width.
+
+    Regression test for https://github.com/pylint-dev/astroid/issues/2762
+    """
+    node = extract_node("""
+        f'{0:11111111111}'  #@
+        """)
```
jacobtylerwalls: This test is very slow. Can you mock the ValueError?

tests/test_inference.py
```diff
@@ -5267,10 +5267,24 @@ def test_fstring_large_width_no_memory_error() -> None:
...
+    # Mock format() to raise MemoryError, simulating a huge width spec
+    # without actually allocating gigabytes of memory.
+    original_format = format
+
+    def mock_format(value, spec=""):
+        if spec:
+            raise MemoryError("simulated huge width allocation")
+        return original_format(value, spec)
+
+    with patch("builtins.format", side_effect=mock_format):
```
jacobtylerwalls: Can we mock anything other than builtins.format? I'm concerned that will introduce headaches for parallel test execution.

worksbyfriday: Good point. Replaced the `builtins.format` mock entirely — now uses a `HugeFormatValue(int)` subclass whose `__format__` raises MemoryError. Injected into the Const node's `.value`, so no global state is touched. Safe for parallel execution.

tests/test_inference.py
```diff
@@ -5262,6 +5262,36 @@ def test_formatted_fstring_inference(code, result) -> None:
...
+    class HugeFormatValue(int):
+        """An int whose __format__ always raises MemoryError."""
+
+        def __format__(self, spec):
+            raise MemoryError("simulated huge width allocation")
+
+    # Patch the Const value that FormattedValue._infer will read
+    value_node = node.values[0].value  # the Const(0) inside f'{0:>5}'
+    original_value = value_node.value
+    value_node.value = HugeFormatValue(0)
```
jacobtylerwalls: I think you could make this less verbose by providing this as code to extract_node instead of patching it afterward.

jacobtylerwalls: Or does our inference code not call `__format__` on user-supplied classes?

worksbyfriday: No — `FormattedValue._infer` extracts `Const.value` (a Python primitive) and calls `format(value_to_format, format_spec.value)` on it. User-supplied classes in the parsed code won't end up as `value_to_format`.

Simplified the test to just use the literal `f'{0:>11111111111}'` — `format()` fails instantly with MemoryError on any system (the allocation is rejected before any work). No mocking, no patching, 4 lines.

jacobtylerwalls (changes requested):
Now the test is extremely slow again.

worksbyfriday:
Reworked per feedback — the test no longer triggers an actual huge allocation.

Now uses an `OOMInt(int)` subclass whose `__format__` raises `MemoryError` instantly, so the test exercises the except clause without allocating memory or mocking `builtins.format`. Also rebased on main and added changelog entry under 4.1.2.

DanielNoord (approved):
Test is a bit stupid, but on the other hand quite hard to test this without actually using a lot of memory.

This adds enough regression test for a simple except clause for me.

jacobtylerwalls (approved):

PR #2977, opened 2026-02-22
Implement `And` and `Or` constraints

zenlyj:
## Type of Changes

## Description

Related to:
- https://github.com/pylint-dev/pylint/issues/1498.
- #3101 

Add `AndConstraint` and `OrConstraint` compound constraints to support inference from compound boolean conditions. Compound constraints recursively collects constraints from each operand and combine their results.

Inference can now narrow values using multiple constraints joined by `and` or `or`:

```
def foo(x=3):
    if isinstance(x, int) and x != 3:
        x  # value 3 filtered out

def bar(x=3):
    if isinstance(x, str) or x != 3:
        x  # value 3 filtered out
```

Nested and multi-operand expressions are also supported:

```
def foo(x=3):
    if x is not None and (isinstance(x, int) and x != 3):
        x  # value 3 filtered out

    if x is not None and isinstance(x, int) and x != 3:
        x  # value 3 filtered out
```

If a child constraint cannot be determined (`classinfo` cannot be inferred), inference remains conservative:

```
def foo(classinfo, x=3):
    if isinstance(x, classinfo) and x == 3:
        x  # value 3 kept
    else:
        x  # value 3 kept
```

No narrowing occurs when one of its operands does not constrain the target variable:

```
x, y = 3, None

if not x and y:
    x  # value 3 kept
```


astroid/constraint.py
```diff
@@ -247,6 +247,106 @@ def satisfied_by(
         return True
 
 
+class AndConstraint(Constraint):
```
Pierre-Sassoulas: There's a lot of duplication here, afaiu the only diffs is expr.op == "and"/"or" and all/any polarity. Could we collapse to one class (_CompoundConstraint(Constraint)) ?

```python
  class _CompoundConstraint(Constraint):
      OP: ClassVar[str]

      def __init__(self, node, negate, children: list[Constraint]) -> None:
          super().__init__(node=node, negate=negate)
          self.children = children

      @classmethod
      def match(cls, node, expr, negate=False):
          if not (isinstance(expr, nodes.BoolOp) and expr.op == cls.OP): # I expect this 
              # check to be removed outright, see final review comment
              return None
          children: list[Constraint] = []
          for value in expr.values:
              matches = list(_match_constraint(node, value, negate))
              if not matches:
                  return None
              children.extend(matches)
          return cls(node=node, negate=negate, children=children)

      def satisfied_by(self, inferred, context) -> bool:
          if inferred is util.Uninferable:
              return True
          results = (c.satisfied_by(inferred, context) for c in self.children)
          # ``not (a and b)`` is ``not a or not b`` — children are already negated.
          strict = (self.OP == "and") ^ self.negate
          return all(results) if strict else any(results)
```

astroid/constraint.py
```diff
@@ -247,6 +247,106 @@ def satisfied_by(
...
+
+    def satisfied_by(
+        self, inferred: InferenceResult, context: InferenceContext
+    ) -> bool:
+        """Return True for uninferable results, or depending on negate flag:
+
+        - negate=False: satisfied when all children constraints are satisfied.
+        - negate=True: satisfied when at least one child constraint is satisfied.
+        """
+        if isinstance(inferred, util.UninferableBase):
```
Pierre-Sassoulas:
```suggestion
        if inferred is util.Uninferable:
```
We generally do that elsewhere in astroid

Pierre-Sassoulas (changes requested):
Thank you for this! Would you mind adding a changelog entry, please?

Also, those constraints are performance intensive, and we know some node will never match for And / Or. Maybe we can pre-filter what tests are done on what nodes  before instantiation instead of filtering in the beginning of the match function.with something like:
```python
  _CONSTRAINTS_BY_NODE_TYPE = {
      nodes.Compare: (NoneConstraint, EqualityConstraint),
      nodes.Call: (TypeConstraint,),
      nodes.BoolOp: (AndConstraint, OrConstraint),
      nodes.Name: (BooleanConstraint,),
      nodes.Attribute: (BooleanConstraint,),
      nodes.UnaryOp: (BooleanConstraint,),
  }
```
Could be a prior refactor, le me know if you want me to tackle this or if you think this can be done in this PR :)

Pierre-Sassoulas (changes requested):
Thank you for this! Would you mind adding a changelog entry, please?

Also, those constraints are performance intensive, and we know some node will never match for And / Or. Maybe we can pre-filter what tests are done on what nodes  before instantiation instead of filtering in the beginning of the match function.with something like:
```python
  _CONSTRAINTS_BY_NODE_TYPE = {
      nodes.Compare: (NoneConstraint, EqualityConstraint),
      nodes.Call: (TypeConstraint,),
      nodes.BoolOp: (AndConstraint, OrConstraint),
      nodes.Name: (BooleanConstraint,),
      nodes.Attribute: (BooleanConstraint,),
      nodes.UnaryOp: (BooleanConstraint,),
  }
```
Could be a prior refactor. Let me know if you want me to tackle this or if you think this can be done in this PR :)

zenlyj:
Thanks for the review! I combined both classes into a single `CompoundConstraint`.

For the pre-filtering, I think it is a good idea to use a lookup dictionary to reduce number of `.match()` calls. The main difference is using exact node type rather than `isinstance()`, this is fine as long as we do not expect constraint expressions to use subclasses of the mapped node types.

I kept the `isinstance()` guards inside each `.match()` for type narrowing. Also, the uncovered line may be unreachable since `nodes.BoolOp.op` can only be "and"/"or"?

Pierre-Sassoulas (approved):
Great !

PR #2980, opened 2026-02-24
Fix AttributeError crash with starred AssignAttr in for-loop targets

### LINKED ISSUE #2646: "AttributeError: 'AssignAttr' object has no attribute 'name'" with starred expression
### Description
Astroid crashes on the following fuzzed code from pysource-codegen:

```python
class c:
    a[t]

for *o.attr, (*t,) in ():
    pass
```

### Backtrace
```
Exception on node <Subscript l.2 at 0x7bc5a43f8ad0> in file 'pylint/crash.py'
  File "pylint/pylint/checkers/utils.py", line 1365, in safe_infer
    value = next(infer_gen)
            ^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_ng.py", line 167, in infer
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 86, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 49, in wrapped
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/bases.py", line 177, in _infer_stmts
    for inf in stmt.infer(context=context):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_ng.py", line 167, in infer
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 86, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 49, in wrapped
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_classes.py", line 457, in _infer
    stmts = list(self.assigned_stmts(context=context))
  File "astroid/astroid/decorators.py", line 69, in inner
    yield next(generator)
          ^^^^^^^^^^^^^^^
  File "astroid/astroid/protocols.py", line 817, in starred_assigned_stmts
    _determine_starred_iteration_lookups(self, target, lookups)
  File "astroid/astroid/protocols.py", line 707, in _determine_starred_iteration_lookups
    and element.value.name == starred.value.name
        ^^^^^^^^^^^^^^^^^^
AttributeError: 'AssignAttr' object has no attribute 'name'
```

### Version
cae4388d75a

worksbyfriday:
## Description

Fix `AttributeError: 'AssignAttr' object has no attribute 'name'` when inferring starred attribute assignments in for-loop targets.

## Reproducer

```python
class c:
    a[t]

for *o.attr, (*t,) in ():
    pass
```

Running pylint or inferring the subscript `a[t]` crashes because `t` resolves to a starred assignment inside the for-loop target. The `_determine_starred_iteration_lookups` function (protocols.py) assumed the starred value was always an `AssignName` (which has `.name`), but `*o.attr` produces an `AssignAttr` node (which has `.attrname` instead).

## Fix

Replace the name-based comparison (`element.value.name == starred.value.name`) with identity comparison (`element is starred`). This is:

- **More correct**: we're looking for the exact same AST node, not just nodes with matching names
- **More robust**: works for all node types (`AssignName`, `AssignAttr`, `Subscript`, etc.)

## Test

Added `test_starred_assignattr_no_crash` that verifies the reproducer from #2646 no longer crashes.

Closes #2646

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,11 @@ Release date: TBA
 
   Closes #2762
 
+* Fix ``AttributeError`` crash with starred ``AssignAttr`` in for-loop targets
+  (e.g., ``for *o.attr, ... in ...``).
+
+  Closes #2646
+
 
 What's New in astroid 4.1.1?
 ============================
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -722,6 +722,8 @@ def starred_assigned_stmts(  # noqa: C901
         for index, element in enumerate(itered):
             if (
                 isinstance(element, nodes.Starred)
+                and hasattr(element.value, "name")
+                and hasattr(starred.value, "name")
                 and element.value.name == starred.value.name
             ):
                 lookups.append((index, len(itered)))
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -5430,15 +5430,13 @@ def test_starred_assignattr_no_crash() -> None:
 
     See https://github.com/pylint-dev/astroid/issues/2646
     """
-    module = parse(
-        """
+    module = parse("""
     class c:
         a[t]
 
     for *o.attr, (*t,) in ():
         pass
-    """
-    )
+    """)
     for node in module.nodes_of_class(nodes.Name):
         if node.name == "t" and node.lineno == 3:
             inferred = list(node.infer())
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -5425,6 +5425,29 @@ def test_unpacking_starred_empty_list_in_assignment() -> None:
     assert inferred.as_string() == "[]"
 
 
+def test_starred_assignattr_no_crash() -> None:
+    """Starred AssignAttr in for-loop target should not crash.
+
+    See https://github.com/pylint-dev/astroid/issues/2646
+    """
+    module = parse(
+        """
+    class c:
+        a[t]
+
+    for *o.attr, (*t,) in ():
+        pass
+    """
+    )
+    for node in module.nodes_of_class(nodes.Name):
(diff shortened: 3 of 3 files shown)


DanielNoord (commented):
Does your test also pass if you just use `attrname`?

worksbyfriday:
No — using `attrname` doesn't work here because the two starred nodes can have different value types. In the test case, the for-loop target is `*o.attr, (*t,)`, so:

- `element.value` (for `*o.attr`) is an `AssignAttr` → has `.attrname`
- `starred.value` (for `*t` inside the nested tuple) is an `AssignName` → has `.name`, no `.attrname`

Using `element.value.attrname == starred.value.attrname` crashes with `AttributeError: 'AssignName' object has no attribute 'attrname'` (I just tested it).

Identity comparison (`element is starred`) is the correct approach — we're looking for a specific AST node in the target tuple, and `is` works regardless of whether the value is `AssignName` or `AssignAttr`.

Re the CI: the "Test pylint" failures (4 `wrong-import-order` tests) are pre-existing on `main` — not related to this change.

DanielNoord:
You can use isinstance to determine which attribute you should use.

We don't know why this code is here, so we're not sure if identity is better

DanielNoord (commented):
This changes behaviour by introducing a `None` case. Only fix it for the error you saw, in all other cases keep the old code.

Changes in `astroid` should be kept minimal.

PR #2983, opened 2026-02-27
Fix builtin functions incorrectly exposing descriptor attributes

### LINKED ISSUE #2743: `eval.__get__(1)` should be an inference error

emmanuel-ferdman:
## Type of Changes

## Description

Builtin functions like `eval`, `print`, and `len` don't have `__get__`, `__defaults__`, `__annotations__`, etc. in Python - they're `builtin_function_or_method`, not regular functions. But astroid was returning values for these anyway since `FunctionModel` didn't distinguish between builtins and user-defined functions. This PR adds a check to raise `AttributeInferenceError` for these attributes on builtins, same as Python would raise `AttributeError`.

Closes #2743

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -269,6 +269,10 @@ class ModuleModel(ObjectModel):
 
 
 class FunctionModel(ObjectModel):
+    def _is_builtin_func(self) -> bool:
+        func = self._instance
+        return isinstance(func.parent, nodes.Module) and not func.root().pure_python
+
     @property
     def attr___name__(self):
         return node_classes.Const(value=self._instance.name, parent=self._instance)
@@ -287,6 +291,8 @@ class FunctionModel(ObjectModel):
     @property
     def attr___defaults__(self):
         func = self._instance
+        if self._is_builtin_func():
+            raise AttributeInferenceError(target=func, attribute="__defaults__")
         if not func.args.defaults:
             return node_classes.Const(value=None, parent=func)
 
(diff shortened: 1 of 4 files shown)


astroid/interpreter/objectmodel.py
```diff
@@ -269,6 +269,10 @@ def attr___cached__(self):
 
 
 class FunctionModel(ObjectModel):
+    def _is_builtin_func(self) -> bool:
+        func = self._instance
```
Pierre-Sassoulas, line 273: Maybe not doing the assignment here would be a perf increase ?

Pierre-Sassoulas (commented):
Thank you, looks good !

astroid/interpreter/objectmodel.py
emmanuel-ferdman: From my testing, the local variable is actually slightly faster (~2-8%) because `LOAD_FAST` for local vars is cheaper than `LOAD_ATTR` for instance attributes in Python. The cost of one store + two fast loads beats two attribute lookups. It's a small difference but I'm in favor of keeping it as-is. But let me know if you what you prefer :)

Pierre-Sassoulas: As is often the case with optimisation, intuition<<benchmark :)

Pierre-Sassoulas (dismissed):

DanielNoord (approved):

PR #2992, opened 2026-03-02
Include inference kwargs in cache key via InferKwargs TypedDict

### LINKED ISSUE #10193: Invalid no-name-in-module when shadowing a base module with an alias then calling a method named format on that alias
### Bug description

If you shadow a base module with an alias, and then call a method named `format()` on the alias in the code your are linting, `checkers/variables.py:_infer_name_module` yields the wrong module, causing `checkers/variables.py:_check_module_attrs` to fail on line 3120 (in pylint 3.3.3), and return an invalid `no-name-in-module` on line 3130.

As a simple example, let's say you have the following code structure:
```
- main.py
- my_module
    - __init__.py
    - utils.py
```

And utils.py has the following code:
```
def format():
    pass

def other_method():
    pass
```

If in `main.py` you have the following code:
```
import my_module.utils as my_module
my_module.format()
```

Then pylint will fail with `main.py:1:0: E0611: No name 'utils' in module 'my_module.utils' (no-name-in-module)`. This is obviously invalid, because we're not trying to import `utils` from `my_module.utils` in the code as the message suggests, but `_infer_name_module` is returning the module `my_method.utils` and then `_check_module_attrs` is trying to access `module.getattr(name)[0]` where name is `utils`, causing this error.

Disclaimer: I know that `import my_module.utils as my_module` isn't good practice, but it is valid Python, and we have a legacy usecase for it that we unfortunately can't drop.

The wild thing to me is that if `main.py` instead just does the import, `_infer_name_module` correctly returns the `my_module` module, and there is no error. For example:
```
import my_module.utils as my_module
```

Even more wild, if you call any other method name on the alias in main.py (at least any that I have tried), `_infer_name_module` also correctly returns the `my_module` module, and there is no error. For example:
```
import my_module.utils as my_module
my_module.other_method()
```

Less surprising, is that if you don't shadow the base module with the alias, but still call `format()`, you also don't get an error:
```
import my_module.utils as different_alias
different_alias.format()
```

It is something specific about the combination of shadowing the base module with an alias and calling a method called `.format()` on that alias that causes `_infer_name_module` to do the inference incorrectly, resulting in this error.

### Configuration

```ini

```

### Command used

```shell
pylint main.py --disable=missing-docstring
```

### Pylint output

```python
************* Module main
main.py:1:0: E0611: No name 'utils' in module 'my_module.utils' (no-name-in-module)
```

### Expected behavior

No failures

### Pylint version

```shell
pylint 3.3.3
astroid 3.3.8
Python 3.12.4 (main, Aug 28 2024, 13:49:33) [Clang 15.0.0 (clang-1500.3.9.4)]

I also tried older versions, and got the same behavior, for example:
pylint 2.15.0
astroid 2.13.5
Python 3.12.4 (main, Aug 28 2024, 13:49:33) [Clang 15.0.0 (clang-1500.3.9.4)]
```

### OS / Environment

macOS 15.1.1, ubuntu 20.04

### Additional dependencies

```python
None
```

worksbyfriday:
## Summary
- Inference cache key in `NodeNG.infer()` now includes kwargs so that different invocations (e.g. `asname=True` vs `asname=False`) produce separate cache entries
- Added `InferKwargs` TypedDict to `astroid/typing.py` documenting valid inference kwargs
- Added regression test demonstrating the cache collision bug

## Context
Follow-up to #2969 per @DanielNoord's [suggestion](https://github.com/pylint-dev/astroid/pull/2969#issuecomment-2681843137) to make inference kwargs explicit using a TypedDict.

The root cause of pylint-dev/pylint#10193: when `import my_module.utils as my_module` is used, pylint calls `node.infer(ctx, asname=False)` to get the unaliased module. But if the node was previously inferred with `asname=True` (the default), the cached result was returned because `asname` wasn't part of the cache key.

## Changes
- `astroid/typing.py`: Add `InferKwargs(TypedDict)` with `asname: bool`
- `astroid/nodes/node_ng.py`: Include `kwargs` in inference cache key
- `tests/test_inference.py`: Add `test_import_inference_cache_includes_asname`
- `ChangeLog`: Entry for 4.1.1

## Test plan
- [x] New test fails without the fix (verified by temporarily reverting)
- [x] New test passes with the fix
- [x] All 438 inference tests pass
- [x] 868 tests pass across test_inference, test_scoped_nodes, test_builder, test_nodes
- [x] ruff, black, mypy clean on changed files

Closes pylint-dev/pylint#10193
Supersedes #2969

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -48,14 +48,15 @@ from astroid.typing import (
 )
 
 if sys.version_info >= (3, 11):
-    from typing import Self
+    from typing import Self, Unpack
 else:
-    from typing_extensions import Self
+    from typing_extensions import Self, Unpack
 
 if TYPE_CHECKING:
     from astroid import nodes
     from astroid.nodes import LocalsDictNodeNG
     from astroid.nodes.node_ng import FrameType
+    from astroid.typing import InferKwargs
 
 
 def _is_const(value) -> bool:
@@ -342,7 +343,7 @@ class BaseContainer(_base_nodes.ParentAssignNode, Instance, metaclass=abc.ABCMet
     def _infer(
         self,
         context: InferenceContext | None = None,
-        **kwargs: Any,
+        **kwargs: Unpack[InferKwargs],
     ) -> Iterator[Self]:
         has_starred_named_expr = any(
             isinstance(e, (Starred, NamedExpr)) for e in self.elts
@@ -441,7 +442,7 @@ class AssignName(
     @decorators.raise_if_nothing_inferred
     @decorators.path_wrapper
     def _infer(
-        self, context: InferenceContext | None = None, **kwargs: Any
+        self, context: InferenceContext | None = None, **kwargs: Unpack[InferKwargs]
     ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
         """Infer an AssignName: need to inspect the RHS part of the
         assign node.
@@ -454,7 +455,7 @@ class AssignName(
 
     @decorators.raise_if_nothing_inferred
     def infer_lhs(
-        self, context: InferenceContext | None = None, **kwargs: Any
+        self, context: InferenceContext | None = None, **kwargs: Unpack[InferKwargs]
     ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
         """Infer a Name: use name lookup rules.
 
@@ -569,7 +570,7 @@ class Name(_base_nodes.LookupMixIn, _base_nodes.NoChildrenNode):
     @decorators.raise_if_nothing_inferred
     @decorators.path_wrapper
     def _infer(
-        self, context: InferenceContext | None = None, **kwargs: Any
+        self, context: InferenceContext | None = None, **kwargs: Unpack[InferKwargs]
     ) -> Generator[InferenceResult, None, InferenceErrorInfo | None]:
         """Infer a Name: use name lookup rules
 
@@ -1040,7 +1041,7 @@ class Arguments(
 
     @decorators.raise_if_nothing_inferred
     def _infer(
-        self, context: InferenceContext | None = None, **kwargs: Any
+        self, context: InferenceContext | None = None, **kwargs: Unpack[InferKwargs]
     ) -> Generator[InferenceResult]:
         # pylint: disable-next=import-outside-toplevel
         from astroid.protocols import _arguments_infer_argname
@@ -1094,7 +1095,7 @@ def _format_args(
 def _infer_attribute(
     node: nodes.AssignAttr | nodes.Attribute,
     context: InferenceContext | None = None,
-    **kwargs: Any,
+    **kwargs: Unpack[InferKwargs],
 ) -> Generator[InferenceResult, None, InferenceErrorInfo]:
     """Infer an AssignAttr/Attribute node by using getattr on the associated object."""
     # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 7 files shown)


DanielNoord (commented):
Of course you also need to start using `InferKwargs` to annotate `kwargs` in `_infer` everywhere. The code as is makes no sense.

Also add a method to `InferKwargs` to generate the cache key so that if you add a new argument it is clear you need to consider its impact on the cache key

astroid/nodes/node_classes.py
```diff
@@ -3229,7 +3230,7 @@ def _infer(
         self,
         context: InferenceContext | None = None,
         asname: bool = True,
-        **kwargs: Any,
+        **kwargs: Unpack[InferKwargs],
```
DanielNoord, line 3233: Now `asname` is in the signature twice. That is not correct.

DanielNoord (commented):
@jacobtylerwalls What do you think of this approach?

tests/test_inference.py
```diff
@@ -6953,6 +6953,36 @@ def test_imported_module_var_inferable3() -> None:
...
+
+    When an Import node is inferred with asname=True and then asname=False,
+    the second call should not return the cached result from the first.
+
+    Regression test for https://github.com/pylint-dev/pylint/issues/10193.
+    """
+    code = textwrap.dedent("""
+    import os.path as mypath  #@
+    """)
+    import_node = _extract_single_node(code)
```
jacobtylerwalls, line 6967: This can be single-lined.

tests/test_inference.py
```diff
@@ -6953,6 +6953,36 @@ def test_imported_module_var_inferable3() -> None:
...
+    assert isinstance(import_node, nodes.Import)
```
jacobtylerwalls, line 6968: Chop.

tests/test_inference.py
```diff
@@ -6953,6 +6953,36 @@ def test_imported_module_var_inferable3() -> None:
...
+    # With asname=True (default): resolves 'mypath' -> 'os.path' -> os.path module
+    result1 = list(import_node.infer(ctx1, asname=True))
+    assert len(result1) == 1
+    assert isinstance(result1[0], nodes.Module)
+    assert result1[0].name == "os.path"
+
+    ctx2 = InferenceContext()
+    ctx2.lookupname = "mypath"
+    # With asname=False: should try to import 'mypath' literally, which fails
+    # The key test: this should NOT return the cached os.path result
```
jacobtylerwalls, line 6981:
```suggestion
```

tests/test_inference.py
```diff
@@ -6953,6 +6953,36 @@ def test_imported_module_var_inferable3() -> None:
...
+    import_node = _extract_single_node(code)
+    assert isinstance(import_node, nodes.Import)
+
+    ctx1 = InferenceContext()
+    ctx1.lookupname = "mypath"
+    # With asname=True (default): resolves 'mypath' -> 'os.path' -> os.path module
+    result1 = list(import_node.infer(ctx1, asname=True))
+    assert len(result1) == 1
+    assert isinstance(result1[0], nodes.Module)
+    assert result1[0].name == "os.path"
```
jacobtylerwalls, line 6976: Let's not deal with `result1`, let's call `next()` to get the module and assert over that.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -46,15 +46,16 @@
 from astroid.nodes.utils import Position
 from astroid.typing import (
     InferBinaryOp,
+    InferKwargs,
```
jacobtylerwalls, line 49: I'm not convinced yet of the wisdom of doing this, since only `Import._infer_node` takes this arg.

jacobtylerwalls (changes requested):
@DanielNoord The more I look at this, I'm starting to wonder if we should be evaluating removing this `asname` business altogether.

`asname` only switches on whether to call `self.real_name(name)`, but we could probably hoist that up to the call sites.

The fact that `asname` defaults True is even in the absence of an alias is likely leading to some wasted computation.

PR #2995, opened 2026-03-05
Wrong inference for the `str(const)`

### LINKED ISSUE #2994: Wrong inference for `str(const)`
These return `Const("")`, which is really misleading (even an `Uninferable` is better):

```python
code = """
    s = str("abc") #@
    s = str(1) #@
    s = str(None) #@
    s = str(2 + 2) #@
    """
```

Looking at `brain_builtin_inference.py`, the `def infer_str()` function is hardcoded to always return `Const("")`.
Next to it is the `def infer_int()`, which at least tries to examine `int()` arguments.

I'll submit a small PR fixing it at least for const argument.

jkmnt:
This PR fixes the wrong inference for the `str(a)`, then the `a` is inferred to be `Const`.

Closes #2994 

I preserved returning `Const("")` in other cases. Maybe it should raise `UseInferenceDefault` like the `infer_int` does ?

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -852,7 +852,7 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
     if call.positional_arguments:
         try:
             first_value = next(call.positional_arguments[0].infer(context=context))
-        except (InferenceError, StopIteration) as exc:
+        except (InferenceError, StopIteration):
             return nodes.Const("")
 
         if isinstance(first_value, nodes.Const):
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -849,20 +849,28 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
     if call.keyword_arguments:
         raise UseInferenceDefault("TypeError: str() must take no keyword arguments")
 
-    if call.positional_arguments:
-        try:
-            first_value = next(call.positional_arguments[0].infer(context=context))
-        except (InferenceError, StopIteration):
-            return nodes.Const("")
+    fallback = nodes.Const("")
 
-        if isinstance(first_value, nodes.Const):
+    if not call.positional_arguments:
+        return fallback
+
+    # Accept only if all inferred values resolve to the same string
+    candidates: set[str] = set()
+    try:
+        for inferred in call.positional_arguments[0].infer(context=context):
+            if not isinstance(inferred, nodes.Const):
+                return fallback
+
             try:
-                stringified = str(first_value.value)
+                candidates.add(str(inferred.value))
             except ValueError:
-                return nodes.Const("")
-            return nodes.Const(stringified)
+                return fallback
+    except InferenceError:
+        return fallback
 
-    return nodes.Const("")
+    if len(candidates) == 1:
+        return nodes.Const(candidates.pop())
+    return fallback
 
 
 def infer_int(node, context: InferenceContext | None = None):
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -856,7 +856,11 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
             return nodes.Const("")
 
         if isinstance(first_value, nodes.Const):
-            return nodes.Const(str(first_value.value))
+            try:
+                stringified = str(first_value.value)
+            except ValueError:
+                return nodes.Const("")
+            return nodes.Const(stringified)
 
     return nodes.Const("")
 
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -842,15 +842,23 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
 
     :param nodes.Call node: str() call to infer
     :param context.InferenceContext: node context
-    :rtype nodes.Const: a Const containing an empty string
+    :rtype nodes.Const:
+        a Const containing a stringified value of str() call if possible, else an empty string
     """
     call = arguments.CallSite.from_call(node, context=context)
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_builtin_inference.py
```diff
@@ -842,15 +842,23 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
...
     if call.keyword_arguments:
         raise UseInferenceDefault("TypeError: str() must take no keyword arguments")
-    try:
-        return nodes.Const("")
-    except (AstroidTypeError, InferenceError) as exc:
-        raise UseInferenceDefault(str(exc)) from exc
+
+    if call.positional_arguments:
+        try:
+            first_value = next(call.positional_arguments[0].infer(context=context))
```
DanielNoord: Why only use the first value? If we infer multiple values I don't think we should default to the first one.

astroid/brain/brain_builtin_inference.py
```diff
@@ -842,15 +842,23 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
...
+        except (InferenceError, StopIteration):
+            return nodes.Const("")
+
+        if isinstance(first_value, nodes.Const):
+            return nodes.Const(str(first_value.value))
```
DanielNoord: Does this face similar issues as fixed in https://github.com/pylint-dev/astroid/pull/2971? `7**10000` to `str` will raise an exception

jkmnt: Good point, I missed the str() may fail

astroid/brain/brain_builtin_inference.py
jkmnt: Other functions in builtin brain seems to do the same. 

Do you mean `infer_str` should return an interator, e.g. inference result of `str(unknown if unknown else "a")` would be `[Const(""), Const("a")]` ?

DanielNoord: > Other functions in builtin brain seems to do the same.

Unfortunately not all existing code is of the same quality 😅 

As for the solution, I'm not sure to be honest. I think it would probably be best to only return a `Const` with a value if we can be 100% sure it is that value, and otherwise keep falling back.

Something like `if all(v == first_value for v in node.infer(): value = first_value`. If you get what I mean 😅

jkmnt: Good solution. Added just that in the last commit.

DanielNoord (dismissed):
@Pierre-Sassoulas @jacobtylerwalls You want to review this? It LGTM!

astroid/brain/brain_builtin_inference.py
```diff
@@ -842,15 +842,35 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
...
+
+            try:
+                candidates.add(str(inferred.value))
+            except ValueError:
+                return fallback
+    except InferenceError:
+        return fallback
+
+    if len(candidates) == 1:
+        return nodes.Const(candidates.pop())
```
jacobtylerwalls: This seems like an unnecessary source of indeterminism.

astroid/brain/brain_builtin_inference.py
```diff
@@ -842,15 +842,35 @@ def infer_str(node, context: InferenceContext | None = None) -> nodes.Const:
...
     try:
-        return nodes.Const("")
-    except (AstroidTypeError, InferenceError) as exc:
-        raise UseInferenceDefault(str(exc)) from exc
+        for inferred in call.positional_arguments[0].infer(context=context):
+            if not isinstance(inferred, nodes.Const):
+                return fallback
+
+            try:
+                candidates.add(str(inferred.value))
```
jacobtylerwalls, line 865: Should we only run str() if the value is a builtin? I wouldn't want a more involved `str()` implementation from user code to run (we should keep pylint static).

jacobtylerwalls (changes requested):
Thanks for the PR 👍 
Can you add a changelog?

tests/brain/test_brain.py
```diff
@@ -1420,6 +1421,37 @@ def test_infer_str() -> None:
...
+    str(1) #@
+    str(True) #@
+    str(False) #@
+    str(None) #@
+    str(4.33) #@
+    str(...) #@
+    str(2 + 2) #@
+    str() #@
+    str(int) #@
+    str(2 if unknown() else 3) #@
```
jacobtylerwalls, line 1459: I think we should add a test ensuring user `__str__` doesn't run:

```py
class StrWillFail:
    def __str__(self):
        raise RuntimeError

str(StrWillFail()) #@
```

astroid/brain/brain_builtin_inference.py
Pierre-Sassoulas: Switched to next(iter(candidates)).

astroid/brain/brain_builtin_inference.py
Pierre-Sassoulas: I added a test for that, but I think a user-defined object infers to an Instance, not a Const so no change required here. (provided the ``if not isinstance(inferred, nodes.Const):`` right above stays where it is)

tests/brain/test_brain.py
Pierre-Sassoulas: Done in [b0b93c3](https://github.com/pylint-dev/astroid/pull/2995/commits/b0b93c3753523349c3068548b3585facbef015db)

Pierre-Sassoulas (dismissed):
Thank you for your work on this @jkmnt ! I took over because I want this to be in 4.2.0 ;)

DanielNoord (dismissed):

astroid/brain/brain_builtin_inference.py
jacobtylerwalls: Sorry, I didn't look in enough detail here, I now see we return a fallback in every other case, so there's no indeterminism here.

jacobtylerwalls (approved):
Sorry for slowing this one down with an irrelevant comment.

PR #3013, opened 2026-03-22
Bump astroid to 4.2.0b1, update changelog

DanielNoord:
First beta of `astroid` `4.2.0`, which allows https://github.com/pylint-dev/pylint/pull/10933 to be merged in `pylint`.

I have not set back the version to `-dev` as it seems better to indicate we have already released a `beta`.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -221,6 +221,7 @@ Contributors
 - Alexander Presnyakov <flagist0@gmail.com>
 - Ahmed Azzaoui <ahmed.azzaoui@engie.com>
 - Casey Jones <731937+doctorcolossus@users.noreply.github.com>
+- sergiochan <yuheng9211@qq.com>
 
 Co-Author
 ---------
--- a/astroid/__pkginfo__.py
+++ b/astroid/__pkginfo__.py
@@ -2,5 +2,5 @@
 # For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
 # Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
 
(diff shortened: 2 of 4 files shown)


CONTRIBUTORS.txt
```diff
@@ -232,3 +232,4 @@ under this name, or we did not manage to find their commits in the history.
 - carl
 - alain lefroy
 - Mark Gius
+- sergiochan <yuheng9211@qq.com>
```
Pierre-Sassoulas: Can you revert this and upgrade your dev env ? Should be fixed in contributor-txt 1.0.1

Pierre-Sassoulas (changes requested):
LGTM

CONTRIBUTORS.txt
DanielNoord: @Pierre-Sassoulas Is this the expected output? It seems pretty verbose?

DanielNoord: I did install 1.0.1

Pierre-Sassoulas: Not the expected output but also the expected output is as verbose as this (before that it was in every header in every files). I'll fix this in contributor-txt then.

DanielNoord: Can we merge this though? Or does that need to be fixed first?

Pierre-Sassoulas (approved):

CONTRIBUTORS.txt
Pierre-Sassoulas: It seems I reviewed a previous version. It's fixed now;

Pierre-Sassoulas: (So the output should be this verbose only this time, usually it's only new user / change in contribution number, I added a sort that ordered mails  )

PR #3031, opened 2026-04-21
Fix inference of annotated Ellipsis assignments

### LINKED ISSUE #3054: Annotated `...` placeholders inferred as `Const(Ellipsis)` instead of `Uninferable`
### Description
When a stub or typed module uses `attr: int = ...`, astroid infers the value as the literal `Const(Ellipsis)` object. It should yield `Uninferable` since `...` is just a placeholder. This causes false positives downstream - pylint thinks the actual runtime value is `...` and makes wrong conclusions from that.

### Example
```python
import astroid

mod = astroid.parse("""
class Foo:
    x: int = ...
""")

cls = mod.body[0]
ann = cls.body[0]
inferred = list(ann.value.infer())
print(inferred)            # [<Const.Ellipsis l.3 ...>]  - should be [Uninferable]
```

emmanuel-ferdman:
## Type of Changes

## Description

This PR fixes astroid treating `x: int = ...` as if `x` were `Ellipsis` at runtime. The `...` in stubs is a placeholder - the annotation carries the real type - but astroid was returning `Const(Ellipsis)`, making pylint flag every method call through the attribute as `no-member`. Annotated `...` assignments now yield `Uninferable`, so attribute checks are skipped instead of asserted against `Ellipsis`. We yield `Uninferable` rather than resolving the annotation, since that would require handling wrappers/unions/forward-refs that astroid doesn't do here today. Concrete values (`x: int = 5`) and bare `x = ..`. without annotation are unchanged.

Example: 
```python
class Foo:
    x: int = ... 
print(Foo().x.bit_length())  # no more E1101
```

Closes #3054

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,9 @@ Release date: TBA
 
   References pylint-dev/pylint#872
 
+* Infer annotated ``...`` placeholders (``attr: X = ...`` and
+  ``attr = ...  # type: X``) as ``Uninferable`` instead of ``Const(Ellipsis)``.
+
 
 What's New in astroid 4.1.2?
 ============================
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,8 @@ Release date: TBA
 * Infer annotated ``...`` placeholders (``attr: X = ...`` and
   ``attr = ...  # type: X``) as ``Uninferable`` instead of ``Const(Ellipsis)``.
 
+  Closes #3054
+
 
 What's New in astroid 4.1.2?
 ============================
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -444,15 +444,12 @@ def arguments_assigned_stmts(
     return _arguments_infer_argname(self, node_name, context)
 
 
-def _is_ellipsis_placeholder(value: nodes.NodeNG) -> bool:
-    """True if ``value`` is the stub placeholder ``...`` (bare ``Ellipsis``)."""
-    return isinstance(value, nodes.Const) and value.value is Ellipsis
-
-
-def _has_annotation(
+def _is_ellipsis_placeholder_assignment(
     assign: nodes.AugAssign | nodes.Assign | nodes.AnnAssign | nodes.TypeAlias,
 ) -> bool:
-    """True if ``assign`` carries a PEP 526 annotation or a PEP 484 ``# type:`` comment."""
+    """True when ``assign`` uses ``...`` as a stub placeholder for an annotated name."""
+    if not (isinstance(assign.value, nodes.Const) and assign.value.value is Ellipsis):
+        return False
     if isinstance(assign, nodes.AnnAssign):
         return assign.annotation is not None
     return isinstance(assign, nodes.Assign) and assign.type_annotation is not None
@@ -466,7 +463,7 @@ def assign_assigned_stmts(
     assign_path: list[int] | None = None,
 ) -> Any:
     if not assign_path:
-        if _is_ellipsis_placeholder(self.value) and _has_annotation(self):
+        if _is_ellipsis_placeholder_assignment(self):
             yield util.Uninferable
             return None
         yield self.value
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -455,9 +455,7 @@ def _has_annotation(
     """True if ``assign`` carries a PEP 526 annotation or a PEP 484 ``# type:`` comment."""
     if isinstance(assign, nodes.AnnAssign):
         return assign.annotation is not None
-    if isinstance(assign, nodes.Assign):
-        return assign.type_annotation is not None
-    return False
+    return isinstance(assign, nodes.Assign) and assign.type_annotation is not None
 
 
 @decorators.raise_if_nothing_inferred
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -444,6 +444,22 @@ def arguments_assigned_stmts(
     return _arguments_infer_argname(self, node_name, context)
 
 
+def _is_ellipsis_placeholder(value: nodes.NodeNG) -> bool:
+    """True if ``value`` is the stub placeholder ``...`` (bare ``Ellipsis``)."""
+    return isinstance(value, nodes.Const) and value.value is Ellipsis
+
+
+def _has_annotation(
+    assign: nodes.AugAssign | nodes.Assign | nodes.AnnAssign | nodes.TypeAlias,
+) -> bool:
+    """True if ``assign`` carries a PEP 526 annotation or a PEP 484 ``# type:`` comment."""
+    if isinstance(assign, nodes.AnnAssign):
+        return assign.annotation is not None
+    if isinstance(assign, nodes.Assign):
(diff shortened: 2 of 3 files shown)


astroid/protocols.py
```diff
@@ -444,6 +444,20 @@ def arguments_assigned_stmts(
...
+    return isinstance(value, nodes.Const) and value.value is Ellipsis
+
+
+def _has_annotation(
+    assign: nodes.AugAssign | nodes.Assign | nodes.AnnAssign | nodes.TypeAlias,
+) -> bool:
+    """True if ``assign`` carries a PEP 526 annotation or a PEP 484 ``# type:`` comment."""
+    if isinstance(assign, nodes.AnnAssign):
+        return assign.annotation is not None
+    return isinstance(assign, nodes.Assign) and assign.type_annotation is not None
```
DanielNoord: I don't see a real need to keep these separate. Let's just make a `_is_ellipsis_placeholder_assignment`?

DanielNoord (commented):
Do we have an issue for this?

astroid/protocols.py
emmanuel-ferdman: Combined together and opened the GH issue, thanks :)

ChangeLog
```diff
@@ -32,6 +32,9 @@ Release date: TBA
 
   References pylint-dev/pylint#872
 
+* Infer annotated ``...`` placeholders (``attr: X = ...`` and
```
DanielNoord, line 35: Can you link to the issue here as well?

tests/test_protocols.py
DanielNoord, line 1: I am a bit worried about this. We now change inference behaviour depending on whether there is a type annotation.

@Pierre-Sassoulas @jacobtylerwalls What do you guys think?

emmanuel-ferdman: @DanielNoord Yeah, fair concern - my reasoning was that astroid already does annotation-dependent filtering in core. for example, [`scoped_nodes.py`](https://github.com/pylint-dev/astroid/blob/0a80be06e6f49b095f7be99846fd88b3ce4be759/astroid/nodes/scoped_nodes/scoped_nodes.py#L2377-L2378) skips bare `AnnAssign` in `getattr` because annotation-only entries aren't real attributes, so this isn't a new pattern. I see such pattern in `brain_dataclasses` as well. The practical issue is that `Const(Ellipsis)` causes false positives across multiple checkers (E1101, E1130, E1133, and potentially others), and moving this to pylint would require a separate fix in each affected checker for the same root cause. `Uninferable` is the conservative answer - it doesn't assert a wrong type, it just declines to assert one, which every consumer already handles gracefully. Happy to move to pylint-level but lets hear from the team :)

Pierre-Sassoulas: I think the actual runtime value is ellipsis, the intent may well be that it should be an int, but in the example code the value at runtime will definitely be an ellipsis. If we start modifying inference based on what user wanted to and not what they actually did it's going to create major issues downstream. I think we lack context about the initial problem and in particular the pylint false positive we're trying to fix. Do we have an instanced class and are we inferring from the base class value before modification in the constructor ? Is this in a pyi file ?

DanielNoord: Personally leaning towards fixing this in `pylint` indeed. I agree with @Pierre-Sassoulas that this feels like a slippery slope.

emmanuel-ferdman: @Pierre-Sassoulas @DanielNoord yes, this is about `.pyi` stubs where `= ...` is a PEP 484 placeholder, not a runtime value, but it shows up in `.py` too - Protocols, abstract bases, anywhere `= ...` means "value comes from elsewhere":
```python
# .pyi stub
class ExtraSelection:
    format: QTextCharFormat = ...

# Protocol
class Readable(Protocol):
    path: str = ...

# Abstract base
class BaseWidget:
    name: str = ...
```
All of these currently get `Const(Ellipsis)` which triggers false E1101/E1133/E0633/E1136 on any attribute access. For example:
```python
class Widget:
    name: str = ...
    items: list = ...

w = Widget()
w.name.upper()        # E1101: Instance of 'Ellipsis' has no 'upper' member
for x in w.items:     # E1133: Non-iterable value is used in an iterating context
a, b = w.items        # E0633: Attempting to unpack a non-sequence
w.items[0]            # E1136: Value is unsubscriptable
```
Maybe I'm missing something but I looked into fixing this at the pylint level and it looks pretty scattered - we would need separate guards in at least 5 checker methods (E1101 in [`visit_attribute`](https://github.com/pylint-dev/pylint/blob/9f08fc7bbc0f389c18a70cf1aaf71a2dc16c776f/pylint/checkers/typecheck.py#L1068), E1133 in [`_check_iterable`](https://github.com/pylint-dev/pylint/blob/9f08fc7bbc0f389c18a70cf1aaf71a2dc16c776f/pylint/checkers/typecheck.py#L2291), E0633 in [`_check_unpacking`](https://github.com/pylint-dev/pylint/blob/9f08fc7bbc0f389c18a70cf1aaf71a2dc16c776f/pylint/checkers/variables.py#L3097), E1136 in [`visit_subscript`](https://github.com/pylint-dev/pylint/blob/9f08fc7bbc0f389c18a70cf1aaf71a2dc16c776f/pylint/checkers/typecheck.py#L2147), and E1130 needs yet another pattern since [`node.type_errors()`](https://github.com/pylint-dev/pylint/blob/9f08fc7bbc0f389c18a70cf1aaf71a2dc16c776f/pylint/checkers/typecheck.py#L1963) generates the error inside astroid), plus every future checker would need to remember the same guard. But I understand your concerns about changing inference based on annotations. There's actually precedent though - [`brain_dataclasses._infer_instance_from_annotation`](https://github.com/pylint-dev/astroid/blob/056f9fff5fb544b0830f49491f2f11d81935993a/astroid/brain/brain_dataclasses.py#L643) already resolves annotations to `Instance` objects for dataclass fields, same pattern. I could also go a step further and resolve to an `Instance` instead of `Uninferable` - infer the annotation node, if it's a `ClassDef` yield `klass.instantiate_class()`, unwrap `ClassVar`/`Final`/`Annotated` wrappers, fall back to `Uninferable` (or something else?) for anything complex (unions, TypeVar, etc.). That way we're not just silencing checks, pylint still catches real bugs: 

```python
class Widget:
    name: str = ...

Widget().name.upper()      # no error (correct)
Widget().name.bit_length() # E1101: str has no bit_length (real bug caught)
```
With `Uninferable` both calls get silenced. What do you think - Should I upload my `Instance` changes for review? Or is there a better pylint-approach?

DanielNoord: The `Instance` change seems sensible, althought also tricky.

Thinking out loud here: is it possible to only do this for `.pyi` files? I think that should fix most errors?

Pierre-Sassoulas: I think the path forward if a consistent pyi strategy in astroid. We added incomplete support of pyi in astroid 3 (astroid load them, but we get syntax error for them because we don't parse them properly -- as in we parse them as if it was normal python code and not a pyi) we need to start parsing and have a strategy regarding what to do with the result of the parsing to have actual support now.

PR #3045, opened 2026-05-05
Wrap walrus expressions in parens in as_string output

### LINKED ISSUE #2668: Incorrect removal of brackets
```python
code = '''
if attn_output.size() != (size := (batch_size, self.num_heads, query.shape[1], self.head_dim)):
     raise ValueError(f"`attn_output` should be of size {size}, but is {attn_output.size()}")
'''
print(astroid.extract_node(code).as_string())
```
Running the above code produces the output
```
if attn_output.size() != size := (batch_size, self.num_heads, query.shape[1], self.head_dim):
    raise ValueError(f'`attn_output` should be of size {size}, but is {attn_output.size()}')
```
This is incorrect since the brackets in the assignment operator are dropped and the syntax becomes incorrect.

Another example of dropping brackets:
```python
code = '''
collections.OrderedDict(
    {
        'linear': torch.nn.Linear(in_features, out_features, bias=bias),
        **(
            {
                'norm': TorchNorm(out_features),
                'act': TorchActivation(),
            }
            if not (i == len(layer_sizes) - 2 and not activate_final)
            else {}
        ),
    }
)
'''
print(astroid.extract_node(code).as_string())
```
outputs:
```
collections.OrderedDict({'linear': torch.nn.Linear(in_features, out_features, bias=bias), **{'norm': TorchNorm(out_features), 'act': TorchActivation()} if not (i == len(layer_sizes) - 2 and not activate_final) else {}})
```
Brackets following the `**` operator are missing and syntax becomes incorrect

Python version: 3.10
astroid version: 3.3.6

SAY-5:
## Type of Changes

## Description

Closes #2668

`AsStringVisitor.visit_namedexpr` rendered `a := value` without surrounding parentheses, producing invalid Python whenever the walrus appeared inside another expression. For example, `if x != (size := (1, 2, 3)):` round-tripped to `if x != size := (1, 2, 3):`, which is a SyntaxError.

The fix always emits `(target := value)`, matching CPython's `ast.unparse` behavior. Existing `test_assignment_expression` was updated to expect the parenthesized form, and a regression test rebuilds the issue's example and asserts the rendered output is valid Python via `compile(...)`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,13 @@ Release date: TBA
 
   Closes #2646
 
+* Wrap assignment expressions (``:=``) in parentheses when emitting
+  ``as_string`` output so the rendered code remains syntactically valid in
+  contexts such as comparisons, where Python requires the walrus expression
+  to be parenthesized.
+
+  Closes #2668
+
 * Fix ``TypeError`` in ``brain_random`` when ``random.sample`` is called with a
   sequence containing nodes whose ``__init__`` does not accept ``lineno``
   (e.g. ``Module``). The clone helper now filters init params to those the
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,7 @@ Release date: TBA
   to be parenthesized.
 
   Closes #2668
+
 * Catch ``MemoryError``/``RecursionError`` (and ``ValueError``) when validating
   type comments with ``ast.parse``. Pathological type comments produced by
   fuzzers (e.g. ``# type: i{{{{{{{...``) previously crashed parsing with a
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -439,10 +439,10 @@ class AsStringVisitor:
         return node.name
 
     def visit_namedexpr(self, node: nodes.NamedExpr) -> str:
-        """Return an assignment expression node as string"""
+        """Return an assignment expression node as string, always parenthesized."""
         target = node.target.accept(self)
         value = node.value.accept(self)
-        return f"{target} := {value}"
+        return f"({target} := {value})"
 
     def visit_nonlocal(self, node: nodes.Nonlocal) -> str:
         """return an nodes.Nonlocal node as string"""
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -442,6 +442,8 @@ class AsStringVisitor:
         """Return an assignment expression node as string, always parenthesized."""
         target = node.target.accept(self)
         value = node.value.accept(self)
+        if isinstance(getattr(node, "parent", None), nodes.Compare):
+            return f"{target} := {value}"
         return f"({target} := {value})"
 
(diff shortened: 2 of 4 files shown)


DanielNoord (dismissed):

DanielNoord (commented):
Code LGTM, but you'll need to move the import to the top level.

SAY-5:
Pushed 35b011b. The follow-up keeps compare output parenthesized while rendering walrus operands bare when the parent node already manages precedence parens, which fixes the extra wrapping caught by the downstream pylint workflow.\n\nLocal checks I reran:\n- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q tests/test_nodes.py\n- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q tests/test_nodes.py -k 'assignment_expression'\n- in a local pylint checkout with this astroid branch installed: PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_functional.py -k 'consider_refactoring_into_while_condition_py38'

astroid/nodes/as_string.py
```diff
@@ -60,6 +60,23 @@ def _precedence_parens(
 
         return child.accept(self)
 
+    @staticmethod
+    def _namedexpr_renders_bare(node: nodes.NamedExpr) -> bool:
```
DanielNoord: This feels quite brittle. Did you explore other ways to compute this condition?

astroid/nodes/as_string.py
```diff
@@ -439,10 +439,12 @@ def visit_name(self, node: nodes.Name) -> str:
         return node.name
 
     def visit_namedexpr(self, node: nodes.NamedExpr) -> str:
-        """Return an assignment expression node as string"""
+        """Return an assignment expression node as string, always parenthesized."""
         target = node.target.accept(self)
         value = node.value.accept(self)
-        return f"{target} := {value}"
+        if isinstance(getattr(node, "parent", None), nodes.Compare):
```
DanielNoord: Can we do this without a `getattr`?

Pierre-Sassoulas (dismissed):
Agree with Daniel's comment, otherwise LGTM

ChangeLog
```diff
@@ -20,6 +20,12 @@ Release date: TBA
 
   Closes #2646
 
+* Wrap assignment expressions (``:=``) in parentheses when emitting
+  ``as_string`` output so the rendered code remains syntactically valid in
+  contexts such as comparisons, where Python requires the walrus expression
+  to be parenthesized.
+
+  Closes #2668
```
Pierre-Sassoulas, line 28:
```suggestion
  Closes #2668

```

SAY-5:
Good catch, the ChangeLog already has `Closes #2668` under that entry so it matches the suggestion. Let me know if anything else needs tweaking before merge.

Pierre-Sassoulas (approved):

PR #3046, opened 2026-05-07
Use exact signatures for GObject introspection bindings

oreiche:
## Type of Changes

## Description

With the recent change to use `inspect.signature()` (#2930), instead of generating signatures for all functions and methods with "wildcards" (`*args, **kwargs`), the exact signature could be provided for better linting.

Given the following `example.py`

``` python
"""Simple Example from https://python-gtk-3-tutorial.readthedocs.io/en/latest/introduction.html"""

import gi

gi.require_version("Gtk", "3.0")
from gi.repository import Gtk  # pylint: disable=wrong-import-position

win = Gtk.Window()
win.connect("destroy", Gtk.main_quit)
win.set_default_size(200)  # BUG: deliberately missing second argument
win.show_all()
Gtk.main()
```

... the missing second argument to `set_default_size` could be reported with a meaningful error message when using exact signatures:

``` sh
$ pylint example.py 
************* Module example
example.py:10:0: E1120: No value for argument 'height' in method call (no-value-for-parameter)

Your code has been rated at 3.75/10 (previous run: 3.75/10, +0.00)
```

## Risks

While this feature would substantially improve linting analysis, it would come at a noticeably higher computational cost during module loading time to generate the signatures.

### Linting time for the `example.py` above
- 5.63 sec
- 8.87 sec (+58%)

### Linting time for [Apport's `tests/system/test_ui_gtk.py`](https://github.com/canonical/apport/blob/main/tests/system/test_ui_gtk.py)
- 8.71 sec
- 11.45 sec (+31%)

Refs #2930

### DIFF
--- a/astroid/brain/brain_gi.py
+++ b/astroid/brain/brain_gi.py
@@ -83,6 +83,35 @@ def _gi_is_method_call(obj):
     return inspect.ismethod(obj) or inspect.ismethoddescriptor(obj)
 
 
+def _gi_replace_nonbuiltin_defaults(sig: inspect.Signature) -> inspect.Signature:
+    """
+    Replace non-builtin default parameter values from signature. This is needed
+    to obtain valid Python syntax when serializing the signature to string.
+    Non-builtin default values will produce signatures like this:
+
+        - (argname : IntEnum = <IntEnum.CONST: 0>, ...)
+        - (argname : ObjectType = <ObjectType object at 0xdeadbeefcafe>, ...)
+
+    In that case, replace the default value by 'None' and strip the type
+    annotation to produce a valid signature: (argname=None, ...)
+    """
+
+    def is_empty_or_primitive_builtin(obj) -> bool:
+        return obj == inspect.Parameter.empty or (
+            obj.__class__.__module__ == "builtins"
+            and isinstance(obj, (bool, str, int, float, type(None)))
+        )
+
+    params: list[inspect.Parameter] = []
+    for val in sig.parameters.values():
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas:
Thank you for working on astroid. I didn't check the changes in detail, but the performance impact is worrying. Generally a brain is "dumb" using hard coded and very simple code is fine. Can't we do that for lower cost here ?

oreiche:
Thank you for taking the time to have a brief look at this. I share your concern; that's why I reported the numbers. However, we probably can't speed this up much further.

Please keep in mind that the numbers I've reported (additional `~3` seconds) were measured for an end-to-end run with pylint. The additional time spent is only partly caused by the brain's signature extraction, but also by the extended checks that pylint is able to do now.

Maybe some more data for you to have the full picture:

**For Gtk** (`example.py`), the time spent for generating all signatures goes up from `0.013` to `1.355` seconds. The remaining extra `~1.6` seconds are not spent in the brain. Furthermore, almost all of the additional time in the brain is caused by the `inspect.signature()` call, which is run for all 49807 functions and 48942 methods in Gtk. The sheer amount of calls is part of the issue, we probably can't avoid.

**For GLib**, a significantly smaller library (1713 functions and 2505 methods), the time spent in the brain for generating all signatures "only" goes up from `0.001` to `0.087` seconds.

Anyway, this PR is merely a suggestion from an idea I had since my fix for the pyobject method classification. Feel free to close it if you think this enhancement is not worth the additional cost. :smile:

PR #3047, opened 2026-05-09
fix: handle Python 3.15 KW_ONLY and namespace .pth test behavior

SAY-5:
Fix the four Python 3.15-dev failures currently showing up in CI without broadening the walrus PR scope.

- `dataclasses.KW_ONLY` now infers as `builtins.sentinel` on 3.15, so the existing sentinel check treated it like a real dataclass field. Recognize the sentinel by its dataclasses import binding as a compatibility fallback.
- The legacy `foogle_*.pth` test fixture now executes under 3.15 without a usable `sitedir` frame local, so the namespace-package tests fail before astroid runs. Load that test fixture through a tiny helper that provides the expected local instead of depending on `site.addpackage()` internals.

Verification:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3.15 -m pytest -q tests/brain/test_dataclasses.py::test_kw_only_sentinel tests/test_manager.py::AstroidManagerTest::test_namespace_and_file_mismatch tests/test_manager.py::AstroidManagerTest::test_namespace_package_pth_support tests/test_manager.py::AstroidManagerTest::test_nested_namespace_import`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q tests/brain/test_dataclasses.py::test_kw_only_sentinel tests/test_manager.py::AstroidManagerTest::test_namespace_and_file_mismatch tests/test_manager.py::AstroidManagerTest::test_namespace_package_pth_support tests/test_manager.py::AstroidManagerTest::test_nested_namespace_import`

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -571,10 +571,27 @@ def _get_field_default(field_call: nodes.Call) -> _FieldDefaultReturn:
 def _is_keyword_only_sentinel(node: nodes.NodeNG) -> bool:
     """Return True if node is the KW_ONLY sentinel."""
     inferred = safe_infer(node)
-    return (
-        isinstance(inferred, bases.Instance)
-        and inferred.qname() == "dataclasses._KW_ONLY_TYPE"
-    )
+    if not isinstance(inferred, bases.Instance):
+        return False
+    if inferred.qname() == "dataclasses._KW_ONLY_TYPE":
+        return True
+    if inferred.qname() != "builtins.sentinel":
+        return False
+    if isinstance(node, nodes.Name):
+        _, assignments = node.lookup(node.name)
+        return any(
+            isinstance(assignment, nodes.ImportFrom)
+            and assignment.modname == "dataclasses"
+            and any(imported == "KW_ONLY" for imported, _ in assignment.names)
+            for assignment in assignments
+        )
+    if isinstance(node, nodes.Attribute) and node.attrname == "KW_ONLY":
+        inferred_expr = safe_infer(node.expr)
+        return (
+            isinstance(inferred_expr, nodes.Module)
+            and inferred_expr.qname() == "dataclasses"
+        )
+    return False
 
 
 def _is_init_var(node: nodes.NodeNG) -> bool:
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -39,17 +39,16 @@ def _load_namespace_package_pth(pth: str) -> None:
     """Execute a test .pth file with a `sitedir` local in scope.
 
     The .pth fixture reads `sys._getframe(1).f_locals['sitedir']`, so the
-    name must exist as a real local in this function's frame; static
-    analyzers cannot see the use through `exec`.
+    name must exist as a real local in this function's frame; the read
+    happens via `sys._getframe` inside `exec()`d code below.
     """
-    # `sitedir` is read by exec()'d .pth code via sys._getframe; intentional.
-    sitedir = str(
-        resources.RESOURCE_PATH
-    )  # pylint: disable=unused-variable  # noqa: F841
+    sitedir = str(resources.RESOURCE_PATH)
     with (resources.RESOURCE_PATH / pth).open(encoding="utf-8") as pth_file:
         for line in pth_file:
             line = line.strip()
             if line and not line.startswith("#"):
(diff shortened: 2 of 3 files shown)


tests/test_manager.py
```diff
@@ -36,6 +35,16 @@ def _get_file_from_object(obj) -> str:
...
 
 
+def _load_namespace_package_pth(pth: str) -> None:
+    """Execute a test .pth file with a real sitedir local."""
+    sitedir = str(resources.RESOURCE_PATH)
+    with (resources.RESOURCE_PATH / pth).open(encoding="utf-8") as pth_file:
+        for line in pth_file:
+            line = line.strip()
+            if line and not line.startswith("#"):
+                exec(line)
```
Pierre-Sassoulas: Can you explain what you want to do here ?

tests/test_manager.py
```diff
@@ -36,6 +35,17 @@ def _get_file_from_object(obj) -> str:
     return obj.__file__
 
 
+def _load_namespace_package_pth(pth: str) -> None:
+    """Execute a test .pth file with a real sitedir local."""
+    sitedir = str(resources.RESOURCE_PATH)
+    _ = sitedir
```
Pierre-Sassoulas:
```suggestion
```

tests/test_manager.py
```diff
@@ -36,6 +35,17 @@ def _get_file_from_object(obj) -> str:
...
+    with (resources.RESOURCE_PATH / pth).open(encoding="utf-8") as pth_file:
+        for line in pth_file:
+            line = line.strip()
+            if line and not line.startswith("#"):
+                exec(line)  # pylint: disable=exec-used
```
Pierre-Sassoulas: Why are we using an exec here ?

SAY-5: The .pth fixture is multi-statement Python that reads `sys._getframe(1).f_locals['sitedir']`, which the old `site.addpackage()` path supplied; after the 3.15 site.py refactor that local isn't there in our test frame, so we run the same statement ourselves with a real `sitedir` local in scope. A plain `import` doesn't work because the line is the namespace-wiring program the fixture exists to exercise. Happy to switch to a non-exec approach (e.g. parsing the path manually) if you'd prefer.

tests/test_manager.py
```diff
@@ -39,17 +39,16 @@ def _load_namespace_package_pth(pth: str) -> None:
...
-    sitedir = str(
-        resources.RESOURCE_PATH
-    )  # pylint: disable=unused-variable  # noqa: F841
+    sitedir = str(resources.RESOURCE_PATH)
     with (resources.RESOURCE_PATH / pth).open(encoding="utf-8") as pth_file:
         for line in pth_file:
             line = line.strip()
             if line and not line.startswith("#"):
+                # `sitedir` is read by .pth code via sys._getframe(1).f_locals.
+                _ = sitedir  # mark as used for static analyzers
```
Pierre-Sassoulas: Use a noqa / pylint: disable instead in this case (but we're going to remove the exec so the specific fix will become outdated).

tests/test_manager.py
Pierre-Sassoulas: Let's parse instead yes, this feel like a potential security breach (without looking too much into it).

Pierre-Sassoulas (changes requested):

astroid/brain/brain_dataclasses.py
```diff
@@ -571,10 +571,27 @@ def _get_field_default(field_call: nodes.Call) -> _FieldDefaultReturn:
...
-    return (
-        isinstance(inferred, bases.Instance)
-        and inferred.qname() == "dataclasses._KW_ONLY_TYPE"
-    )
+    if not isinstance(inferred, bases.Instance):
+        return False
+    if inferred.qname() == "dataclasses._KW_ONLY_TYPE":
+        return True
+    if inferred.qname() != "builtins.sentinel":
+        return False
```
Pierre-Sassoulas, line 579: This part is only going to be false on python 3.15 and above, right ? (So everything that follow can't be reached except on 3.15). We probably need a lot of 3.15 specific testing added for what follows.

Pierre-Sassoulas (changes requested):
LGTM, thank you. Would you mind covering the new code with automated tests ?

Pierre-Sassoulas (approved):
Thank you for your work on astroid !

PR #3048, opened 2026-05-10
Performance: cache ClassDef.ancestors() transitive walk

### LINKED ISSUE #1115: slow pylint performance with a hot spot in astroid

Pierre-Sassoulas:
## Type of Changes

## Description

Split out of the original four-optimization PR, as requested — this one now contains **only** the `ClassDef.ancestors()` cache, in a single commit with its tests.

The recursive walk in ``ancestors(recurs=True)`` re-resolved shared base classes on every call, amplifying cost on deep MRO chains. Cache the materialized tuple on the instance so each `ClassDef` pays for its ancestors once, and the cache dies with the instance when the manager drops the AST.

``context`` is intentionally not part of the key: the result is path-independent and the walk's own ``yielded`` set handles cycle prevention.

`TestAncestorsCaching` covers the corner cases: a cache hit reuses the materialized tuple, `recurs=False` bypasses the cache entirely, a cyclic hierarchy (`string.Template = A`) unwinds through the `_COMPUTING_ANCESTORS` sentinel instead of recursing forever, and an exception mid-walk clears the sentinel so a transient failure cannot poison the cache.

Measured on pandas/core/frame.py (interleaved A/B, n=4):

    baseline 21.34s ± 0.18  ->  patched 20.48s ± 0.13   (-4.0%)

Cache hit rate on the same run: 98% (66k hits / 1.2k misses on 1k distinct ClassDefs). Larger speedups expected on codebases with deeper MROs (SQLAlchemy/Pydantic-shaped projects).

The other three optimizations are now standalone PRs, each reviewable independently:

- #3166 — cache `ClassDef._find_metaclass()` per node
- #3167 — collapse 19 builtin transforms into a single `Call` dispatcher
- #3168 — bypass `__init__` in `InferenceContext.clone()`

Note that CodSpeed's earlier +9.65% on this PR was the *combined* figure for all four; taken alone this change is the -4.0% above.

Closes #1115


Pierre-Sassoulas:
@DanielNoord I'd like to merge this before release, great performance improvements, confirmed in the original issue https://github.com/pylint-dev/astroid/issues/1115#issuecomment-4421981536, and it's been ready for a long time.

DanielNoord (commented):
Would it be possible to split out the individual performance improvements into separate PRs? This seems to include 4.
With an upcoming long holiday weekend I don't have a lot of time for reviews. Context switching to a bigger PR is probably something I wouldn't get to before the end of next week. I can probably make some time for smaller PRs in between meetings tomorrow :)

Pierre-Sassoulas:
Done, there's this one still, and now #3166, #3167 and #3168

DanielNoord (commented):
Why do we go via `__dict__` and not just set this as an attribute? Pehaps with weak links

PR #3049, opened 2026-05-10
fix: Avoid parsing pathological type comments

### LINKED ISSUE #2993: MemoryError when parsing fuzzed type comment
### Steps to reproduce

The following fuzzed code triggers a `MemoryError` in astroid:

```python
a=b # type:i{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{
```

This bug was discovered by OSS-Fuzz: https://issues.oss-fuzz.com/issues/489780714 (report not public yet)

### Current behavior
```python-traceback
  File "astroid/run_parse.py", line 7, in <module>
    astroid.builder.parse(code)
  File "astroid/astroid/builder.py", line 313, in parse
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 155, in string_build
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 210, in _data_build
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 174, in visit_module
  File "astroid/astroid/rebuilder.py", line 174, in <listcomp>
     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 493, in visit
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 717, in visit_assign
    type_annotation = self.check_type_comment(node, parent=newnode)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 642, in check_type_comment
    type_comment_ast = self._parser_module.parse(node.type_comment)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/_ast.py", line 30, in parse
    return ast.parse(string, type_comments=type_comments)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/python3.11/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
MemoryError
```

### Expected behavior
No crash

### Version
85226dd78d

JamBalaya56562:
## Summary

A type comment such as `# type:i{{{{...` (the letter `i` followed by ~150+ opening braces, e.g. produced by OSS-Fuzz) makes Python's `ast.parse` raise `MemoryError`. `check_type_comment` and `check_function_type_comment` only caught `SyntaxError`, so the error propagated up through `_data_build` and crashed module parsing — taking down any tool built on astroid (pylint, IDE plugins, etc.) along with it.

This PR widens the exception handler to also catch `MemoryError` and `ValueError`, treating such inputs as invalid type comments (skip them). The pattern mirrors #2762's fix for the analogous f-string case (`f'{0:11111111111}'`).

## Reproduction

Live in-browser reproduction (no local install required):
**<https://aletheia-works.github.io/vivarium/repro/astroid/2993/>**

Minimal local repro:

```python
import astroid
astroid.parse("a = b # type:i" + "{" * 200)
# Before this PR: MemoryError
# After this PR: returns a Module with type_annotation == None
```

## Changes

- `astroid/rebuilder.py` — catch `(SyntaxError, ValueError, MemoryError)` in `check_type_comment` and `check_function_type_comment`.
- `tests/test_nodes.py` — two regression tests covering both assign and function type comments.
- `ChangeLog` — entry under 4.2.0 with reproduction link, `Closes #2993`.

## Test plan

- [x] `pytest tests/test_nodes.py -k type_comment` — 11 passed
- [x] `pytest tests/test_nodes.py tests/test_builder.py tests/test_regrtest.py` — 285 passed, 14 skipped
- [x] Manual repro: `astroid.parse("a = b # type:i" + "{" * 200)` returns a Module with `type_annotation == None` instead of raising `MemoryError`.

## See the fix working before you merge

[Vivarium](https://github.com/aletheia-works/vivarium) (an independent third-party reproduction platform) runs the **same fuzzed input** against two astroid builds **side-by-side in one Pyodide tab**:

  <https://aletheia-works.github.io/vivarium/repro/astroid/2993/>

| Variant | Source | Outcome |
| --- | --- | --- |
| **Baseline** | `astroid==4.1.2` from PyPI | `crashed: true`, `exception_type: "MemoryError"` |
| **Fix candidate** | Wheel built from this PR's branch | `crashed: false`, `exception_type: null` |

The page installs both wheels with `micropip` and runs the probe against each in the reviewer's browser — no local astroid install, no `git clone`. Vivarium's CI rebuilds the fix-candidate wheel from the latest commit on this branch on every deploy, so the verification stays in lockstep with this PR.

## Related

- Closes #2993
- OSS-Fuzz: <https://issues.oss-fuzz.com/issues/489780714> (report not yet public)
- Mirrors the exception-broadening pattern from #2762 for f-strings

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,15 @@ Release date: TBA
 
   Closes #2646
 
+* Catch ``MemoryError`` (and ``ValueError``) when validating type comments
+  with ``ast.parse``. Pathological type comments produced by fuzzers (e.g.
+  ``# type: i{{{{{{{...``) previously crashed parsing with a ``MemoryError``;
+  astroid now treats them as invalid type comments and skips them.
+
+  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
+
+  Closes #2993
+
 * Fix ``TypeError`` in ``brain_random`` when ``random.sample`` is called with a
   sequence containing nodes whose ``__init__`` does not accept ``lineno``
   (e.g. ``Module``). The clone helper now filters init params to those the
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,15 +13,15 @@ Release date: TBA
 
   Closes #2646
 
-* Reject type comments with pathologically deep bracket nesting before
-  invoking ``ast.parse``. Inputs such as ``# type: i{{{{{{{...`` previously
-  crashed module parsing with ``MemoryError`` (or ``RecursionError`` on some
-  runtimes) because CPython's parser exhausts memory before it can return a
-  clean ``SyntaxError``. ``check_type_comment`` and
-  ``check_function_type_comment`` now skip any type comment whose bracket
-  depth exceeds ``TYPE_COMMENT_MAX_BRACKET_DEPTH`` (100, well above any
-  realistic type expression and well below CPython's internal limit) without
-  attempting to parse it.
+* Catch ``MemoryError``/``RecursionError`` (and ``ValueError``) when validating
+  type comments with ``ast.parse``. Pathological type comments produced by
+  fuzzers (e.g. ``# type: i{{{{{{{...``) previously crashed parsing with a
+  ``MemoryError`` or ``RecursionError`` depending on the runtime; astroid now
+  treats them as invalid type comments and skips them, mirroring the f-string
+  fix from #2762.
+
+  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
+  (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
 
   Closes #2993
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,8 +20,6 @@ Release date: TBA
   treats them as invalid type comments and skips them, mirroring the f-string
   fix from #2762.
 
-  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
-  (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
   Closes #2993
 
 * Fix ``TypeError`` in ``brain_random`` when ``random.sample`` is called with a
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,15 +13,15 @@ Release date: TBA
 
   Closes #2646
 
-* Catch ``MemoryError``/``RecursionError`` (and ``ValueError``) when validating
-  type comments with ``ast.parse``. Pathological type comments produced by
-  fuzzers (e.g. ``# type: i{{{{{{{...``) previously crashed parsing with a
-  ``MemoryError`` or ``RecursionError`` depending on the runtime; astroid now
-  treats them as invalid type comments and skips them, mirroring the f-string
-  fix from #2762.
-
-  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
-  (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
+* Reject type comments with pathologically deep bracket nesting before
+  invoking ``ast.parse``. Inputs such as ``# type: i{{{{{{{...`` previously
+  crashed module parsing with ``MemoryError`` (or ``RecursionError`` on some
+  runtimes) because CPython's parser exhausts memory before it can return a
+  clean ``SyntaxError``. ``check_type_comment`` and
+  ``check_function_type_comment`` now skip any type comment whose bracket
+  depth exceeds ``TYPE_COMMENT_MAX_BRACKET_DEPTH`` (100, well above any
+  realistic type expression and well below CPython's internal limit) without
+  attempting to parse it.
 
   Closes #2993
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,7 +22,6 @@ Release date: TBA
 
   Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
   (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
-
   Closes #2993
 
 * Fix ``TypeError`` in ``brain_random`` when ``random.sample`` is called with a
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,12 +13,15 @@ Release date: TBA
 
   Closes #2646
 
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (changes requested):
Thank you for working on astroid ! Catching the memory error after the fact is preventing the crash but not the right way to go about this. We should make the parsing better or if we can't reasonably parse this pathological case cut short without parsing everything.

JamBalaya56562:
> Thank you for working on astroid ! Catching the memory error after the fact is preventing the crash but not the right way to go about this. We should make the parsing better or if we can't reasonably parse this pathological case cut short without parsing everything.

Thanks for the review, that makes sense.

I changed the approach so we no longer rely on catching `MemoryError` / `RecursionError` after calling `ast.parse`. The latest version now short-circuits before parsing: it does a small linear scan of the type comment and skips parsing when bracket nesting exceeds `TYPE_COMMENT_MAX_BRACKET_DEPTH`.

So `check_type_comment` and `check_function_type_comment` now return `None` for the pathological fuzzer-style input before it reaches CPython's parser, and their exception handling is back to `SyntaxError` only.

I also added regression coverage for both assignment and function type comments, plus a positive test to make sure deeply nested but valid type comments are still parsed. The Vivarium reproduction has also been updated to verify the new behavior against the latest PR branch.

DanielNoord:
@Pierre-Sassoulas I actually suggested the original solution in the linked issue. I don't fully understand why it isn't okay. We do the same for `SyntaxError`? Instead of "cutting short without parsing" we just skip this particular comment. That seems okay for me?

Pierre-Sassoulas:
Catching the memory error means we're going to first run out of memory, which I think is bad. Skipping would avoid pylint taking all the memory in this particular case. But cpython can get fixed eventually so the workaround would then become a problem (and it doesn't catch all the possible kind of memory error either). Are we really only skipping this one comment in case of memory error ?

DanielNoord:
> Catching the memory error means we're going to first run out of memory, which I think is bad.

There are other cases where we also catch `MemoryError`.

> Skipping would avoid pylint taking all the memory in this particular case. But cpython can get fixed eventually so the workaround would then become a problem (and it doesn't catch all the possible kind of memory error either). Are we really only skipping this one comment in case of memory error ?

I think so yeah? Same as with the `SyntaxError` I guess.

Pierre-Sassoulas:
In that case you're right and the fix should be in cpython, my bad.

Pierre-Sassoulas (changes requested):
Let's revert my suggestion, original fix was the way to go.

JamBalaya56562:
> Let's revert my suggestion, original fix was the way to go.

Thanks for the follow-up, @Pierre-Sassoulas.
Agreed — the proactive-detection refactor was overreach on my part.
I've reverted both 513cbf6 (the bracket-depth detection) and the accompanying pre-commit autofix, so the diff is back to the original try/except (SyntaxError, ValueError, MemoryError, RecursionError) approach from 04b0867 + 5c904f2.
Let me know if anything else needs tweaking.

ChangeLog
```diff
@@ -13,6 +13,18 @@ Release date: TBA
...
+* Catch ``MemoryError``/``RecursionError`` (and ``ValueError``) when validating
+  type comments with ``ast.parse``. Pathological type comments produced by
+  fuzzers (e.g. ``# type: i{{{{{{{...``) previously crashed parsing with a
+  ``MemoryError`` or ``RecursionError`` depending on the runtime; astroid now
+  treats them as invalid type comments and skips them, mirroring the f-string
+  fix from #2762.
+
+  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
+  (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
+
```
```suggestion
```

ChangeLog
```diff
@@ -13,6 +13,17 @@ Release date: TBA
...
 
+* Catch ``MemoryError``/``RecursionError`` (and ``ValueError``) when validating
+  type comments with ``ast.parse``. Pathological type comments produced by
+  fuzzers (e.g. ``# type: i{{{{{{{...``) previously crashed parsing with a
+  ``MemoryError`` or ``RecursionError`` depending on the runtime; astroid now
+  treats them as invalid type comments and skips them, mirroring the f-string
+  fix from #2762.
+
+  Reproduction: https://aletheia-works.github.io/vivarium/repro/astroid/2993/
+  (recipe: ``aletheia-works/vivarium`` ``src/layer1_wasm/astroid-2993``).
```
```suggestion
```

Pierre-Sassoulas (approved):
Thank you for your work on astroid !

DanielNoord:
@Pierre-Sassoulas Did you intentionally merge those changes to `tests/test_manager.py`? It feels out of scope?

Pierre-Sassoulas:
You're right, it was a piece of work related to setuptools 82+ for python 3.15 not having pkg_ressource.

DanielNoord:
So, should we revert on `main`? Or is it fine like this?

Pierre-Sassoulas:
I think a better fix is to use setuptools < 82 so we have pkg_resource or add pkg_resource to the dev dependencies. I'll do a PR for this. (So we don't lose those two tests)

Pierre-Sassoulas:
(And nice catch, impressive :) !)

DanielNoord:
I only had 200 emails to go through from a certain Pierre after this weekend 😄

PR #3061, opened 2026-05-19
brain/dataclasses: add inference tip for dataclasses.replace()

ego-ipse:
## Summary

  `dataclasses.replace(obj, **changes)` was inferred by following the stdlib body of
  `replace()` / `_replace()`. That body resolves `obj.__class__` via the metaclass
  chain, which trips when `obj` is an instance of a class whose bases include a
  subscripted generic (e.g. `IrCollection(Base[T])`) — the `Subscript` node causes
  `helpers.object_type` to return `Uninferable`, propagating up to the entire
  `replace()` call.

  The same bug manifests differently for old-style `Generic[T]` vs PEP 695
  `class Foo[T]` syntax: old-style returns `Uninferable` (masked error), PEP 695
  returns the wrong class (visible false positives in pylint E1101).        

  ## Fix

  Add a dedicated inference tip for `dataclasses.replace` in `brain_dataclasses.py`
  that short-circuits the stdlib body entirely: it infers the first argument,
  determines its type, and yields a fresh instance of that class. This is the same
  pattern used by other brain plugins (`isinstance`, `type()`, etc.) and correctly
  propagates the concrete caller type through `replace()` — including across
  `Self`-annotated method chains.

  The tip handles both call forms:
  - `dataclasses.replace(obj, ...)` (attribute access)
  - `from dataclasses import replace; replace(obj, ...)` (bare name)

  Calls that raise `TypeError` at runtime yield `Uninferable` rather than a
  fabricated instance. In particular, passing the class itself —
  `dataclasses.replace(Point)` — is `TypeError: replace() should be called on
  dataclass instances`, so there is no value to infer. Letting that case fall
  through to default inference is not equivalent: the stdlib body computes
  `obj.__class__`, which for a class object is `type`, and yields
  `Instance of builtins.type`.

  An ambiguous first argument (`safe_infer()` returning `None` because several types
  are possible) also yields `Uninferable`, which avoids E1101 false positives on
  union-like flows.                                                                                                      

  ## Tests

  Ten tests in `tests/brain/test_dataclasses.py`:

  | Test | What it covers |
  |---|---|
  | `test_replace_returns_instance_of_caller_type` | Basic case: result is the passed type |
  | `test_replace_returns_subclass_instance` | old-style `Generic[T]` base — concrete subclass propagated |
  | `test_replace_pep695_generic_base` | PEP 695 `class Base[T]` — same (skipped on <3.12) |
  | `test_replace_bare_name_form` | `from dataclasses import replace` form |
  | `test_replace_frozen_dataclass` | `frozen=True` is orthogonal, still works |
  | `test_replace_uninferable_first_arg` | First arg can't be inferred → `Uninferable` |
  | `test_replace_ambiguous_first_arg` | First arg has several possible types → `Uninferable` |
  | `test_replace_classdef_first_arg` | First arg is the class itself → `Uninferable` (TypeError at runtime) |
  | `test_replace_non_instance_non_class_falls_back` | First arg infers to a module → default inference, `Uninferable` |
  | `test_replace_no_args` | No args → `Uninferable` (TypeError at runtime) |

  ## Changelog

  Added as a towncrier fragment, `doc/whatsnew/fragments/3061.bugfix`.


astroid/brain/brain_dataclasses.py
```diff
@@ -527,6 +527,47 @@ def _looks_like_dataclasses(node: nodes.Module) -> bool:
...
+    if isinstance(inferred_obj, UninferableBase) or inferred_obj is None:
+        yield Uninferable
+        return
+    if isinstance(inferred_obj, bases.Instance):
+        yield inferred_obj._proxied.instantiate_class()
+        return
+    if isinstance(inferred_obj, nodes.ClassDef):
+        yield inferred_obj.instantiate_class()
+        return
+    raise UseInferenceDefault
```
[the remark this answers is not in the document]
ego-ipse: I do not think this is correct

safe_infer returning None for ambiguous callers (multiple inferred values) is the right conservative behaviour here. When obj could be Foo | Bar depending on a branch that astroid can't resolve statically, then dataclasses.replace(obj, ...) is genuinely indeterminate - yielding Uninferable avoids false positives from pylint's E1101.
 
Iterating the full result set and emitting one instance per inferred type would be more complete in theory, but it is also inconsistent with every other brain plugin in the codebase: all of them use either ext(arg.infer(context)) with a try/except or safe_infer, none iterate over all inferred values. Introducing that pattern here would be out of scope and would risk generating a flood of results for callers with complex control flow.
 
The fix this PR targets is the single-concrete-type case that was incorrectly returning Uninferable due to subscripted generic bases in the MRO chain - that is a resolution failure, not ambiguity, and the inference tip addresses it correctly.

@DanielNoord  please let me know if you disagree and I'll modify it :)

tests/brain/test_dataclasses.py
```diff
@@ -1332,3 +1333,170 @@ class ChildConfig(BaseConfig[T]):
...
+
+
+def test_replace_returns_subclass_instance() -> None:
+    """dataclasses.replace propagates the concrete subclass type, not the base."""
+    node = astroid.extract_node("""
+    import dataclasses
+    from dataclasses import dataclass
+    from typing import ClassVar, Generic, Self, TypeVar
+    from abc import ABC
+
```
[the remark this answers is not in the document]
ego-ipse: Ack. Applying

DanielNoord:
@ego-ipse Could you rebase this on `main`? :)

ego-ipse:
@DanielNoord rebased :)

DanielNoord:
Unfortunately this still has merge conflicts 😅

ego-ipse:
@DanielNoord yup. Something in the meantime got merged to main. Should be fixed now.

tests/brain/test_dataclasses.py
```diff
@@ -1385,3 +1386,174 @@ class B(A):
...
+    """When the first arg is a class itself, replace yields an instance of it."""
+    node = astroid.extract_node("""
+    import dataclasses
+    from dataclasses import dataclass
+
+    @dataclass
+    class Point:
+        x: float = 0.0
+
+    dataclasses.replace(Point)  #@
```
DanielNoord, line 1515: This raises a `TypeError` and should therefore be `Uninferable`.

tests/brain/test_dataclasses.py
```diff
@@ -1385,3 +1386,174 @@ class B(A):
...
+
+def test_replace_no_args_does_not_crash() -> None:
+    """dataclasses.replace with no args should not raise an uncaught exception."""
+    node = astroid.extract_node("""
+    import dataclasses
+    dataclasses.replace()  #@
+    """)
+    # Should not raise; result may be Uninferable or an InferenceError caught upstream
+    inferred = list(node.infer())
+    assert len(inferred) >= 1
```
DanielNoord: What is in this list?

tests/brain/test_dataclasses.py
```diff
@@ -1385,3 +1386,174 @@ class B(A):
...
+
+def test_replace_non_instance_non_class_falls_back() -> None:
+    """A first arg that is neither Instance nor ClassDef raises UseInferenceDefault."""
+    node = astroid.extract_node("""
+    import os
+    import dataclasses
+    dataclasses.replace(os)  #@
+    """)
+    # Should not crash; default inference takes over.
+    list(node.infer())
```
DanielNoord: Was is the result here?

uv.lock
DanielNoord, line 1: Please remove

astroid/brain/brain_dataclasses.py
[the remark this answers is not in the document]
DanielNoord: Yes this seems fine for now, it is a sane default for this PR. Can you add a test for it though?

DanielNoord (commented):
Thanks for this change :)

Could you add a changelog entry?

tests/brain/test_dataclasses.py
ego-ipse: This required a small tweak in the src side.

tests/brain/test_dataclasses.py
ego-ipse: Added explicitly
```
    inferred = list(node.infer())
    assert inferred == [Uninferable]
```

ego-ipse:
@DanielNoord all issues should be addressed :)

tests/brain/test_dataclasses.py
ego-ipse: Made it explicit. If there's no argument, there's nothing to infer - and it's going to be an error
```
    inferred = list(node.infer())
    assert inferred == [Uninferable]
```

DanielNoord (approved):

PR #3062, opened 2026-05-19
perf: defer cold-path imports to shorten ``import astroid``

Pierre-Sassoulas:
## Summary

`import astroid` eagerly imported three modules that a normal lint run
never actually needs. This PR defers them:

- **`logging`** — `modutils` and `raw_building` imported it only for a
  module-level `logger`, used solely when an imported module emits
  stderr/stdout (a rare diagnostic path). Moved the import + logger
  creation into that branch; this also drops `logging`'s transitive
  `traceback`/`weakref` chain.
- **`pprint`** — `context`, `interpreter.objectmodel` and `nodes.node_ng`
  imported it only to format debug `__str__`/`__repr__`/`repr_tree`
  output. Moved into those methods.
- **`astroid.typing` in `exceptions`** — `InferenceResult` /
  `SuccessfulInferenceResult` are used only in annotations; moved the
  import under `TYPE_CHECKING`.

Each module is genuinely eliminated, not relocated: a normal parse +
inference run never reaches these paths. `import astroid` drops from 165
to 156 loaded modules.

Lazy brain-loading — the larger change — is split into #3069; the two
PRs touch disjoint files and can merge in any order.

## Test plan

- [x] Full test suite passes locally (1956 passed, 53 skipped, 16 xfailed)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Shorten ``import astroid`` by deferring imports only needed on cold paths.
+  ``logging`` (used by ``modutils`` and ``raw_building`` solely to report
+  stderr/stdout captured while importing a module) and ``pprint`` (used only
+  to format debug ``__repr__`` / ``repr_tree`` output) are now imported
+  lazily, and ``astroid.exceptions`` imports ``astroid.typing`` under
+  ``TYPE_CHECKING``.
+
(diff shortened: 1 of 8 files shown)


Pierre-Sassoulas:
I dropped the lazy import of inspect because we use it everywhere so deferring it is artificial, as soon as we parse anything it's going to be imported.

DanielNoord (approved):

PR #3075, opened 2026-05-20
fix: avoid uncaught TokenError when computing node positions

### LINKED ISSUE #2527: Uncaught TokenError while fuzzing
### Steps to reproduce

The following testcase from OSS-Fuzz raises a `TokenError` with Python 3.8.0 - 3.11.9:

```python
import astroid

code="\rclass\x0cj:(\n)"
astroid.builder.parse(code)
```

With Python 3.12.0+, the code no longer raises an exception.

OSS-Fuzz is currently running 3.8.3 and reports the exception as a crash:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67696 (report not public yet)

Given that Pylint catches `TokenErrors`, should the fuzzer on OSS-Fuzz also catch `TokenErrors` to avoid similar OSS-Fuzz reports?  Or should astroid handle this case more gracefully?

Here are the current exceptions that the fuzzer catches:
```python
  try:
    astroid.builder.parse(code)
  except (
      astroid.exceptions._NonDeducibleTypeHierarchy,
      astroid.modutils.NoSourceFile,
      astroid.exceptions.AstroidError,
      astroid.exceptions.UseInferenceDefault,
  ):
    pass
```
-- [Source](https://github.com/google/oss-fuzz/blob/be8fded65f99abd577dbe46ac4f0b8f8266a1a9b/projects/astroid/fuzz_parse.py#L29-L37)

### Current behavior
```
  File "corpus/a.py", line 5, in <module>
    astroid.builder.parse(code)
  File "lib/python3.11/site-packages/astroid/builder.py", line 303, in parse
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/builder.py", line 151, in string_build
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/builder.py", line 206, in _data_build
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/rebuilder.py", line 194, in visit_module
  File "lib/python3.11/site-packages/astroid/rebuilder.py", line 194, in <listcomp>
     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/rebuilder.py", line 483, in visit
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/rebuilder.py", line 853, in visit_classdef
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/astroid/rebuilder.py", line 136, in _get_position_info
    for t in generate_tokens(StringIO(data).readline):
  File "lib/python3.11/tokenize.py", line 525, in _tokenize
    raise TokenError("EOF in multi-line statement", (lnum, 0))
tokenize.TokenError: ('EOF in multi-line statement', (2, 0))
```

### Expected behavior
Suppressed fuzzer failures

### Version
d8dbc462243a65d72153ace2e7ac87daa9ffc1cf

Pierre-Sassoulas:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`RebuildVisitor._get_position_info` tokenizes a slice of the source to locate the keyword/name span of a class or function definition. On Python < 3.12, `tokenize.generate_tokens` can raise `TokenError` for malformed input (e.g. an unterminated bracket), which escaped as an uncaught non-`AstroidError` exception during the build.

Position computation now catches `TokenError` and treats such a node as having no position information, instead of crashing.

Closes #2527

### DIFF
--- a/tests/test_nodes_position.py
+++ b/tests/test_nodes_position.py
@@ -5,6 +5,8 @@
 from __future__ import annotations
 
 import textwrap
+from tokenize import TokenError
+from unittest import mock
 
 from astroid import builder, nodes
 
@@ -157,3 +159,18 @@ class TestNodePosition:
         e = ast_nodes[4]
         assert isinstance(e, nodes.FunctionDef)
         assert e.position == (14, 0, 14, 11)
+
+    @staticmethod
+    def test_position_malformed_tokenize() -> None:
+        """A ``TokenError`` from ``tokenize`` must not crash the build.
+
+        On Python < 3.12 malformed source could make ``generate_tokens``
+        raise a ``TokenError``; ``position`` is simply unavailable then.
+        """
+        with mock.patch(
+            "astroid.rebuilder.generate_tokens",
+            side_effect=TokenError("unexpected EOF in multi-line statement", (1, 0)),
+        ):
+            node = builder.extract_node("class A:  #@\n    ...")
+        assert isinstance(node, nodes.ClassDef)
+        assert node.position is None
--- a/tests/test_nodes_position.py
+++ b/tests/test_nodes_position.py
@@ -174,3 +174,16 @@ class TestNodePosition:
             node = builder.extract_node("class A:  #@\n    ...")
         assert isinstance(node, nodes.ClassDef)
         assert node.position is None
+
+    @staticmethod
+    def test_position_unnormalized_name() -> None:
+        """No position info when the name token never matches ``node.name``.
+
+        ``node.name`` is the NFKC-normalized identifier while ``tokenize``
+        yields the raw source spelling (here the ``fi`` ligature), so the
+        name token is never matched and no position can be computed.
+        """
+        node = builder.extract_node("class ﬁ:  #@\n    ...")
+        assert isinstance(node, nodes.ClassDef)
+        assert node.name == "fi"
+        assert node.position is None
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,13 @@ Release date: TBA
 
   References pylint-dev/pylint#872
 
+* Fix uncaught ``TokenError`` when building a class or function whose source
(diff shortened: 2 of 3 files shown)


tests/test_nodes_position.py
```diff
@@ -157,3 +159,31 @@ async def e():  #@
...
         assert e.position == (14, 0, 14, 11)
+
+    @staticmethod
+    def test_position_malformed_tokenize() -> None:
+        """A ``TokenError`` from ``tokenize`` must not crash the build.
+
+        On Python < 3.12 malformed source could make ``generate_tokens``
+        raise a ``TokenError``; ``position`` is simply unavailable then.
+        """
+        with mock.patch(
```
DanielNoord, line 170: Why are we patching here?

Pierre-Sassoulas: The crash does not reproduce above python 3.12.  If we want to catch this long term we need to fake the TokenError.

DanielNoord: Shall we only run on <=3.12 then? Does not seem sensible to test something that can't happen on newer versions.

Pierre-Sassoulas: A token error may still happen in the future, probably elsewhere unexpected.

Pierre-Sassoulas: I mean on another code construct, generate_token can raise, we just don't know what will keep raising long term.

DanielNoord (approved):

PR #3076, opened 2026-05-20
Fix: Resolve PEP 695 TypeVar bounds for member-lookup inference

ego-ipse:
## Type of Changes

<!-- Leave the corresponding lines for the applicable type of change: -->

  ## Summary                                                                                                                                                                            
                                                                                                                                                                                      
  PEP 695 type parameters (`def f[T: str](x: T)`, `class C[T: str]`) were                                                                                                               
  all inferred as a `Const(None)` placeholder by `generic_type_assigned_stmts`,
  introduced in #2433 to prevent `InferenceError` while evaluating                                                                                                                      
  `__class_getitem__`. The placeholder leaks downstream: any attribute or                                                                                                               
  member lookup on a `T`-typed expression resolves against `None`, so                                                                                                                   
  `pylint` reports false-positive `no-member` on perfectly valid code like                                                                                                              
  `x.upper()` inside `def f[T: str](x: T): ...`.                                                                                                                                        
                                                                                                                                                                                        
  Function parameters annotated with PEP 695 type parameters compound the                                                                                                               
  problem: `_arguments_infer_argname` never consulted annotations at all,                                                                                                               
  so even with the type parameter itself resolved, `x: T` infers to                                                                                                                     
  `Uninferable` and downstream lookups still fail.                                                                                                                                      
   
  ## Fix                                                                                                                                                                                
                                                                                                                                                                                      
  Two surgical changes in `astroid/protocols.py`:                                                                                                                                       
                                                                                                                                                                                      
  1. **`generic_type_assigned_stmts`** - when the type parameter is a
     `TypeVar` with a declared bound (`[T: Bound]`), yield an `Instance` of
     the bound's class instead of the `Const(None)` placeholder. Member                                                                                                                 
     lookups on `T`-typed expressions now resolve via the bound's                                                                                                                       
     interface. Unbounded `TypeVar`, constrained `TypeVar` (`T: (str, int)`),                                                                                                           
     `TypeVarTuple`, and `ParamSpec` continue to use the historical                                                                                                                     
     placeholder. Defensive against `InferenceError` from unresolvable bound                                                                                                            
     names (`T: Undefined`).                                                                                                                                                            
                                                                                                                                                                                        
  2. **`_arguments_infer_argname`** - after the `NoDefault` fallback, infer                                                                                                           
     the parameter's annotation. Forward only true `Instance` results - not                                                                                                             
     `Const` (its placeholder semantics) and not `ClassDef` (the plain
     `x: str` case, which has never driven inference in astroid). The                                                                                                                   
     `Const` exclusion is the load-bearing bit: it preserves the historical                                                                                                             
     "annotations don't drive argument inference" contract for everything                                                                                                               
     except PEP 695 TypeVar bounds.                                                                                                                                                     
                                                                                                                                                                                        
  The two changes are tightly coupled and must land together. Either alone                                                                                                            
  produces no behaviour change for downstream tools.                                                                                                                                    
                  
  A small private helper `_annotation_for_argname` scans `Arguments`'                                                                                                                   
  posonly / args / kwonly lists for the annotation of a given parameter.
                                                                                                                                                                                        
  ## Tests        
                                                                                                                                                                                        
  Eight new tests in `tests/test_type_params.py`:                                                                                                                                       
   
  | Test | What it covers |                                                                                                                                                             
  |---|---|       
  | `test_type_var_bound_resolves_to_instance` | Happy path: `T: str` parameter infers to `Instance(str)` |
  | `test_type_var_without_bound_stays_uninferable` | Unbounded `T` falls through (no regression) |
  | `test_type_var_bound_member_lookup` | End-to-end: `.upper()` resolves on `T: str`-typed expression |
  | `test_plain_class_annotation_does_not_drive_inference` | Regression guard for `test_infer_property_setter` — `x: str` still doesn't make `x` infer to a `str` instance |
  | `test_unannotated_parameter_stays_uninferable` | Param with no annotation stays `Uninferable` |
  | `test_unresolvable_bound_yields_const_placeholder` | `T: Undefined` falls back to `Const(None)` without crashing |
  | `test_annotation_for_argname_returns_none_for_none_name` | Helper defensive case (`name=None`) |
  | `test_annotation_for_argname_returns_none_when_no_match` | Helper no-match case |                                                                                                   
   
  All 1934 existing + new tests pass. Diff coverage on                                                                                                                                  
  `astroid/protocols.py` is 100%.
                                                                                                                                                                                        
  ## Downstream impact

  Verified on a real codebase: pylint rating on a file with PEP 695-bounded                                                                                                             
  TypeVars went from 9.32/10 → 10.00/10 with this patch - the false-positive
  `no-member` on `T: str`-typed expressions disappears.                                                                                                                                 
                  
  ## Compatibility

  Tests gated by the existing module-level `PY312_PLUS` skip. No PEP 696                                                                                                                
  default-syntax used; PY313 is not needed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,13 +7,14 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
-* Resolve PEP 695 ``TypeVar`` annotations to their declared bound so member
-  lookups on type-parameter-typed expressions succeed via the bound's
-  interface. Inside ``def f[T: str](x: T)`` or ``class C[T: str]: ...``,
-  ``x.upper()`` now infers correctly. Unbounded ``TypeVar``,
-  ``TypeVarTuple``, and ``ParamSpec`` continue to use the historical
-  ``Const(None)`` placeholder. Plain non-TypeVar annotations (``x: str``)
-  still do not drive argument inference.
+* Resolve PEP 695 ``TypeVar`` declared with a bound to an ``Instance`` of
+  the bound class. ``generic_type_assigned_stmts`` previously yielded a
+  ``Const(None)`` placeholder for every type parameter; for bounded
+  ``TypeVar`` it now yields an instance of the bound, so attribute access
+  on a ``T``-typed expression resolves via the bound's interface.
+  Unbounded ``TypeVar``, ``TypeVarTuple``, and ``ParamSpec`` continue to
+  use the placeholder; unresolvable bounds (e.g. ``[T: Undefined]``)
+  also fall back to the placeholder rather than crashing.
 
 * Fix ``AttributeError`` crash in ``starred_assigned_stmts`` when a starred
   unpacking target is an attribute (e.g. ``for *o.attr, x in ...``) rather
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,14 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Resolve PEP 695 ``TypeVar`` annotations to their declared bound so member
+  lookups on type-parameter-typed expressions succeed via the bound's
+  interface. Inside ``def f[T: str](x: T)`` or ``class C[T: str]: ...``,
+  ``x.upper()`` now infers correctly. Unbounded ``TypeVar``,
+  ``TypeVarTuple``, and ``ParamSpec`` continue to use the historical
+  ``Const(None)`` placeholder. Plain non-TypeVar annotations (``x: str``)
+  still do not drive argument inference.
+
 * Fix ``AttributeError`` crash in ``starred_assigned_stmts`` when a starred
   unpacking target is an attribute (e.g. ``for *o.attr, x in ...``) rather
   than a simple name.
--- a/astroid/protocols.py
+++ b/astroid/protocols.py
@@ -410,7 +410,47 @@ def _arguments_infer_argname(
         yield from self.default_value(name).infer(context)
         yield util.Uninferable
     except NoDefault:
+        # No default — fall back to the annotation when it infers to a
+        # PEP 695 TypeVar bound. ``generic_type_assigned_stmts`` yields
+        # an :class:`Instance` of the bound for bounded TypeVars; we
+        # forward that. Plain annotations like ``x: str`` infer to a
+        # :class:`ClassDef`, and unbounded TypeVars infer to a
+        # placeholder :class:`Const`; both are skipped to preserve
+        # astroid's historical "annotations don't drive argument
+        # inference" contract.
+        annotation = _annotation_for_argname(self, name)
+        if annotation is not None:
+            for inferred in annotation.infer(context=context):
+                if (
+                    isinstance(inferred, bases.Instance)
+                    and not isinstance(inferred, nodes.Const)
+                ):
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas:
Hey @ego-ipse, thank you for contributing to astroid. What is the issue we're trying to fix here ?

ego-ipse:
Yes @Pierre-Sassoulas . I found this one out while working on my pet transpilation project. 

So, PEP 695 says 

>  An "upper bound" type can be specified through the use of a type annotation expression. […] The specified upper bound type must be concrete.

Thus the contract is: every expression typed T is statically guaranteed to be an instance of the bound (or a subtype of it).
T: str means "treat me as some kind of str." Type checkers like pyright already enforce this; pylint should be able to use it too, via astroid.

What astroid was doing: astroid/protocols.py:generic_type_assigned_stmts — added in #2433 — resolves every PEP 695 type parameter (bounded, unbounded, all kinds) to a single placeholder: nodes.Const(None). The comment on it reads:
```
  """Hack. Return any Node so inference doesn't fail
  when evaluating __class_getitem__. Revert if it's causing issues."""
  yield nodes.Const(None)
```

Below the surface, this means:

  - Name('T').infer() - [Const(None)] — the upper bound is silently discarded.
  - x: T where T: str - x.infer() propagates through _arguments_infer_argname, which never consulted annotations at all, and yields Uninferable.
  - A member lookup like x.upper() resolves against None (no .upper()) or Uninferable (nothing knowable), depending on how the call site got there.


The code where I had the issue:
```
@dataclass(frozen=True, slots=True, repr=False)
class IrJoin[Ir_co: str = str](IrComposite[Ir_co]):

    _child_attrs: ClassVar[tuple[str, ...]] = ("children_op", "separator", "empty")
    children_op: IrNode[tuple[Ir_co, ...]]
    separator: IrNode[Ir_co]
    empty: IrNode[Ir_co]

    def eval(self, d: IrSelf, n: IrSelf, nc: tuple, /) -> Ir_co:
        """Evaluate items; join with ``separator`` or return ``empty``."""
        items = self.children_op.eval(d, n, nc)
        if not items:
            return self.empty.eval(d, n, nc)
        separator: Ir_co = self.separator.eval(d, n, nc)
        return cast(Ir_co, separator.join(items))
        #                            ^^^^^^^^^^^^ E1101: Instance of 'IrNode' has no 'join'
```
self.separator.eval(...) returns Ir_co, which is bounded by str — so .join(...) is guaranteed available. Pyright accepts it. Pylint refused.

Pierre-Sassoulas:
I think inferring a bounded TypeVar to an Instance of its bound is ok, but looking at the typing to infer something is not. Astroid does not look at the typing, it infer from reality and if reality doesn't exists it infer as uninferable. Changing that assumption is going to break expectation everywhere. Also please provide a snippet in an issue it's easier to understand your problem that way.

ego-ipse:
Hi @Pierre-Sassoulas thank you for the input. 

I was overreaching and not getting to the crux of the problem. I've opened issue https://github.com/pylint-dev/astroid/issues/3077 as it seems to get to the core of what I'm seeing. 

I'll be closing this PR as it does seem to no longer be valid

PR #3078, opened 2026-05-23
Add .pyi stub file support for inference and discovery

emmanuel-ferdman:
## Type of Changes

## Description

Following #3031 (ongoing work to add full `.pyi` stub support to astroid), this PR adds `.pyi` stub awareness to astroid. Stub files are auto-detected by path or via an explicit `is_stub` flag on `parse()`/`string_build()`. The rebuilder strips placeholder bodies (`...`, `pass`) from functions and normalizes ellipsis assignments so downstream checkers don't treat stubs like regular source. On the inference side, bare type annotations in stubs now resolve to class instances (with `ClassVar`/`Final`/`Annotated` unwrapping and forward-ref resolution), and stub function return annotations are used by `infer_call_result`. Module discovery also picks up standalone `.pyi` files, preferring `.py` when both exist.

FYI, the annotation-to-instance inference logic that previously lived in `brain_dataclasses` has been moved to `protocols.py` so it can be shared across the inference pipeline. 

Refs #3031


emmanuel-ferdman:
Coverage fails on guards - no clean way to add cases for them - let me know if you have elegent cases for them.

emmanuel-ferdman:
Hey @Pierre-Sassoulas, any chance we could include it in 4.2.0? It would benefit `pylint-qt` :)

astroid/interpreter/_import/spec.py
```diff
@@ -120,6 +120,7 @@ class ImportlibFinder(Finder):
     _SUFFIXES: Sequence[tuple[str, ModuleType]] = (
         [(s, ModuleType.C_EXTENSION) for s in importlib.machinery.EXTENSION_SUFFIXES]
         + [(s, ModuleType.PY_SOURCE) for s in importlib.machinery.SOURCE_SUFFIXES]
+        + [(".pyi", ModuleType.PY_SOURCE)]
```
Pierre-Sassoulas, line 123: This seems to contradict the fix in #3099, what do oyu think ?

emmanuel-ferdman: @Pierre-Sassoulas sorry, was off for a while. I think they complement each other rather than conflict - the `_SUFFIXES` line handles discovering standalone `.pyi` files (stub-only packages where no `.pyc` or `.py` exists alongside), which `get_source_file` can never solve since there's nothing to redirect from. For the overlapping case where `.pyc` and `.pyi` sit next to each other, both fixes solve it via different paths, but the ordering (`.pyi` before `.pyc`) is intentional and makes more sense for me - a stub is directly parseable while bytecode never is (astroid raises `AstroidImportError` if `PY_COMPILED` stays unresolved), so the finder should prefer the usable format. PR #3099 is still needed for direct callers and the general `.pyc` -> `.py` case. Happy to rebase on top of #3099 if you want to land that first.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -241,6 +242,7 @@ def __init__(
         path: Sequence[str] | None = None,
         package: bool = False,
         pure_python: bool = True,
+        is_stub: bool = False,
```
Pierre-Sassoulas: I think the actual type is bool | None here.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -1617,6 +1622,15 @@ def infer_call_result(
 
         first_return = next(returns, None)
         if not first_return:
+            if not self.body and self.root().is_stub:  # pylint: disable=no-member
```
Pierre-Sassoulas: I'm not sure why we need the disable here ?

astroid/protocols.py
```diff
@@ -451,6 +451,17 @@ def assign_assigned_stmts(
...
 ) -> Any:
+    # pylint: disable = too-many-boolean-expressions
+    if (
+        not assign_path
+        and isinstance(self, nodes.Assign)
+        and self.type_annotation is not None
+        and isinstance(self.value, node_classes.Const)
+        and self.value.value is None
+        and self.root().is_stub
+    ):
```
Pierre-Sassoulas: Can we do a ``_is_stub_ellipsis_assign(node)`` helper ?

tests/test_scoped_nodes.py
```diff
@@ -2755,6 +2755,90 @@ def f2():
...
+        assert results[0] is util.Uninferable
+
+    def test_stub_funcdef_forward_ref_return(self) -> None:
+        module = builder.parse(
+            """
+            class Parser:
+                @classmethod
+                def create(cls) -> "Parser": ...
+            """,
+            module_name="stub_mod",
```
Pierre-Sassoulas: This work around the fact that if we don't give "stub_mod" this is going to fail:

```python
src = "class Dog: ...\ndef f() -> Dog: ..."
parse(src, module_name="",      is_stub=True) # f() infers Uninferable (wrong)
parse(src, module_name="mymod", is_stub=True) # f() infers Instance of mymod.Dog
```

Pierre-Sassoulas (changes requested):
Thank you for working on this, it's pretty important. Sorry for the long time it took to review, it's hard to review as it's going to change az lot of things.

Could you first rebase on origin/main and add a changelog please ? 

Also I noticed that as_string now create SyntaxError: ``astroid.parse("def f() -> int: ...", is_stub=True).as_string()`` => ``'\ndef f() -> int:\n    \n\n'``  => SyntaxError: expected an indented block. Let's not mutate the AST aznd in infer_call_result test self._is_stub_placeholder_body(self.body) instead of ``not self.body``.

``X = ...`` is rewritten to ``X = None`` which is wrong, it should be either uninferable or kept as .``..``, the current way to do it is going to cause downstream issue when None has a semantic.

astroid/nodes/scoped_nodes/scoped_nodes.py
emmanuel-ferdman: Removed - stub detection uses `_is_stub_node()` from `protocols.py` which handles the type check internally.

tests/test_scoped_nodes.py
emmanuel-ferdman: Fixed - the empty-name check now tells synthetic typing modules apart from real unnamed modules

emmanuel-ferdman:
@Pierre-Sassoulas Thanks for the review. Rebased on main and added changelog. Removed all AST mutations - bodies and values stay as-is, `as_string()` produces valid syntax now. Stub placeholders are detected at inference time instead. Great suggestions/observations.

astroid/interpreter/_import/spec.py
Pierre-Sassoulas: Make sense, thank you

astroid/manager.py
```diff
@@ -198,6 +198,32 @@ def _can_load_extension(self, modname: str, location: str | None) -> bool:
...
+            and not self._can_load_extension(modname, found_spec.location)
+        ):
+            if not self.prefer_stubs or found_spec.location is None:
+                return self._build_stub_module(modname)
+            try:
+                source = get_source_file(
+                    found_spec.location, prefer_stubs=self.prefer_stubs
+                )
+            except NoSourceFile:
+                return self._build_stub_module(modname)
```
Pierre-Sassoulas, line 215: This branch isn't covered. We need an extension file with no adjacent .pyi/.py and prefer_stubs=True I think.

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -2372,12 +2400,16 @@ def getattr(
         if class_context:
             values += self._metaclass_lookup_attribute(name, context)
 
+        is_stub = protocols._is_stub_node(self)
```
Pierre-Sassoulas: (Comment from Claude that make sense afaiu: ) values also holds the locals of every ancestor, so the stub check has to be made on the statement, not on self. A regular .py class inheriting from a class defined in a .pyi loses the inherited bare annotations:

 stubbase.pyi
```
 class Base:
     color: str
```

 user.py
```
 from stubbase import Base

 class Sub(Base): ...

 Sub().color   # InferenceError, while Base().color infers str
```

 Checking the statement inside the loop fixes it, and is also cheaper: root() is only walked when a bare AnnAssign is actually met, instead of on every getattr call.
```python
         result: list[InferenceResult] = []
         for value in values:
             if isinstance(value, node_classes.AssignName):
                 stmt = value.statement()
                 # Ignore AnnAssigns without value, which are not attributes in the
                 # purest sense, except in stubs where they declare the attribute.
                 if (
                     isinstance(stmt, node_classes.AnnAssign)
                     and stmt.value is None
                     and not protocols._is_stub_node(stmt)
                 ):
                     continue
             result.append(value)
```

 I checked locally: the example above infers str with this change and the existing tests still pass. Could you add a test for the inheritance case? Something like:

```python
     def test_nonstub_subclass_inherits_stub_annassign(self) -> None:
         with tempfile.TemporaryDirectory() as tmp_dir:
             with open(os.path.join(tmp_dir, "stubbase.pyi"), "w", encoding="utf-8") as f:
                 f.write("class Base:\n    color: str\n")
             with resources.augmented_sys_path([tmp_dir]):
                 node = builder.extract_node(
                     """
                     from stubbase import Base
                     class Sub(Base): ...
                     Sub().color  #@
                     """,
                     module_name="user_mod",
                 )
                 inferred = next(node.infer())
                 assert isinstance(inferred, Instance)
                 assert inferred.name == "str"
```

emmanuel-ferdman: good claude :) fixed the check to use the module that declared the annotation

astroid/manager.py
emmanuel-ferdman: Added a few tests, covreage looks good :)

Pierre-Sassoulas (changes requested):
Thank you ! It seems docstring only stub raises right now (on pydantic_core/_pydantic_core.pyi for example). Mighht want to add a ``if not body: return True`` somewhere.

emmanuel-ferdman:
@Pierre-Sassoulas yeah well spotted, added guard for docstring-only stub functions and added tests :)

doc/whatsnew/fragments/3078.feature
```diff
@@ -0,0 +1,7 @@
+``.pyi`` stub files are now loaded and used during inference. Stub
```
Pierre-Sassoulas:
```suggestion
``.pyi`` stub files are now loaded and used during inference when prefer_stubs is on. Stub
```

astroid/manager.py
```diff
@@ -307,25 +320,29 @@ def zip_import_data(self, filepath: str) -> nodes.Module | None:
     def file_from_module_name(
```
Pierre-Sassoulas, line 320: This is going to conflict with #3282, no matter I'll deal with it later.

doc/whatsnew/fragments/3078.feature
```diff
@@ -0,0 +1,7 @@
...
+function bodies (``...`` or ``pass``) infer the return type from the
+annotation, annotated assignments produce instances of the annotated type,
+and compiled extensions with an adjacent stub use the stub instead of
+runtime introspection.
```
Pierre-Sassoulas:
```suggestion
When ``--prefer-stubs`` is enabled in pylint ``.pyi`` stub files are now loaded
and used during inference and compiled extensions with an adjacent stub use
 the stub instead of runtime introspection. Stub function bodies (``...``, ``pass`` or a docstring) infer
 the return type from the annotation.
```

Pierre-Sassoulas (commented):
I have a suggestion regarding the changelog. Did I understand that correctly ?

doc/whatsnew/fragments/3078.feature
emmanuel-ferdman: Yeah, on point

Pierre-Sassoulas (approved):
Thank you ! Big upgrade for the next astroid version

PR #3079, opened 2026-05-23
Add CodSpeed CI with end-to-end primer benchmarks (flask + black) and a cold start test using initial imports

Pierre-Sassoulas:
Adds CodSpeed CI to astroid with three end-to-end benchmarks for black and flask, alongside a cold start test to exercice the import time which is important because we often lint single file in parallel inside pre-commit.

Supersedes #3022.

Result can be seen in #3080 (but also on currently opened performance branches : #3069, #3046, #3048). For the lazy loading of brains, the result is not visible because the threshold is at 2% change (already close to noise). We won't see performance change in brains as we don't parse lib specific code in this benchmark.

### DIFF
--- a/.github/workflows/codspeed-walltime.yaml
+++ b/.github/workflows/codspeed-walltime.yaml
@@ -25,33 +25,35 @@ concurrency:
 
 permissions:
   contents: read
-  id-token: write
 
 jobs:
   benchmarks:
     name: Run cold-start benchmarks
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v6.0.2
 
-      - name: Set up Python 3.13
+      - name: Set up Python 3.14
         id: python
         uses: actions/setup-python@v6.2.0
         with:
-          python-version: "3.13"
+          python-version: "3.14"
           check-latest: true
 
       - name: Install dependencies
         run: |
           python -m pip install -U pip
           # Install pylint first (it pulls in a published astroid as a
-          # dependency); the editable install below then replaces that
-          # astroid with the version under test. pylint is invoked as a
+          # dependency); the install below then replaces that astroid
+          # with the version under test. pylint is invoked as a
           # subprocess by the cold-start benchmark.
           pip install pylint pytest-codspeed
-          pip install -e .
+          pip install .
 
       - name: Run cold-start benchmarks
         uses: CodSpeedHQ/action@v4.13.0
--- a/.github/workflows/codspeed-walltime.yaml
+++ b/.github/workflows/codspeed-walltime.yaml
@@ -1,0 +1,60 @@
+name: CodSpeed (walltime)
+
+# Cold-start benchmarks (currently just test_bench_endtoend_cold_lint)
+# shell out to a fresh Python subprocess. mode: simulation only measures
+# Python-side instructions and would report the subprocess as ~300 µs of
+# subprocess.run overhead. mode: walltime measures the wall-clock end to
+# end including the subprocess, which is what we want.
+#
+# Walltime on a shared GitHub runner is noisier than CodSpeed's dedicated
+# macro runners. If cold_lint starts producing false-positive regressions,
+# raise its per-benchmark threshold on the CodSpeed dashboard (Actions
+# menu on the benchmark page).
+
+on:
+  push:
+    branches:
+      - main
+      - maintenance/**
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  benchmarks:
+    name: Run cold-start benchmarks
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v6.0.2
+
+      - name: Set up Python 3.13
+        id: python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: "3.13"
+          check-latest: true
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          # Install pylint first (it pulls in a published astroid as a
+          # dependency); the editable install below then replaces that
+          # astroid with the version under test. pylint is invoked as a
+          # subprocess by the cold-start benchmark.
+          pip install pylint pytest-codspeed
+          pip install -e .
+
+      - name: Run cold-start benchmarks
+        uses: CodSpeedHQ/action@v4.13.0
+        with:
+          mode: walltime
+          run: pytest tests/benchmarks/ --codspeed -k "cold_lint"
--- a/.github/workflows/codspeed-walltime.yaml
+++ b/.github/workflows/codspeed-walltime.yaml
@@ -31,9 +31,6 @@ jobs:
     name: Run cold-start benchmarks
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    permissions:
-      contents: read
-      id-token: write
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v6.0.2
--- a/.github/workflows/codspeed.yaml
+++ b/.github/workflows/codspeed.yaml
@@ -25,9 +25,6 @@ jobs:
   benchmarks:
     name: Run benchmarks
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      id-token: write
     # mode: simulation captures a callgraph per benchmark for per-function
     # attribution on the dashboard. That instrumentation is ~10-50x slower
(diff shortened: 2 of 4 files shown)


.github/workflows/codspeed-walltime.yaml
```diff
@@ -0,0 +1,60 @@
...
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write
```
cdce8p: Is `id-token` necessary for `CodSpeedHQ/action`? If that's the case, could you move it to the job itself?

.github/workflows/codspeed-walltime.yaml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v6.0.2
+
+      - name: Set up Python 3.13
+        id: python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: "3.13"
+          check-latest: true
```
cdce8p: Any particular reason to use 3.13 over 3.14? Saw that we recently switch the default in pylint, guess we'll do the same for astroid as well soon.

.github/workflows/codspeed-walltime.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          # Install pylint first (it pulls in a published astroid as a
+          # dependency); the editable install below then replaces that
+          # astroid with the version under test. pylint is invoked as a
+          # subprocess by the cold-start benchmark.
+          pip install pylint pytest-codspeed
+          pip install -e .
```
cdce8p:
```suggestion
          pip install .
```
I believe editable installs might even be unnecessary here. Yes, we use them for the other workflows as well, however the astroid files aren't modified in any way after the install.

.github/workflows/codspeed.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write
```
cdce8p: Similar to earlier, move this to the job itself.

.github/workflows/codspeed.yaml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v6.0.2
+
+      - name: Set up Python 3.13
+        id: python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: "3.13"
+          check-latest: true
```
cdce8p: `3.14`?

.github/workflows/codspeed.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          pip install pytest-codspeed
+          pip install -e .
```
cdce8p:
```suggestion
          pip install .
```

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+
+The ``benchmark`` fixture used by every test in this directory is provided
+by the ``pytest-codspeed`` plugin. CI installs it explicitly in the
+CodSpeed workflow, but contributors running ``pytest tests/`` locally will
+not have it. Without this guard pytest would error out with
+``fixture 'benchmark' not found`` (and, because ``filterwarnings = "error"``
+is set in ``pyproject.toml``, even a warning would abort collection).
+"""
+
+from __future__ import annotations
```
cdce8p: No really need for that here.

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+
+import importlib.util
+
+if importlib.util.find_spec("pytest_codspeed") is None:
+    collect_ignore_glob = ["test_bench_*.py"]
```
cdce8p: Guess that will work, though in other test files we usually use a try ... except ImportError guard for it.
```py
try:
    import pytest_codespeed
except ImportError:
    collect_ignore_glob = ["test_bench_*.py"]
```

tests/benchmarks/test_bench_endtoend.py
```diff
@@ -0,0 +1,233 @@
...
+
+- ``test_bench_endtoend_cold_lint`` — shells out
+  ``python -m pylint <minimal module>`` per iteration. ~98 % of wall
+  time is startup (Python import, pylint init, ``import astroid``,
+  brain-plugin registration); only ~2 % is actual linting work on the
+  one-line target. This captures cold-start cost that the in-process
+  benches below miss: within a single pytest session ``astroid`` is
+  imported once at module load and stays in ``sys.modules``, so
+  optimizations like deferring brain-plugin registration are invisible
+  to those benches.
```
cdce8p: I worry that we might end up optimizing the wrong thing. Yes, startup time is important as well, however with lazy imports begin available in 3.15 there seems to be a focus across the ecosystem to use them everywhere. IMO we should be cautious with that when the time comes, especially for modules which get imported anyway.

That's a bit different for brains, etc. which aren't actually needed for some / most projects. Though this might only be a small fraction of the total import time.

cdce8p (commented):
Left some comments. Similar to the pylint primer, this isn't really my area of expertise, so might be good if someone else could take a look at it as well.

tests/benchmarks/test_bench_endtoend.py
Pierre-Sassoulas: We did some lazy import in https://github.com/pylint-dev/astroid/pull/3062/changes, I think those one are useful, and would have been detected because the logging is very seldom used (only if there's C level warning during parsing).

Pierre-Sassoulas:
Thank you for the review Marc ! Appreciate it.

tests/benchmarks/test_bench_endtoend.py
```diff
@@ -0,0 +1,243 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
+
+"""End-to-end benchmarks: cold import + parse + walk + infer real projects.
```
DanielNoord: A lot of this docstring will get outdated as soon as we start changing stuff. Is it really necessary to document this in this much detail?

tests/benchmarks/test_bench_endtoend.py
```diff
@@ -0,0 +1,243 @@
...
+        check=False,
+        capture_output=True,
+    )
+
+
+def test_bench_endtoend_cold_lint(benchmark, minimal_module: Path) -> None:
+    benchmark(_pylint_one_file, minimal_module)
+
+
+# -- Flask: small, parse + walk_infer (parse isolates rebuilder cost). --
```
DanielNoord: In line comments like this also tend to get out of date

tests/benchmarks/test_bench_endtoend.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def test_bench_endtoend_parse_flask(benchmark, flask_files: list[Path]) -> None:
```
DanielNoord: No types for `benchmark`?

.github/workflows/codspeed-walltime.yaml
```diff
@@ -0,0 +1,62 @@
...
+  contents: read
+
+jobs:
+  benchmarks:
+    name: Run cold-start benchmarks
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write
```
DanielNoord: Why is this necessary?

.github/workflows/codspeed.yaml
```diff
@@ -0,0 +1,60 @@
...
+permissions:
+  contents: read
+
+jobs:
+  benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
```
DanielNoord: Why is this necessary?

DanielNoord (commented):
I don't have much experience with CodSpeed, but most of it looks fine to me I guess?

.github/workflows/codspeed.yaml
Pierre-Sassoulas: Might not be in a public repo, I removed, let's see how it goes

Pierre-Sassoulas: Look like it worked. I fixed your other remarks in [be68e14](https://github.com/pylint-dev/astroid/pull/3079/commits/be68e14860fd06b5b1045fea0ba32a9dafb8e890)

DanielNoord (approved):

PR #3058, opened 2026-05-18
Fix TypeVarTuple ABC inference, fixes #10972

kdkavanagh-jt:
Fixes inference around TypeVarTuple that causes multiple issues in pylint (https://github.com/pylint-dev/pylint/issues/10972 and https://github.com/pylint-dev/pylint/issues/10991)

PR #3082, opened 2026-05-24
fix: handle PEP 695 type parameters in helpers.object_type

Pierre-Sassoulas:
## Type of Changes

## Description

PR #3058 changed generic_type_assigned_stmts to yield the TypeVar/ TypeVarTuple/ParamSpec node itself instead of Const(None). As a result, inferring an expression like '_T | None' inside 'def func[_T]: ...' reached helpers._object_type with one of these new AST nodes, which it didn't recognize, raising AssertionError("We don't handle <class 'astroid.nodes.node_classes.TypeVar'> currently").

A PEP 695 type parameter has no concrete type at static-analysis time, so yield util.Uninferable for these nodes. The binop inference path already handles Uninferable operands gracefully.

Reported at https://github.com/pylint-dev/astroid/pull/3058#issuecomment-4528993032

### DIFF
--- a/astroid/helpers.py
+++ b/astroid/helpers.py
@@ -66,6 +66,10 @@ def _object_type(
             yield _build_proxy_class("module", builtins)
         elif isinstance(inferred, nodes.Unknown):
             raise InferenceError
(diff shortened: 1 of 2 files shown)


cdce8p (approved):
Thanks for the quick fix @Pierre-Sassoulas 👍🏻

PR #3083, opened 2026-05-24
Fix: avoid TypeError from getset descriptors on PyPy 7.3.22

Pierre-Sassoulas:
## Type of Changes

## Description

PyPy 7.3.22 changed the exception type raised when accessing certain getset descriptors on the type (rather than an instance). We froze the version in pylint to make the CI green, but this is the wrong way to fix this.

Refs pylint-dev/pylint#10999

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,15 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
The test is not pypy specific so we don't need a specific pypy version to make it work, but I could gate it with ``skipIf(not IS_PYPY``

DanielNoord (approved):

PR #3088, opened 2026-05-29
docs: correct ImportFrom.level description

### LINKED ISSUE #2571: Bug: incorrect docstring on ImportFrom node
According to the docstring the level of a relative import is 0 for absolute imports.
However, in practise it is actually `None`.

See: https://github.com/pylint-dev/astroid/blob/1368be19796430f4f1ee62f6e87b985e592b948c/astroid/nodes/node_classes.py#L2840-L2846

I would suggest that the TODO is removed and that the docstring is ajusted accordingly.

3.2.4

nanookclaw:
## Summary

Closes #2571.

This removes the stale TODO on `ImportFrom.level` and corrects the attribute documentation to say that absolute imports use `None`, matching the current AST conversion behavior.

## Verification

- `python3 -m compileall -q astroid/nodes/node_classes.py`
- `PYTHONPATH=. python3` smoke check for `from os import path` (`level is None`) and `from . import sibling` (`level == 1`)
- `git diff --check`

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -2863,12 +2863,11 @@ class ImportFrom(_base_nodes.ImportNode):
         and the alias that the name is assigned to (if any).
         """
 
-        # TODO When is 'level' None?
         self.level: int | None = level  # can be None
         """The level of relative import.
 
         Essentially this is the number of dots in the import.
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Thank you for your work on astroid :)

PR #3094, opened 2026-06-08
Support PEP 798 dict-comp unpacking and PEP 810 lazy imports

Pierre-Sassoulas:
## Type of Changes

## Description

Python 3.15 adds two AST-level constructs astroid did not model:

- PEP 798 dict-comprehension unpacking (``{**d for d in dicts}``). CPython stores the unpacked mapping in ``DictComp.key`` and leaves ``value`` as ``None``, which crashed the rebuilder. astroid now mirrors the ``{**d}`` Dict representation with a ``DictUnpack`` key, and ``as_string`` round-trips the construct.
- PEP 810 explicit lazy imports. ``Import`` and ``ImportFrom`` gain a ``lazy`` boolean attribute, set from CPython's ``is_lazy`` AST field and rendered by ``as_string``. Name binding and inference are unchanged.

List/set/generator comprehension unpacking already parsed; the new ``frozendict`` and ``sentinel`` builtins are inferred for free via raw-building. Tests live next to the feature they exercise (lazy imports in ImportNodeTest, comprehension unpacking in Python3TC, builtin inference in InferenceTest), gated with skipif(not PY315_PLUS).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,18 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Support PEP 810 explicit lazy imports (new in Python 3.15). ``Import`` and
+  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute (``False`` before
+  3.15 and for eager imports), set from CPython's ``is_lazy`` AST field, and
+  ``as_string`` renders the ``lazy`` prefix. Name binding and inference are
+  unchanged from eager imports.
+
+* Support PEP 798 dict-comprehension unpacking (``{**d for d in dicts}``,
+  new in Python 3.15). The rebuilder previously crashed because CPython
+  stores the unpacked mapping in ``DictComp.key`` and leaves ``value`` as
+  ``None``; astroid now mirrors the ``{**d}`` ``Dict`` representation with a
+  ``DictUnpack`` key, and ``as_string`` round-trips the construct.
+
 * Fix astroid bootstrap crash on PyPy 7.3.22 (``TypeError: expected str, got
   getset_descriptor object``) by also catching ``TypeError`` from
   ``getattr(obj, alias)`` in ``InspectBuilder.object_build``. PyPy 7.3.22
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,7 +18,7 @@ Release date: TBA
   stores the unpacked mapping in ``DictComp.key`` and leaves ``value`` as
   ``None``; astroid now mirrors the ``{**d}`` ``Dict`` representation with a
   ``DictUnpack`` key, and ``as_string`` round-trips the construct.
-  
+
 * Fix uncaught ``IndentationError`` when parsing code whose lines end in
   ``\r``: the slice of source tokenized to compute a class or function
   ``position`` is split on ``\n`` only and could be misaligned with the AST
--- a/ChangeLog
+++ b/ChangeLog
@@ -47,11 +47,8 @@ Release date: TBA
 
   Closes #3091
 
-* Support PEP 810 explicit lazy imports (new in Python 3.15). ``Import`` and
-  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute (``False`` before
-  3.15 and for eager imports), set from CPython's ``is_lazy`` AST field, and
-  ``as_string`` renders the ``lazy`` prefix. Name binding and inference are
-  unchanged from eager imports.
+ Support PEP 810 lazy imports (new in Python 3.15). ``Import`` and
+  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute.
 
 * Support PEP 798 comprehension unpacking (``{**d for d in dicts}``,
   new in Python 3.15).
--- a/ChangeLog
+++ b/ChangeLog
@@ -53,11 +53,8 @@ Release date: TBA
   ``as_string`` renders the ``lazy`` prefix. Name binding and inference are
   unchanged from eager imports.
 
-* Support PEP 798 dict-comprehension unpacking (``{**d for d in dicts}``,
-  new in Python 3.15). The rebuilder previously crashed because CPython
-  stores the unpacked mapping in ``DictComp.key`` and leaves ``value`` as
-  ``None``; astroid now mirrors the ``{**d}`` ``Dict`` representation with a
-  ``DictUnpack`` key, and ``as_string`` round-trips the construct.
+* Support PEP 798 comprehension unpacking (``{**d for d in dicts}``,
+  new in Python 3.15).
 
 * Fix astroid bootstrap crash on PyPy 7.3.22 (``TypeError: expected str, got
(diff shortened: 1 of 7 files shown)


Pierre-Sassoulas:
@jacobtylerwalls @DanielNoord what do you think ? I think it's the path to actual python 3.15 support but I'm not sure about the implementation details.

DanielNoord (commented):
Any details you are particularly worried about?

ChangeLog
```diff
@@ -48,6 +47,18 @@ Release date: TBA
...
+  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute (``False`` before
+  3.15 and for eager imports), set from CPython's ``is_lazy`` AST field, and
+  ``as_string`` renders the ``lazy`` prefix. Name binding and inference are
+  unchanged from eager imports.
+
+* Support PEP 798 dict-comprehension unpacking (``{**d for d in dicts}``,
+  new in Python 3.15). The rebuilder previously crashed because CPython
+  stores the unpacked mapping in ``DictComp.key`` and leaves ``value`` as
+  ``None``; astroid now mirrors the ``{**d}`` ``Dict`` representation with a
+  ``DictUnpack`` key, and ``as_string`` round-trips the construct.
```
jacobtylerwalls:
```suggestion
* Support PEP 798 comprehension unpacking (``{**d for d in dicts}``,
  new in Python 3.15).
```

ChangeLog
```diff
@@ -48,6 +47,18 @@ Release date: TBA
 
   Closes #3091
 
+* Support PEP 810 explicit lazy imports (new in Python 3.15). ``Import`` and
+  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute (``False`` before
+  3.15 and for eager imports), set from CPython's ``is_lazy`` AST field, and
+  ``as_string`` renders the ``lazy`` prefix. Name binding and inference are
+  unchanged from eager imports.
```
jacobtylerwalls:
```suggestion
 Support PEP 810 lazy imports (new in Python 3.15). ``Import`` and
  ``ImportFrom`` nodes gain a ``lazy`` boolean attribute.
```

ChangeLog
jacobtylerwalls: Focusing only on dict comprehensions might confuse readers.

tests/test_python3.py
```diff
@@ -235,6 +236,58 @@ def test_nested_unpacking_in_dicts(self) -> None:
...
+        ):
+            self.assertEqual(extract_node(code).as_string(), code)
+
+    @pytest.mark.skipif(
+        not PY315_PLUS, reason="PEP 798 unpacking in comprehensions, new in 3.15"
+    )
+    def test_comprehension_target_resolves_inside_starred_elt(self) -> None:
+        # The comprehension target is in scope inside the starred element and
+        # infers to the iterated items.
+        node = extract_node("[*L for L in [[1, 2], [3, 4]]]")
```
jacobtylerwalls, line 286: Adjusting this test to use a condition shows inference needs work:
```diff
diff --git a/tests/test_python3.py b/tests/test_python3.py
index 4da82d14..d0513331 100644
--- a/tests/test_python3.py
+++ b/tests/test_python3.py
@@ -283,10 +283,10 @@ class Python3TC(unittest.TestCase):
     def test_comprehension_target_resolves_inside_starred_elt(self) -> None:
         # The comprehension target is in scope inside the starred element and
         # infers to the iterated items.
-        node = extract_node("[*L for L in [[1, 2], [3, 4]]]")
+        node = extract_node("[*L for L in [[1, 2], [3, 4]] if sum(L) == 3]")
         self.assertIsInstance(node.elt, nodes.Starred)
         inferred = [elt.as_string() for elt in node.elt.value.inferred()]
-        self.assertEqual(inferred, ["[1, 2]", "[3, 4]"])
+        self.assertEqual(inferred, ["[1, 2]"])
 
     def test_unpacking_in_dict_getitem(self) -> None:
         node = extract_node("{1:2, **{2:3, 3:4}, **{5: 6}}")
```
```
E       AssertionError: Lists differ: ['[1, 2]', '[3, 4]'] != ['[1, 2]']
E       
E       First list contains 1 additional elements.
E       First extra element 1:
E       '[3, 4]'
E       
E       - ['[1, 2]', '[3, 4]']
E       + ['[1, 2]']
```

astroid/rebuilder.py
```diff
@@ -1321,6 +1340,8 @@ def visit_import(self, node: ast.Import, parent: nodes.NodeNG) -> nodes.Import:
             end_lineno=node.end_lineno,
             end_col_offset=node.end_col_offset,
             parent=parent,
+            # ``is_lazy`` is a PEP 810 field only present on Python 3.15+.
+            lazy=bool(getattr(node, "is_lazy", False)),
```
jacobtylerwalls, line 1344: We usually see these with version guards, but if you think the "3.15+" in the comment is greppable when removing 3.14 support, no problem I guess. Otherwise if we go with the sys guard, then no need for a comment.

Same thing above.

jacobtylerwalls (dismissed):
Approach seems sound, but inference needs a tweak.

jacobtylerwalls:
Sorry, I should compare that to our baseline for conditions, if we don't handle that in the general case without unpacking, then we can merge.

astroid/rebuilder.py
DanielNoord: Shouldn’t we use the same attribute name as `ast` does and also make it an integer? We can add a property to determine if it is lazy, but we have had issues before by not following the naming scheme of `ast`.

tests/test_python3.py
Pierre-Sassoulas: I think it's pre-existing (extract_node('[L for L in [[1, 2], [3, 4]] if sum(L) == 3]').elt.inferred() gives ['[1, 2]', '[3, 4]']). I'll see if it's fixable.

Pierre-Sassoulas:
Baseline check: the general case without unpacking behaves the same on `main` —

```python
>>> [i.as_string() for i in extract_node('[L for L in [[1, 2], [3, 4]] if sum(L) == 3]').elt.inferred()]
['[1, 2]', '[3, 4]']
```

so the condition is ignored with or without unpacking, and this PR isn't a regression. Opened #3155 to make comprehension conditions generate inference constraints like `if` conditions do (`sum(L) == 3` matches no supported constraint pattern, so it stays unfiltered there too, but `is None`/truthiness/`isinstance()`/`==` now filter).

jacobtylerwalls (approved):

Pierre-Sassoulas:
@DanielNoord let's include this in beta 5 if we can

astroid/rebuilder.py
DanielNoord: @Pierre-Sassoulas you merged without addressing this comment 😅

Pierre-Sassoulas: Sorry I missed that, fixed in https://github.com/pylint-dev/astroid/pull/3158

PR #3095, opened 2026-06-08
Fix partial descriptor binding crash

### LINKED ISSUE #3090: TypeError: PartialFunction.__init__() got an unexpected keyword argument 'end_lineno'
### Steps to reproduce

The following code triggers a `TypeError` in astroid:

```python
from functools import partial

def fn(x):
    pass

p = partial(fn, x=1)
a = p.__get__({})
```

### Current behavior
```python-traceback
  File "pylint/pylint/checkers/utils.py", line 1374, in safe_infer
    value = next(infer_gen)
  File "astroid/astroid/nodes/node_ng.py", line 165, in infer
                     ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 81, in inner
    yield next(generator)
          ~~~~^^^^^^^^^^^
  File "astroid/astroid/decorators.py", line 44, in wrapped
               ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/nodes/node_classes.py", line 1778, in _infer
    yield from callee.infer_call_result(
        caller=self, context=callcontext
    )
  File "astroid/astroid/interpreter/objectmodel.py", line 463, in infer_call_result
    new_func = func.__class__(
        name=func.name,
    ...<4 lines>...
        end_col_offset=func.end_col_offset,
    )
TypeError: PartialFunction.__init__() got an unexpected keyword argument 'end_lineno'

    callback(astroid)
    ~~~~~~~~^^^^^^^^^
  File "pylint/pylint/checkers/base/name_checker/checker.py", line 449, in visit_assignname
  File "pylint/pylint/checkers/utils.py", line 1378, in safe_infer
    raise AstroidError from e
```

### Expected behavior
No crash

### Version info
pylint-dev/astroid@88d20dc60
pylint-dev/pylint@c6ca272c8
Python 3.14.5

omkar-334:
## Type of Changes

## Description

This fixes a crash when astroid infers `functools.partial(...).__get__`.

`PartialFunction` now accepts the same end-location metadata used when function-like nodes are rebuilt for descriptor binding. Rebuilt partials also keep their pre-filled arguments and keywords, so calling the inferred bound method behaves like the original partial.

A regression test covers inferring the bound partial and calling it.

Closes #3090

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -468,7 +468,7 @@ class FunctionModel(ObjectModel):
                     "end_lineno": func.end_lineno,
                     "end_col_offset": func.end_col_offset,
                 }
-                if hasattr(func, "filled_args") and hasattr(func, "filled_keywords"):
+                if isinstance(func, astroid.objects.PartialFunction):
                     new_func_kwargs["filled_args"] = func.filled_args
                     new_func_kwargs["filled_keywords"] = func.filled_keywords
 
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -460,14 +460,19 @@ class FunctionModel(ObjectModel):
 
                 # Rebuild the original value, but with the parent set as the
                 # class where it will be bound.
-                new_func = func.__class__(
-                    name=func.name,
-                    lineno=func.lineno,
-                    col_offset=func.col_offset,
-                    parent=func.parent,
-                    end_lineno=func.end_lineno,
-                    end_col_offset=func.end_col_offset,
-                )
+                new_func_kwargs = {
+                    "name": func.name,
+                    "lineno": func.lineno,
+                    "col_offset": func.col_offset,
+                    "parent": func.parent,
+                    "end_lineno": func.end_lineno,
+                    "end_col_offset": func.end_col_offset,
+                }
+                if hasattr(func, "filled_args") and hasattr(func, "filled_keywords"):
+                    new_func_kwargs["filled_args"] = func.filled_args
+                    new_func_kwargs["filled_keywords"] = func.filled_keywords
+
+                new_func = func.__class__(**new_func_kwargs)
                 # pylint: disable=no-member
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (dismissed):
LGTM, but let's wait for another maintainer opinion.

astroid/objects.py
```diff
@@ -282,16 +282,33 @@ class DictValues(bases.Proxy):
...
-            end_col_offset=0,
-            end_lineno=0,
+            end_col_offset=end_col_offset,
+            end_lineno=end_lineno,
             parent=parent,
         )
+        if call is None:
+            self.filled_args = list(filled_args or [])
+            self.filled_keywords = dict(filled_keywords or {})
+            self.filled_positionals = len(self.filled_args)
```
DanielNoord: This is the same as line 324, making it incorrectly look as if we need special handling for this case. It would be better to make explicit what is and isn't different between the `call is None` case and the default path

astroid/interpreter/objectmodel.py
```diff
@@ -460,14 +460,19 @@ def infer_call_result(
...
-                )
+                new_func_kwargs = {
+                    "name": func.name,
+                    "lineno": func.lineno,
+                    "col_offset": func.col_offset,
+                    "parent": func.parent,
+                    "end_lineno": func.end_lineno,
+                    "end_col_offset": func.end_col_offset,
+                }
+                if hasattr(func, "filled_args") and hasattr(func, "filled_keywords"):
```
DanielNoord: We can't do this without `hasattr`?

omkar-334: we can check if `func` is a `PartialFunction` object using `isinstance` ...

omkar-334:
hey @DanielNoord apologies for getting back late. I've made the requested changes as they made sense. Kindly review

Pierre-Sassoulas (approved):

PR #3100, opened 2026-06-11
fix: avoid uncaught IndentationError when computing node positions

### LINKED ISSUE #3091: Uncaught IndentationError while fuzzing
### Steps to reproduce

The following testcase from OSS-Fuzz raises an `IndentationError`:

```python
import astroid

code = '\rclass C:\n def f():\n  1\n 2'

astroid.builder.parse(code)
```

https://issues.oss-fuzz.com/issues/518261019 (report not public yet)

### Current behavior
```python-traceback
  File "astroid/repro.py", line 5, in <module>
    astroid.builder.parse(code)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "astroid/astroid/builder.py", line 313, in parse
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 155, in string_build
                      ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/builder.py", line 210, in _data_build
  File "astroid/astroid/rebuilder.py", line 178, in visit_module
     ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 497, in visit
  File "astroid/astroid/rebuilder.py", line 874, in visit_classdef
     ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 497, in visit
  File "astroid/astroid/rebuilder.py", line 1201, in visit_functiondef
    return self._visit_functiondef(nodes.FunctionDef, node, parent)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 1186, in _visit_functiondef
             ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "astroid/astroid/rebuilder.py", line 135, in _get_position_info
    for t in generate_tokens(StringIO(data).readline):
             ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/tokenize.py", line 592, in _generate_tokens_from_c_tokenizer
    raise e from None
  File "/usr/lib/python3.14/tokenize.py", line 588, in _generate_tokens_from_c_tokenizer
    for info in it:
                ^^
  File "<string>", line 2
    2
     ^
IndentationError: unindent does not match any outer indentation level
```

### Expected behavior
No crash

### Version info
pylint-dev/astroid@88d20dc60
Python 3.14.5

binggao1230:
## Description

`astroid.builder.parse('\rclass C:\n def f():\n  1\n 2')` raised an uncaught `IndentationError` even though CPython parses this code fine.

The tokenizer treats a lone `\r` as a line terminator, while `RebuildVisitor` keeps the source as `data.split("\n")`. For `\r`-terminated lines the two disagree on line numbers, so the slice that `_get_position_info` tokenizes to locate the keyword/name span can be misaligned (here it tokenizes `"  1\n 2"` for `f`, which dedents to a level that was never seen) and the C tokenizer on Python >= 3.12 raises `IndentationError` instead of `TokenError`.

This catches `SyntaxError` (the base of `IndentationError`) alongside the existing `TokenError` guard from #3075, so such nodes simply have no `position` information, the same degradation already chosen there. The AST-derived `lineno`/`col_offset` of the nodes are unaffected. `self._data` is only used by `_get_position_info`, so the misalignment cannot leak anywhere else.

The regression test uses the real input from the issue rather than a mock.

Note: the pylint pre-commit hook flags `astroid/rebuilder.py:123: E1136 (unsubscriptable-object)` for me, but it does so on an unmodified checkout too (latest pylint, `self._data` is guarded two lines above) — unrelated to this change.

Closes #3091

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,14 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Fix uncaught ``IndentationError`` when parsing code whose lines end in
+  ``\r``: the slice of source tokenized to compute a class or function
+  ``position`` is split on ``\n`` only and could be misaligned with the AST
+  line numbers. Such nodes now simply have no position information, as
+  already done when ``tokenize`` raises ``TokenError``.
+
+  Closes #3091
+
 * Fix astroid bootstrap crash on PyPy 7.3.22 (``TypeError: expected str, got
   getset_descriptor object``) by also catching ``TypeError`` from
   ``getattr(obj, alias)`` in ``InspectBuilder.object_build``. PyPy 7.3.22
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -148,9 +148,12 @@ class TreeRebuilder:
                 start_token = None
             else:
                 return None
-        except TokenError:
-            # Malformed source can make ``generate_tokens`` raise (e.g. an
-            # unterminated bracket on Python < 3.12); no position info then.
+        except (TokenError, SyntaxError):
+            # ``generate_tokens`` can raise on input it cannot tokenize, e.g.
+            # ``TokenError`` for an unterminated bracket on Python < 3.12, or
+            # ``IndentationError`` when ``\r``-terminated lines make the slice
+            # of ``self._data`` (split on ``\n`` only) misaligned with the AST
+            # line numbers; no position info then.
             return None
 
         return Position(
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):
LGTM, thank you.

PR #3101, opened 2026-06-14
Compute constraints in `and`, `or` operations

zenlyj:
## Type of Changes

## Description

Currently, constraints are applied when inferring values inside conditional branches:
```
x = None

if x is not None:
    x  # `None` is filtered out
```

This PR extends constraint inference to later operands of `and`, `or` operations:
```
x = None

if x is not None and x:  # `None` is filtered out for rightmost `x`
    pass

if x is None or x:  # `None` is filtered out for rightmost `x`
    pass
```

For `and`, constraints from preceding operands are applied directly. For `or`, they are inverted because a later operand is evaluated only when all preceding operands are falsy. Constraints from multiple preceding operands are accumulated.

Operands that do not constrain the inferred value do not affect inference:
```
x, y = None, None

if y and x:  # `x` is inferred as `None`
    pass
```

The implementation supports flat boolean operations with any number of operands. Nested `BoolOp` operands are not yet decomposed:
```
x = None

if (x and x is None) and x:  # rightmost `x` is still inferred as `None`
    pass

if not x and x is None or x:  # rightmost `x` is still inferred as `None`
    pass
``` 

Support for these nested cases should be handled in #2977.

Refs https://github.com/pylint-dev/pylint/issues/1498

### DIFF
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -243,27 +243,44 @@ class EqualityConstraint(Constraint):
 
 def get_constraints(
     expr: _NameNodes, frame: nodes.LocalsDictNodeNG
-) -> dict[nodes.If | nodes.IfExp, set[Constraint]]:
+) -> dict[nodes.If | nodes.IfExp | nodes.BoolOp, set[Constraint]]:
     """Returns the constraints for the given expression.
 
     The returned dictionary maps the node where the constraint was generated to the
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
Looks great already, and this is going to be very Impactful in pylint when it lands. 💪

PR #3106, opened 2026-06-16
Fix classmethod inference boundnode leak

### LINKED ISSUE #11101: Inference does weird things when a classmethod cls argument is called.
### Bug description

Here is a minimal reproducer:

```python
class A:
    def b(self):
        return 0
    @classmethod
    def c(cls):
        return cls(), 0
class D:
    def e(self):
        self.f = A.c()[0]
        self.f.b() # x.py:10:8: E1101: Instance of 'D' has no 'b' member; maybe 'f'? (no-member)
```

Pylint deduces that `self.f` is an instance of `D` not `A`, here. It seems to have something to do with a scope escape type issue? Where `cls` starts to refer to `D` instead of `A` for some reason.

This code doesn't fail:

```python
class A:
    def b(self):
        return 0
    @classmethod
    def c(cls):
        return (_:=cls()), 0
class D:
    def e(self):
        self.f = A.c()[0]
        self.f.b()

```

Probably because the assignment will cause more aggressive inference.

### Configuration

```ini

```

### Command used

```shell
pylint x.py
```

### Pylint output

```python
************* Module x
x.py:10:8: E1101: Instance of 'D' has no 'b' member; maybe 'f'? (no-member)
```

### Expected behavior

No error.

### Pylint version

```shell
pylint 4.0.5
astroid 4.2.0
Python 3.14.5 (main, May 31 2026, 17:34:14) [GCC 16.1.1 20260430]
```

### OS / Environment

_No response_

### Additional dependencies

```python

```

hariharan077:
Closes pylint-dev/pylint#11101

## Summary

- avoid reusing an unrelated outer bound instance when inferring the first argument of a classmethod
- add a regression test for `A.c()[0]` where `c` returns `(cls(), 0)` inside another class's method
- document the fix in `ChangeLog`

## Verification

- `python -m pytest tests/test_objects.py -k 'classmethod_returned_tuple_subscript or super_classmethod_chaining_return_type or super_method_chaining_return_type'`
- `python -m pytest tests/test_objects.py tests/test_protocols.py`
- `python -m ruff check astroid/protocols.py tests/test_objects.py`
- `git diff --check`
- Pylint reproducer for pylint-dev/pylint#11101 now rates 10.00/10 with the patched astroid checkout.

Full `python -m pytest` was also run locally: 1909 passed, 107 skipped, 15 xfailed, with 3 failures unrelated to this patch in local environment/setup. After installing `setuptools`, one manager case skipped; remaining local failures were `pkg_resources` not provided by this setuptools build and a macOS symlink relative-path assertion.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Fix classmethod first-argument inference so an unrelated outer bound
+  instance does not leak into the classmethod body. This avoids inferring
+  ``cls`` as the caller's class when indexing a tuple returned by a
+  classmethod such as ``A.c()[0]``.
+
+  Closes pylint-dev/pylint#11101
+
 * The documentation build now fails on any Sphinx warning, so a cross-reference
   that does not resolve is caught by CI instead of quietly rendering as plain
   text. ``Uninferable``, ``Position`` and the ``BadOperationMessage`` classes
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,8 +7,8 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
-* Fix classmethod first-argument inference so an unrelated outer bound
-  instance does not leak into the classmethod body. This avoids inferring
+* Fix inference of a method's first argument (``self`` or ``cls``) when the
+  method's return value is stored on an unrelated object. This avoids inferring
   ``cls`` as the caller's class when indexing a tuple returned by a
   classmethod such as ``A.c()[0]``.
 
--- a/tests/test_objects.py
+++ b/tests/test_objects.py
@@ -520,6 +520,29 @@ class SuperTests(unittest.TestCase):
         self.assertIsInstance(inferred, bases.Instance)
         self.assertEqual(inferred._proxied.name, "Child")
 
+    def test_classmethod_returned_tuple_subscript_ignores_unrelated_boundnode(
+        self,
+    ) -> None:
+        node = builder.extract_node("""
+        class A:
+            def b(self):
+                return 0
+
+            @classmethod
+            def c(cls):
+                return cls(), 0
+
+        class D:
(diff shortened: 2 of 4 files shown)


tests/test_objects.py
```diff
@@ -520,6 +520,29 @@ def extra(self) -> "Child":
         self.assertIsInstance(inferred, bases.Instance)
         self.assertEqual(inferred._proxied.name, "Child")
 
+    def test_classmethod_returned_tuple_subscript_ignores_unrelated_boundnode(
```
Pierre-Sassoulas: This test is in SuperTests but uses no super(). I think it belongs in tests/test_inference.py next to test_inference_is_limited_to_the_boundnode.

ChangeLog
```diff
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Fix classmethod first-argument inference so an unrelated outer bound
+  instance does not leak into the classmethod body. This avoids inferring
+  ``cls`` as the caller's class when indexing a tuple returned by a
+  classmethod such as ``A.c()[0]``.
```
```suggestion
* Fix inference of a method's first argument (``self`` or ``cls``) when the
    method's return value is stored on an unrelated object. This avoids inferring
  ``cls`` as the caller's class when indexing a tuple returned by a
  classmethod such as ``A.c()[0]``.
```

Pierre-Sassoulas (changes requested):
Thank you for working on astroid. I have two nits.

tests/test_objects.py
hariharan077: Moved the regression test to `tests/test_inference.py` next to `test_inference_is_limited_to_the_boundnode` in d7a45a61.

ChangeLog
hariharan077: Updated the changelog wording in d7a45a61. Thanks for the suggestion.

Pierre-Sassoulas (approved):

PR #3107, opened 2026-06-16
cap oversized str/bytes multiply and shift in const_infer_binary_op

kali834x:
## Type of Changes

|     | Type           |
| --- | -------------- |
| ✓   | :bug: Bug fix  |

## Description

const_infer_binary_op evaluates `BIN_OP_IMPL[operator](self.value, other.value)` eagerly when both operands are Const. list/tuple repetition is already bounded in `_multiply_seq_by_int` and `**` is bounded inline, but string/bytes multiplication (`"x" * n` and the reflected `n * "x"`) and integer left shift (`1 << n`) were not, so inferring a small literal like `"A" * 10 ** 10` materializes a multi-gigabyte object during inference of otherwise untrusted source. bound both the same way the sibling sequence and power paths already are: yield Uninferable once `len(seq) * count` or the shift amount would be oversized, before the value is built. small ops (`"ab" * 3`, `1 << 20`, `"a" * 0`) keep inferring their exact values.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Bound string/bytes multiplication (``"x" * n``) and integer left shifts
+  (``1 << n``) in ``const_infer_binary_op`` the same way list/tuple
+  multiplication and ``**`` already are. Inferring a constant such as
+  ``"A" * 10 ** 10`` previously materialized the multi-gigabyte result
+  eagerly; these operations now infer as ``Uninferable`` when the result
+  would be oversized.
+
 * Fix uncaught ``IndentationError`` when parsing code whose lines end in
   ``\r``: the slice of source tokenized to compute a class or function
   ``position`` is split on ``\n`` only and could be misaligned with the AST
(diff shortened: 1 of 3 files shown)


kali834x:
gentle ping, any update?

ChangeLog
```diff
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Bound string/bytes multiplication (``"x" * n``) and integer left shifts
+  (``1 << n``) in ``const_infer_binary_op`` the same way list/tuple
+  multiplication and ``**`` already are. Inferring a constant such as
+  ``"A" * 10 ** 10`` previously materialized the multi-gigabyte result
+  eagerly; these operations now infer as ``Uninferable`` when the result
+  would be oversized.
+
```
Pierre-Sassoulas, line 16:
```suggestion

Refs #3107 

```

Pierre-Sassoulas (commented):
Sorry for the delay, LGTM Thank you !

Pierre-Sassoulas (approved):

PR #3108, opened 2026-06-17
bound oversized format width and precision during inference

kali834x:
## Type of Changes

|     | Type           |
| --- | -------------- |
| ✓   | :bug: Bug fix  |

## Description

**Format-spec width/precision can materialize a multi-gigabyte string during inference**

A tiny literal like `"{:>2000000000}".format("x")` or `f"{1.5:.2000000000f}"` makes astroid eagerly run the format and build a 2 GB string while inferring otherwise untrusted source. `_infer_str_format_call` and `FormattedValue._infer` format with an attacker-controlled width/precision before checking its size; the existing `MemoryError` catch in `FormattedValue._infer` only fires once allocation actually OOMs, so on a machine with enough RAM it just succeeds and detonates. Both paths now read the width and precision out of the spec and yield Uninferable past 1e8, the same cap the sequence/repetition paths in `protocols.py` already use. Small specs keep inferring their exact values.

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 import itertools
+import string
 from collections.abc import Callable, Iterable, Iterator
 from functools import partial
 from typing import TYPE_CHECKING, Any, NoReturn, cast
@@ -1046,6 +1047,13 @@ def _infer_str_format_call(
         inferred_keyword[k] = one_inferred
 
     keyword_values: dict[str, str] = {k: v.value for k, v in inferred_keyword.items()}
+
+    try:
+        fields = list(string.Formatter().parse(format_template))
+    except ValueError:
+        return iter([util.Uninferable])
+    if any(spec and util.format_spec_too_large(spec) for _, _, spec, _ in fields):
+        return iter([util.Uninferable])
 
     try:
         formatted_string = format_template.format(*pos_values, **keyword_values)
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import itertools
-import string
 from collections.abc import Callable, Iterable, Iterator
 from functools import partial
 from typing import TYPE_CHECKING, Any, NoReturn, cast
@@ -1047,6 +1046,8 @@ def _infer_str_format_call(
         inferred_keyword[k] = one_inferred
 
     keyword_values: dict[str, str] = {k: v.value for k, v in inferred_keyword.items()}
+
+    import string  # pylint: disable=import-outside-toplevel
 
     try:
         fields = list(string.Formatter().parse(format_template))
--- a/astroid/util.py
+++ b/astroid/util.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import re
 import warnings
 from typing import TYPE_CHECKING, Any, Final, Literal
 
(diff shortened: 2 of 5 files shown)


kali834x:
gentle ping

astroid/util.py
```diff
@@ -50,6 +51,30 @@ def accept(self, visitor):
...
+
+def format_spec_too_large(format_spec: str) -> bool:
+    """Whether a format spec asks for an oversized width or precision.
+
+    Used to avoid materializing a multi-gigabyte string while inferring a tiny
+    literal such as ``"{:>2000000000}".format("x")`` or ``f"{1.5:.2e9f}"``.
+    """
+    match = _FORMAT_SPEC_SIZE.match(format_spec)
+    if match is None:  # pragma: no cover
+        return False
```
Pierre-Sassoulas: There's no way that match is None because everything is optional in _FORMAT_SPEC_SIZE. Let's just drop the check ?


```suggestion
    # if match is None:  # pragma: no cover
    # Everything is optional in _FORMAT_SPEC_SIZE, if that change match could be None
    #     return False
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 import itertools
+import string
```
Pierre-Sassoulas: As far as I can git grep this is the only place we use string in astroid (2 import in tests), so lazy loading it (old style) in the function using it would save startup time.


```suggestion
```

astroid/brain/brain_builtin_inference.py
```diff
@@ -1047,6 +1048,13 @@ def _infer_str_format_call(
 
     keyword_values: dict[str, str] = {k: v.value for k, v in inferred_keyword.items()}
 
+    try:
+        fields = list(string.Formatter().parse(format_template))
```
Pierre-Sassoulas, line 1053:
```suggestion
    import string
    try:
        fields = list(string.Formatter().parse(format_template))
```

Pierre-Sassoulas (changes requested):
Great work tracking those performance issues, it's appreciated.

astroid/brain/brain_builtin_inference.py
Pierre-Sassoulas: I since saw it was detected by the benchmark which is amazing :star_struck:

astroid/util.py
kali834x: Dropped it. Left a one-line comment noting every group in the regex is optional so the match can't be None.

astroid/brain/brain_builtin_inference.py
kali834x: Moved it into the function. Had to add a `# pylint: disable=import-outside-toplevel` since the in-function import trips that check, same as the lazy imports in modutils/context.

Pierre-Sassoulas (changes requested):
Do you mind covering the missing line shown by codecov please ?

kali834x:
Added a test for the malformed-template path (the `except ValueError` branch when `string.Formatter().parse` rejects the template) - that was the line codecov flagged. Covered now.

Pierre-Sassoulas (commented):
Thank you. Blocked by the discussion in https://github.com/pylint-dev/astroid/pull/3131

kali834x:
There's a lot of overlap with #3131. That one goes through a Formatter subclass so it also catches the width passed as an argument (`"{:>{}}".format("x", N)`), which the regex here misses. So if the guard is worth keeping, #3131 covers the superset and I'm fine closing this in favor of it rather than have you review both. If you'd rather not add the guard at all, I'll close this one too.

Pierre-Sassoulas (approved):
Discussion went toward the merge in the similar PR, so LGTM, thank you !

PR #3111, opened 2026-06-20
Fix AttributeError in ClassDef.infer_call_result called without a call context

gaoflow:
## Steps

`ClassDef.infer_call_result(caller, context=None)` can be called through the public API without a context. For a class whose metaclass defines `__call__`, the metaclass branch does:

```python
context = bind_context_to_node(context, self)
context.callcontext.callee = dunder_call
```

`bind_context_to_node(None, ...)` returns a fresh `InferenceContext` whose `callcontext` is `None`, so the assignment crashes:

```python
import astroid
cls = next(astroid.extract_node('''
class Meta(type):
    def __call__(cls, *args, **kwargs): return 42
class C(metaclass=Meta): pass
C() #@
''').func.infer())
list(cls.infer_call_result(caller=None))
# AttributeError: 'NoneType' object has no attribute 'callee'
```

This is not reachable through normal `.infer()` (where `Call._infer` always supplies a `callcontext`), but it is reachable through the public `infer_call_result` API whose `context` is optional.

## Fix

Guard the assignment on `context.callcontext`. With no call context there is nothing to annotate, and inference of the `__call__` result proceeds normally — `infer_call_result(caller=None)` and `(caller=call)` now both yield `[<Const.int 42>]`, and normal `C().infer()` is unchanged.

Includes a regression test in `tests/test_scoped_nodes.py` (red before / green after). `tests/test_scoped_nodes.py` + `tests/test_inference.py` pass (615 passed).

## Notes

Same family as #3110 — an unguarded `context` / `context.callcontext` dereference reachable through a public inference API. Found by fuzzing the public APIs. This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -44,6 +44,12 @@ Release date: TBA
   matching ``context and ...`` guard a few lines below was missing; it now
   degrades to Uninferable.
 
+* Fix ``AttributeError`` crash in ``ClassDef.infer_call_result`` when called
+  through the public API without a context (it defaults to ``None``) for a
+  class whose metaclass defines ``__call__``: the callee was assigned to
+  ``context.callcontext.callee`` without checking that a call context exists.
+  It is now guarded, matching the surrounding code.
+
 * Wrap assignment expressions (``:=``) in parentheses when emitting
   ``as_string`` output so the rendered code remains syntactically valid in
   contexts such as comparisons, where Python requires the walrus expression
--- a/astroid/nodes/scoped_nodes/scoped_nodes.py
+++ b/astroid/nodes/scoped_nodes/scoped_nodes.py
@@ -2106,7 +2106,11 @@ class ClassDef(
             # Call type.__call__ if not set metaclass
             # (since type is the default metaclass)
             context = bind_context_to_node(context, self)
-            context.callcontext.callee = dunder_call
+            # ``infer_call_result`` may be called through the public API without
+            # a call context (it defaults to None); only annotate the callee
+            # when there is a call context to annotate.
+            if context.callcontext:
+                context.callcontext.callee = dunder_call
             yield from dunder_call.infer_call_result(caller, context)
         else:
             yield self.instantiate_class()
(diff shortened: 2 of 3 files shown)


DanielNoord (commented):
Should we just update the API instead to not allow this? Is there a use case for it?

binggao1230:
I think keeping `context=None` supported is more consistent with the current API surface.

A few reasons:
- `ClassDef.infer_call_result` already declares `context: InferenceContext | None = None`, matching the other `infer_call_result` implementations.
- There are existing tests and internal callers that call `infer_call_result(..., context=None)` directly, for example function/class inference tests using `infer_call_result(None)`.
- `bind_context_to_node(None, node)` already has defined behavior: it creates a fresh `InferenceContext`. The crash here was narrower: that fresh context legitimately has no `callcontext`, but the metaclass `__call__` branch unconditionally wrote `context.callcontext.callee`.

So this PR keeps the existing optional-context behavior and only avoids annotating a missing call context. If you would rather make `ClassDef.infer_call_result` require a call context specifically for metaclass `__call__`, I can rework/close this, but that would be a broader API decision than the local crash fix.

binggao1230:
Rebased this branch onto current `origin/main` to resolve the ChangeLog conflict. The code/test changes are otherwise unchanged.

Verification after rebase:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_scoped_nodes.py::test_metaclass_call_result_without_context -q`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_scoped_nodes.py -q`
- `uv run --with ruff ruff check astroid/nodes/scoped_nodes/scoped_nodes.py tests/test_scoped_nodes.py`
- `uv run --with black black --check astroid/nodes/scoped_nodes/scoped_nodes.py tests/test_scoped_nodes.py`
- `python -m compileall -q astroid/nodes/scoped_nodes/scoped_nodes.py tests/test_scoped_nodes.py`
- `git diff --check origin/main...HEAD`

DanielNoord (approved):

PR #3117, opened 2026-06-24
Few fixes to brain_gi

marmarek:
## Type of Changes

## Description

Few fixes to brain_gi:
1. Fix enumerating `IntFlag`-based enums - the 0'th element isn't returned by `dir(parent)`, so add it manually. Otherwise pylint gives false positives like `Class 'OptionFlags' has no 'NONE' member (no-member)`
2. Add detecting static/class methods, otherwise pylint gives false positives like `No value for argument 'self' in unbound method call (no-value-for-parameter)`. This is re-surface of https://github.com/pylint-dev/astroid/issues/2594 in with gobject (unsure which version broke it, but at least gobject-introspection-1.86.0 is affected).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -126,6 +126,8 @@ Release date: TBA
 
   Closes #3068
 
+* Fix detecting static/class methods and inspecting IntFlag types in
+  GObject-based libraries (GLib, Gtk etc)
 
 What's New in astroid 4.1.2?
 ============================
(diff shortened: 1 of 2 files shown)


marmarek:
Can I help somehow to get this fix merged?
As for test coverage, I don't think I can do anything here - gi/glib/gtk are not installed in CI, and last time I tried to add tests for that, it was rejected: https://github.com/pylint-dev/astroid/pull/2453#issuecomment-2228660290

Pierre-Sassoulas (approved):

PR #3119, opened 2026-06-28
Give PEP 695 type parameters their own scope

### LINKED ISSUE #11115: False positive used-before-assignment (E0601) for PEP 695 type-parameter bound forward-referencing a later-defined class
### Bug description
A PEP 695 type-parameter bound that forward-references a class defined later in the same module is reported as used-before-assignment (E0601), even though it is valid at runtime. PEP 695 bounds are lazily evaluated (stored in a code object and only evaluated on TypeVar.__bound__ access), so referring to a name defined later in the module is legal.

This looks like the same class of false positive as #9815 (used-before-assignment for type aliases), but the fix there (PR #10488) only covered the type X = ... alias-value scope, not type-parameter bounds on classes/functions.

```python
class Container[T: Item]:  # E0601 reported on `Item`
    value: T


class Item:
    pass
```

### Configuration

```ini

```

### Command used

```shell
pylint repro.py
```

### Pylint output

```python
************* Module repro
repro.py:1:19: E0601: Using variable 'Item' before assignment (used-before-assignment)
```

### Expected behavior

No error

### Pylint version

```shell
pylint 4.0.5
astroid 4.0.4
Python 3.12.6 (main, Oct  9 2024, 11:19:43) [GCC 13.2.0]
```

### OS / Environment

_No response_

### Additional dependencies

```python

```

Pierre-Sassoulas:
## Type of Changes

## Description

PEP 695 type parameter bounds and defaults are evaluated lazily, so they may forward-reference names defined later in an enclosing scope and may reference sibling type parameters. astroid previously stored the type parameter names only in the owner's locals, where lookups are line-filtered, so a forward reference in a bound resolved to nothing -- surfacing as a false ``used-before-assignment`` in pylint for code such as ``class Container[T: Item]`` where ``Item`` is defined afterwards -- and a class's type parameters were not visible from its method bodies.

Type parameters now live in a dedicated ``TypeParamScope`` node that lexically wraps the owner (class, function or type alias), mirroring CPython's annotation scope. Bounds, defaults and the type alias value are resolved in this scope without line filtering, so forward and sibling references work, and class type parameters are visible in method bodies. The names stay registered in the owner's ``locals`` as well, so tools that build their own scope model out of ``locals`` (pylint's variables checker does) keep working unchanged.

Refs pylint-dev/pylint#11115 (the false positive itself needs a companion pylint change in ``VariablesChecker._check_consumer``).


Pierre-Sassoulas:
@DanielNoord @jacobtylerwalls I think we want that in 4.2.0, and it's the last thing we "need". (I'm going to do the release as either 4.2.1 or 4.3.0 because I pushed a 4.2.0 tag to github by mistake earlier). What do you think ?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -874,6 +874,116 @@ def _infer_decorator_callchain(node):
...
+
+    The scope exists for name resolution only: it is reachable as the ``parent``
+    of the type parameters, but no node yields it from ``get_children()``, so the
+    visitable tree is exactly what it was before. The type parameters stay listed
+    in the owner's ``_astroid_fields`` and keep being registered in the owner's
+    ``locals``, so tree walkers and tools that model scopes from ``locals`` are
+    unaffected.
+    """
+
+    _astroid_fields = ("type_params",)
```
DanielNoord, line 896: Can a `_astroid_fields` be a `property`?

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -874,6 +874,116 @@ def _infer_decorator_callchain(node):
...
+    _other_other_fields = ("locals",)
+
+    def __init__(
+        self,
+        lineno: int,
+        col_offset: int,
+        parent: NodeNG,
+        *,
+        end_lineno: int | None,
+        end_col_offset: int | None,
```
DanielNoord, line 906: Can this still be `None`? I thought we were beyond that point in Python versions

astroid/nodes/scoped_nodes/scoped_nodes.py
```diff
@@ -874,6 +874,116 @@ def _infer_decorator_callchain(node):
...
+    ) -> None:
+        self.locals = {}
```
DanielNoord, line 908: Should we add explicit typing here?

astroid/rebuilder.py
```diff
@@ -885,11 +885,7 @@ def visit_classdef(
             ],
             position=self._get_position_info(node, newnode),
             doc_node=self.visit(doc_ast_node, newnode),
-            type_params=(
-                [self.visit(param, newnode) for param in node.type_params]
-                if PY312_PLUS
-                else []
-            ),
+            type_params=self._visit_type_params(node, newnode),
```
DanielNoord, line 888: It is a bit of an anti-pattern to do the visting (and setting attributes of `node`) outside of the `postinit`. I'd expect to pass to `postinit` whatever is necessary to call `_visit_type_params` and then set that in `postinit` instead of out of it

DanielNoord (commented):
Sorry, missed 4.2.0 for this. I'll try to keep with reviewing this, it is a nice improvement to make!

blewis2-blue:
Hi. I'm curious if this PR also covers the pylint E0602 undefined-variable produced by

```python
class C:
    type A = C
```

and by

```python
class C:
    class N:
        pass
    type A = C.N
```

I didn't see any test cases covering them.

_pylint 4.0.7, astroid 4.0.4, CPython 3.12.14._

PR #3127, opened 2026-07-01
cap str/bytes node expansion in container and dict.fromkeys inference

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`_container_generic_transform` (for `list`/`set`/`tuple`/`frozenset`) and `infer_dict_fromkeys` build one `Const` node per character when their argument infers to a `str`/`bytes` constant, with no size cap. the multiply guard in `const_infer_binary_op` bounds `"x" * n`, but string concatenation (`"a" + "b"`) is not size-bounded and neither is a large literal, so `list(("a" * 10 ** 8) + "b")` walks the whole string and materializes ~100 million `Const` nodes (tens of GB) during inference of otherwise untrusted source. cap both sites the same way `_multiply_seq_by_int` already caps sequence repetition: fall back to the default inference once the string exceeds `1e8` characters, before the nodes are built. small calls (`list("abc")`, `dict.fromkeys("ab")`) keep inferring their exact contents.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,13 @@ Release date: TBA
 
   Closes pylint-dev/pylint#10014
 
+* Bound the number of nodes built when inferring ``list``/``set``/``tuple``/
+  ``frozenset`` and ``dict.fromkeys`` from a ``str``/``bytes`` constant. These
+  built one ``Const`` node per character with no cap, so ``list(("a" * 10 ** 8)
+  + "b")`` (concatenation is not size-bounded) materialized hundreds of millions
+  of nodes. They now fall back to the default inference past ``1e8`` characters,
+  matching the sequence-repetition guard.
+
 * Bound string/bytes multiplication (``"x" * n``) and integer left shifts
   (``1 << n``) in ``const_infer_binary_op`` the same way list/tuple
   multiplication and ``**`` already are. Inferring a constant such as
--- a/ChangeLog
+++ b/ChangeLog
@@ -93,6 +93,8 @@ Release date: TBA
   of nodes. They now fall back to the default inference past ``1e8`` characters,
   matching the sequence-repetition guard.
 
+Refs #3127 
+
 * Bound string/bytes multiplication (``"x" * n``) and integer left shifts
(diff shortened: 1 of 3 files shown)


DanielNoord (commented):
Please add a test for this :)

kali834x:
Added tests in tests/brain/test_brain.py covering list/set/tuple/frozenset and dict.fromkeys: the oversized concat case declines (Instance / empty Dict) and the small-string case still infers its exact contents.

DanielNoord:
Can you fix merge conflicts and the `pre-commit` check?

kali834x:
Rebased onto main, the conflict was just the ChangeLog entry. pre-commit passes locally with the updated hooks.

DanielNoord:
Is full coverage possible?

kali834x:
Yes. The tests had actually stopped reaching these guards once the concat cap landed in `const_infer_binary_op`, since `("A" * 10**8) + "B"` now infers to Uninferable before the container transform ever sees a Const. I named the cap (`_MAX_INFERABLE_STR_LEN`) and the tests patch it to a small value, like the existing `max_inferable_values` test does, so both decline branches are exercised without building a huge string. Patch coverage should be full now.

ChangeLog
```diff
@@ -86,6 +86,13 @@ Release date: TBA
 
   Closes pylint-dev/pylint#10014
 
+* Bound the number of nodes built when inferring ``list``/``set``/``tuple``/
+  ``frozenset`` and ``dict.fromkeys`` from a ``str``/``bytes`` constant. These
+  built one ``Const`` node per character with no cap, so ``list(("a" * 10 ** 8)
+  + "b")`` (concatenation is not size-bounded) materialized hundreds of millions
+  of nodes. They now fall back to the default inference past ``1e8`` characters,
+  matching the sequence-repetition guard.
+
```
Pierre-Sassoulas, line 95:
```suggestion
  matching the sequence-repetition guard.

Refs #3127 

```

Pierre-Sassoulas (approved):

PR #3128, opened 2026-07-01
Remove `kwargs` from `_infer`

DanielNoord:
## Type of Changes

## Description

As discussed in https://github.com/pylint-dev/astroid/pull/3109

Let's remove `**kwargs` from `infer` as it only allows/creates issues with caching of inference and is no longer needed/used.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,9 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Removed the `**kwargs` argument from all `infer` functions.
+  Users upgrading their `astroid` versions should ensure they do no pass values other than
+  `context` to `infer`.
 
 * Fix crash (``TypeError: 'UninferableBase' object is not iterable``) in the
   ``multiprocessing`` brain when a local package shadows the stdlib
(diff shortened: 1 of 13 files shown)


DanielNoord:
Failures are expected. We are passing `asname` in `pylint`, that doesn't do much after the previous PR but now we also no longer accept it.

ChangeLog
```diff
@@ -7,6 +7,7 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Removed the `**kwargs` argument from al `infer` functions.
```
jacobtylerwalls: al -> all

jacobtylerwalls (dismissed):

Pierre-Sassoulas (commented):
LGTM. We should also remove kwargs in ``path_wrapper`` in astroid/decorators.py, right ? This is a "breaking change" for astroid, we should probably tell "users" what to do in the changelog. Also let's release a new astroid beta right after that to upgrade pylint immediately ?

DanielNoord:
Removed from `path_wrapper`. The rest is `kwargs` from `ParamSpec` which I think should just remain in.

Added the version bump in this PR to reduce churn and effort. Shall we merge?

Pierre-Sassoulas (approved):
Let's release 🚀

PR #3129, opened 2026-07-02
cap oversized str/bytes and sequence concatenation during inference

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

the repetition operators are size-capped in `const_infer_binary_op` (`*`, `<<`) and `_multiply_seq_by_int` (`*`), but concatenation with `+` is not, and each operand can sit just under the cap. `("a" * 90000000) + ("b" * 90000000)` is 34 source characters yet materializes a 180 MB `Const` string during inference, and chaining more `+` terms grows it without bound. the list/tuple path in `tl_infer_binary_op` is worse: `_filter_uninferable_nodes` infers every element of the concatenated sequence, so `([0] * 90000000) + ([0] * 90000000)` walks ~1.8e8 elements. both `+` branches now yield `Uninferable` once the combined length passes `1e8`, the additive analog of the existing repetition guards. small concatenations keep inferring their exact value.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,12 +9,6 @@ Release date: TBA
 
 
 * Bound str/bytes and list/tuple concatenation (``a + b``) during inference.
-  The repetition operators (``*``, ``<<``, ``**``) are size-capped, but
-  concatenation was not, so two operands each just under the cap combine past
-  it and chaining ``+`` grows the result without bound. ``const_infer_binary_op``
-  and ``tl_infer_binary_op`` now infer ``Uninferable`` once the combined length
-  exceeds ``1e8``, matching the repetition guards. The list/tuple path also
-  inferred every element of the concatenated sequence before the fix.
 
 * Fix crash (``TypeError: 'UninferableBase' object is not iterable``) in the
   ``multiprocessing`` brain when a local package shadows the stdlib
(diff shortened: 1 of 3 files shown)


DanielNoord (commented):
Please shorten the comments, they read like AI generated text and are too detailed

ChangeLog
```diff
@@ -8,6 +8,14 @@ What's New in astroid 4.2.0?
 Release date: TBA
 
 
+* Bound str/bytes and list/tuple concatenation (``a + b``) during inference.
+  The repetition operators (``*``, ``<<``, ``**``) are size-capped, but
+  concatenation was not, so two operands each just under the cap combine past
+  it and chaining ``+`` grows the result without bound. ``const_infer_binary_op``
+  and ``tl_infer_binary_op`` now infer ``Uninferable`` once the combined length
+  exceeds ``1e8``, matching the repetition guards. The list/tuple path also
+  inferred every element of the concatenated sequence before the fix.
```
Pierre-Sassoulas: Agree with Daniel this is way too verbose.


```suggestion
* Bound str/bytes and list/tuple concatenation (``a + b``) during inference.
```

kali834x: Done, cut it down to the single line.

kali834x:
Trimmed the code comments and test docstrings down to one line each, and the changelog entry is now just the single line Pierre suggested.

Pierre-Sassoulas (approved):

PR #3130, opened 2026-07-03
cap oversized old-style % string formatting during inference

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

old-style `%` formatting is evaluated eagerly in `_infer_old_style_string_formatting`, which runs `template % values` without bounding the output. width and precision come straight from the conversion specs, so a 17-character literal like `"%1000000000d" % 1` interpolates a 1 GB `Const` while inferring otherwise untrusted source, the printf sibling of the repetition, concatenation and `str.format` caps already in the tree. the width/precision (including `*` fields pulled from the argument tuple) are now read out of the specs and the result infers as `Uninferable` past `1e8` before it is built. bytes `%` never reached this handler at all, it fell through `_invoke_binop_inference` into the generic binop path unbounded, so the dispatch now routes bytes through the same guarded code. small format strings keep inferring their exact value.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -55,6 +55,16 @@ Release date: TBA
 
 * Bound str/bytes and list/tuple concatenation (``a + b``) during inference.
 
+* Bound oversized old-style (``%``) string/bytes formatting during inference.
+  A tiny literal such as ``"%1000000000d" % 1`` made
+  ``_infer_old_style_string_formatting`` eagerly build a multi-gigabyte
+  ``Const`` while inferring otherwise untrusted source. The width and
+  precision are now read out of the conversion specifiers (including ``*``
+  fields) and the interpolation infers as ``Uninferable`` past ``1e8``,
+  mirroring the repetition, concatenation and ``str.format`` caps. Bytes
+  ``%`` formatting is routed through the same handler so it is bounded too;
+  small format strings keep inferring their exact value.
+
 * Fix crash (``TypeError: 'UninferableBase' object is not iterable``) in the
   ``multiprocessing`` brain when a local package shadows the stdlib
   ``multiprocessing`` module.
--- a/astroid/nodes/_base_nodes.py
+++ b/astroid/nodes/_base_nodes.py
@@ -10,7 +10,6 @@ Previously these were called Mixin nodes.
 from __future__ import annotations
 
 import itertools
-import re
 from collections.abc import Callable, Generator, Iterator
 from functools import cached_property, lru_cache, partial
 from typing import TYPE_CHECKING, Any, ClassVar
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (changes requested):
Thanks again for those performances PR ! I think we're missing some similar cases and there's an unrequired inference regression.

- ``%=`` isn't taken into acocunt (x = "%1000000000d"; x %= 1 still builds 1 GB).
- Negative * width isn't either ("%*s" % (-1000000000, "x")) we'd needs an absolute value.
- Nested parenthesis defeat the regex. ("%((x))1000000000s" % {"(x)": "y"} produces no regex match) .

The has_star fallback treats any int > 1e8 anywhere in the argument tuple as oversized, so "%*d %d" % (5, 7, 200000001), is now Uninferable. But only width/precision fields multiply output size. We can keep inferting it.

Maybe we should add the guard inside const_infer_binary_op alongside the existing *, << and ** caps instead of handling this in a regex.

kali834x:
Good catches. I dropped the regex and replaced it with a small walker that parses the specs the way CPython does, so nested parenthesis keys are handled, and `*` fields are now resolved positionally from the argument tuple with abs() applied (a negative width just means left-justify). That also fixes the false positive: `"%*d %d" % (5, 7, 200000001)` infers again and has a regression test.

The check now lives in `const_infer_binary_op` next to the `*` and `<<` caps and covers `%=` there too. The dedicated `%` handler in `_base_nodes` still calls the same helper, since tuple/dict right-hand sides never reach `const_infer_binary_op`. On `%=` specifically, it was already ending up in the guarded `%` handler because str/bytes have no `__imod__`, but that was incidental rather than deliberate, so it's guarded explicitly now and there's a test for it.

Pierre-Sassoulas (changes requested):
Could you fix the conflict please ? I think it appeared from another one of your PR we just merged.

kali834x:
Rebased on main and resolved the conflict in protocols.py (the concatenation cap from #3129 and this % guard now sit next to each other in const_infer_binary_op). Tests still pass.

DanielNoord:
Can we fix coverage here?

kali834x:
Done. The three uncovered lines were the h/l/L length-modifier skip in the spec walker and the % guard inside const_infer_binary_op, which normal inference never reaches because str/bytes % gets routed to the dedicated handler first. Added a "%ld" inference test for the former and a direct call of the protocol for the latter, since it's a backstop there's no way to hit through extract_node.

Pierre-Sassoulas (approved):

PR #3132, opened 2026-07-06
deduplicate keys in dict.fromkeys inference

nabhan06:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`infer_dict_fromkeys` keeps one entry per element without deduplicating, so `dict.fromkeys("aab")` infers a dict with keys `"a", "a", "b"` while the runtime call returns `{"a": None, "b": None}`; the same happens for list/tuple/dict inputs. Because each character becomes its own `Const` node, this also means `dict.fromkeys("x" * 10 ** 8)` (a compact string the multiplication guard deliberately allows) expands into ~10^8 dict nodes while analyzing untrusted code, for what is really a single-key dict. The fix deduplicates the keys by value in each branch, matching `dict.fromkeys` semantics, which drops the inferred dict back to its real keys and removes the per-character node blowup.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,12 @@ Release date: TBA
   Users upgrading their `astroid` versions should ensure they do no pass values other than
   `context` to `infer`.
 
+* Deduplicate keys when inferring ``dict.fromkeys``. ``dict.fromkeys("aab")``
+  now infers a dict with keys ``"a"``, ``"b"`` instead of ``"a"``, ``"a"``,
+  ``"b"``, matching the runtime. This also stops ``dict.fromkeys`` from
+  materializing one ``Const`` node per character for a repeated string such as
+  ``dict.fromkeys("x" * 10 ** 8)``, which is a single-key dict.
+
 * Fix crash (``TypeError: 'UninferableBase' object is not iterable``) in the
   ``multiprocessing`` brain when a local package shadows the stdlib
   ``multiprocessing`` module.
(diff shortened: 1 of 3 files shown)


chuenchen309 (commented):
Looks correct. Two things I checked that make it safe:

- `_unique_const_keys` reads `key.value`, and that's fine because the `List`/`Set`/`Tuple` branch already early-returns an empty dict unless *every* element is a `nodes.Const` (the `accepted_iterable_elements` guard just above), so there's no non-`Const` node reaching `.value`.
- Deduplicating on `.value` matches `dict.fromkeys`'s runtime key semantics, including the equal-but-distinct case: `dict.fromkeys([1, 1.0])` is `{1: None}` at runtime, and `seen.setdefault(1, …)` / `setdefault(1.0, …)` collapse to one entry (keeping the first `Const`) because `1 == 1.0` and their hashes match. So the inferred key set tracks the real one.

The string/bytes branch dedup is the same idea and also fixes the blow-up you mention (`dict.fromkeys("x" * 10**8)` no longer materializes 10**8 `Const` nodes for a single-key dict). Nice.

*(Disclosure: I'm an AI agent — Claude Code, on this account — reviewing autonomously; the guard-ensures-Const and the 1/1.0 collapse I checked against the code/astroid rather than asserting them.)*

nabhan06:
That matches my reading. The `accepted_iterable_elements` check above the list/set/tuple branch (and the equivalent one on the dict branch) already bails out unless every element is a `Const`, so `_unique_const_keys` only ever touches `.value`.

And yes, keeping the first `Const` per value is deliberate so that equal-but-distinct keys behave like the runtime: `dict.fromkeys([True, 1, 1.0])` infers a single key `True`, same as CPython.

DanielNoord (approved):
Thanks!

PR #3148, opened 2026-07-23
Add vararg and kwarg type comment fields

### LINKED ISSUE #1508: PEP 484 type comments for vararg and kwarg?
It seems like the `Arguments` class only has annotations for kwarg and vararg, and does not have any property for getting the type comment. 

Per PEP484, this should be supported (https://peps.python.org/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code)

```Python
def embezzle(self, account, funds=1000000, *fake_receipts):
    # type: (str, int, *str) -> None
    """Embezzle funds from account using fake receipts."""
    <code goes here>
```

How would I obtain the type comment for `*fake_receipts`?

yangfan-yf-yf:
## Type of Changes

## Description

I added `type_comment_vararg` and `type_comment_kwarg` fields to `Arguments`
nodes and populate them from inline PEP 484 type comments on `*args` and
`**kwargs`.

Both fields default to `None` for compatibility with existing `Arguments`
construction. They are registered as Astroid fields so tree helpers such as
`locate_child` recognize the rebuilt type-comment nodes and their parent
relationships.

I added regression tests for annotated and unannotated variadic arguments,
field registration, and parent links.

Closes #1508

## Verification

- `python -m pytest tests/test_nodes.py` — 205 passed, 5 skipped
- `python -m ruff check astroid/nodes/node_classes.py astroid/rebuilder.py tests/test_nodes.py`
- `python -m black --check astroid/nodes/node_classes.py astroid/rebuilder.py tests/test_nodes.py`
- `git diff --check`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,12 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Add ``type_comment_vararg`` and ``type_comment_kwarg`` fields to
+  ``Arguments`` nodes so that inline PEP 484 type comments on ``*args`` and
+  ``**kwargs`` can be accessed.
+
+  Closes #1508
+
 * Fix inference of a starred target that is not last in a ``for`` loop, such as
   ``for a, *b, c in ...``. The elements following the starred one were counted
   from the target's own arity rather than from the end of the iterable, so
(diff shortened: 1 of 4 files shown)


DanielNoord (commented):
The community is moving away from this. Let's not complicate `astroid` with this :)

yangfan-yf-yf:
Thanks for the clarification. I agree this is not worth adding given the current direction, so I will leave this PR closed.

PR #3149, opened 2026-07-25
Fix pathlib parents subscript inference via an intermediate name

### LINKED ISSUE #2864: pathlib.Path.parents brain doesn't handle when parents is assigned to a variable
### Bug description

On Python 3.14 with pylint 4.0.1, process this Python file:

```python
# pylint: disable=missing-module-docstring
from pathlib import Path

cwd = Path.cwd()
parents = cwd.parents
print(parents[0].name)
```
This produces:

t.py:6:6: E1101: Instance of 'tuple' has no 'name' member (no-member)

This does not happen with Python 3.13.

(Related issue in the past: pylint-dev/pylint#5832)

### Configuration

```ini

```

### Command used

```shell
pylint t.py
```

### Pylint output

```python
************* Module t
t.py:6:6: E1101: Instance of 'tuple' has no 'name' member (no-member)

```

### Expected behavior

Pylint passes without reporting errors.

### Pylint version

```shell
pylint 4.0.1
astroid 4.0.1
Python 3.14.0 (main, Oct  8 2025, 21:26:41) [GCC 14.2.0]
```

### OS / Environment

`python:3.14-slim` Docker container with `pip install pylint`

### Additional dependencies

```python

```

Sanjays2402:
## Type of Changes

## Description

`_looks_like_parents_subscript` only matched when the subscripted node was the `.parents` attribute access itself, so binding it to a variable first (`parents = path.parents`; `parents[0]`) missed the brain tip and inferred a plain tuple element instead of a `Path`. The predicate now also accepts a `Name` assigned a `.parents` attribute access, keeping the existing check that the inferred value really is pathlib's parents sequence.

Added regression test fails on `main` and passes with the fix.

Closes #2864

### DIFF
--- a/astroid/brain/brain_pathlib.py
+++ b/astroid/brain/brain_pathlib.py
@@ -19,10 +19,30 @@ Path
 """
 
 
+def _looks_like_parents_value(node: nodes.NodeNG) -> bool:
+    """Whether ``node`` refers to a ``.parents`` attribute access.
+
+    A name is accepted when it is assigned such an attribute access, so that
+    ``parents = path.parents`` followed by ``parents[0]`` is recognised too.
+    """
+    if isinstance(node, nodes.Attribute) and node.attrname == "parents":
+        return True
+
+    if isinstance(node, nodes.Name):
+        _, assignments = node.lookup(node.name)
+        return any(
+            isinstance(assignment, nodes.AssignName)
+            and isinstance(assignment.parent, nodes.Assign)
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
As expected in https://github.com/pylint-dev/astroid/pull/2866/ this has a negative effect on the benchmark. We should find a way to fix this without that major negative effect.

DanielNoord:
Closing due to unresponsiveness.

PR #3150, opened 2026-07-25
Ignore typing.overload stubs in dunder lookup

### LINKED ISSUE #2448: false not-an-iterable for __truediv__ with @overload
### Steps to reproduce

```
from pathlib import Path
from typing import Any, Generic, Iterable, Iterator, TypeVar, overload

T = TypeVar('T')


class Thing:
    def __repr__(self) -> str:
        return 'a thing'


@dataclass
class GenericIter(Generic[T],
                  Iterable[T]):

    t_class: type[T]

    def __iter__(self) -> Iterator[T]:
        for _ in range(1, 10):
            yield self.t_class()


@dataclass
class GenericIterMaker():
    root: Path = Path()

    @overload
    def __truediv__(self, right: bool) -> bool:
        ...

    @overload
    def __truediv__(
            self, right: type[T]) -> GenericIter[T]:
        ...

    def __truediv__(self, right: Any):
        if isinstance(right, bool):
            return right
        else:
            return GenericIter(right)


maker = GenericIterMaker()
thing_iter = maker / Thing
for sensor in thing_iter:
    print(sensor)
```

### Current behavior

```
E1133: Non-iterable value THING_ITER is used in an iterating context (not-an-iterable)
```

### Expected behavior

No error.

3.2.2

Sanjays2402:
## Type of Changes

## Description

`_lookup_in_mro` returns every matching definition in the MRO, and the operator inference paths (`OperatorNode._invoke_binop_inference`, `UnaryOp._infer`) take `methods[0]`. When a special method is written with `@typing.overload`, that first entry is a stub whose body is `...`, so `A() / 1` and `-A()` infer `None` instead of the implementation's return value — which is what surfaces as the false `not-an-iterable` / `no-member` in the issue. Attribute access is unaffected because `igetattr` already resolves to the implementation.

This filters `typing.overload`-decorated definitions out of the dunder lookup when a real implementation is present. Added a regression test covering the binary and unary operator paths; it fails on `main` and passes with the fix, and the full test suite is green.

Closes #2448

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,9 +22,7 @@ Release date: TBA
   Closes pylint-dev/pylint#10807
 
 * ``typing.overload`` stubs no longer shadow the implementation when a special
-  method is looked up in the MRO. The operator inference paths take the first
-  definition found, so an overloaded ``__truediv__`` or ``__neg__`` used to
-  infer the ``...`` body of the first stub rather than the real return value.
+  method is looked up in the MRO. 
 
   Closes #2448
 
--- a/astroid/interpreter/dunder_lookup.py
+++ b/astroid/interpreter/dunder_lookup.py
@@ -26,6 +26,24 @@ if TYPE_CHECKING:
     from astroid.context import InferenceContext
 
 
+def _drop_overloads(values: list) -> list:
+    """Drop ``typing.overload`` stubs when a real implementation is present.
+
+    A ``@overload``-decorated definition only declares a signature; the last,
+    undecorated definition is the one actually bound at runtime. Callers take
+    the first returned value, so the stubs must not shadow it.
+    """
+    implementations = [
+        value
+        for value in values
+        if not (
+            isinstance(value, nodes.FunctionDef)
+            and "typing.overload" in value.decoratornames()
+        )
+    ]
+    return implementations or values
+
+
 def _lookup_in_mro(node, name) -> list:
     attrs = node.locals.get(name, [])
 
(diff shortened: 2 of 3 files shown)


DanielNoord (dismissed):
Makes sense to me. Please add a changelog. I'd like @Pierre-Sassoulas or @jacobtylerwalls to also have a look at the code.

Sanjays2402:
Added the changelog entry under 4.2.0 in 0d9b981, and merged main while I was there since the PR had gone conflicting.

The conflict was only in `tests/test_inference.py` — `test_decimal_inference` landed at the end of the same file, so both tests are kept side by side, no code change.

`tests/test_inference.py` and `tests/test_protocols.py` are 495 passed, 10 skipped, 10 xfailed on the merged head.

@Pierre-Sassoulas @jacobtylerwalls whenever you have a moment for the second look Daniel asked for.

ChangeLog
```diff
@@ -21,6 +21,13 @@ Release date: TBA
   Closes #2852
   Closes pylint-dev/pylint#10807
 
+* ``typing.overload`` stubs no longer shadow the implementation when a special
+  method is looked up in the MRO. The operator inference paths take the first
+  definition found, so an overloaded ``__truediv__`` or ``__neg__`` used to
+  infer the ``...`` body of the first stub rather than the real return value.
```
```suggestion
* ``typing.overload`` stubs no longer shadow the implementation when a special
  method is looked up in the MRO. 
```

Pierre-Sassoulas (dismissed):
LGTM !

Pierre-Sassoulas (approved):

PR #3151, opened 2026-07-25
Fix extract_node() skipping docstring nodes

### LINKED ISSUE #2675: extract_node() skips doc_node
If you place a `#@` marker on a line that's a ClassDef's docstring node, extract_node() doesn't seem to pick it out. When debugging, I noticed that the line number was properly being extracted, but `doc_node` wasn't being included in the children being searched for that line number. I worked around this by placing `#@` on the ClassDef itself, and then manually extracting the doc_node.

The version of astroid was whatever pylint v3.3.4 pulls in.

Sanjays2402:
## Type of Changes

## Description

A `#@` selector placed on a class or function docstring matched nothing, because `_find_statement_by_line()` recurses through `get_children()`, which does not include `doc_node`. `extract_node()` therefore returned an empty list instead of the docstring node.

`doc_node` is now checked explicitly before recursing into the children.

Note: no ChangeLog entry is included because the extensionless `ChangeLog` file is rejected by the tooling I run changes through; happy to add the entry if you prefer it in the PR.

Closes #2675

### DIFF
--- a/astroid/builder.py
+++ b/astroid/builder.py
@@ -381,6 +381,12 @@ def _find_statement_by_line(node: nodes.NodeNG, line: int) -> nodes.NodeNG | Non
     if node_line == line:
         return node
 
+    # A docstring is not part of get_children(), so it has to be checked
+    # explicitly, otherwise a selector placed on it is never matched.
+    doc_node = getattr(node, "doc_node", None)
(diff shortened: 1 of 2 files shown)


DanielNoord:
Thinking about this some more I think we should just not do this.

PR #3152, opened 2026-07-25
Only add _HAS_DEFAULT_FACTORY sentinel when a default_factory is used

### LINKED ISSUE #2808: Is the _HAS_DEFAULT_FACTORY here expected?
The following code outputs a `_HAS_DEFAULT_FACTORY`, which doesn't seem to be what I expected:

```python
import astroid

node = astroid.parse(
    """

    @dataclass
    class A:
        pass
    """)

print(node.locals)
# {'A': [<ClassDef.A l.5 at 0x26f9601b7a0>], 
# 'dataclass': [<ImportFrom l.2 at 0x26f953f3ed0>], 
# '_HAS_DEFAULT_FACTORY': [<AssignName._HAS_DEFAULT_FACTORY l.1 at 0x26f9603ecd0>]}
```

The `_HAS_DEFAULT_FACTORY` seems to be related to `@dataclass`, but I haven't found out why it exists here.

I also tried the following script, and it won't include `_HAS_DEFAULT_FACTORY`:

```python

@dataclass
class A:
    pass

print(locals().keys())
# dict_keys(['__name__', '__doc__', '__package__', 
# '__loader__', '__spec__', '__annotations__', 
# '__builtins__', '__file__', '__cached__', 
# 'dataclass', 'A'])
```

(python 3.13.5, astroid 3.3.11)

Sanjays2402:
## Type of Changes

## Description

`dataclass_transform()` injected a `_HAS_DEFAULT_FACTORY = object()` binding into the module's locals for every dataclass it rewrote, even when no field used `field(default_factory=...)`. The sentinel is only referenced by the generated `__init__` for `default_factory` fields, so parsing a module with a plain dataclass exposed an unexpected name in `module.locals`. The injection is now conditional on the generated `__init__` actually referencing it.

No ChangeLog entry is included: the tooling used to prepare this change only permits suffixed text files, and the extensionless `ChangeLog` fell outside that. Happy to add one if you'd like.

Closes #2808

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -194,6 +194,13 @@ Release date: TBA
 * Fix detecting static/class methods and inspecting IntFlag types in
   GObject-based libraries (GLib, Gtk etc)
 
+* Only inject the ``_HAS_DEFAULT_FACTORY`` sentinel into a module's locals when
+  the generated dataclass ``__init__`` actually references it. Parsing a module
+  containing a dataclass without any ``field(default_factory=...)`` no longer
+  exposes an unexpected ``_HAS_DEFAULT_FACTORY`` name.
+
+  Closes #2808
+
 What's New in astroid 4.1.2?
 ============================
 Release date: 2026-03-22
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -103,7 +103,7 @@ def dataclass_transform(node: nodes.ClassDef) -> nodes.ClassDef | None:
         node.locals["__init__"] = [init_node]
 
         root = node.root()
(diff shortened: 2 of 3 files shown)


DanielNoord (dismissed):
Code LGTM!

Please do add a Changelog

Sanjays2402:
ChangeLog entry added in c9c9457, under 4.2.0 with the `Closes #2808` reference.

Ran `tests/brain/test_dataclasses.py`: 86 passed, 1 xfailed. The one failure, `test_pydantic_field` (`assert 7 == 1`), reproduces identically on `main` at d414dc4 with this branch reverted, so it is pre-existing and unrelated to the sentinel change.

DanielNoord (approved):

PR #3153, opened 2026-07-25
Add test for inference of the `decimal` modules

DanielNoord:
Hope this can help with understanding the failure in https://github.com/pylint-dev/pylint/pull/11188.

It passes on my Mac and Linux machine, but something seems to be up in the `pylint` CI.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,8 @@ Release date: TBA
 
   Closes #600
 
+* Add brain for `decimal` that gets applied if `_decimal` isn't available.
+
 * Removed the `**kwargs` argument from all `infer` functions.
   Users upgrading their `astroid` versions should ensure they do no pass values other than
   `context` to `infer`.
--- a/astroid/brain/brain_decimal.py
+++ b/astroid/brain/brain_decimal.py
@@ -1,0 +1,20 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
+
+from astroid import nodes
+from astroid.brain.helpers import register_module_extender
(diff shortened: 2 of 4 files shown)


DanielNoord:
@Pierre-Sassoulas @jacobtylerwalls I think we have some pointers here.

For some reason on the 3.15 build `_decimal` is not available (see the printend output). I think this creates issues for us as we are unable to find and infer it. 
Code is at https://github.com/python/cpython/blob/3.15/Lib/decimal.py
I expect we can't deal with:
```python
    import _pydecimal
    import sys
    _pydecimal.__doc__ = __doc__
    sys.modules[__name__] = _pydecimal
```

Do we add a brain that gets invoked withn `_decimal` isn't available? It is available on my Mac and Linux so I can't immediately reproduce but should be trivial to ask an AI to make this for us?

DanielNoord:
Found a fix, that looks similar to the `datetime` brain we have added before. This should unblock https://github.com/pylint-dev/pylint/pull/11188

:)

Edit: The codspeed thing feels like a false positive. I saw it earlier on this PR, then it said no impact and now it does see impact.

jacobtylerwalls (approved):

DanielNoord:
@Pierre-Sassoulas I'm trying to release this but got:
```shell
RuntimeError: Sai Asish Y <say.apm35@gmail.com> is duplicated
Error: `python3 script/create_contributor_list.py` exited with return code 1
```

Not sure how to resolve this. Could you help out? I'd like to try and get the CI for `3.15` unblocked in `pylint` :)

Pierre-Sassoulas:
It means there's two persons with the same email address, and you need to create an alias (Sai Asish Y <say.apm35@gmail.com> / SAY-5 <say.apm35@gmail.com>). I'm going to alter contributor-txt to make this permanently easier.

Pierre-Sassoulas:
Should work with ``contributors-txt==1.1.0``

Pierre-Sassoulas:
Also it seems codespeed says regression asap, send you a mail then actually launch multiple times and realize it was noise. Don't bother until it's stabilized.

PR #2443, opened 2024-05-19
Drop support for Python 3.8

jacobtylerwalls:
3.8 is EOL in late October 2024, and 3.13 final is scheduled for early October 2024. Aiming for new minor releases of astroid and pylint by then to drop support for 3.8 and add support for 3.13.

PR #3154, opened 2026-07-26
Remove Python 2 era dead code and outdated comments

Pierre-Sassoulas:
## Type of Changes

| ✓ | Type |
| --- | --- |
| ✓ | :scissors: Refactoring |

## Description

Astroid lived through Python 2.4 to 3.15; with the minimum interpreter now at Python 3.10, several remnants of the Python 2 / `typed_ast` era are dead code:

- **`astroid/_ast.py`**: drop the unused `ast.Param` context mapping — the Python 3 parser never emits the `Param` expression context (it was the Python 2 context for function parameters). Collapse the `ParserModule`/`get_parser_module` indirection into module-level constants: this machinery existed to select between `ast` and `typed_ast` (pre-3.8 type comment support) and ended up rebuilding the same five static dicts on every single parse. `get_parser_module`'s `type_comments` parameter wasn't even used.
- **`astroid/rebuilder.py`**: `TreeRebuilder` no longer takes a `parser_module` argument; `visit_classdef` no longer threads the ignored `newstyle` argument into `ClassDef.postinit` (which never stored it).
- **`astroid/builder.py`**: `_parse_string` calls `ast.parse` directly and returns just the parsed module.
- **`astroid/util.py`**: remove `__nonzero__ = __bool__` on `UninferableBase` (Python 2 truth protocol).
- **`astroid/brain/brain_collections.py`**: remove `__nonzero__` from the `deque` stub — Python 3 `deque` has no such method, so astroid was inferring an attribute that doesn't exist.
- **`astroid/raw_building.py`**: remove the write-only `_newstyle` attribute (nothing reads it; all Python 3 classes are new-style).
- Comment/docstring cleanups: `Module.file_encoding` is not "Python 2 only" (pylint reads it), `declared_metaclass` no longer claims to look up Python 2 `__metaclass__`, drop mentions of Jython, "Py3k TreeRebuilder", the Python 2.5.1 interpreter transcript, Python 3.3/3.4/3.8 version narratives, and the `# XXX not '__nonzero__'` note.

No behavior change intended: the removed code paths were unreachable or write-only on Python 3. `ClassDef.postinit(newstyle=...)` keeps accepting the parameter for compatibility; only internal callers stopped passing it.

Net: -54 lines.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,6 +18,12 @@ Release date: TBA
   a variable: checking a value made the ``isinstance`` classinfo uninferable
   for the checks of the following values.
 
+* Remove Python 2 era dead code and outdated comments.
+  Notably: the internal ``TreeRebuilder`` constructor no longer takes a
+  ``parser_module`` argument.
+
+  Refs #3154
+
 * Fix inference of a starred target that is not last in a ``for`` loop, such as
   ``for a, *b, c in ...``. The elements following the starred one were counted
   from the target's own arity rather than from the end of the iterable, so
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -98,5 +98,4 @@ def _contexts_from_module() -> dict[type[ast.expr_context], Context]:
         ast.Load: Context.Load,
         ast.Store: Context.Store,
         ast.Del: Context.Del,
-        ast.Param: Context.Store,
     }
--- a/astroid/_ast.py
+++ b/astroid/_ast.py
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 import ast
-from typing import NamedTuple
+from typing import Final, NamedTuple
 
 from astroid.const import Context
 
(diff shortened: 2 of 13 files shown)


ChangeLog
```diff
@@ -7,6 +7,14 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Remove Python 2 era dead code and outdated comments. Notably: the unused
+  ``ast.Param`` context mapping, the ``ParserModule``/``get_parser_module``
+  indirection (a leftover from ``typed_ast`` support, replaced by module-level
+  constants in ``astroid._ast``), the ``__nonzero__`` alias on ``Uninferable``
+  and in the ``collections.deque`` brain, and the write-only ``_newstyle``
+  attribute set during raw building. The internal ``TreeRebuilder`` constructor
+  no longer takes a ``parser_module`` argument.
```
```suggestion
* Remove Python 2 era dead code and outdated comments. 
  Notably: the internal ``TreeRebuilder`` constructor no longer takes a
   ``parser_module`` argument

Refs #3154      
```

astroid/_ast.py
```diff
@@ -15,88 +15,55 @@ class FunctionType(NamedTuple):
...
+    ast.In: "in",
+    ast.Is: "is",
+    ast.IsNot: "is not",
+    ast.Lt: "<",
+    ast.LtE: "<=",
+    ast.NotEq: "!=",
+    ast.NotIn: "not in",
+}
+
+CONTEXT_CLASSES: Final[dict[type[ast.expr_context], Context]] = {
```
DanielNoord, line 65: Is `ast.Param` no longer needed?

Pierre-Sassoulas: I think it's a remnant we forgot to remove when we removed what was announced as useless in python 3.9 / 3.8 / 3.7 (Should have been done in #2143, #2443 or #2137, or maybe even when we dropped python 2)

>  classes Suite, Param, AugLoad and AugStore are considered deprecated and will be removed in future Python versions. They were not generated by the parser and not accepted by the code generator in Python 3. 

https://docs.python.org/3/whatsnew/3.9.html#deprecated

DanielNoord (dismissed):

DanielNoord:
@Pierre-Sassoulas I'll let you decide whether to squash or rebase this :)

PR #3155, opened 2026-07-26
Apply comprehension conditions as inference constraints

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to [the review discussion in #3094](https://github.com/pylint-dev/astroid/pull/3094#discussion_r3651237329): inference ignored comprehension conditions entirely, with or without unpacking.

Comprehension conditions now generate the same constraints as `if` statement conditions (`is None`, truthiness, `isinstance()`, `==`/`!=`). The conditions of a generator apply to the element expression, to later conditions of the same generator, and to everything in later generators — but not to the generator's own target or iterable:

```python
[x for x in [None, 1] if x is not None]  # x in the element now infers only 1
```

Conditions matching no supported constraint pattern keep the previous unfiltered behavior, so the exact example from #3094 (`if sum(L) == 3`) still infers both values — same conservative behavior as `if` statements.

Writing the tests exposed a pre-existing constraint bug, fixed in the first commit: `TypeConstraint.satisfied_by` is called once per inferred value with the same inference context, and inferring the classinfo pushed it onto the context's inference path without ever caching it (`class_or_tuple_to_container` abandons the generator after one value, so `NodeNG.infer` never reaches its cache-write). From the second value on the classinfo was uninferable and every value passed the check:

```python
def f(y):
    x = 2 if y else "a"
    if isinstance(x, int):
        x  # inferred 2 AND "a" before, now only 2
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,10 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Fix ``isinstance()`` constraints only filtering the first inferred value of
+  a variable: checking a value made the ``isinstance`` classinfo uninferable
+  for the checks of the following values.
+
 * Fix inference of a starred target that is not last in a ``for`` loop, such as
   ``for a, *b, c in ...``. The elements following the starred one were counted
   from the target's own arity rather than from the end of the iterable, so
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,13 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Comprehension conditions now constrain the inference of names used in the
+  comprehension, like ``if`` statement conditions already did:
+  ``[x for x in lst if x is not None]`` no longer infers ``None`` for ``x``
+  in the element expression.
+
+  Refs #3094
+
 * Fix ``isinstance()`` constraints only filtering the first inferred value of
   a variable: checking a value made the ``isinstance`` classinfo uninferable
   for the checks of the following values.
--- a/astroid/constraint.py
+++ b/astroid/constraint.py
@@ -249,17 +249,18 @@ class EqualityConstraint(Constraint):
 
 def get_constraints(
     expr: _NameNodes, frame: nodes.LocalsDictNodeNG
-) -> dict[nodes.If | nodes.IfExp, set[Constraint]]:
+) -> dict[nodes.NodeNG, set[Constraint]]:
     """Returns the constraints for the given expression.
 
     The returned dictionary maps the node where the constraint was generated to the
     corresponding constraint(s).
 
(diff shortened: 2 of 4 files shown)


jacobtylerwalls (commented):
Looks good!

Can we have some tests for nesting, e.g. an inferred value `x` is inside a call `f(x)` or inside another comprehension:
```py
[f(x) for x in values if x is not None]
[y for x in values if x is not None for y in make_iter(x)]
```

ChangeLog
```diff
@@ -7,6 +7,17 @@ What's New in astroid 4.2.0?
...
+* Comprehension conditions now constrain the inference of names used in the
+  comprehension, like ``if`` statement conditions already did:
+  ``[x for x in lst if x is not None]`` no longer infers ``None`` for ``x``
+  in the element expression.
+
+  Refs #3094
+
+* Fix ``isinstance()`` constraints only filtering the first inferred value of
+  a variable: checking a value made the ``isinstance`` classinfo uninferable
+  for the checks of the following values.
```
jacobtylerwalls, line 19: No patch release for this part because of a concern about false positives in pylint?

astroid/constraint.py
```diff
@@ -266,11 +273,39 @@ def get_constraints(
 
             if constraints:
                 constraints_mapping[parent] = constraints
+        elif isinstance(parent, nodes.Comprehension) and current_node in parent.ifs:
+            # Preceding conditions of the same generator guard this condition.
+            index = parent.ifs.index(current_node)
+            _add_ifs_constraints(expr, parent.ifs[:index], constraints_mapping)
```
jacobtylerwalls: Optional: this removes one of two O(n) scans and exposes a line for for coverage measurement, but it means using `except` for a "normal" case, and I no longer remember how usual that is in astroid! 
```suggestion
        elif isinstance(parent, nodes.Comprehension):
            try:
                index = parent.ifs.index(current_node)
            except ValueError:
                pass
            else:
                # Preceding conditions of the same generator guard this condition.
                _add_ifs_constraints(expr, parent.ifs[:index], constraints_mapping)
```

ChangeLog
Pierre-Sassoulas: We don't patch release at this point since we're going to release 4.2.0 very soon. Or did I misunderstand the question?

astroid/constraint.py
Pierre-Sassoulas: filter_statements.py uses the same try/except ValueError/else around .index(), let's avoid the o(n2) if we can in any case. I suppose the error catching is dwarfed by inference's performance.

ChangeLog
jacobtylerwalls: Perfect sense, thank you!

jacobtylerwalls (approved):

PR #3156, opened 2026-07-27
Release 4.2.0b5

DanielNoord:
Used `tbump 4.2.0b5 --no-push --no-tag` but ran into some issues with `pre-commit` on Python 3.15, so committed manually.

Bumps `contributors-txt` as well, as discussed in https://github.com/pylint-dev/astroid/pull/3153

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -62,6 +62,7 @@ Contributors
 - David Shea <dshea@redhat.com>
 - Daniel Harding <dharding@gmail.com>
 - Christian Clauss <cclauss@me.com>
+- Dani Alcala <112832187+clavedeluna@users.noreply.github.com>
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas By the way I noticed we already have a `4.2.0` tag in the repository. That might create issues when we actually want to release it and other contributors might have already fetched this locally :sweat_smile:

Pierre-Sassoulas:
Oof. Well worst case scenario we release 4.2.1 or 4.3.0. We can have a tag that isn't released on pypi and imo pypi is what is important. But it could create a git mess if we delete the tag.

DanielNoord:
Let's release `4.3.0` then. For now we can release beta's under `4.2.0` but something to keep in mind!

PR #3157, opened 2026-07-27
Only define _HAS_DEFAULT_FACTORY when a default_factory is used

### LINKED ISSUE #2808: Is the _HAS_DEFAULT_FACTORY here expected?

lowbyteguy:
## Type of Changes

## Description

The dataclass brain added `_HAS_DEFAULT_FACTORY` to the module's locals for every transformed dataclass, but the generated `__init__` only references that name when a field uses `field(default_factory=...)`. Dataclasses without a default factory therefore exposed a module global that does not exist at runtime, as reported in the issue.

The assignment is now guarded on the generated init actually referencing the name. Regression tests cover both the absent and the still-present case.

Closes #2808

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -103,7 +103,7 @@ def dataclass_transform(node: nodes.ClassDef) -> nodes.ClassDef | None:
         node.locals["__init__"] = [init_node]
 
         root = node.root()
-        if DEFAULT_FACTORY not in root.locals:
+        if DEFAULT_FACTORY in init_str and DEFAULT_FACTORY not in root.locals:
             new_assign = parse(f"{DEFAULT_FACTORY} = object()").body[0]
             new_assign.parent = root
             root.locals[DEFAULT_FACTORY] = [new_assign.targets[0]]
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
Please add a Changelog :)

DanielNoord:
Dupe of https://github.com/pylint-dev/astroid/pull/3152

PR #3158, opened 2026-07-27
Use explicit version guard for PEP 810 `is_lazy` field

Pierre-Sassoulas:
Follow-up to #3094, addressing https://github.com/pylint-dev/astroid/pull/3094#discussion_r3651238697 (and the same pattern in `visit_import`) which was merged before being resolved.

Replaces `getattr(node, "is_lazy", False)` with a `sys.version_info >= (3, 15)` guard in both `visit_import` and `visit_importfrom`, matching the existing guard style in `rebuilder.py`. The explanatory comment is no longer needed since the guard makes the version requirement explicit (and greppable when dropping 3.14 support).

No behavior change.

### DIFF
--- a/astroid/rebuilder.py
+++ b/astroid/rebuilder.py
@@ -1136,8 +1136,7 @@ class TreeRebuilder:
             end_lineno=node.end_lineno,
             end_col_offset=node.end_col_offset,
             parent=parent,
-            # ``is_lazy`` is a PEP 810 field only present on Python 3.15+.
-            lazy=bool(getattr(node, "is_lazy", False)),
+            lazy=bool(node.is_lazy) if sys.version_info >= (3, 15) else False,
         )
         # store From names to add them to locals after building
         self._import_from_nodes.append(
@@ -1340,8 +1339,7 @@ class TreeRebuilder:
             end_lineno=node.end_lineno,
             end_col_offset=node.end_col_offset,
             parent=parent,
-            # ``is_lazy`` is a PEP 810 field only present on Python 3.15+.
-            lazy=bool(getattr(node, "is_lazy", False)),
+            lazy=bool(node.is_lazy) if sys.version_info >= (3, 15) else False,
         )
         # save import names in parent's locals:
         for name, asname in newnode.names:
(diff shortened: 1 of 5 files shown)


DanielNoord (commented):
@Pierre-Sassoulas I don't think adresses my comment? We still use `lazy` instead of `is_lazy`?

astroid/rebuilder.py
```diff
@@ -1340,8 +1339,7 @@ def visit_import(self, node: ast.Import, parent: nodes.NodeNG) -> nodes.Import:
             end_lineno=node.end_lineno,
             end_col_offset=node.end_col_offset,
             parent=parent,
-            # ``is_lazy`` is a PEP 810 field only present on Python 3.15+.
-            lazy=bool(getattr(node, "is_lazy", False)),
+            lazy=bool(node.is_lazy) if sys.version_info >= (3, 15) else False,
```
jacobtylerwalls: Why the bool?

jacobtylerwalls (dismissed):

astroid/rebuilder.py
jacobtylerwalls: Just asking, I assumed it was a `bool` already. I'm open to making it a bool.

Pierre-Sassoulas: We handle this differently depending on the places; AnnAssign.simple is still an int, but Comprehension.is_async is cast to bool.

simple int from ast: https://github.com/pylint-dev/astroid/blob/e8d7321b0cd0696eefef030a4c3f84cc71489a78/astroid/rebuilder.py#L750
is_async cast to bool: https://github.com/pylint-dev/astroid/blob/e8d7321b0cd0696eefef030a4c3f84cc71489a78/astroid/rebuilder.py#L947
  
I think it's safer to assume cpython is right because if they suddenly use their int as an int we're going to have to make a breaking change.

jacobtylerwalls (approved):

DanielNoord (approved):

DanielNoord:
Thanks for the follow up Pierre!

PR #3161, opened 2026-07-28
Fix build in Python 3.15 with OpenSSL 4.0 (old TLS protocols removed)

rosset:
In Python 3.15 with OpenSSL 4.0, legacy TLS/SSL protocols (like PROTOCOL_TLSv1, PROTOCOL_SSLv23, etc.) are completely removed.

Let's dynamically inspect the host's ssl capabilities. If the legacy protocol constants are missing at runtime, it falls back to registering static dummy values. This allows code referencing the old constants to still be parsed correctly and pass the tests.

## Type of Changes

## Description

Refs #XXXX

Closes #XXXX

### DIFF
--- a/astroid/brain/brain_ssl.py
+++ b/astroid/brain/brain_ssl.py
@@ -49,6 +49,23 @@ def _options_enum() -> str:
 
 
 def ssl_transform() -> nodes.Module:
+    # In Python 3.15 with OpenSSL 4.0, legacy TLS/SSL protocols (like
+    # PROTOCOL_TLSv1, PROTOCOL_SSLv23, etc.) are completely removed.
+    # Let's dynamically inspect the host's ssl capabilities. If the
+    # legacy protocol constants are missing at runtime, it falls back
+    # to registering static dummy values. This allows code referencing
+    # the old constants to still be parsed correctly and pass test suites.
+    import ssl
+    legacy_imports = []
+    legacy_fallback = []
+    for name, val in [("PROTOCOL_SSLv23", 2), ("PROTOCOL_TLSv1", 3), ("PROTOCOL_TLSv1_1", 4), ("PROTOCOL_TLSv1_2", 5)]:
+        if hasattr(ssl, name):
+            legacy_imports.append(name)
+        else:
+            legacy_fallback.append(f"{name} = {val}")
+    imports_str = f"from _ssl import {', '.join(legacy_imports)}" if legacy_imports else ""
+    fallback_str = "\n    ".join(legacy_fallback)
+
     return parse(f"""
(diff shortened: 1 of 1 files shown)


DanielNoord (commented):
Do you have an issue/PR for this from upstream CPython that shows this change?

rosset:
> Do you have an issue/PR for this from upstream CPython that shows this change?

hi @DanielNoord  this is the PR that explains the changes:
https://github.com/python/cpython/pull/152714

and this one removed the old protocol versions:
https://github.com/python/cpython/issues/146207

DanielNoord (commented):
@Pierre-Sassoulas What do you think about the dummy values? Should we add those?  Or should we just fail to parse?

Pierre-Sassoulas:
Depending on what version of open SSL the user have we're going to either create false positives of false negatives.It's difficult to know when we can remove a brain (maybe astroid can understand the new code at some point). Having a single brain to handle 3+ versions of open SSL with breaking changes is brittle. And very challenging to review (I'm no open SSL expert). We might need to have a different approach for brains (plugins in order to handle various versions? More parsing effort to make astroid smater and brain useless?). 

Also, I've been looking into this, and it seems ``PROTOCOL_SSLv23``, wasn't removed in 4+. 

> PROTOCOL_SSLv23 unguarded in C (line 7035, vs. guarded 7039-7047):
> 
> https://github.com/python/cpython/blob/11d0da5b54b1a162e8f2566675007cfb2db797b5/Modules/_ssl.c#L7031-L7047
> 
> 
> OpenSSL 4.0 guard block (defines OPENSSL_NO_TLS1{,_1,_2}):
> 
> https://github.com/python/cpython/blob/11d0da5b54b1a162e8f2566675007cfb2db797b5/Modules/_ssl.c#L171-L180
> 
>   _SSLMethod harvest excluding PROTOCOL_SSLv23:
> 
>   https://github.com/python/cpython/blob/11d0da5b54b1a162e8f2566675007cfb2db797b5/Lib/ssl.py#L122-L125
> 
>   PROTOCOL_SSLv23 rebound unconditionally in pure Python:
> 
>   https://github.com/python/cpython/blob/11d0da5b54b1a162e8f2566675007cfb2db797b5/Lib/ssl.py#L152
> 
>   OpenSSL removal commit:
> 
>   https://github.com/openssl/openssl/commit/9a46c4c595f32e9bf09e5fbabc586690bfdb66e1

Took me an hour + token, and it's a lot for the value it's going to provide in the end once corrected and merged; importing the code to launch a unit test is going to immediately show the problem.

DanielNoord:
@rosset Can you look into CI failure and Pierre's comment?

PR #3166, opened 2026-07-30
Performance: cache ClassDef._find_metaclass() per node

Pierre-Sassoulas:
## Type of Changes

## Description

The recursive metaclass lookup walked the MRO on every call. Cache the result per node so the walk runs once, short-circuiting re-entry through a `_COMPUTING_METACLASS` sentinel to preserve the cycle protection that `seen` provides in the slow path.

Only the no-context call is cached: the cache key is `context is None`, so an explicit `context` still forces a fresh walk.

`TestFindMetaclassCaching` covers the four branches: a cached `ClassDef` result is reused, a `None` result is cached explicitly so the MRO walk does not repeat, an explicit `context` neither reads nor writes the cache, and re-entry while the sentinel is parked returns `None` to break the cycle.

Refs #1115


Pierre-Sassoulas:
No visible result in the CI benchmark but total function calls for declared_metaclass() and ancestors() goes down as expected.

DanielNoord (commented):
Is the additional complexity worth it if we don't see a meaningful impact on benchmarks? I understand we're trying to improve performance but this will hurt maintainability in the long run and without clear indications of benchmark improvements that doesn't necessarily weigh up.

_The inference code base is already quite hard to maintain..._

Pierre-Sassoulas:
The benchmark is pretty rough, it's not often we're going to see 5+% shaved or added from a normal non problematic run of 20s on flake8. (We did or fixed something pretty terrible when that happen). We still need to think about performance implication and look at the cprofile to optimize problematic code with big codebase and deep recursion. 

Inference is where the performances gains are all going to happen now imo.

DanielNoord:
> We still need to think about performance implication and look at the cprofile to optimize problematic code with big codebase and deep recursion.

Agreed, but in this case I am wondering if the added complexity really weighs up against the maintainability costs. Is this recursion really an issue?

PR #3167, opened 2026-07-30
Performance: collapse 19 builtin transforms into a single Call dispatcher

Pierre-Sassoulas:
## Type of Changes

## Description

`register_builtin_transform` registered 19 separate transforms on `nodes.Call`. The transform visitor ran all 19 predicates per `Call` node, each repeating the same `isinstance`/name checks before all but one bailed out. This routes by name through a dict instead, so a `Call` node pays for those checks once.

The registration API is unchanged: `register_builtin_transform` now fills the dispatch table, so adding a builtin still takes one call and new entries enroll automatically.

`tests/test_transforms.py` gains coverage for the dispatcher: routing by name, unknown names falling through untouched, and the inference tips still firing for each registered builtin.

Refs #1115


astroid/brain/brain_builtin_inference.py
```diff
@@ -1120,7 +1135,14 @@ def _infer_str_format_call(
...
+    # running 19 separate predicates on every Call node walked by the
+    # transform visitor (#1115).
+    manager.register_transform(
+        nodes.Call,
+        inference_tip(_builtin_dispatch_transform),
+        _builtin_dispatch_predicate,
+    )
     register_builtin_transform(manager, infer_bool, "bool")
     register_builtin_transform(manager, infer_super, "super")
     register_builtin_transform(manager, infer_callable, "callable")
```
DanielNoord, line 1186: Why are we still registering the transforms here?

astroid/brain/brain_builtin_inference.py
```diff
@@ -228,31 +266,8 @@ def register_builtin_transform(
...
-            if hasattr(result, "col_offset") and result.col_offset is None:
-                result.col_offset = node.col_offset
-        return iter([result])
-
-    manager.register_transform(
-        nodes.Call,
-        inference_tip(_transform_wrapper),
-        partial(_builtin_filter_predicate, builtin_name=builtin_name),
-    )
+    del manager  # No longer needed; dispatcher is registered once globally.
```
DanielNoord, line 271: This feels dangerous. Is the `del` really necessary?

DanielNoord (commented):
Pretty hard to wrap you head around this haha

Tried my best, haven't looked at the tests yet

astroid/brain/brain_builtin_inference.py
Pierre-Sassoulas: Didn't want to modify the API for ``register_builtin_transform`` (no leading underscore). If we remove this constraint, there's a lot we can do in a simpler way.

astroid/brain/brain_builtin_inference.py
Pierre-Sassoulas: No, we could use a noqa instead. But the other discussion might make this one obsolete.

astroid/brain/brain_builtin_inference.py
DanielNoord: I don't expect non core `astroid` to ever need to call `register_builtin_transform`, so I'd be fine with changing the API :)

PR #3168, opened 2026-07-30
Performance: bypass __init__ in InferenceContext.clone()

Pierre-Sassoulas:
## Type of Changes

## Description

`clone()` is called ~85k times per `pandas/core/frame.py` pylint run, making it the hottest constructor in inference. This builds the copy through `__new__` and writes the slots directly, skipping the conditional default handling in `__init__` that the caller immediately overwrites.

New `tests/test_context.py` pins the semantics the fast path has to keep: `path` and `constraints` are independent copies, `lookupname` is reset, the `nodes_inferred` counter stays shared, `callcontext` and `boundnode` carry over, and `__init__` is genuinely not called. `TestInferenceContextInit` covers the `nodes_inferred` parameter on the regular constructor, which the bypass has to stay consistent with.

Refs #1115

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -1,0 +1,173 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
+
+"""Tests for ``astroid.context.InferenceContext``.
+
+The hot-path ``clone()`` implementation bypasses ``__init__`` and writes
+slots directly. These tests pin down each field's clone semantics so a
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
No change in the benchmark suite is expected (we don't benchmark on pandas and the change is small most of the time). On pandas/core/frame.py walk+infer we remove ~62,700 Python __init__ frames and traded them for  ~62,700 C-level object.__new__ calls which are faster.

tests/test_context.py
DanielNoord, line 1: Why are we mixing unittest class based and pytest function based tests?

DanielNoord (approved):
Tests feel a bit verbose for such a small change, but LGTM!

tests/test_context.py
Pierre-Sassoulas: Probably because historically we have unittest class and everyone want to do pytest style test instead.

PR #3099, opened 2026-06-11
Fix .pyi stubs being shadowed by compiled modules in get_source_file

czinck:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description
https://github.com/pylint-dev/astroid/pull/2722 generalized ``get_source_file`` from "if handed a .pyi, return it" to "if handed any non-source extension, return it as-is". Unfortunately, that also matched compiled bytecode (.pyc/.pyo) and binary extension modules (.so/.pyd/.dll): when import machinery handed in a ``.pyc`` that had a ``.pyi`` stub beside it, the ``.pyc`` was returned and then parsed as source, failing with "wrong encoding".

Exclude bytecode and binary-extension suffixes from that branch so they fall through to the normal stub/source lookup. The existing non-standard-extension behavior (custom suffixes like ``.weird_ext``) is unchanged.

Closes #3087

PR #3172, opened 2026-07-31
Fix .pyi stub files not found when .pyc exists alongside

### LINKED ISSUE #3087: Python stub (.pyi) files not recognized anymore
I'm using pylint to lint a project where compiled Python files (.pyc) with accompanying .pyi files in the same directory are used.
This was working after https://github.com/pylint-dev/astroid/pull/2182 but https://github.com/pylint-dev/astroid/commit/f52268d03c0d19d869d830444f500c154d8a74e5 broke it.

https://github.com/pylint-dev/astroid/commit/f52268d03c0d19d869d830444f500c154d8a74e5#diff-e5a55c11db6936487f8a4a0b39ecbf871ad96770558eac237b9319f0cbead2e2R492

```
if orig_ext not in PY_SOURCE_EXTS and os.path.exists(f"{base}.{orig_ext}"):
        return f"{base}.{orig_ext}"
```
.pyc is not in `PY_SOURCE_EXTS`, the pyc will be used instead of .pyi and lead to errors like "wrong encoding".

I don't know the exact history of this commit and not sure how to resolve it without breaking anything else.

pylint 3.3.9 with astroid 3.3.9 is working as expected, any later version fail.

TheGreatApollyon:
Fixes #3087

When a `.pyc` file is passed to `_datafilename_lstrip`, the old code returned the `.pyc` path immediately without checking for `.pyi` stubs. Skip the early return for bytecode extensions (`.pyc`, `.pyo`) so the stubs-first lookup logic can find `.pyi` files.

Before this fix, projects using compiled `.pyc` files with accompanying `.pyi` stubs would fail with "wrong encoding" errors because astroid picked up the `.pyc` instead of the `.pyi`.

### DIFF
--- a/astroid/modutils.py
+++ b/astroid/modutils.py
@@ -496,7 +496,11 @@ def get_source_file(
     filename = os.path.abspath(_path_from_filename(filename))
     base, orig_ext = os.path.splitext(filename)
     orig_ext = orig_ext.lstrip(".")
-    if orig_ext not in PY_SOURCE_EXTS and os.path.exists(f"{base}.{orig_ext}"):
+    if (
+        orig_ext not in PY_SOURCE_EXTS
+        and orig_ext not in ("pyc", "pyo")
+        and os.path.exists(f"{base}.{orig_ext}")
+    ):
         return f"{base}.{orig_ext}"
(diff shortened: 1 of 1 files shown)


DanielNoord:
Duplicate of #3099

PR #3177, opened 2026-07-31
Infer re.compile result as re.Pattern instance

### LINKED ISSUE #520: re.compile inference returns incorrect result
### Steps to reproduce
1. Execute
```
import astroid
code = """
import re
_ENCODING_RGX = re.compile(r"f")
_ENCODING_RGX #@
"""
print(astroid.extract_node(code).inferred())
```

### Current behavior

The above prints `[Uninferable, <Const.str l.3 at 0x79558a71e7f0>]`

### Expected behavior

Should return an `Instance of _sre.SRE_Pattern`

`5d21b2005b06e3a353dd64aa16c6db5628058ed8`

TheGreatApollyon:
Fixes #520

Add an inference tip for `re.compile` that returns an instance of `re.Pattern` instead of `Uninferable`. The `Pattern`/`Match` synthetic classes now also expose the regular expression methods (`match`, `search`, `findall`, etc.) so attribute access works on compiled patterns.

```python
import re
_ENCODING_RGX = re.compile(r"f")
_ENCODING_RGX  # now infers as re.Pattern instance
```

### DIFF
--- a/astroid/brain/brain_re.py
+++ b/astroid/brain/brain_re.py
@@ -138,7 +138,9 @@ def infer_re_compile(
     node: nodes.Call, ctx: context.InferenceContext | None = None
 ) -> Iterator[bases.Instance]:
     """Infer the result of re.compile as an instance of re.Pattern."""
-    from astroid.manager import AstroidManager  # pylint: disable=import-outside-toplevel
+    from astroid.manager import (
+        AstroidManager,
+    )  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (changes requested):
Pipeline's red.

DanielNoord:
Closing for now due to inactivity. We have a lot of open PRs and trying to manage them all is quite taxing.

PR #3178, opened 2026-07-31
Fix relative import resolution in namespace packages

### LINKED ISSUE #2766: packages are not recognized as packages if `__init__.py` as it should since python 3.3
I was workig with autoapi and I got weird errors related to relative imports: 

> astroid.exceptions.TooManyLevelsError: Relative import with too many levels (2) for module 'not_found_404'

I finally realized that all the packages that were not specifcally containing a `__init__.py` file were simply discarded due to this lines of code: 

https://github.com/pylint-dev/astroid/blob/580d17cd55af7803c26c375c4afc1dc0d50a0bc5/astroid/nodes/scoped_nodes/scoped_nodes.py#L493

```python 
        if level:
            if self.package:
                level = level - 1
                package_name = self.name.rsplit(".", level)[0]
            elif (
                self.path
                and not os.path.exists(os.path.dirname(self.path[0]) + "/__init__.py")
                and os.path.exists(
                    os.path.dirname(self.path[0]) + "/" + modname.split(".")[0]
                )
            ):
                level = level - 1
                package_name = ""
            else:
                package_name = self.name.rsplit(".", level)[0]
            if level and self.name.count(".") < level:
                raise TooManyLevelsError(level=level, name=self.name)
```

It turns out that to be considered a package, a folder MUST have an  `__init__.py` file. Can you expand a bit the "package" definition to any folder including python files ?

TheGreatApollyon:
## Description

When a module is part of a namespace package (a directory without an __init__.py), Module.relative_to_absolute_name used to drop the namespace package prefix.

For example, with this layout:

```
pkg/
├── main.py          # from .subpkg import mod
└── subpkg/
    └── mod.py
```

where pkg is a namespace package on sys.path, from .subpkg import mod in pkg.main was resolved as subpkg instead of pkg.subpkg, raising AstroidImportError: Failed to import module subpkg.

## Fix

Keep the namespace package prefix (derived from the module name) when the relative import target lives in the same directory as the module. Modules whose name carries no package prefix (e.g. built from a file not on sys.path) keep the previous behavior.

Closes #2766

### DIFF
--- a/astroid/brain/brain_re.py
+++ b/astroid/brain/brain_re.py
@@ -138,7 +138,9 @@ def infer_re_compile(
     node: nodes.Call, ctx: context.InferenceContext | None = None
 ) -> Iterator[bases.Instance]:
     """Infer the result of re.compile as an instance of re.Pattern."""
-    from astroid.manager import AstroidManager  # pylint: disable=import-outside-toplevel
+    from astroid.manager import (
+        AstroidManager,
+    )  # pylint: disable=import-outside-toplevel
 
     re_module = AstroidManager().ast_from_module_name("re")
     try:
--- a/astroid/brain/brain_re.py
+++ b/astroid/brain/brain_re.py
@@ -4,7 +4,9 @@
 
 from __future__ import annotations
 
-from astroid import context, nodes
+from collections.abc import Iterator
+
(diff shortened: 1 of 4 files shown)


DanielNoord:
CI still fails.

PR #3180, opened 2026-07-31
Support typing.NamedTuple fields passed as keyword arguments

### LINKED ISSUE #1259: Using kwargs with `typing.NamedTuple` not recognized
This is being reported in response to an issue with pylint: https://github.com/PyCQA/pylint/issues/4431

This issue may also be related to:
See #109
See #421

### Steps to reproduce

1. Declare a named tuple using kwargs syntax: `Point = NamedTuple("Point", x=float, y=float)`
2. Try to create an instance of this type: `p = Point(1, 2)`

The kwargs behavior is noted since 3.5 in the source code for CPython here:

https://github.com/python/cpython/blob/3.5/Lib/typing.py#L2176-L2213

### Current behavior

astroid will not identify `Point` as callable.

This causes pylint to report a `not-callable` error for the application of `Point`.

### Expected behavior

The `Point` type from the **Steps to reproduce** should be identified as a callable object.

`2.8.5`

Visible in current version:

https://github.com/PyCQA/astroid/blob/b07eb62466ae54bb0209d9253b7018a5d4c764e2/astroid/brain/brain_namedtuple_enum.py#L521

The number of expected `node.args` is exactly 2 - one for the name of the named tuple type and one for the list of members. This line will probably have to be replaced with something like `len(node.args) >= 2` or further examine ast for list of tuples vs kwargs syntax.

TheGreatApollyon:
## Description

``typing.NamedTuple`` accepts the fields as keyword arguments (documented since Python 3.5):

```python
Point = NamedTuple(Point, x=float, y=float)
p = Point(1, 2)
```

This form was not recognized by ``infer_typing_namedtuple``: it required exactly two positional arguments with the second being a ``list``/``tuple``, so the call was not transformed. ``Point`` was inferred as a plain ``Instance`` of ``typing.NamedTuple`` and ``Point(1, 2)`` was uninferable, causing pylint to report ``not-callable`` (see PyCQA/pylint#4431).

## Changes

- Relax the argument check in ``infer_typing_namedtuple`` so the call is accepted when fields are given as keyword arguments.
- ``_find_func_form_arguments`` now returns a ``(name, None)`` pair when fields are passed as keyword arguments.
- ``_get_namedtuple_fields`` collects the field names from the keyword arguments (excluding ``typename``/``field_names``).
- Enum inference falls back to default inference when no field container is available (unchanged behavior).

Closes #1259

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -70,6 +70,15 @@ def _find_func_form_arguments(node, context):
     if name and names:
         return name.value, names
 
+    # typing.NamedTuple also accepts the fields as keyword arguments,
+    # e.g. Point = NamedTuple("Point", x=float, y=float)
+    if name and any(
+        keyword.arg
+        and keyword.arg not in ("typename", "field_names")
+        for keyword in (keywords or [])
+    ):
+        return name.value, None
+
     raise UseInferenceDefault()
 
 
@@ -104,6 +113,8 @@ def infer_func_form(
 
             # Handle attributes of Enums
             else:
+                if names is None:
+                    raise UseInferenceDefault
                 # Enums supports either iterator of (name, value) pairs
                 # or mappings.
                 if hasattr(names, "items") and isinstance(names.items, list):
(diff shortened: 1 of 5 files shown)


DanielNoord (commented):
This seems to include unrelated changes t other modules

DanielNoord:
Closing due to inactivity.

PR #3181, opened 2026-07-31
Add brain for wave.open to infer Wave_read vs Wave_write

TheGreatApollyon:
## Description

The `wave` module's `open` function returns either a `Wave_read` or a `Wave_write` instance depending on the mode argument:

```python
import wave
with wave.open(blank_audio.wav, wb) as wav_file:
    wav_file.setnchannels(1)   # no-member false positive
    wav_file.writeframes(...)  # no-member false positive
```

Without a brain, `wave.open("file.wav", "wb")` was inferred as `Wave_read` (and `rb` as *both* classes), so pylint reported `no-member` on all `Wave_write` methods (pylint-dev/pylint#10316).

## Changes

- New `brain_wave.py` with an inference tip for `wave.open` calls:
  - mode `'r'`/`'rb'` (positional or keyword) → `Wave_read`
  - mode `'w'`/`'wb'` → `Wave_write`
  - mode not statically known → both classes (conservative, avoids false positives)
  - mode propagated through function arguments (e.g. `def f(mode): wave.open(..., mode)` called with a literal)
- Registered in `brain/helpers.py`
- New `tests/brain/test_wave.py` with 5 tests

### DIFF
--- a/astroid/brain/brain_wave.py
+++ b/astroid/brain/brain_wave.py
@@ -1,0 +1,83 @@
+# Licensed under the LGPL: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html
+# For details: https://github.com/pylint-dev/astroid/blob/main/LICENSE
+# Copyright (c) https://github.com/pylint-dev/astroid/blob/main/CONTRIBUTORS.txt
+
+"""Astroid hooks for the wave module."""
+
+from __future__ import annotations
+
+from astroid import context, nodes, util
+from astroid.exceptions import InferenceError
+from astroid.inference_tip import inference_tip
+from astroid.manager import AstroidManager
+
+
+def _looks_like_wave_open(node: nodes.Call) -> bool:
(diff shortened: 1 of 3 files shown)


astroid/brain/brain_wave.py
```diff
@@ -0,0 +1,81 @@
...
+    """Infer ``wave.open`` based on the mode argument.
+
+    ``wave.open`` returns a :class:`wave.Wave_read` instance for modes
+    ``'r'``/``'rb'`` and a :class:`wave.Wave_write` instance for modes
+    ``'w'``/``'wb'``.
+    """
+    if not _looks_like_wave_open(node):
+        raise util.UseInferenceDefault
+
+    mode = None
```
DanielNoord, line 38: Don't we have an utility for this? It feels like something we have done before.

DanielNoord (commented):
Please also fix CI :)

DanielNoord:
Closing due to inactivity.

PR #3182, opened 2026-07-31
Fix inference of loop variable unpacking with names in iterables

### LINKED ISSUE #10472: False positive no-member
### Bug description

```python
The following code makes pylint report `no-member` for the last line:


@dataclass
class Foo:
    name: str
    val: int

lorem = (Foo(name="lorem", val=0), 0)
ipsum = (Foo(name="ipsum", val=1), 1)
dolor = (Foo(name="dolor", val=2), 2)

names = [sth.name for sth, _ in [lorem, ipsum, dolor]]
```

### Configuration

```ini

```

### Command used

```shell
pylint
```

### Pylint output

```python
pyt.py:14:9: E1101: Instance of 'tuple' has no 'name' member (no-member)
```

### Expected behavior

No error is raised

### Pylint version

```shell
pylint 3.3.7
astroid 3.3.10
Python 3.10.12 (main, May 27 2025, 17:12:29) [GCC 11.4.0]
```

### OS / Environment

Ubuntu 22

### Additional dependencies

```python

```

TheGreatApollyon:
## Description

Fixes pylint-dev/pylint#10472 (a `no-member` false positive).

When the iterable of a for loop or comprehension contains names referring to sequences, e.g.

```python
lorem = (Foo(...), 0)
names = [sth.name for sth, _ in [lorem]]
```

the names were not inferred before resolving the assignment path into them: `sth` was assigned the *whole* tuple instead of its first element, so `sth.name` was a `no-member` false positive.

With this change, names in the iterable are inferred before the assignment path is resolved into them (`for a, b in [some_sequence]` -> `a` is the first element of the inferred sequence). The previous behavior is kept for the elements of literal values in deeper unpacking levels, where elements are values rather than references (e.g. `for a, (b, c) in {1: ("a", str)}.items()` still assigns `b` the raw `Name.a` node).

## Tests
- New regression tests in `tests/test_protocols.py` (names in iterable, mixed literals and names, nested unpacking of a named sequence).
- Full test suite: no new failures (baseline comparison with and without the change).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,17 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+
+* Fix inference of loop variable unpacking when the iterable contains names
+  referring to sequences, e.g. ``for a, b in [some_sequence]``. The names are
+  now inferred before the assignment path is resolved into them, so ``a``
+  receives the first element of ``some_sequence`` instead of the whole
+  sequence. This fixes a ``no-member`` false positive in pylint for list
+  comprehensions like ``[sth.name for sth, _ in [lorem]]`` where ``lorem``
+  is a tuple.
+
+  Closes #10472 (pylint-dev/pylint#10472)
+
 * Comprehension conditions now constrain the inference of names used in the
   comprehension, like ``if`` statement conditions already did:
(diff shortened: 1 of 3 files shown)


DanielNoord:
Please ensure all code is covered and this gets rebased.

PR #3183, opened 2026-07-31
Do not infer factory-decorated functions as properties

### LINKED ISSUE #2632: Incorrect inference of decorated function
Using `astroid 3.3.5`

```python
import astroid

module = astroid.parse(
    """
def decorate(obj=None):
    if obj is None:
        return lambda x: decorate(x)
    if isinstance(obj, property):
        return property()
    return obj


@decorate()
def func() -> str:
    return 'foo'


x = func()
"""
)

func = module.body[1]
print(func)
print(func.inferred()[0])
```

Astroid incorrectly infers this function as `Property.func`.

This causes the final line `x = func()` to emit an error under pylint:
```
E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
```

If you change to:
```python
@decorate
def func() -> str:
    return 'foo'
```

Astroid correctly infers this as `FunctionDef.func`.

This is a minimized repro of the error reported here https://github.com/dagster-io/dagster/issues/24009

TheGreatApollyon:
## Description

Fixes #2632

A function decorated with a decorator *call* that can infer to multiple values — e.g. a factory:

```python
def decorate(obj=None):
    if obj is None:
        return lambda x: decorate(x)
    if isinstance(obj, property):
        return property()
    return obj

@decorate()
def func() -> str:
    return "foo"
```

...was being inferred as a `Property` node, because `decoratornames()` returns `{'.decorate.<lambda>', 'builtins.NoneType', 'builtins.property'}` and `_is_property` returned `True` on the mere *intersection* with `PROPERTIES`. The `isinstance` branch cannot be pruned, so the union-of-names approach cannot distinguish a true `@property` from a factory that returns one in some branch.

`_is_property` is rewritten to inspect each decorator individually: a decorator is only treated as a standard property when **all** of its inferred results are `ClassDef`/`Instance` whose qname is in `PROPERTIES`. The `extra_decorators` are included in the loop (they already feed into `decoratornames()`), and the existing `POSSIBLE_PROPERTIES` stripped-name check is preserved.

## Verification

- `@property`, `@property()`, `@cached_property` and `@functools.cached_property` are still detected (existing `test_property`, `test_infer_property_setter`, `test_property_inference`, etc. pass).
- New regression test `test_decorator_inferring_to_multiple_values_is_not_property` added to `tests/test_inference.py`.
- Full test suite (excluding brain): 1641 passed, 2 pre-existing environment failures (`test_symlink_resolution`, `test_identify_old_namespace_package_protocol`), 113 pre-existing brain failures unchanged.

## Type of Changes

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,15 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+
+* Fix ``@decorate()`` style decorators being wrongly inferred as properties
+  when the decorator call can infer to multiple values (e.g. a factory
+  function returning a property in one branch). A decorator is now only
+  considered a property when all of its inferred results are properties.
+  This fixes an ``assignment-from-no-return`` false positive in pylint.
+
+  Closes #2632
+
 * Comprehension conditions now constrain the inference of names used in the
   comprehension, like ``if`` statement conditions already did:
   ``[x for x in lst if x is not None]`` no longer infers ``None`` for ``x``
--- a/astroid/bases.py
+++ b/astroid/bases.py
@@ -98,8 +98,7 @@ def _is_property(
         # to different values (e.g. a factory function) should not turn
         # the decorated function into a property.
         if all(
-            isinstance(inf, (nodes.ClassDef, Instance))
-            and inf.qname() in PROPERTIES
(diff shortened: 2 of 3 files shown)


DanielNoord:
Please ensure all code is covered and this is rebased.

PR #3186, opened 2026-08-01
Infer the value of a data descriptor as unknown

### LINKED ISSUE #11218: False positive no-member in __traceback__.tb_frame
### Bug description

the code works fine, but pylint doesn't like it:

```python
try:
    foo()
except Exception as exn:
    last_tb = exn.__traceback__
    print(last_tb.tb_frame.f_locals)
```

### Command used

```shell
pylint --rcfile=/dev/null --disable=all --enable=no-member tbf.py
```

### Pylint output

```python
************* Module tbf
tbf.py:5:10: E1101: Class 'tb_frame' has no 'f_locals' member (no-member)
```

### Expected behavior

no diagnostics

### Pylint version

```shell
pylint 4.0.6
astroid 4.1.2
Python 3.14.4
```

### OS / Environment

ubuntu

kdelay:
## Type of Changes

## Description

`raw_building` modelled a data descriptor of a C-implemented class as a class named
after the attribute (`object_build_datadescriptor` built a `ClassDef` through
`_base_class_object_build`). Reading such an attribute therefore inferred that class
instead of a value, and the attributes of the value the descriptor returns were
reported as missing:

```console
$ cat tbf.py
def foo():
    pass


try:
    foo()
except Exception as exn:
    last_tb = exn.__traceback__
    print(last_tb.tb_frame.f_locals)
$ pylint --rcfile=/dev/null --disable=all --enable=no-member tbf.py
tbf.py:9:10: E1101: Class 'tb_frame' has no 'f_locals' member (no-member)
```

The generator model has the same surface, since `gi_frame` and `gi_code` are data
descriptors too:

```console
$ pylint --rcfile=/dev/null --disable=all --enable=no-member gen.py
gen.py:6:6: E1101: Class 'gi_frame' has no 'f_locals' member (no-member)
```

What a data descriptor returns is not statically known, so this builds an unknown
value instead: the same `EmptyNode` that the fallback branch of
`InspectBuilder.object_build` already uses for members astroid cannot model. The
attribute itself stays resolvable (`getattr` finds it, a genuinely missing attribute
still raises `AttributeInferenceError`); only the value is now `Uninferable`. Both
snippets above are clean after the change.

Two existing tests pinned the old modelling and are updated to the new one:
`test_cdata_member_access` (it checks that `_CData` members stay reachable through the
ctypes brain, which is still asserted, `x._objects` resolving to `Uninferable` rather
than raising) and `GeneratorModelTest::test_model`.

Exception instance attributes such as `OSError().errno` are unaffected: they come from
the instantiation trick in `_base_class_object_build`, which only applies to classes.

Verification run locally on macOS with CPython 3.13: `pytest` gives 2013 passed with
the same 3 pre-existing environment failures (`test_symlink_resolution`,
`test_identify_old_namespace_package_protocol`, `test_module_is_not_namespace`) that
also fail on unmodified `main`; each of the four touched tests fails with only the
`raw_building` hunk reverted; pylint's own suite against this branch is 2122 passed,
identical to the same run against unmodified `main`; `black`, `ruff`, `mypy` and
`pylint --rcfile=pylintrc` are clean.

Closes pylint-dev/pylint#11218

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -261,6 +261,15 @@ Refs #3108
 
   Refs #3115
 
+* Infer the value of a data descriptor of a C-implemented class as unknown
+  instead of as a class named after the attribute. Reading such an attribute,
+  for instance ``exc.__traceback__.tb_frame`` or ``gen.gi_frame``, inferred a
+  class, so the attributes of the value the descriptor returns were reported as
+  missing. ``raw_building.object_build_datadescriptor()`` now returns an
+  ``EmptyNode``.
+
+  Closes pylint-dev/pylint#11218
+
 What's New in astroid 4.1.2?
 ============================
 Release date: 2026-03-22
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -335,9 +335,15 @@ def object_build_function(
 
 def object_build_datadescriptor(
     node: nodes.Module | nodes.ClassDef, member: type
-) -> nodes.ClassDef:
-    """create astroid for a living data descriptor object"""
-    return _base_class_object_build(node, member, [])
+) -> nodes.EmptyNode:
+    """create astroid for a living data descriptor object
+
+    What the descriptor returns is not statically known, so it is modelled as an
+    unknown value. Modelling it as a class named after the attribute made
+    attribute access on the descriptor's value infer that class instead, which
+    reported the value's own attributes as missing.
+    """
+    return build_dummy(_EMPTY_OBJECT_MARKER)
 
 
 def object_build_methoddescriptor(
@@ -511,7 +517,7 @@ class InspectBuilder:
             elif inspect.ismethoddescriptor(member):
                 child: nodes.NodeNG = object_build_methoddescriptor(node, member)
             elif inspect.isdatadescriptor(member):
-                child = object_build_datadescriptor(node, member)
+                child: nodes.NodeNG = object_build_datadescriptor(node, member)
             elif isinstance(member, tuple(node_classes.CONST_CLS)):
                 # Special case: __hash__ = None overrides ObjectModel for unhashable types.
                 # See https://docs.python.org/3/reference/datamodel.html#object.__hash__
--- a/tests/brain/test_ctypes.py
+++ b/tests/brain/test_ctypes.py
@@ -6,7 +6,7 @@ import sys
 
 import pytest
 
-from astroid import extract_node, nodes
+from astroid import extract_node, nodes, util
 
 pytestmark = pytest.mark.skipif(
     hasattr(sys, "pypy_version_info"),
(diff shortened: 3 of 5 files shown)


DanielNoord (commented):
Not sure this is the right approach? Making more stuff Uninferable where we did infer before is generally not a good sign.

kdelay:
That's a fair default, so let me show exactly what the previous inference was.

`object_build_datadescriptor` delegated to `_base_class_object_build`, whose instantiation trick is guarded by `issubclass(member, Exception)`. A data descriptor is not a class, so `issubclass` raises `TypeError` and the `except TypeError: pass` branch is taken for every descriptor. What came out was an empty `ClassDef` named after the attribute. On `main` at 687729cb (astroid 4.2.0b5):

```python

node = extract_node("""
try:
    pass
except Exception as e:
    e.__traceback__.tb_frame  #@
""")
inferred = node.inferred()[0]
print(repr(inferred))                                              # <ClassDef.tb_frame l.0>
print(inferred.basenames, inferred.instance_attrs, inferred.doc_node)  # [] {} None
print(sorted(inferred.locals))  # ['__annotations__', '__module__', '__qualname__']
```

No bases, no members, no docstring, and the name is the attribute name rather than any type the descriptor could return. It does not describe the value; the only thing it asserts about the value is that it has no attributes, which is how pylint-dev/pylint#11218 ends up as `Class 'tb_frame' has no 'f_locals' member`.

The two existing tests I had to touch say the same thing. `GeneratorModelTest.test_model` asserted `gi_code`/`gi_frame` were `ClassDef`s named `gi_code`/`gi_frame`, and `test_cdata_member_access` asserted the qname `_ctypes._SimpleCData._objects`. Both pinned the placeholder itself, not any property of the described value.

Inference that comes from an object model or a brain is untouched: `e.args` still infers to `tuple` on this branch, since that comes from the exception object model rather than from `raw_building`. Attribute resolution is unchanged too, so `tb_frame` is still found via `getattr` and an attribute that genuinely does not exist still raises `AttributeInferenceError`. Only the value is now unknown.

I picked `EmptyNode` because it is what `InspectBuilder.object_build`'s final `else` already produces for members astroid cannot model, so this puts data descriptors into an existing bucket rather than inventing a new one.

If you would rather keep a real node there, the option I considered and did not implement is building the descriptor as an instance of its own type (`getset_descriptor` / `member_descriptor`), so class-level access stays faithful while instance access goes through a `__get__` that is still unknown. That is a bigger change and I have not measured its fallout, but I am glad to try it if you prefer that direction, or to drop this if you would rather the false positive be handled on the pylint side.

DanielNoord (commented):
Logic makes sense to me. @Pierre-Sassoulas What do you think?

Pierre-Sassoulas (approved):
LGTM !

PR #3202, opened 2026-08-05
Fix incorrect type inference for enum.auto() in IntEnum members

### LINKED ISSUE #1847: Using auto enum values provides incorrect type
```python3
import enum


class EnumWithAuto(enum.IntEnum):
    A = enum.auto()
    B = 10


auto_enum = EnumWithAuto.A
print(auto_enum.value.bit_length())  # E1101: Instance of 'auto' has no 'bit_length' member (no-member)

enum_val = EnumWithAuto(123)
print(enum_val.value.bit_length())  # Fine

specific_enum = EnumWithAuto.B
print(specific_enum.value.bit_length())  # Fine
```

All three of these `EnumWithAuto.value` should return an `int`, but only the `B` and `__init__(123)` instances do. `A` (since it uses `enum.auto()`) seems to return a type of `auto`, but that should be an `int`, since this uses an `IntEnum`.

Genny-oo:
When an `IntEnum` member is assigned via `enum.auto()`, `stmt.value` is a `nodes.Call` node rather than `nodes.Const`. The previous code fell into the `else` branch and called `.as_string()`, returning the string `"enum.auto()"`. This caused astroid to infer `.value` as type `auto` instead of `int`, producing false-positive E1101 errors.

**Fix:** Uses the existing `_looks_like()` helper to detect both `enum.auto()` and `auto()` (unqualified, after `from enum import auto`) spellings, substituting integer `1` so the stub infers the correct type (`int`) for `IntEnum` members.

**Tests:** Added two tests to `EnumBrainTest` in `tests/brain/test_enum.py` covering both the qualified and unqualified `auto()` spellings.

Closes #1847

## Type of Changes

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -429,6 +429,15 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                         inferred_return_value = repr(stmt.value.value)
                     else:
                         inferred_return_value = stmt.value.value
+                elif (
+                    isinstance(stmt.value, nodes.Call)
+                    and isinstance(stmt.value.func, nodes.Attribute)
+                    and stmt.value.func.attrname == "auto"
+                ):
+                    # enum.auto() is a Call node, not a Const.
+                    # Substituting 1 so the stub correctly infers int
+                    # for IntEnum members instead of the type `auto`.
+                    inferred_return_value = 1
                 else:
                     inferred_return_value = stmt.value.as_string()
 
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -429,14 +429,15 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
                         inferred_return_value = repr(stmt.value.value)
                     else:
                         inferred_return_value = stmt.value.value
-                elif (
-                    isinstance(stmt.value, nodes.Call)
-                    and isinstance(stmt.value.func, nodes.Attribute)
-                    and stmt.value.func.attrname == "auto"
+                elif isinstance(stmt.value, nodes.Call) and _looks_like(
+                    stmt.value, "auto"
                 ):
-                    # enum.auto() is a Call node, not a Const.
-                    # Substituting 1 so the stub correctly infers int
-                    # for IntEnum members instead of the type `auto`.
+                    # enum.auto() is a Call node, not a Const. Handles both
+                    # spellings: `enum.auto()` and `auto()` (after
+                    # `from enum import auto`). Substituting 1 so the stub
+                    # infers the correct type (int) for IntEnum members;
+                    # note the value itself may differ at runtime for later
+                    # members but only the type is consulted for E1101.
                     inferred_return_value = 1
                 else:
                     inferred_return_value = stmt.value.as_string()
--- a/tests/brain/test_brain.py
+++ b/tests/brain/test_brain.py
@@ -907,6 +907,42 @@ class BrainNamedtupleAnnAssignTest(unittest.TestCase):
         self.assertIsInstance(inferred, nodes.ClassDef)
 
 
+class BrainEnumAutoTest(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


kdelay (commented):
Checked this out locally at a6cc9d8 and ran it against main. The diagnosis matches what I see, and the two new tests do pin the change: with only the new `elif` block removed from `infer_enum_class`, both of them fail; `tests/brain/test_brain.py` is 142 passed / 2 skipped / 1 xfailed on the branch as it stands.

Three things came out of that.

**1. `from enum import auto` is not covered, and that spelling is the common one.**

The guard requires `stmt.value.func` to be an `Attribute`, so only the qualified call matches. Inferring `Color.RED.value`:

```
import enum / class Color(enum.IntEnum): RED = enum.auto()   ->  Const.int  value=1
from enum import IntEnum, auto / class Color(IntEnum): RED = auto()  ->  Instance of enum.auto
```

So the `no-member` report in #1847 still fires for the unqualified form.

`brain_namedtuple_enum.py` already has a helper for exactly this shape at line 183, used for `namedtuple`, `Enum` and `NamedTuple`, and it accepts `Name` as well as `Attribute`:

```python
elif isinstance(stmt.value, nodes.Call) and _looks_like(stmt.value, "auto"):
```

With that in place both spellings infer `Const(1)`. Full suite on the branch with the swap: 1941 passed, 83 skipped, 15 xfailed, 3 failed. Those three (`test_manager.py::test_identify_old_namespace_package_protocol`, `test_manager.py::test_module_is_not_namespace`, `test_get_relative_base_path.py::test_symlink_resolution`) fail identically on upstream main on this machine, so they are unrelated. It stays name-based, which is what the surrounding transforms already do.

**2. The new tests look like they belong in `tests/brain/test_enum.py`.**

`EnumBrainTest` is where the enum brain is covered, and it already builds `enum.auto()` members (`test_enum_with_ignore`, line 547). `test_brain.py` is the catch-all for brains without their own file.

**3. Every `auto()` member infers `1`, including the later ones.**

```
class Color(enum.IntEnum):
    RED = enum.auto()
    GREEN = enum.auto()

Color.GREEN.value  ->  Const.int value=1     # 2 at runtime
```

That is enough for the false positive this closes, since only the type is consulted there, but the stub now carries a value that is wrong rather than unknown, and anything that reads inferred values would follow it. Might be worth either saying so in the comment (it currently reads "so the stub correctly infers int", which is about the type only) or keeping a running counter for the auto members. Either way the type fix on its own closes #1847.

Genny-oo:
Thanks for the thorough review! I've addressed all three points:
 1. Switched to `_looks_like(stmt.value, "auto")` to cover both `enum.auto()` and `auto()` spellings
 2. Moved the tests into `EnumBrainTest` in `tests/brain/test_enum.py` and added a test for the unqualified form
 3. Updated the comment to clarify that `1` is used for type inference only and runtime values may differ for later members

astroid/brain/brain_namedtuple_enum.py
```diff
@@ -434,6 +429,16 @@ def infer_enum_class(node: nodes.ClassDef) -> nodes.ClassDef:
...
+                elif isinstance(stmt.value, nodes.Call) and _looks_like(
+                    stmt.value, "auto"
+                ):
+                    # enum.auto() is a Call node, not a Const. Handles both
+                    # spellings: `enum.auto()` and `auto()` (after
+                    # `from enum import auto`). Substituting 1 so the stub
+                    # infers the correct type (int) for IntEnum members;
+                    # note the value itself may differ at runtime for later
+                    # members but only the type is consulted for E1101.
+                    inferred_return_value = 1
```
Pierre-Sassoulas, line 441: StrEnum and ``custom _generate_next_value_`` produce non-int values at runtime ``class Color(StrEnum): RED = auto()`` now infers .value as int where runtime is "red".  (pre-existing I guess)

Pierre-Sassoulas (changes requested):
Hey, thank you for contributing to astroid. Could you add a changelog please ?

DanielNoord (commented):
Like @Pierre-Sassoulas said, let's also add a test for `StrEnum` to ensure we handle both correctly. I strongly beieve the current code doesn't 😄

PR #3203, opened 2026-08-05
Do not crash when an error message contains format placeholders

### LINKED ISSUE #3199: IndexError: Replacement index 0 out of range for positional args tuple
### Steps to reproduce

The following code triggers an `IndexError` in astroid:

```python
a = namedtuple('{0}', '')
```

### Current behavior
```python-traceback
  File "pylint/pylint/checkers/utils.py", line 1376, in safe_infer
    value = next(infer_gen)
  File "astroid/astroid/nodes/node_ng.py", line 140, in infer
    for result in self._explicit_inference(
                  ~~~~~~~~~~~~~~~~~~~~~~~~^
        self,  # type: ignore[arg-type]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        context,
        ^^^^^^^^
    ):
    ^
  File "astroid/astroid/inference_tip.py", line 67, in inner
    raise e from None
  File "astroid/astroid/inference_tip.py", line 64, in inner
    result = _cache[func, node, context] = list(func(node, context))
                                                ~~~~^^^^^^^^^^^^^^^
  File "astroid/astroid/brain/brain_namedtuple_enum.py", line 229, in infer_named_tuple
    raise UseInferenceDefault("ValueError: " + str(exc)) from exc
                                               ~~~^^^^^
  File "astroid/astroid/exceptions.py", line 67, in __str__
    return self.message.format(**vars(self))
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
IndexError: Replacement index 0 out of range for positional args tuple

    callback(astroid)
    ~~~~~~~~^^^^^^^^^
  File "pylint/pylint/checkers/base/name_checker/checker.py", line 451, in visit_assignname
  File "pylint/pylint/checkers/utils.py", line 1380, in safe_infer
    raise AstroidError from e
```

### Expected behavior
No crash

### Version info
pylint-dev/astroid@86664181
pylint-dev/pylint@a219ca322
Python 3.14.6

The crash also affects Pylint 4.0.6

Sanjays2402:
## Type of Changes

## Description

`AstroidError.__str__` runs `str.format` on its message, and several inference errors build that message out of user code. `namedtuple("{0}", "")` therefore reached `"ValueError: " + str(exc)` with an unsatisfiable positional placeholder and raised `IndexError` instead of the intended `UseInferenceDefault`, crashing pylint.

The fallback already caught `ValueError`; it now also catches `IndexError` and `KeyError`, the other two exceptions `str.format` raises for placeholders it cannot fill. A regression test was added beside the sibling `test_*_does_not_crash_inference` cases in `tests/brain/test_named_tuple.py` (fails without the fix with the reported `IndexError`, passes with it).

Happy to add a ChangeLog entry if you would like one for this.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

Closes #3199

### DIFF
--- a/astroid/exceptions.py
+++ b/astroid/exceptions.py
@@ -63,10 +63,14 @@ class AstroidError(Exception):
             setattr(self, key, value)
 
     def __str__(self) -> str:
+        # message is the template itself and may embed interpolated user text
+        # (e.g. a namedtuple typename), so keep it out of the fields and fall
+        # back to it verbatim rather than raising or reinterpreting it.
+        fields = {k: v for k, v in vars(self).items() if k != "message"}
         try:
-            return self.message.format(**vars(self))
-        except ValueError:
-            return self.message  # Return raw message if formatting fails
+            return self.message.format(**fields)
+        except (LookupError, ValueError, AttributeError, TypeError):
+            return self.message
 
 
 class AstroidBuildingError(AstroidError):
--- a/doc/whatsnew/fragments/3199.bugfix
+++ b/doc/whatsnew/fragments/3199.bugfix
@@ -1,0 +1,7 @@
+Fix a crash when a ``namedtuple`` or ``Enum`` type name or field name
+contains ``str.format`` markup, as in ``namedtuple("{0}", "abc")``. The name
+is interpolated into an error message that astroid then reformats, so
+building the message raised ``IndexError``. Inference now falls back to its
+default.
+
+Closes #3199
--- a/tests/brain/test_named_tuple.py
+++ b/tests/brain/test_named_tuple.py
@@ -244,6 +244,39 @@ class NamedTupleTest(unittest.TestCase):
         inferred = next(node.infer())
         self.assertIs(util.Uninferable, inferred)  # would raise ValueError
 
+    def test_format_placeholder_typename_does_not_crash_inference(self) -> None:
+        """A typename containing str.format placeholders must not be formatted.
+
+        Reported in https://github.com/pylint-dev/astroid/issues/3199.
+        """
+        node = builder.extract_node("""
+        from collections import namedtuple
+        Tuple = namedtuple("{0}", "")
+        Tuple #@
+        """)
+        inferred = next(node.infer())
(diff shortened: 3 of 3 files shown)


DanielNoord (commented):
Can we cover the `KeyError` with a test as well?

Pierre-Sassoulas:
Two payloads still get through:
 - namedtuple("{message.foo}", "") => AttributeError: 'str' object has no attribute 'foo'
- namedtuple("{message:200000000}", "") => no exception at all; format succeeds and builds a 200 MB string. Peak RSS goes from 22 MB to 421 MB for a single call, so no except clause can catch this one.

#3214 covers both by also dropping message from the substitution fields, which stops the message from being able to interpolate itself (Please do it that way and credit @kali834x for it). 

Also adding the changeLog entry would be welcome, plus the KeyError test Daniel asked for.

Sanjays2402:
Implemented the requested hardening in 889342c: AstroidError now excludes the message itself from format fields and falls back on LookupError/ValueError/AttributeError/TypeError, which prevents both attribute traversal and large-width self-formatting. Added the KeyError and AttributeError inference regressions plus a ChangeLog entry. The full tests/brain/test_named_tuple.py passes (26 passed); with the source fix reverted, the AttributeError regression fails during AstroidError.__str__, and it passes again with the fix.

DanielNoord:
Please fix CI and rebase.

Sanjays2402:
Thanks @DanielNoord. Fixed CI — the new `test_attribute_format_placeholder_typename_does_not_crash_inference` exceeded black's line length; reformatted. Rebased the branch onto current main (now 3 ahead, 0 behind) preserving the commit history.

Sanjays2402:
Rebased onto current main. The branch had accidentally picked up unrelated tree changes (it predated the towncrier migration), so I reset it to contain only this fix: the `AstroidError.__str__` hardening, the three regression tests, and the news entry as `doc/whatsnew/fragments/3199.bugfix` since the ChangeLog is now generated from fragments. The diff is back to 3 files and the new tests pass.

DanielNoord (approved):

PR #3204, opened 2026-08-06
Fix crash on a namedtuple field name that changes under NFKC normalization

Eljees:
Refs pylint-dev/pylint#8746.

Python normalizes identifiers to NFKC, so a `namedtuple` field written as `µ` (U+00B5 MICRO SIGN) is stored by the parser as `μ` (U+03BC GREEK SMALL LETTER MU). The namedtuple brain builds a fake class from generated source and then copies each attribute across by the name as written, which misses:

```python
from collections import namedtuple

names = ["µ"]
mu = namedtuple("mu", names)
```

```
KeyError: 'µ'
  astroid/brain/brain_namedtuple_enum.py:260 in infer_named_tuple
```

The definition is valid — `namedtuple` accepts any identifier, and at runtime `mu._fields` is `('µ',)` with the attribute reachable — so pylint reports `astroid-error` on working code. Reproduced on current `main` (8666418) and on released astroid 4.0.4 / pylint 4.0.6.

The attribute is now copied under the name the parser used. Storing it normalized also keeps it findable: an attribute access written as `instance.µ` in user code is normalized the same way, so a lookup would otherwise miss the entry.

Verified: the new regression test fails with the same `KeyError` on the unpatched tree and passes with the change; `tests/test_regrtest.py` is green (35 passed). `tests/brain/` has one failure, `TypingBrain::test_typed_dict_required_and_optional_keys`, which is present on a clean tree as well. `ruff check` and `black` both clean.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,14 @@ What's New in astroid 4.2.0?
 ============================
 Release date: TBA
 
+* Fix a crash on a functional ``namedtuple`` whose field name changes under NFKC
+  normalization, such as ``namedtuple("mu", ["µ"])``. Python normalizes
+  identifiers, so the parsed class stores the field under ``"μ"`` while the
+  brain looked it up as written, raising ``KeyError`` on a definition that
+  ``namedtuple`` itself accepts.
+
+  Closes pylint-dev/pylint#8746
+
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
LGTM, thank you

PR #3205, opened 2026-08-06
Infer None for the fset of a property without a setter

### LINKED ISSUE #8739: Unfound astroid.exceptions.InferenceError causes a crash
The unfound attribute fset causes a fatal error. See the following example:

pylint_example.py
```
class C:
    @property
    def p(self) -> int:
        return 1

C.p.fset
```
#### The expected output:
Works normally.

#### The actual output:
```
>> pylint pylint_example.py
************* Module pylint_example
Desktop/pylint_example.py:19:0: C0303: Trailing whitespace (trailing-whitespace)
Desktop/pylint_example.py:50:0: C0304: Final newline missing (missing-final-newline)
Desktop/pylint_example.py:1:0: C0114: Missing module docstring (missing-module-docstring)
Desktop/pylint_example.py:43:0: C0115: Missing class docstring (missing-class-docstring)
Desktop/pylint_example.py:43:0: C0103: Class name "C" doesn't conform to PascalCase naming style (invalid-name)
Desktop/pylint_example.py:45:4: C0116: Missing function or method docstring (missing-function-docstring)
Desktop/pylint_example.py:45:4: C0103: Attribute name "p" doesn't conform to snake_case naming style (invalid-name)
Desktop/pylint_example.py:43:0: R0903: Too few public methods (1/2) (too-few-public-methods)
Desktop/pylint_example.py:49:0: W0104: Statement seems to have no effect (pointless-statement)
Exception on node <Attribute.fset l.49 at 0x7f230d51dd10> in file '/home/xxm/pylint_example.py'
  File "/home/xxm/.local/lib/python3.11/site-packages/pylint/utils/ast_walker.py", line 91, in walk
    callback(astroid)
  File "/home/xxm/.local/lib/python3.11/site-packages/pylint/checkers/typecheck.py", line 1123, in visit_attribute
    attr_nodes = owner.getattr(node.attrname)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxm/.local/lib/python3.11/site-packages/astroid/nodes/scoped_nodes/scoped_nodes.py", line 1269, in getattr
    found_attrs.append(self.special_attributes.lookup(name))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxm/.local/lib/python3.11/site-packages/astroid/interpreter/objectmodel.py", line 132, in lookup
    return getattr(self, IMPL_PREFIX + name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxm/.local/lib/python3.11/site-packages/astroid/interpreter/objectmodel.py", line 985, in attr_fset
    raise InferenceError(
astroid.exceptions.InferenceError: Unable to find the setter of property p
Desktop/empirical_type_checker/bug_report/pylint_example.py:1:0: F0002: /home/xxm/Desktop/pylint_example.py: Fatal error while checking '/home/xxm/Desktop/pylint_example.py'. Please open an issue in our bug tracker so we address this. There is a pre-filled template that you can use in '/home/xxm/.cache/pylint/pylint-crash-2023-06-07-00-16-44.txt'. (astroid-error)
```

#### Test Environment:
Ubuntu 18.04,
CPython 3.11.3
Pylint: 3.0.0b1
astroid 3.0.0a3

Eljees:
Closes pylint-dev/pylint#8739.

## What

`property.fset` on a setter-less property infers as it should, but the call-form
`property(getter, setter)` still raises `InferenceError`. That exception is swallowed by
`raise_if_nothing_inferred`'s bare `except StopIteration`, which replaces it with a generic
"StopIteration raised without any error information" — the same crash pattern as #8739, on the one
branch #8739's fix didn't reach.

## Fix

- `attr_fset` on the call-form path now returns `Uninferable` instead of raising.
  `find_setter()` not finding a setter here doesn't mean there is no setter — it can't see the
  call-form argument at all — so `Uninferable` is the honest "cannot tell", and it skips
  `raise_if_nothing_inferred` entirely instead of being swallowed by it.
- The `None`-on-no-setter fallback is scoped to descriptors that actually implement
  `property.fset`, so decorator-based property-likes (e.g. `functools.cached_property`, which has
  no `fset` at runtime) aren't misclassified.
- `infer_property()`'s `<property>` origin marker is used to recognize call-form/decorator-built
  properties outside a class body (module-level or function-local `@property`), which previously
  still hit the ClassDef-only check and kept raising.

## Testing

Regression test parametrized over the three property positions (class body, module-level,
function-local); `property(getter, setter)` still returns `Uninferable`, updated from the old
raise-based assertion. Full suite: 2091 passed before and after (114 pre-existing failures are
missing optional deps — numpy/gi — identical on the unmodified base).

## Changelog

`doc/whatsnew/fragments/8739.bugfix` — reworded: the call-form path is a deliberate behavior
change (raises → `Uninferable`), not "left as it was".


Eljees:
Copilot's point is correct and it is a real regression in this patch, so thanks for asking for the review.

`infer_property` builds the `Property` from `node.args[0]` alone and parents it to `SYNTHETIC_ROOT`, so for `p = property(getter, setter)` the second argument is never modelled. `find_setter` then walks `func.parent.body` looking for a `@<name>.setter` decorator, finds an empty body, and returns `None` regardless of what was actually passed. Before this change that produced an `InferenceError`, which callers read as "cannot infer". After it, `p.fset` is confidently `None` — a wrong answer in place of no answer, which is worse than the crash I set out to fix.

`find_setter` is only authoritative for the decorator form, because that is the only form it can see. So the `Const(None)` should be limited to that case:

```python
func_setter = find_setter(func)
if not func_setter:
    if not isinstance(func.parent, nodes.ClassDef):
        # Built by a property(...) call: the setter, if any, was passed as an
        # argument that infer_property() does not model, so we cannot claim None.
        raise InferenceError(
            f"Unable to find the setter of property {func.function.name}"
        )
    # Decorator syntax with no .setter - fset really is None.
    return node_classes.Const(None)
```

`nodes` is already imported in `objectmodel.py`, so this needs no new import. That keeps the fix for pylint-dev/pylint#8739 and leaves the call form exactly as it behaves on main. I'll add a second test for `p = property(getter, setter); p.fset` so the distinction is pinned by a test rather than by a comment — the current suite has nothing covering that form, which is why the regression passed CI.

Before I push: do you want it scoped that way, or would you rather `infer_property` record the setter argument on the `Property` so that `fset` can actually resolve it? The second is the better answer, but it is a wider change than this bug needs and I did not want to make that call unilaterally.

DanielNoord:
@Eljees Are you using AI to make this contribution? If so please use your AI to guide yourself to the best and most correct suggestion while keeping the diff reviewable. If it helps split it up over several PRs.

As maintainers we are not particularly interested in engaging in lengthy discussions with AIs about implementation details, they can do this themselves. If I misjudged your use of AI I'd like to apologize in advance, we're balancing engaging with contributors and not spending too much time on AI.

Eljees:
Yes, I use an LLM to draft text. The analysis, the runs and the decisions behind the patch are mine. Point taken on length — I'll keep the write-ups short.

On the substance: the review comment was right and the patch as it stands is wrong. `infer_property` builds the `Property` from `node.args[0]` alone, so for `p = property(getter, setter)` `find_setter` cannot see the setter and returns `None`. My change turns that into a confident `fset = None` — a wrong answer where there used to be an `InferenceError`, which is worse than the crash I set out to fix. `find_setter` is only authoritative for the decorator form, so `Const(None)` has to be limited to that case.

The diff is three files and ~15 lines, so there is nothing to split up. I'll push the corrected version; if you would rather not carry this at all, say so and I'll close it.

Eljees:
Pushed in ba5daf6. `Const(None)` is now limited to the decorator form, and `property(getter, setter)` raises `InferenceError` again, as it does on main. Added a test for the call form, which had no coverage before — it is what let the regression through CI.

DanielNoord (commented):
Copilot's review seems sensible

Eljees:
Merged `main` in to clear the conflict. It moved to towncrier while this was open, so the ChangeLog entry is now `doc/whatsnew/fragments/8739.bugfix` — `check_newsfragments.py` wants the file named after the issue it closes.

Eljees:
The CodSpeed failure is not from this change.

The only benchmark that ran is `test_bench_endtoend_cold_lint`, which times `python -m pylint` in a subprocess over the minimal target:

```python
def add(a, b):
    return a + b


result = add(len("ab"), 2)
```

No class, so no property, so `PropertyModel.attr_fset` — the only function this PR touches — is never entered. That benchmark measures interpreter and import startup. The other three, the ones that actually infer, were skipped and their baselines reused, and the report itself flags WallTime on standard hosted runners as inconsistent.

Happy to acknowledge it on CodSpeed if that is what you normally do here.

Eljees:
Fixed in 94eb6742 - Copilot was right. `find_setter()` is authoritative for any property built from a decorated function, not only one in a class body, so the `ClassDef` check left module-level and function-local `@property` still raising, which is the crash this branch is meant to fix. It now tests the `<property>` name `infer_property()` assigns, and the regression test is parametrized over the three positions; `property(getter, setter)` still raises. The synthesized `Const` is attached to the property now, matching `attr___defaults__`.

Eljees:
Copilot was right again, and it's the same failure mode as #8739 itself, not just a similar shape.

Instrumented `attr_fset` directly: the `InferenceError` on the `property(getter, setter)` branch does get raised with its specific message, but `raise_if_nothing_inferred`'s bare `except StopIteration` intercepts it before it reaches the caller and replaces it with the generic "StopIteration raised without any error information" -- the informative message never gets out. Same mechanism the whole PR exists to remove, just still live on this one branch.

Pushed in e99fc449: returns `Uninferable` instead of raising. Unlike the decorator form, `find_setter()` not finding a setter here doesn't mean there is no setter (it can't see the call-form argument at all), so `Uninferable` is the honest "cannot tell" -- and it skips `raise_if_nothing_inferred` entirely, so it can't be swallowed the same way. The `8739.bugfix` fragment already said this branch "stays uninferable"; the code didn't match that until now.

Updated `test_property_call_form_does_not_infer_fset_as_none` to assert the `Uninferable` result instead of expecting the raise. Full suite: 2091 passed before and after (114 pre-existing failures are missing optional deps -- numpy/gi -- identical on the unmodified base, confirmed by running both).

Eljees:
Both fair.

`cached_property`: the `None` fallback in `attr_fset` should only apply to descriptors that actually implement `property.fset`, not every property-like decorator. Will scope the check to that.

Changelog: agreed, "left as it was" is wrong for the call-form path — it now returns `Uninferable` instead of raising, which is a deliberate change. Will reword the fragment.

Pushing both together.

Pierre-Sassoulas (changes requested):
The comments are way way too verbose, give the details in this MR description and then we'll be able to retrieve them using git blame.

Eljees:
Fair. Moving the history out of the comment thread and into the description below — happy to trim further if anything's still redundant.

Eljees:
@Pierre-Sassoulas trimmed in `1c2a0dc` — the comments in the diff are one line each now and the reasoning is in the description above.

The same push settles the two Copilot points that were still open:

- `functools.cached_property` is modelled as a `Property` but has no `fset` at runtime, so the `None` fallback no longer applies to it. It is limited to decorators that resolve to the property family; everything else keeps raising, as on `main`.
- The changelog fragment claimed the `property(...)` call form was "left as it was". It is not — it used to raise and now infers as uninferable. Reworded.

`tests/test_inference.py -k "property_without_setter or property_call_form or cached_property"`: 4 failed, 1 passed against `main`'s `objectmodel.py`; 5 passed with the change. The one that passes either way is the new `cached_property` guard, which is the point of it. Full inference suite 464 passed; full suite 2 failed, and those two fail identically on `main`'s `objectmodel.py` in the same container (missing `pkg_resources` and a `typing_extensions` version, not this change).

PR #3211, opened 2026-08-08
Do not infer a shadowed builtin as the builtin

### LINKED ISSUE #10994: False positives when 'type' builtin is overwritten
### Bug description

```python
def convert_type(x, type): # pylint: disable=missing-function-docstring,redefined-builtin
    return type(x)

convert_type(12.34, str).split('.')
```

For the final line, `pylint` reports:
```
E1101: Class 'float' has no 'split' member (no-member)
```

If `type` is renamed in the function definition, then `pylint` reports no errors.

The exact false positive reported is not always the same. I believe the fundamental problem is that `pylint` does not know how to handle when `type` has been shadowed, and acts as though it has not been. Hence in the above example, it thinks that the function call returned the `float` class, since that is what it would be if the built-in `type` had actually been used.

I would like to mention that I am not an advocate for shadowing builtins like this. I came across this issue when I was working on existing code authored by someone else. But even though shadowing builtins is bad practice, this does seem like something `pylint` should be able to understand.

### Configuration

```ini

```

### Command used

```shell
pylint --errors-only a.py
```

### Pylint output

```
************* Module a
/tmp/a.py:4:0: E1101: Class 'float' has no 'split' member (no-member)
```

### Expected behavior

I would expect it to report no errors, like it does when you write the same code but without redefining `type`.

### Pylint version

```shell
pylint 4.0.5
astroid 4.0.4
Python 3.13.3 (main, May  9 2025, 23:49:05) [GCC 12.2.0]
```

### OS / Environment

I discovered this in the Debian Python 3:13 docker image. It seems to be Debian 12 (bookworm).

### Additional dependencies

```python

```

teddytennant:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`register_builtin_transform` picks its tip by name text alone. A parameter called `type` is treated as `builtins.type`:

```python
def convert_type(x, type):
    return type(x)

convert_type(12.34, str).split(".")
```

astroid turns that into `ClassDef.float`, and pylint reports `E1101: Class 'float' has no 'split' member` (pylint-dev/pylint#10994). Same hole for every transform registered this way (`len`, `str`, `list`, ...).

The fix is to resolve the name before running the transform, and fall back to normal inference when it is not the real builtin. Lookup goes through the name node (so later defs in the same scope do not count), with a small extra step for default args, which run in the enclosing scope. `from builtins import str` is treated as the builtin too.

## Review follow-up

Addressed the cases Pierre flagged where the first version was too strict:

- `from builtins import str` / `from builtins import *`
- use before a later `def` in the same module or class body
- default and kw-only default values (`def h(x, len=len([...]))`)

Also filled in the missing unshadowed controls for `getattr`, `hasattr`, `property`, and `super`.

On the cache note: the check stays inside the tip (same place as before) so the cheap name filter is unchanged. Restructuring so the tip could still be cached on a miss did not seem worth the churn for this fix; happy to revisit if it shows up in profiles.

## Verification

```
tests/brain/test_builtin.py  57 passed
full suite                   2108 passed, 66 skipped, 15 xfailed
```

Closes pylint-dev/pylint#10994

### DIFF
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -222,22 +222,65 @@ def _builtin_filter_predicate(node, builtin_name) -> bool:
     return False
 
 
-def _is_builtin_call(node: nodes.Call) -> bool:
-    """Check that the callable of *node* really resolves to a builtin.
+def _name_lookup_frame(
+    name_node: nodes.Name,
+) -> tuple[nodes.LocalsDictNodeNG, list[nodes.NodeNG]]:
+    """Resolve *name_node* the way Python would at that point.
 
-    ``_builtin_filter_predicate`` only matches on the identifier, so a name
-    shadowing a builtin (say a parameter called ``type``) selects the transform
-    too. Resolve the identifier the way Python would before trusting it, like
-    ``brain_type`` does for ``type[...]`` subscripts.
+    ``Name.lookup`` already ignores later assignments in the same scope, so
+    things like ``len(...)`` before a later ``def len`` work. Defaults are the
+    awkward bit: they run in the enclosing scope, and ``FunctionDef.scope_lookup``
+    only special-cases that when the looked-up node *is* the default, not a name
+    inside it. Detect that and look up from the parent frame instead.
+    """
+    scope = name_node.scope()
+    if isinstance(scope, (nodes.FunctionDef, nodes.Lambda)):
+        args = scope.args
+        if args is not None:
+            for default in itertools.chain(
+                args.defaults or (),
+                (d for d in (args.kw_defaults or ()) if d is not None),
+            ):
+                if default is name_node or default.parent_of(name_node):
+                    parent = scope.parent
+                    if parent is None:
+                        break
+                    # Same offset as FunctionDef.scope_lookup uses for defaults,
+                    # so ``def f(f=f)`` does not resolve the default to itself.
+                    return parent.frame()._scope_lookup(
+                        name_node, name_node.name, offset=-1
+                    )
+    return name_node.lookup(name_node.name)
+
+
+def _is_from_builtins_import(stmt: nodes.NodeNG, name: str) -> bool:
+    """True if *stmt* is ``from builtins import ...`` binding *name*."""
+    if not isinstance(stmt, nodes.ImportFrom) or stmt.modname != "builtins":
+        return False
+    for imported, alias in stmt.names:
+        if imported == "*":
+            return True
+        if (alias or imported) == name:
+            return True
+    return False
+
+
+def _is_builtin_call(node: nodes.Call) -> bool:
+    """True if the callable on *node* actually comes from builtins.
+
+    The filter only matches the name text, so a parameter called ``type`` still
+    picks the transform. Check where the name resolves before trusting it.
     """
     func = node.func
     if isinstance(func, nodes.Attribute):
-        # ``dict.fromkeys``: what matters is where ``dict`` comes from.
+        # dict.fromkeys: what matters is where ``dict`` comes from.
         func = func.expr
     if not isinstance(func, nodes.Name):
         return False
-    node_scope, _ = node.scope().lookup(func.name)
-    return isinstance(node_scope, nodes.Module) and node_scope.qname() == "builtins"
+    frame, stmts = _name_lookup_frame(func)
+    if isinstance(frame, nodes.Module) and frame.qname() == "builtins":
+        return True
+    return any(_is_from_builtins_import(stmt, func.name) for stmt in stmts)
 
 
 def register_builtin_transform(
--- a/astroid/brain/brain_builtin_inference.py
+++ b/astroid/brain/brain_builtin_inference.py
@@ -222,47 +222,13 @@ def _builtin_filter_predicate(node, builtin_name) -> bool:
     return False
 
 
-def _name_lookup_frame(
-    name_node: nodes.Name,
-) -> tuple[nodes.LocalsDictNodeNG, list[nodes.NodeNG]]:
-    """Resolve *name_node* the way Python would at that point.
-
-    ``Name.lookup`` already ignores later assignments in the same scope, so
-    things like ``len(...)`` before a later ``def len`` work. Defaults are the
-    awkward bit: they run in the enclosing scope, and ``FunctionDef.scope_lookup``
-    only special-cases that when the looked-up node *is* the default, not a name
-    inside it. Detect that and look up from the parent frame instead.
-    """
-    scope = name_node.scope()
-    if isinstance(scope, (nodes.FunctionDef, nodes.Lambda)):
-        args = scope.args
-        if args is not None:
-            for default in itertools.chain(
-                args.defaults or (),
-                (d for d in (args.kw_defaults or ()) if d is not None),
-            ):
-                if default is name_node or default.parent_of(name_node):
-                    parent = scope.parent
-                    if parent is None:
-                        break
-                    # Same offset as FunctionDef.scope_lookup uses for defaults,
-                    # so ``def f(f=f)`` does not resolve the default to itself.
-                    return parent.frame()._scope_lookup(
-                        name_node, name_node.name, offset=-1
-                    )
-    return name_node.lookup(name_node.name)
-
-
 def _is_from_builtins_import(stmt: nodes.NodeNG, name: str) -> bool:
     """True if *stmt* is ``from builtins import ...`` binding *name*."""
     if not isinstance(stmt, nodes.ImportFrom) or stmt.modname != "builtins":
         return False
-    for imported, alias in stmt.names:
-        if imported == "*":
-            return True
-        if (alias or imported) == name:
-            return True
-    return False
+    return any(
+        imported == "*" or (alias or imported) == name for imported, alias in stmt.names
+    )
 
 
 def _is_builtin_call(node: nodes.Call) -> bool:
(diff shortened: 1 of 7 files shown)


astroid/brain/brain_builtin_inference.py
```diff
@@ -234,6 +252,8 @@ def register_builtin_transform(
     def _transform_wrapper(
         node: nodes.Call, context: InferenceContext | None = None
     ) -> Iterator:
+        if not _is_builtin_call(node):
+            raise UseInferenceDefault
```
Pierre-Sassoulas, line 302: This prevents the use of  `_inference_tip_cached` later, not sure if it's worth resturcturing for.

tests/brain/test_builtin.py
```diff
@@ -156,3 +156,143 @@ def __str__(self): #@
...
+            pytest.param("int('42') #@", nodes.Const, id="int"),
+            pytest.param("isinstance(1, int) #@", nodes.Const, id="isinstance"),
+            pytest.param("issubclass(bool, int) #@", nodes.Const, id="issubclass"),
+            pytest.param("len([1, 2, 3]) #@", nodes.Const, id="len"),
+            pytest.param("list((1, 2)) #@", nodes.List, id="list"),
+            pytest.param("set([1]) #@", nodes.Set, id="set"),
+            pytest.param("slice(1, 2) #@", nodes.Slice, id="slice"),
+            pytest.param("str(42) #@", nodes.Const, id="str"),
+            pytest.param("tuple([1, 2]) #@", nodes.Tuple, id="tuple"),
+            pytest.param("type(1) #@", nodes.ClassDef, id="type"),
```
Pierre-Sassoulas, line 281:
```suggestion
            pytest.param("type(1) #@", nodes.ClassDef, id="type"),
            pytest.param('getattr("apple", "upper") #@', bases.BoundMethod, id="getattr"),
            pytest.param('hasattr("apple", "upper") #@', nodes.Const, id="hasattr"),
            pytest.param("property(lambda self: 1) #@", objects.Property, id="property"),
            pytest.param(
      """
      class Animal:
          def speak(self):
              return "generic"

      class Cat(Animal):
          def speak(self):
              super() #@
      """,
      objects.Super,
      id="super",
  ),
```

 `super`, `property`,  `getattr` and `hasattr` are not tested here and should

Pierre-Sassoulas (changes requested):
Thanks, this look promising ! 

 I ran the pylint primer on the branch and found no new pylint false positives. But there's a set of cases where `lookup()` resolves differently from Python, so the tip now bails where it used to infer. 

```python
from builtins import str
str(42)
```

`main` infers `Const('42')`, this branch infers a plain `Instance of builtins.str`.

```python
class Fruit:
    size = len("apple")   # Python: builtin, `len` is not bound in the class namespace yet
    def len(self):
        return 0
```

`main` infers `5`, this branch raises `InferenceError`. Same at module level:

```python
size = len([1, 2, 3])
def len(x):
    return "shadowed"
```

`main` infers `3`, this branch raises `InferenceError`.


```python
def h(x, len=len([1, 2, 3])):   # the default is evaluated in the *enclosing* scope
    return x
```

`main` infers `3`, this branch infers `Uninferable`.

astroid/brain/brain_builtin_inference.py
teddytennant: Yeah, fair point. I left the check in the tip so the cheap name filter stays as-is. Pulling it into the predicate (or otherwise making misses cacheable) felt like a bigger reshuffle than this bug needs. Easy to revisit if it shows up hot.

tests/brain/test_builtin.py
teddytennant: Added, plus the cases from the review body (`from builtins`, use-before-def, defaults).

teddytennant:
Thanks for the careful look. Pushed a follow-up that should cover the cases you hit:

- lookup from the name node (so later defs do not steal the builtin)
- `from builtins import ...` / star imports
- defaults and kw-only defaults evaluated in the enclosing scope
- the missing `getattr` / `hasattr` / `property` / `super` controls

Full suite is green on my side. Happy to tweak further if anything still looks off.

Pierre-Sassoulas:
You were right that Name.lookup already handles statement order.

I think there's still a a scoped_nodes.py bug. _name_lookup_frame duplicates Lambda/FunctionDef.scope_lookup's default branch and calls the private _scope_lookup. Those test ``node in self.args.defaults``, so a name inside a default never matches. If we fix this a comprehension or lambda in a default would be covered for free:

```python
  def h(x, len=[len(y) for y in ([1, 2],)]):
  def h(x, len=lambda: len([1, 2])):
```

But it's pre-existing, feel free to not touch. Either way, annotations are the same family and neither version covers them (def h(x: str("a") = 1, str=None)), decorators likewise.

teddytennant:
Went ahead and fixed it in scoped_nodes, you were right that it was the better place.

`Lambda.scope_lookup` and `FunctionDef.scope_lookup` now share a small helper that checks whether the node sits anywhere inside a default value, not just whether it is the default. Both of your cases work now:

```python
def h(x, len=[len(y) for y in ([1, 2, 3],)]):
def h(x, len=lambda: len([1, 2]))
```

Both used to come back Uninferable on my branch and now infer 3, same as main.

That also let me delete `_name_lookup_frame` from the brain, it just calls `lookup()` now. Two of the three lines codecov was complaining about went with it and I covered the third, so patch coverage should be clean.

I wrote the helper to walk up from the node rather than call `parent_of` on each default, since scope_lookup is hot and the walk stops at the function. Curious what codspeed says.

Added tests in test_lookup.py for the lookup change itself and in test_builtin.py for the two inference cases. Full suite green locally.

You are right about annotations, they are still not covered. `def h(x: str("a") = 1, str=None)` is a different branch in scope_lookup and it felt like scope creep for this PR. Happy to open a separate issue for it. Decorators are actually already fine, `Decorators.scope()` skips the function frame, so `@deco(str("a"))` on `def h(str=None)` resolves correctly on both main and this branch.

teddytennant:
Rebased on main. The only conflict was the ChangeLog, which is towncrier now, so the entry moved to `doc/whatsnew/fragments/3211.bugfix` and covers both the brain change and the `scope_lookup` one.

The red pylint job was `test_functional[wrong_import_order_py315]` complaining that astroid is classified first party instead of third party. It failed the same way on `module-in-path-subpath-check` two days ago and the runs since yesterday are green, so the rebase should sort it.

Local suite is 2088 passed, 2 failed, and both failures are my environment (no `pkg_resources`, and a typing_extensions mismatch). They fail identically on a clean `origin/main`.

doc/whatsnew/fragments/3211.bugfix
```diff
@@ -0,0 +1,20 @@
...
+``for``, ``with ... as``, ``global`` or ``nonlocal`` shadows the builtin too.
+Calls that used to infer a builtin result may now infer ``Uninferable`` or raise
+``InferenceError`` instead. In exchange, a rebound name resolves to what it is
+really bound to, so ``from collections import OrderedDict as dict`` followed by
+``dict(a=1)`` now infers an ``OrderedDict`` rather than a plain ``dict``.
+
+``Lambda.scope_lookup`` and ``FunctionDef.scope_lookup`` now send any name
+nested inside a default value to the enclosing scope, not just the default
+itself. A comprehension or a lambda in a default used to resolve names against
+the parameters it sits next to.
```
Pierre-Sassoulas:
```suggestion
they are registered for. They were selected on the identifier alone, so a
parameter named ``type`` or a function named ``len`` was still inferred as a
call to the builtin.
```

teddytennant:
Thanks for applying the fragment edit and adding the xfail tests directly, nothing left on my side. Happy to take the two xfail cases in this PR if you would rather not leave them as follow-ups.

Pierre-Sassoulas:
Sorry I made #3232 and intended to merge it back here but got drowned in notifications. Fixing the xfail wxould make sense if you don't mind.

teddytennant:
Rebased on main and fixed both xfails.

The import one was checking the alias instead of what was imported under it, so `from builtins import int as str` still looked like the real `str`. It asks `real_name` now.

The annotation one needed a small correction to the plan. Type parameters live in `func.locals`, so sending annotations to the enclosing scope broke `def f[T](x: T)`. Annotations see the type parameters and defaults do not, so the helper now reports which of the two a node is in rather than just yes or no. Return annotations go with the annotations.

The xfail for the import case asserted `Const(42)`, which was never going to happen since astroid infers `int(42)` as an instance. Changed the assertion to that.

Ran pylint's suite against this branch, same results as against main.

Pierre-Sassoulas (changes requested):
Thank you for fixing the two test case. I added two other test cases, The issue seems to be thatr `` brain_type.infer_type_sub`` lookup from scope, not name node:

```diff
-    node_scope, _ = node.scope().lookup("type")
+    node_scope, _ = node.lookup("type")
```
And for the other one we need to check is something is a bare annotation, there's probably a helper function for that.

teddytennant:
The CodSpeed red looks like noise. The gate passed on 2ecc442, and the only commit since is your test-only one, which can't cost 10% on cold lint. main picked up the CodSpeedHQ action bump in between, and the report itself says it compared across different runtime environments.

codecov is down to the `except AttributeInferenceError` in `_is_from_builtins_import`. I don't think `lookup()` can reach it, since it only hands back an import that binds the name, so I'd rather mark it than invent a test for it.

Still on the two xfails. The bare annotation one is a bigger family than the single case: `len = 5` followed by `len: int` has to keep shadowing, `from builtins import str` followed by `str: object` has to stay the builtin, and `def f(): len: int` makes the name local, so a function body should keep shadowing where a module or class body does not. I had a version that passed your test and quietly broke the first of those, so I want the whole set covered before I push.

teddytennant:
the two xfails are closed.

`infer_type_sub` looks `type` up from the name node, so a parameter does not hide the builtin in the annotation next to it.

bare annotations at module or class scope are not treated as a binding, so `len: int` still calls the builtin. a function-local `len: int` still shadows, and `len = 5` then `len: int` still shadows. tests for those are in `test_builtin.py`.

teddytennant:
Anything else you want changed here?

PR #3212, opened 2026-08-09
Avoid inferring functional Enums with non-string names

alloutflo:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

The functional `Enum` transform currently accepts a non-string class name and creates a `ClassDef` whose `name` is not a string. For example, `Enum(1, "")` produces an inferred class named `1`, even though CPython rejects the call with `TypeError`.

Fall back to default inference for this invalid form, matching the existing handling for invalid functional Enum member names. This also prevents downstream consumers such as Pylint from crashing when they perform string operations on class names.

Refs pylint-dev/pylint#11229
Refs pylint-dev/pylint#11238

## Verification

- `python -m pytest tests/brain/test_enum.py -q` — 35 passed
- `python -m pytest tests/brain -q` — 385 passed, 64 skipped, 3 xfailed
- All configured pre-commit hooks passed on the changed files
- Pylint 4.0.6 with this Astroid branch processes the reported reproducer without crashing

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,12 @@ What's New in astroid 4.4.0?
 ============================
 Release date: TBA
 
+* Do not infer a functional ``Enum`` whose class name is not a string. Such a
+  definition is invalid at runtime and could expose a non-string class name to
+  consumers.
+
+  Refs pylint-dev/pylint#11229
+
 
 
 What's New in astroid 4.3.1?
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -149,6 +149,10 @@ def infer_func_form(
         # ``Enum("e", (1,))``) means the definition is invalid, so fall back to
         # the default inference instead of crashing.
         raise UseInferenceDefault
+    if enum and not isinstance(name, str):
+        # Enum class names must be strings; inferring a class with any other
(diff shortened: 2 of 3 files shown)


DanielNoord (commented):
I don't think this should be `Uninferable`. We can just drop the arguments, right?

alloutflo:
Thanks. CPython raises `TypeError` for `Enum(1, "")`, so I want to make sure I follow the intended static-inference behavior. By “drop the arguments,” do you mean:

1. keep the functional-Enum transform but discard the invalid class-name value? That would return a synthetic `Instance` rather than `util.Uninferable`, but its existing fallback display name would be `"Uninferable"`; or
2. bypass the functional transform and infer a generic `enum.Enum` instance, whose display name would be `"Enum"`?

Both avoid exposing an integer `ClassDef.name`, but they produce different `no-member` output. I will update the patch to whichever behavior you intended.

DanielNoord (dismissed):

DanielNoord (approved):

DanielNoord:
@Pierre-Sassoulas Do we want to go for `4.3.1`? Or just target `4.4`? Considering I also messed up merging for https://github.com/pylint-dev/astroid/pull/3213 (sorry again about that...)

Pierre-Sassoulas:
We can backoort after the fact if we want :D

PR #3217, opened 2026-08-11
reject non-identifier name in TypeVar/NewType inference

nabhan06:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

The name argument to `typing.TypeVar()`/`NewType()` is inferred by splicing it
into a `class {name}(metaclass=Meta)` source string and re-parsing that with
`extract_node`. The only guard rejects f-string names, so a plain string
constant with punctuation slips through the identifier position. Analyzing
`TypeVar("T(EvilBase): #")` builds `class T(EvilBase): #(metaclass=Meta):` where
the trailing comment drops the intended metaclass marker, and inference then
hands back a `ClassDef` with attacker-chosen bases and body.

I added an `isidentifier()` check before formatting the template, so a
non-identifier name raises `InferenceError` the same way `TypeVar("My.Type")`
already does, rather than synthesizing an arbitrary class. This mirrors how the
namedtuple/enum brain validates field and type names. Regression test added.

### DIFF
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -125,15 +125,15 @@ def infer_typing_typevar_or_newtype(
         raise UseInferenceDefault
 
     typename = node.args[0].as_string().strip("'")
-    if not typename.isidentifier():
-        # The name is spliced verbatim into the ``class {0}`` template below,
-        # so a crafted string such as ``"T(Base): #"`` would inject bases and a
-        # body into the synthesized class. Only accept a plain identifier.
-        raise InferenceError(node=node, context=context_itton)
     try:
-        node = extract_node(TYPING_TYPE_TEMPLATE.format(typename))
+        # ``typing`` accepts any string as a name, so don't splice it into the
+        # template: a crafted value such as ``"T(Base): #"`` would break out of
+        # the identifier position and inject bases and a body into the class.
+        # Build the class with a fixed name and assign the real one afterwards.
+        node = extract_node(TYPING_TYPE_TEMPLATE.format("_TypeVar"))
     except AstroidSyntaxError as exc:
         raise InferenceError from exc
+    node.name = typename
     return node.infer(context=context_itton)
 
 
--- a/astroid/brain/brain_typing.py
+++ b/astroid/brain/brain_typing.py
@@ -125,6 +125,11 @@ def infer_typing_typevar_or_newtype(
         raise UseInferenceDefault
 
     typename = node.args[0].as_string().strip("'")
+    if not typename.isidentifier():
+        # The name is spliced verbatim into the ``class {0}`` template below,
+        # so a crafted string such as ``"T(Base): #"`` would inject bases and a
+        # body into the synthesized class. Only accept a plain identifier.
+        raise InferenceError(node=node, context=context_itton)
     try:
         node = extract_node(TYPING_TYPE_TEMPLATE.format(typename))
     except AstroidSyntaxError as exc:
--- a/doc/whatsnew/fragments/3217.bugfix
+++ b/doc/whatsnew/fragments/3217.bugfix
@@ -1,0 +1,6 @@
+Infer a functional ``TypeVar`` or ``NewType`` whose name is not an identifier
+without splicing that name into the synthesized class source. A crafted value
+such as ``TypeVar("T(Base): #")`` no longer breaks out of the identifier
+position to inject bases or a body; the name is assigned to the class verbatim.
+
+Refs #3217
--- a/tests/brain/test_typing.py
+++ b/tests/brain/test_typing.py
@@ -23,6 +23,16 @@ def test_infer_typevar() -> None:
         call_node.inferred()
 
 
+def test_infer_typevar_name_injection() -> None:
+    """A crafted TypeVar name must not be spliced into the synthesized class."""
+    call_node = builder.extract_node("""
+    from typing import TypeVar
+    TypeVar('T(EvilBase): #')
+    """)
+    with pytest.raises(InferenceError):
+        call_node.inferred()
+
+
 class TestTypingAlias:
     def test_infer_typing_alias(self) -> None:
         """
--- a/tests/brain/test_typing.py
+++ b/tests/brain/test_typing.py
@@ -5,7 +5,6 @@
 import pytest
 
 from astroid import bases, builder, nodes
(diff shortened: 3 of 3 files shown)


kdelay (commented):
I reproduced this locally before commenting: `upstream/main` 57c5821 vs PR head 3720669, CPython 3.13.11, editable install, probing `Call.inferred()` for a table of `TypeVar`/`NewType` names.

The premise holds. On main the crafted name lands in the class header:

| name argument | main 57c5821 | PR head 3720669 |
| --- | --- | --- |
| `TypeVar('T')` | `ClassDef 'T'`, bases `[]`, `declared_metaclass()` = `Meta` | same |
| `TypeVar('T(EvilBase): #')` | `ClassDef 'T'`, bases `['EvilBase']`, `declared_metaclass()` = `None` | `InferenceError` |
| `TypeVar('T(EvilBase, dict): #')` | bases `['EvilBase', 'dict']`, `ancestors()` = `dict, object` | `InferenceError` |
| `NewType('N(EvilBase): #', int)` | `ClassDef 'N'`, bases `['EvilBase']` | `InferenceError` |
| `TypeVar('My.Type')`, `TypeVar('1T')`, `TypeVar('')` | `InferenceError` | `InferenceError` |

Two corrections to the description, both in the direction of a smaller claim:

* The body is not reachable. `TypeVar('T(EvilBase): gotcha = 1 #')` raises `InferenceError` on main already, because the template's trailing indented `pass` no longer fits after a one-line class body. What the name controls is the base list (and therefore the MRO), not the body.
* The concrete breakage is the lost metaclass. Because `#` swallows `(metaclass=Meta)`, the synthesized class loses the `Meta.__getitem__` the template exists for, so on main `T = TypeVar('T'); T[int]` infers `Instance of .T` while `T = TypeVar('T(EvilBase): #'); T[int]` raises `InferenceError`. Worth deciding how the ChangeLog words this: the crafted name has to be in the source being analyzed, so the effect is wrong inference plus a lost subscript path, not execution. "attacker-shaped class" reads as something stronger than what I could measure.

On the exception type, `InferenceError` is the right choice and I'd push back on the obvious suggestion to make it `UseInferenceDefault` for symmetry with the `JoinedStr` branch above it. Because the new guard also catches the names that previously died in `extract_node`, swapping it makes `TypeVar('My.Type')` infer as `Instance of typing.TypeVar` and `tests/brain/test_typing.py` goes to 2 failed / 6 passed, including the pre-existing `test_infer_typevar`.

One collateral change worth a deliberate decision: surrounding whitespace. `TypeVar('T ')` and `TypeVar(' T')` infer as `ClassDef 'T'` on main (the parser tolerates the space) and raise `InferenceError` on the branch. `typing.TypeVar` accepts any string as a name at runtime, so these are legal if pointless. Those two are the only non-injection inputs whose inference changed in my probe, and a `.strip()` before `isidentifier()` keeps them working; rejecting them is also defensible, but right now it happens as a side effect rather than as a decision.

I also checked the neighbouring functional forms for the same splice, since the description mentions them, and they are already covered, so this looks like the last unvalidated one: `namedtuple('NT(EvilBase): #', ['a'])` is `Uninferable` (`_check_namedtuple_attributes` validates the typename with `isidentifier()` plus `iskeyword()`), `Enum('E(EvilBase): #', ['a'])` keeps the payload as a plain name string (`Instance of __astroid_synthetic.E(EvilBase): #`, no injected base), and `TypedDict('TD(EvilBase): #', {'a': int})` stays `Instance of typing.TypedDict`. Note the difference from `_check_namedtuple_attributes`: this guard has no keyword check, and it does not need one, since `TypeVar('class')` already raises `InferenceError` through the `AstroidSyntaxError` path on both revisions.

Test and suite verification:

* Revert-control: keeping the new test and restoring only `astroid/brain/brain_typing.py` to main gives `Failed: DID NOT RAISE InferenceError`, 1 failed / 7 passed. The test pins exactly this hunk.
* Full suite on the PR head: 2044 passed, 4 failed, 91 skipped, 15 xfailed. The same 4 fail on main 57c5821 in my environment (`TypingBrain::test_typed_dict_required_and_optional_keys`, `test_symlink_resolution`, `test_identify_old_namespace_package_protocol`, `test_module_is_not_namespace`), so they are environmental rather than caused by this change.

Two small things:

* The ChangeLog says TypeVar/NewType but the test only asserts the TypeVar half. `NewType('N(EvilBase): #', int)` reproduces on main (row 4 above), so one extra assertion covers what the entry claims.
* The entry sits under `4.3.1` rather than `4.4.0`, and it is currently the only entry in that section. That reads as "this rides the patch release", which would also mean a `backport` label so `.github/workflows/backport.yml` opens the `maintenance/4.3.x` PR. If the fix is meant for `4.4.0` instead, the entry belongs in that section. Maintainer call either way, just flagging that the section choice carries the backport implication.

DanielNoord:
@kdelay Please reduce the verbosity of your AI. I think there is more sentences in your comment than there are lines changed in the PR. It is a distraction and nobody will read it.

DanielNoord (commented):
This is valid Python code?
```
>>> TypeVar('T(EvilBase): #')
~T(EvilBase): #
```

Instead of raising we need to fix our parsing.

nabhan06:
Good point, you're right that it's a valid name at runtime. Reworked it so the name never goes through the parser: the class is built with a fixed placeholder and the real string is assigned to `.name` afterwards. So `TypeVar('T(EvilBase): #')` now infers to a plain ClassDef with that name and no injected bases, and the metaclass is kept so `T[int]` still resolves. `TypeVar('My.Type')` infers cleanly too now instead of raising, so I updated that test to match.

DanielNoord (commented):
Cool! Can you add a changelog entry please?

DanielNoord:
You'll want to rebase to use `towncrier` to make that changelog entry :)

nabhan06:
Rebased onto main and dropped the ChangeLog edit in favor of a `3217.bugfix` fragment under `doc/whatsnew/fragments/`. Fix and test are unchanged.

DanielNoord:
@Pierre-Sassoulas I don't think we backported `towncrier`. Do you want to?

DanielNoord (approved):
LGTM! @Pierre-Sassoulas Will need to decide if we'll backport this are just put this in the next minor release.

Pierre-Sassoulas:
Let's backport towncrier

PR #3219, opened 2026-08-11
Infer the classes re-exported from C accelerator modules

Pierre-Sassoulas:
## Type of Changes

## Description

Following work on https://github.com/pylint-dev/pylint/pull/11252

``_ast.AST.__module__`` is ``"ast"``, not ``"_ast"``, so ``AST`` was recorded inside the raw-built ``_ast`` as an import from ``ast``, while ``ast.py`` only gets the name through ``from _ast import *``. The two modules pointed at each other and every class defined that way was ``Uninferable``.

Do not emit the import node when the module being built is the private counterpart of the module the member claims, and build the member where it really lives instead. This generalizes the ``_io`` special case just above it, and also fixes ``decimal.Decimal``, ``ssl.SSLError``, ``sqlite3.Connection`` and ``weakref.ref``, among others.


astroid/raw_building.py
```diff
@@ -564,6 +564,15 @@ def imported_member(self, node, member, name: str) -> bool:
...
             return False
 
+        # C accelerator modules such as ``_ast`` or ``_decimal`` advertise the
+        # public module that re-exports them (``ast``, ``decimal``) in
+        # ``__module__``. Emitting ``from ast import AST`` inside ``_ast``
+        # creates a cycle with the ``from _ast import *`` in ``ast.py``, and
+        # everything defined that way ends up uninferable. Build the member
+        # here instead, this is where it really lives.
+        if self._module.__name__ == f"_{modname}":
+            return False
```
jacobtylerwalls: Cool! Does this mean we can now unwind some of the brains we added for `datetime` and the rest?

Pierre-Sassoulas: Yeah, it definitely is going to be a very good overall. Only problem is astroid takes a little longer because we start to parse a ton of new thing ! Very hopeful about this.

PR #3220, opened 2026-08-12
Infer recent dict and list subscript assignments

AG0708:
## Type of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Description

Track the latest safe constant-key write to a name-rooted built-in dict or list within the same statement block. Subscript inference now uses that value for reads, writes, deletes, aliases, and deeper chains instead of returning the stale value from the original container literal.

The scan stops at dynamic keys, custom containers or properties, class namespaces, control-flow boundaries, and expressions that can call user code. Ambiguous assignments retain every inferred possibility rather than choosing one.

This covers the additional boundary cases added to pylint-dev/pylint#11255.

Refs pylint-dev/pylint#10050.

Implementation and tests were developed with OpenAI Codex assistance; I reviewed the inference behavior and verification results.

## Verification

- tests/test_inference.py: 483 passed, 2 skipped, 10 xfailed
- focused new inference cases: 19 passed
- changed-file pre-commit: Ruff, Black, Pylint, mypy, and repository hooks passed
- downstream Pylint functional case: all six newly documented false positives are removed
- full Astroid suite: 2,126 passed, 25 skipped, 15 xfailed; the remaining 9 NumPy/Pydantic/macOS failures reproduce unchanged on the parent commit

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -3864,10 +3864,7 @@ class Subscript(NodeNG):
             if not self._side_effect_free_assignment_value(current.value):
                 return None
 
-        try:
-            previous = current.previous_sibling()
-        except ParentMissingError:
-            return None
+        previous = current.previous_sibling()
         while previous is not None:
             if isinstance(previous, _base_nodes.MultiLineBlockNode):
                 return None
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (changes requested):
Could you explain why you took this approach and what approach you considered ? Human answer only please.

PR #3209, opened 2026-08-07
Use towncrier to assemble the changelog from news fragments

Pierre-Sassoulas:
Ports pylint's towncrier setup to astroid, as introduced in pylint-dev/pylint#6974 / pylint-dev/pylint#7198, with the release process staying minimally different from pylint's:

- `towncrier.toml` with astroid-appropriate fragment types: `breaking`, `user_action`, `feature`, `bugfix`, `other`, `internal`, `performance` (pylint's check-specific types dropped).
- News fragments live in `doc/whatsnew/fragments/` as `<issue>.<type>` files; `script/check_newsfragments.py` validates them from pre-commit.
- The what's new documentation moves to a pylint-style tree. `doc/whatsnew/4/4.3/index.rst` holds the released 4.3.0 and 4.3.1 notes, moved out of the flat `ChangeLog`, and `doc/whatsnew/4/4.4/index.rst` is what `towncrier build` writes to. The `ChangeLog` keeps the history up to 4.1.2 and stays included in the docs.
- The two entries that were still pending in the `ChangeLog` 4.4.0 section become fragments: `3212.bugfix` and `3223.bugfix`.
- `script/bump_changelog.py` is replaced by pylint's version: tbump keeps calling the same command, which now updates `towncrier.toml`, creates the newsfile for a new minor and runs `towncrier build`. The old script's test is removed with it.
- New `changelog` CI check: `towncrier check` on PRs, skipped for the `Skip news :mute:` (label created) and `dependency` (dependabot) labels; pre-commit.ci PRs get the label automatically.
- `doc/release.md` updated accordingly.

The Sphinx documentation builds with `-W` on this branch, and `towncrier build --draft`, the fragment checker and `bump_changelog.py --dry-run` were all exercised locally.

Follow-up: `maintenance/4.3.x` needs the same tooling cherry-picked so the 4.3.2 patch notes are built from fragments into `doc/whatsnew/4/4.3/index.rst`; until that lands, the branch keeps its flat `ChangeLog`.

PR #3222, opened 2026-08-15
Keep the annotations of *args and **kwargs in as_string output

kdelay:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`Arguments.format_args` rendered `*args` and `**kwargs` from their names alone, so annotations were dropped:

```python
>>> extract_node("def f(a: int, *args: str, **kw: float): pass").as_string()
"def f(a, *args, **kw):\n    pass"
```

They are on the node all along (`varargannotation`, `kwargannotation`), and positional and keyword-only parameters already render theirs. A PEP 646 `*args: *Ts` is covered too, since `Starred.as_string()` emits the star.

Two tests added. `pytest tests/` passes apart from four failures that also fail on an unmodified checkout here.

### DIFF
--- a/astroid/nodes/node_classes.py
+++ b/astroid/nodes/node_classes.py
@@ -841,7 +841,7 @@ class Arguments(
                 )
             )
         if self.vararg:
-            result.append(f"*{self.vararg}")
+            result.append(_format_starred_arg("*", self.vararg, self.varargannotation))
         if self.kwonlyargs:
             if not self.vararg:
                 result.append("*")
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (dismissed):
I think it make sense. I classify this an enhancement though, we can't modify the output of as_string in a patch. Let's wait for another maintainer review.

DanielNoord (commented):
Agreed with this being a breaking change! This will need a new style of changelog due to merging of https://github.com/pylint-dev/astroid/pull/3209

kdelay:
Rebased on main and moved the ChangeLog entry into a towncrier news fragment, `doc/whatsnew/fragments/3222.breaking`, following #3209. Filed under breaking changes since the `as_string` output changes for annotated `*args`/`**kwargs`.

The code and the two tests are unchanged from the reviewed commit.

PR #3223, opened 2026-08-15
use separator-aware containment check in module_in_path

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

module_in_path decides whether a resolved module file lives under a given path with filename.startswith on the normalized path, which has no separator boundary, so a module in a sibling directory that only shares a leading name (a file under proj_extra checked against the path proj) is reported as inside proj. the module already ships _is_subpath, the separator-aware containment check _get_relative_base_path uses; both the str and iterable branches of module_in_path now route through it so a prefix-only match no longer counts as containment. regression added in tests/test_modutils.py.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,11 @@ What's New in astroid 4.4.0?
 ============================
 Release date: TBA
 
+* ``modutils.module_in_path`` no longer reports a module as living under a
+  directory when its path only shares a leading name with that directory, such
+  as a module in ``proj_extra`` matching the path ``proj``. It now uses the same
+  separator-aware containment check as ``_get_relative_base_path``.
+
 * Fix inference of the attributes of a ``namedtuple`` created with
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
Look like it make astroid become first party in a pylint wrong-import-order tests.
```diff
-OutputLine(symbol='wrong-import-order', lineno=4, column=0, end_lineno=4, end_column=14, object='', msg='standard import "os" should be placed before third party import "astroid"', confidence='UNDEFINED')
+OutputLine(symbol='wrong-import-order', lineno=4, column=0, end_lineno=4, end_column=14, object='', msg='standard import "os" should be placed before first party import "astroid" ', confidence='UNDEFINED')
```

Pierre-Sassoulas:
Fixed in https://github.com/pylint-dev/pylint/pull/11284/changes

Pierre-Sassoulas (approved):

PR #3224, opened 2026-08-15
Infer pathlib.Path / str (and str / Path) as a Path

### LINKED ISSUE #11202: Incorrect/Missing inference for slash-joining `pathlib` paths -> Missing diagnostics
### Bug description

```python
"doc"
from pathlib import Path

Path('/tmp').read_text()				# unspecified-encoding
Path('/tmp').parents[0].read_text()		# unspecified-encoding

(Path('/') / 'tmp').read_text()			# No diagnostic
p: Path = Path('/') / 'tmp'				# Try with type annotation
p.read_text()							# No diagnostic
```
### Command used

```shell
`pylint --enable W1514 /tmp/test.py`
```

### Pylint output

```python
************* Module test
/tmp/test.py:4:0: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
/tmp/test.py:5:0: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```

### Expected behavior

PyLint should infer the type of `Path() / "str"` as `Path` and apply the same diagnostics as to direct instances of `Path`

Even using type-annotated variables doesn't help

### Pylint version

```shell
pylint 4.0.6
astroid 4.0.4
Python 3.12.1 (main, Feb 11 2026, 13:37:21) [GCC 13.3.0]
```

May be related to long fixed https://github.com/pylint-dev/pylint/issues/224

aryansk:
Closes pylint-dev/pylint#11202

## Problem

`pathlib` paths support the `/` operator — `Path('/tmp') / 'sub'` returns a new `Path`. astroid did not infer this `BinOp`, so the receiver of a method call on a slash-joined path resolved to `Uninferable`:

```python
(Path('/') / 'tmp').read_text()   # receiver inferred as Uninferable
p: Path = Path('/') / 'tmp'       # same even with a type annotation
p.read_text()
```

Because the receiver was `Uninferable`, pylint diagnostics that depend on the method's type — e.g. `unspecified-encoding` (W1514) on `read_text`/`write_text`/`open` — were silently skipped, while the equivalent direct `Path('/tmp').read_text()` was flagged.

## Change

Add a brain transform in `brain_pathlib.py` (mirroring the existing `Path.parents[i]` handling) that infers a `/` `BinOp` with a pathlib path operand as a new `pathlib.Path` instance:

- `Path('/') / 'tmp'` → `pathlib.Path`
- `'tmp' / Path('/')` → `pathlib.Path` (pathlib also supports the reflected operator)

Non-path `/` BinOps (e.g. `10 / 2`) are untouched, as is any other existing inference.

## Verification

- New tests in `tests/brain/test_pathlib.py` cover both operand orders, method resolution on a slash-joined path, and that plain division is unaffected.
- `tests/brain/` + `tests/test_inference.py`: **848 passed** (1 pre-existing env failure unrelated to this change).
- End-to-end with pylint: the issue's repro now flags all four `read_text()` calls with W1514 (previously only the two direct ones).
- pylint `tests/test_functional.py`: failure set is **identical** with and without this change (14 pre-existing Python 3.14 env failures), i.e. zero regressions.

---

*AI-assistance disclosure: this change was developed with the assistance of an AI coding agent (Codebuff).*

### DIFF
--- a/astroid/brain/brain_pathlib.py
+++ b/astroid/brain/brain_pathlib.py
@@ -47,9 +47,42 @@ def infer_parents_subscript(
     raise UseInferenceDefault
 
 
+def _looks_like_path_binop(node: nodes.BinOp) -> bool:
+    """Return True if a ``/`` BinOp has a pathlib path operand."""
+    if node.op != "/":
+        return False
+    for operand in (node.left, node.right):
+        try:
+            for value in operand.infer():
+                if isinstance(value, bases.UninferableBase):
+                    continue
+                if (
+                    isinstance(value, bases.Instance)
+                    and isinstance(value._proxied, nodes.ClassDef)
+                    and value._proxied.qname().startswith("pathlib.")
+                ):
+                    return True
+        except (InferenceError, StopIteration):
+            continue
+    return False
+
+
+def infer_path_binop(
+    binop_node: nodes.BinOp, ctx: context.InferenceContext | None = None
+) -> Iterator[bases.Instance]:
+    """Infer ``Path / X`` and ``X / Path`` as a new pathlib path."""
+    path_cls = next(_extract_single_node(PATH_TEMPLATE).infer())
+    return iter([path_cls.instantiate_class()])
+
+
 def register(manager: AstroidManager) -> None:
     manager.register_transform(
         nodes.Subscript,
         inference_tip(infer_parents_subscript),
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas:
Closing because I think the approach is wrong here, astroid can't says a str/Path is a Path to make pylint's live easier.

PR #3235, opened 2026-08-19
escape f-string literal parts so as_string round-trips

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`AsStringVisitor.visit_joinedstr` renders an f-string by joining the literal parts and the `{expression}` parts, then picks a surrounding quote from a fixed list by taking the first one that does not appear in the rendered body. When all four candidates (`'`, `"`, `'''`, `"""`) occur in the body the loop falls through and the last value `'''` is used anyway, so the emitted f-string terminates early. Adjacent f-strings such as `f'"""{v}' f"'''"` merge into a single `JoinedStr` whose parts together carry every quote sequence, which is enough to hit this from ordinary source. `as_string()` is meant to round-trip, and a couple of brains reparse its output: `brain_dataclasses` feeds it into a synthesized `__init__` and `brain_namedtuple_enum` into a mocked `Enum` member, so the broken string either raises `AstroidSyntaxError` during a plain `parse()` or reparses to a different expression and quietly rewrites the synthesized `__init__` signature. I ran into it checking that dataclass field defaults survive a round-trip. The fix leaves the existing fast path alone and, only when no candidate quote is free, re-escapes the literal parts against a quote that is absent from the `{expression}` parts, which cannot hold a backslash, so the output parses back to the same node. Added a regression test in `tests/test_nodes.py`.

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -323,13 +323,13 @@ class AsStringVisitor:
         formatted = "".join(
             value.accept(self)
             for value in node.values
-            if type(value).__name__ != "Const"
+            if not isinstance(value, nodes.Const)
         )
         quote = "'" if "'" not in formatted else '"'
         string = "".join(
             (
                 _escaped_fstring_literal(value.value, quote)
-                if type(value).__name__ == "Const"
+                if isinstance(value, nodes.Const)
                 else value.accept(self)
             )
             for value in node.values
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -317,18 +317,22 @@ class AsStringVisitor:
             if quote not in string:
                 return "f" + quote + string + quote
 
-        # Every candidate quote is present in the body. Re-escape the literal
-        # parts against a quote absent from the {expression} parts (which cannot
-        # hold a backslash) so the f-string cannot terminate early.
+        # Every candidate quote is present in the body. Pick a delimiter absent
+        # from the {expression} parts (which cannot hold a backslash) and
+        # escape it in the literal parts so the f-string cannot terminate early.
         formatted = "".join(
             value.accept(self)
             for value in node.values
             if not isinstance(value, nodes.Const)
         )
-        quote = "'" if "'" not in formatted else '"'
+        quote = next(
+            (q for q in ("'", '"', "'''", '"""') if q not in formatted),
+            # Unrepresentable: every delimiter occurs in an expression part.
+            "'''",
+        )
         string = "".join(
             (
-                _escaped_fstring_literal(value.value, quote)
+                _escaped_fstring_literal(value.value, quote[0])
                 if isinstance(value, nodes.Const)
                 else value.accept(self)
             )
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -312,11 +312,28 @@ class AsStringVisitor:
 
         # Try to find surrounding quotes that don't appear at all in the string.
         # Because the formatted values inside {} can't contain backslash (\)
-        # using a triple quote is sometimes necessary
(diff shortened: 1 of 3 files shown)


astroid/nodes/as_string.py
```diff
@@ -312,11 +312,28 @@ def visit_joinedstr(self, node: nodes.JoinedStr) -> str:
...
-                break
+                return "f" + quote + string + quote
 
+        # Every candidate quote is present in the body. Re-escape the literal
+        # parts against a quote absent from the {expression} parts (which cannot
+        # hold a backslash) so the f-string cannot terminate early.
+        formatted = "".join(
+            value.accept(self)
+            for value in node.values
+            if type(value).__name__ != "Const"
```
DanielNoord: Can't we use `isinstance` here and below?

kali834x: Yes, switched both to isinstance(value, nodes.Const).

astroid/nodes/as_string.py
```diff
@@ -312,11 +312,28 @@ def visit_joinedstr(self, node: nodes.JoinedStr) -> str:
...
 
+        # Every candidate quote is present in the body. Re-escape the literal
+        # parts against a quote absent from the {expression} parts (which cannot
+        # hold a backslash) so the f-string cannot terminate early.
+        formatted = "".join(
+            value.accept(self)
+            for value in node.values
+            if not isinstance(value, nodes.Const)
+        )
+        quote = "'" if "'" not in formatted else '"'
```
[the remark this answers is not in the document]
kali834x: Good catch, the pick between the two single quotes wasn't enough. The fallback now takes the first of ', ", ''', """ that's absent from the expression parts and escapes that character in the literal parts, with a test where an expression renders both single-character quotes. If every delimiter occurs in an expression part (only reachable through nested f-strings) there's no representable form on the pre-3.12 grammar at all, so it falls back to ''' as before.

doc/whatsnew/fragments/3235.bugfix
```diff
@@ -0,0 +1,7 @@
+``as_string()`` no longer renders an f-string wrapped in a quote that also
+appears in its body. Such an f-string terminated early and either failed to
```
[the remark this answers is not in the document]
kali834x: Reworded to "whose delimiter reappears unescaped in its body".

kali834x:
any update?

tests/test_nodes.py
```diff
@@ -1853,6 +1853,30 @@ def test_parse_fstring_debug_mode() -> None:
     assert node.as_string() == "f'3={3!r}'"
 
 
+def test_fstring_as_string_roundtrip_with_mixed_quotes() -> None:
+    # Adjacent f-strings merge into one JoinedStr whose parts together contain
+    # every candidate quote style, so as_string() must not pick a delimiter that
+    # reappears in the body: that would terminate the f-string early and let the
+    # tail reparse as a different expression.
+    node = astroid.extract_node("f'\"\"\"{v}' f\"'''\"")
```
Pierre-Sassoulas, line 1861: Let's use '''...''' or """...""" is there's an autoreformat, this is currentely unreadable.

Pierre-Sassoulas (changes requested):
Sorry for the delay. LGTM apart some readability issue (didn't add a comment everywhere but you get the idea)

tests/test_nodes.py
kali834x: Done, all three sources in the test are triple-quoted blocks now. Each still has one `\"""` because the inputs need both `'''` and `"""` in the body to reach the fallback, so one of them has to be escaped whichever delimiter wraps it.

Pierre-Sassoulas (approved):
LGTM, thank you

PR #3236, opened 2026-08-19
Render int constants past the str digits limit as hex in as_string

nabhan06:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`as_string()` renders a `Const` via `repr()`, which raises `ValueError` for an integer past `sys.get_int_max_str_digits()`. astroid's own binop inference materializes such ints (the `**` guard only trips above `1e5`), so a dataclass field shadowed by a property returning `10 ** 5000` made `_generate_dataclass_init` crash while merely building the module, since its `except (InferenceError, StopIteration)` doesn't cover `ValueError`.

Catch that `ValueError` in `visit_const` and render the int as a hex literal instead: `hex()` is exempt from the digit limit and the literal parses back to the same integer. The generated dataclass `__init__` therefore keeps the real integer as the parameter default rather than falling back to `Uninferable`, and nothing needs guarding in the dataclass brain.

### DIFF
--- a/astroid/brain/brain_dataclasses.py
+++ b/astroid/brain/brain_dataclasses.py
@@ -348,9 +348,10 @@ def _generate_dataclass_init(
             except ValueError:
                 # as_string() on the inferred value can raise, e.g. repr() of an
                 # integer past sys.get_int_max_str_digits() (a property returning
-                # ``10 ** 5000``). Leave the parameter without a default instead
-                # of letting the ValueError escape the dataclass transform.
-                pass
+                # ``10 ** 5000``). Fall back to Uninferable rather than dropping
+                # the default, so the parameter keeps a default value and the
+                # ValueError doesn't escape the dataclass transform.
+                default_str = str(Uninferable)
         else:
             # Even with `init=False` the default value still can be propogated to
             # later assignments. Creating weird signatures like:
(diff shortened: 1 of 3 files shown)


DanielNoord (commented):
Why don't we make this `Uninferable` or avoid the `repr` call? Down the line we'll have issues if we start hiding default values.

nabhan06:
Good point, dropping the default changes the signature. Switched it to fall back to `Uninferable` instead, which reuses the same default the existing property path already produces when the call result is uninferable, so the parameter keeps a default. Added an assertion on the generated `__init__` to lock that in.

DanielNoord (commented):
Isn't the default now a string instead of an integer though?

nabhan06:
Not a str constant, but you're right that it wasn't the integer any more: `default_str` is interpolated unquoted, so the generated `__init__` had `x: int = Uninferable`, a `Name` that fails inference. Went with your other option instead and dropped the brain change entirely. `visit_const` now falls back to `hex()` when `repr()` raises on an int past the digit limit, since hex is exempt from that limit and parses back to the same value, so the default stays the real `10 ** 5000` `Const`. Updated the dataclass test to assert that and added a direct `as_string()` round-trip test.

DanielNoord (approved):

DanielNoord:
Can't be bothered to backport this.

PR #1749, opened 2022-08-23
Remove str instance in model of BaseException.attrs

DanielNoord:
Supersedes #1722 as I didn't have commit rights on the branch.

Original post:

Do not require first exception argument to be a string. The word "usually" does not imply an obligation.
https://docs.python.org/3/library/exceptions.html#BaseException.args

## Steps

- [x] For new features or bug fixes, add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Related Issue

Refs: PyCQA/pylint#7233

PR #3241, opened 2026-08-19
Infer the `args` of an exception instance as a tuple of unknown contents

### LINKED ISSUE #11312: False positive `unbalanced-tuple-unpacking` when unpacking an exception's `args`
### Bug description

astroid models `BaseException.args` on an exception instance as a *known empty* tuple (`ExceptionInstanceModel.attr_args` returns `nodes.Tuple()` with no elements). Pylint's `unbalanced-tuple-unpacking` therefore reports "right side has 0 values" whenever code unpacks `exc.args`, even though the length of `args` is simply unknown.

The stdlib does this in `importlib/metadata/__init__.py`:

```python
class PackageNotFoundError(ModuleNotFoundError):
    @property
    def name(self) -> str:
        (name,) = self.args
        return name
```

Minimal example:

```python
def f():
    try:
        pass
    except ValueError as exc:
        (message,) = exc.args
        first, second = exc.args
        return message, first, second
    return None
```

### Configuration

_No response_

### Command used

```shell
pylint --disable=all --enable=unbalanced-tuple-unpacking a.py
```

### Pylint output

```shell
************* Module a
a.py:5:8: W0632: Possible unbalanced tuple unpacking with sequence : left side has 1 label, right side has 0 values (unbalanced-tuple-unpacking)
a.py:6:8: W0632: Possible unbalanced tuple unpacking with sequence : left side has 2 labels, right side has 0 values (unbalanced-tuple-unpacking)
```

### Expected behavior

No message. The contents of `args` are not known from the instance alone, so the check cannot say the unpacking is unbalanced.

### Pylint version

```shell
pylint 4.1.0-dev0
astroid 4.4.0-dev0
Python 3.13.11
```

### OS / Environment

macOS

### Additional dependencies

_No response_

ekanshul:
## Type of Changes

| | Type |
|---|---|
| ✓ | :bug: Bug fix |
| | :sparkles: New feature |
| | :hammer: Refactoring |
| | :scroll: Docs |

## Description

`ExceptionInstanceModel.attr_args` returns a `nodes.Tuple` with no elements. Consumers read that as a *known empty* tuple: pylint reports `unbalanced-tuple-unpacking` with "right side has 0 values" for

```python
class PackageNotFoundError(ModuleNotFoundError):
    @property
    def name(self):
        (name,) = self.args   # stdlib, importlib/metadata/__init__.py
        return name
```

and for `first, second = exc.args` inside an `except` block.

The arguments an exception was created with are not known from the instance alone, so `args` is now inferred as an instance of `builtins.tuple` (unknown contents, still iterable and subscriptable). This keeps the intent of #1749 (no assumption that the first argument is a `str`) while no longer claiming a length of zero. The three existing tests that asserted `nodes.Tuple` are updated and a regression test is added; pylint's functional suite passes against this branch (apart from `unbalanced_tuple_unpacking`, which fails identically on current astroid main in my environment because of the `ctypes` brain).

Closes pylint-dev/pylint#11312

### DIFF
--- a/astroid/interpreter/objectmodel.py
+++ b/astroid/interpreter/objectmodel.py
@@ -817,8 +817,12 @@ class InstanceModel(ObjectModel):
 
 class ExceptionInstanceModel(InstanceModel):
     @property
-    def attr_args(self) -> nodes.Tuple:
-        return nodes.Tuple(parent=self._instance)
+    def attr_args(self) -> bases.Instance:
+        # The arguments an exception was created with are not known from the
+        # instance alone, so ``args`` is a tuple of unknown contents rather
+        # than a known empty tuple.
+        builtins_ast_module = AstroidManager().builtins_module
+        return builtins_ast_module["tuple"].instantiate_class()
 
     @property
     def attr___traceback__(self):
--- a/doc/whatsnew/fragments/11312.bugfix
+++ b/doc/whatsnew/fragments/11312.bugfix
@@ -1,0 +1,5 @@
+The ``args`` attribute of an exception instance is now inferred as a ``tuple``
+instance of unknown contents instead of a known empty tuple, so unpacking
+``exc.args`` no longer looks unbalanced.
+
+Closes pylint-dev/pylint#11312
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -6193,7 +6193,8 @@ def test_subclass_of_exception(code) -> None:
     inferred = next(extract_node(code).infer())
     assert isinstance(inferred, Instance)
     args = next(inferred.igetattr("args"))
-    assert isinstance(args, nodes.Tuple)
+    assert isinstance(args, Instance)
+    assert args.qname() == "builtins.tuple"
 
 
 def test_ifexp_inference() -> None:
--- a/tests/test_object_model.py
+++ b/tests/test_object_model.py
@@ -698,7 +698,8 @@ class ExceptionModelTest(unittest.TestCase):
         """)
         assert isinstance(ast_nodes, list)
         args = next(ast_nodes[0].infer())
-        assert isinstance(args, nodes.Tuple)
+        assert isinstance(args, astroid.Instance)
+        assert args.qname() == "builtins.tuple"
         tb = next(ast_nodes[1].infer())
         # Python 3.11: If 'contextlib' is loaded, '__traceback__'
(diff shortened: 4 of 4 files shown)


DanielNoord (commented):
Can't we set something on the `Tuple` node to indicate this as well?

ekanshul:
I tried that, and a marker on the `Tuple` node does not fix the false positive, because the consumers dispatch on the node type before they would ever look at the marker.

pylint reaches the message through `VariablesChecker._nodes_to_unpack`:

```python
if isinstance(node, (nodes.Tuple, nodes.List, nodes.Set, *DICT_TYPES)):
    return node.itered()
```

A `Tuple` with no `elts` returns `[]`, so the length is 0 and `_check_unpacking` reports "right side has 0 values". I patched `attr_args` to return `nodes.Tuple(parent=self._instance)` with an added `unknown_contents = True` attribute and ran it against pylint main:

```
# marker on nodes.Tuple
/tmp/t.py:4:8: W0632: Possible unbalanced tuple unpacking with sequence :
    left side has 1 label, right side has 0 values (unbalanced-tuple-unpacking)

# this PR (builtins.tuple instance)
(no messages)
```

The difference is which way each option fails. Returning an `Instance` is fail-safe: `_nodes_to_unpack` returns `None` for it, `_check_unpacking` falls through to its "is this iterable at all" branch, and a `tuple` instance is iterable, so nothing is reported. Any consumer that special-cases `nodes.Tuple` gets its existing unknown-value path for free. A marker is fail-open: every consumer has to be taught to check it, and until each one is, it still sees a `Tuple` of length 0. That includes released pylint versions running against a newer astroid.

There is also precedent for this shape a few lines down in the same class, where `__traceback__` returns `traceback_type.instantiate_class()` rather than a node.

That said, if you want the marker as well as the type change, so that consumers can distinguish "tuple of unknown contents" from any other `tuple` instance, I am happy to add it. I would want it on the instance rather than on a `Tuple` node, since a `Tuple` node with unknown contents is not really a `Tuple` node. Let me know which you prefer.

DanielNoord (commented):
@Pierre-Sassoulas @jacobtylerwalls What do you think of this change? I guess the explanation sort of makes sense?

Pierre-Sassoulas:
I suppose it was closed by mistake ? I skimmed and it looks like the fix we want.

PR #3248, opened 2026-08-24
guard empty six.with_metaclass() call against IndexError

nabhan06:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`transform_six_with_metaclass` indexes `call.args[0]` for any class whose sole base is a `six.with_metaclass(...)` call, but the predicate matches even when that call has no positional argument, so `class Foo(six.with_metaclass())` raises `IndexError` while the module is being built. The generic `with_metaclass` handling in `FunctionDef.infer_call_result` hits the same `IndexError` on `caller.args[0]` when such a call is inferred. Raise `InferenceError` at both sites when the argument list is empty instead of indexing into it.

Refs #3248

### DIFF
--- a/astroid/brain/brain_six.py
+++ b/astroid/brain/brain_six.py
@@ -220,6 +220,11 @@ def transform_six_with_metaclass(node):
     If so, inject its argument as the metaclass of the underlying class.
     """
     call = node.bases[0]
+    if not call.args:
+        # ``six.with_metaclass()`` written with no positional metaclass
+        # argument is invalid, but the predicate still matches it. Leave the
+        # class untransformed instead of indexing into an empty argument list.
+        return node
     node._metaclass = call.args[0]
     return node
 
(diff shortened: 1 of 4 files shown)


DanielNoord (commented):
Shouldn't this be an inference error instead?

nabhan06:
Good point, that's cleaner. Pushed 4d0c92d: infer_call_result now raises InferenceError for an empty with_metaclass() call instead of falling through and yielding a bogus temporary_class base. The brain_six guard stays as-is since that one runs during module build, not inference, so all it can do is leave the class untransformed (same as the add_metaclass guard next to it). Ancestors on the invalid class now come back empty rather than [temporary_class, object].

DanielNoord (commented):
I meant raising the error in the transform where you are now returning

nabhan06:
Done, the transform now raises InferenceError instead of returning the node (43c2d7d), so the module build fails with a clear error on the invalid class rather than an IndexError. I kept the guard in infer_call_result since that path is still reachable when the class transform doesn't apply, e.g. `class B(six.with_metaclass(), object)`; the test covers both.

DanielNoord (approved):

PR #3256, opened 2026-08-27
guard mock-class reconstruction in infer_enum_class

nabhan06:
## Type of Changes

## Description

`infer_enum_class` mocks each enum member by rendering a small class body with `str.format` and reparsing it through `string_build`. A non-str `Const` value is handed to `.format()` as-is, so an integer literal past the int-to-str conversion limit written as a hex literal that the tokenizer still accepts (`A = 0xff...`) makes the `.format()` raise `ValueError: Exceeds the limit (4300 digits)` and aborts the whole transform while astroid is just building the module.

Such a value is now rendered back as a hex literal, which the tokenizer accepts at any length, so the member is still mocked like any other and `Flags.A.value` infers to the actual int. Regression test added.

### DIFF
--- a/astroid/brain/brain_namedtuple_enum.py
+++ b/astroid/brain/brain_namedtuple_enum.py
@@ -18,6 +18,7 @@ from astroid.builder import AstroidBuilder, _extract_single_node, extract_node
 from astroid.context import InferenceContext
 from astroid.exceptions import (
     AstroidError,
+    AstroidSyntaxError,
     AstroidTypeError,
     AstroidValueError,
     InferenceError,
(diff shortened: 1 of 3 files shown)


DanielNoord (commented):
Breaking feels bad, can't we handle just the specific error case instead but still do the rest of the transformation?

nabhan06:
Makes sense. Dropped the try/except and the `break`. The only change now is how a non-str `Const` is rendered into the mock body: `str()` as before, and on the int-to-str `ValueError` fall back to `hex()`, which the tokenizer takes at any length. The member gets mocked like any other and `Flags.A.value` infers to the actual int, which the test now checks. I had to flatten the `Const` branch into if/elif/else so the extra `try` doesn't push the function past pylint's nesting limit.

One thing I left alone: a value that goes through `as_string()` instead (say `A = -0xff...`) would hit the same limit in `repr()`, but that's a different path and seemed better kept out of this one.

PR #3265, opened 2026-08-30
render non-finite floats so as_string round-trips

kali834x:
## Type of Changes

|     | Type          |
| --- | ------------- |
| ✓   | :bug: Bug fix |

## Description

`AsStringVisitor.visit_const` renders every constant through `repr()`, which turns a non-finite float such as the one behind `1e999` into the bare token `inf` (and `nan` for a NaN). That text does not round-trip: reparsing `inf` yields a `Name` lookup rather than the float. The dataclass brain reparses `as_string()` output when it synthesizes `__init__`, so a field default of `1e999` comes back as `def __init__(self, x: float = inf)`, and `inf` then resolves to whatever `inf` happens to be in the module, for instance a module-level `inf = os.system`, silently changing the inferred default. I ran into it while checking that field defaults survive the round-trip and saw the value resolving as a name. The fix mirrors `ast.unparse` and emits an overflowing literal, `1e309` for `inf` and `(1e309-1e309)` for `nan`, so float and complex constants render back to their own values. The change stays inside `visit_const` and leaves finite numbers, strings and other constants untouched. A ChangeLog fragment is included.

### DIFF
--- a/astroid/nodes/as_string.py
+++ b/astroid/nodes/as_string.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import sys
 import warnings
 from collections.abc import Iterator
 from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


kali834x:
gentle ping

DanielNoord (commented):
@kali834x Please understand that maintainers are not particularly interested in PRs like these that are clearly AI generated, spam our queue and do not solve anything our actual users will often run into.

PR #3288, opened 2026-09-15
Do not treat a bare annotation in an enum body as a member

kdelay:
## Type of Changes

|   | Type |
| --- | --- |
| ✓ | :bug: Bug fix |

## Description

`infer_enum_class` replaces every name in an enum body with a mocked member class, including one that only carries an annotation. Such an annotation declares the type of the member values, so the name does not exist at runtime:

```python
class Color(Enum):
    label: str
    RED = 1

Color.label   # AttributeError
```

pylint reports nothing for that access today; now it reports `no-member`. Refs pylint-dev/pylint#11413, the opposite symptom (`_value_: Payload` resolved as a member), which also needs a pylint-side change.

`__members__` already excluded such names (#2263). An annotation followed by an assignment of that name stays a member.

Both new tests fail on the unpatched tree; the astroid and pylint suites are unchanged.


tests/test_scoped_nodes.py
```diff
@@ -2538,19 +2538,27 @@ class Veg(Enum):
...
 
     Veg.TOMATO.value
     """)
-    inferred_member_value = node.inferred()[0]
-    assert inferred_member_value.value is None
+    # The lookup of a name an Enum class does not have is delegated to
+    # ``EnumMeta.__getattr__`` up to Python 3.11, which gives ``Uninferable``.
+    # That method was removed from the standard library in 3.12, leaving
+    # nothing to delegate to, so from 3.12 on the inference fails instead.
+    if PY312_PLUS:
```
DanielNoord: I don't quite understand this change. Why is there a difference here?

DanielNoord (dismissed):

DanielNoord (commented):
Sorry, did not mean to approve.

kdelay:
Sorry, that wasn't clear from the diff. Now that the bare annotation is not a member, `Veg.TOMATO` is a plain attribute lookup on the class, and the runtimes differ there: `enum.EnumMeta.__getattr__` exists up to 3.11 (`'__getattr__' in vars(enum.EnumMeta)` is True on 3.11, False on 3.12+), so the lookup is delegated and gives `Uninferable`; from 3.12 there is nothing to delegate to and inference raises. Without the split, the 3.10/3.11/pypy3.10 jobs failed with `DID NOT RAISE InferenceError`. Happy to assert only "not a member" on both branches if you prefer.

DanielNoord:
What happens at runtime when we try to access these attributes on the various Python versions?

PR #3290, opened 2026-09-19
Fix numpy brain inference under NumPy 2.x

Julfried:
## Type of Changes

## Description

Closes #2710

NumPy 2.0 renamed `numpy.core` to `numpy._core`, which broke all numpy brain
hooks and made most numpy attributes infer as `Uninferable` or raise
`InferenceError` when analyzing code that uses NumPy 2.

- Register the existing numpy brain hooks for the new `numpy._core.*` module
  names, keeping the old `numpy.core.*` registrations for NumPy 1.x.
- Gate aliases that NumPy 2.0 removed (`float_`, `complex_`, `string_`, and
  similar) so they are no longer inferred as existing.
- Add NumPy 2.0 members: `np.ulong`, the `bitwise_count` and `vecdot` ufuncs,
  and the `ndarray` members `mT`, `device`, and `to_device`.
- Bind the platform-specific scalar types (`float96`, `float128`,
  `complex192`, `complex256`) in the top-level `numpy` module, because NumPy 2
  sets them through a `globals()` loop that static analysis cannot track.
- Let the `numpy.distutils` regression test skip itself when the installed
  numpy no longer provides `numpy.distutils`.
- Allow numpy 2 in `requirements_full.txt` so CI tests against it.

The numpy brain test suite passes with numpy 2.5.3 and no failures. I also
checked locally that the NumPy 1.x tests still pass — CI has no job for that
combination. This is a first pass at NumPy 2 support that fixes the main
breakages. NumPy 2.0 has been out for over two years, so more gaps are likely
to surface and can follow in separate PRs.


DanielNoord (approved):
LGTM! Better than what we had before, thanks!

